PR #15, opened 2017-06-28
Feature/route navigation

Ishadijcks:
Route navigation works now, supports multiple regions.

### DIFF
--- a/src/scripts/GameConstants.ts
+++ b/src/scripts/GameConstants.ts
@@ -89,5 +89,59 @@ namespace GameConstants {
         "mediumfast",
         "fast"
     }
+
+    // Map navigation
+    export const AMOUNT_OF_ROUTES = 25;
+
+    /**
+     * Each route contains a list of routenumbers that need to be completed
+     * Access with routeRequirements.region.routeNumber
+     */
+    export const routeRequirements = {
+        0: {
+            1: [],
+            2: [1],
+            3: [2],
+            4: [3],
+            5: [4],
+            6: [5],
+            7: [5, 10],
+            8: [5, 6, 7],
+            9: [4],
+            10: [9],
+            11: [6, 12],
+            12: [7, 10, 11],
+            13: [11, 12],
+            14: [13],
+            15: [14],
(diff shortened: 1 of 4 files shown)


src/scripts/GameConstants.ts
```diff
@@ -74,5 +90,59 @@ namespace GameConstants {
...
+            22: [1],
+            23: [22],
+            24: [4],
+            25: [24],
+        }
+    };
+
+    export const routeBadgeRequirements = {
+        0: {
+            3: GameConstants.Badge.Boulder,
```
dennism1997: the accesToRoute method does not seem to work when the index is not specified (so for example route 1,2,4 etc)

`Uncaught TypeError: Cannot read property '1' of undefined
    at Function.RouteRequirement.accessToRoute (script.min.js:2900)
    at <anonymous>:1:18`

src/scripts/Player.ts
```diff
@@ -3,22 +3,27 @@
...
+    private static _dungeonTokens: number = 0;
     private static _caughtPokemonList = [];
-    private static _route: number;
-    private static _region: GameConstants.Region;
-    private static _pokeballs: number[] = [3, 2, 4, 1];
+    private static _route: number = 1;
+    private static _routeKills: number[] = Array.apply(null, Array(GameConstants.AMOUNT_OF_ROUTES)).map(Number.prototype.valueOf, 0);
+    private static _routeKillsNeeded: number = 10;
+    private static _region: GameConstants.Region = GameConstants.Region.kanto;
+    private static _gymBadges: GameConstants.Badge[] = [GameConstants.Badge.Boulder];
```
dennism1997: Maybe an array of booleans is easier? Then we can use
`if(_gymbadges[0]) do stuff`

Ishadijcks: It's easier and faster, but with an array of booleans we can't catch the undefined badge.
`Player.hasBadge(badge)` works correctly now, and can be used.

dennism1997: That's right

PR #19, opened 2017-06-29
Change container width on large screens to 80%

### LINKED ISSUE #16: Bootstrap should ignore windows scale setting
UI at scale 125%
image

UI at scale 100%
image

This is not the browser zoom, but a windows setting.
At ishadijcks.github.io we did not have this problem, so it is possible.

Aegyo:
To fix #16

### DIFF
--- a/src/styles/main.css
+++ b/src/styles/main.css
@@ -10,7 +10,7 @@ body {
 
 @media (min-width: 1200px) {
     .container {
-        width: 80%;
+        width: 70%;
     }
 }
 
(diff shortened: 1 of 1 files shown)


Ishadijcks:
Can confirm it works, but can you change the width to 70%? I think that looks a bit better

Ishadijcks:
(I tried to push to my branch yourself, but it didn't work, not sure why) 
image

Ishadijcks:
Thanks!

PR #20, opened 2017-06-29
Get Routekills working

Aegyo:
This appears to work, however the incrementing of kills on a route is rather ugly. I'm not very familiar with knockout, but it seems that you can't just ++. It might be worth defining a function to increment an observable if this case comes up a lot and there is not a nicer way of doing it.


Ishadijcks:
Amazing, thanks!

PR #25, opened 2017-06-30
Make caughtPokemonList observable

### LINKED ISSUE #24: The PokemonAttackObservable does not get updated
The value that gets returned by the `calculatePokemonAttack()` method changes, as can be seen in the console. However, the value in the browser remains the same

Aegyo:
The KnockoutComputed value needs to have an observable to tell it when it has changed.

In the future, this PokemonCaughtList probably needs to be `KnockoutObservableArray<KnockoutObservable<CaughtPokemon>>`, so that the computed value will also update when pokemon level up and gain a higher attack stat, but this is sufficient for now.

Fixes #24

### DIFF
--- a/src/scripts/Player.ts
+++ b/src/scripts/Player.ts
@@ -8,7 +8,7 @@ class Player {
 
     private static _money: KnockoutObservable<number> = ko.observable(0);
     private static _dungeonTokens: number = 0;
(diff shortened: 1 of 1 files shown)


Ishadijcks:
Looks great!

PR #30, opened 2017-07-02
Progress on pokemonList ui feature

Aegyo:
List of Pokemon now updates with pokemon info when they are caught.

I've checked that the level updates when the pokemon gains exp by manually changing one of the exp values.

Pokemon attack still needs to be calculated, also there seems to be an issue with the pokemon table on narrow screens where it takes up more space than the background box.

### DIFF
--- a/src/scripts/pokemons/PokemonHelper.ts
+++ b/src/scripts/pokemons/PokemonHelper.ts
@@ -21,9 +21,9 @@ class PokemonHelper {
 
     public static calculateLevel(pokemon: CaughtPokemon): number {
         let level;
-        switch (PokemonHelper.getPokemonByName(pokemon.name).levelType.replace(' ', '')) {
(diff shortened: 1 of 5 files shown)


src/scripts/pokemons/PokemonHelper.ts
```diff
@@ -19,4 +19,35 @@ class PokemonHelper {
         return GameConstants.PokemonType[id];
     }
 
+    public static calculateLevel(pokemon: CaughtPokemon): number {
```
dennism1997: moving the method here is a great idea

dennism1997:
Nice job! 💯

src/scripts/pokemons/PokemonHelper.ts
Ishadijcks: Ha, agreed

PR #32, opened 2017-07-02
Finish Pokemon List

dennism1997:
The template wasn't necessary after all. But I do think this code is more clean and easier to maintain.
This PR also implements less and all the required dependencies.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -11,6 +11,7 @@ const browserSync = require('browser-sync');
 const del = require('del');
 const runSequence = require('run-sequence');
 const bsConfig = require("gulp-bootstrap-configurator");
+const less = require('gulp-less');
 
 /**
  * Push build to gh-pages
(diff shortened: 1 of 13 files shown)


gulpfile.js
```diff
@@ -24,7 +25,7 @@ const srcs = {
     buildArtefacts: 'build/**/*',
     scripts: 'src/scripts/**/*.ts',
     html: ['src/*.html', 'src/templates/*.html'],
-    styles: 'src/styles/**/*.css',
+    styles: 'src/styles/**/*.less',
```
Ishadijcks: All css is also less right?

Ishadijcks:
image
On half screen the list overflows.

Also it seems like the attack and level listeners aren't working?

dennism1997:
On half screen the columns should collapse right?

gulpfile.js
dennism1997: yes. But also all less is css

dennism1997:
The listeners did work already I think. Just did some refactoring. The values don't get computed yet so that's why they don't change

Ishadijcks:
Ah, I see.
Note that attackBonus is the bonus gained from breeding/items.
And it's attack is calculated using the base Pokémon and the above factors.

PR #33, opened 2017-07-02
Feature/map

Ishadijcks:
Add basic routenavigation using the worldmap.


src/index.html
```diff
@@ -132,7 +131,364 @@
             </div>
 
             <div class="page-item" style="background-color: green; min-height: 350px">
-                <h2>Map</h2>
+                <svg id="map" viewBox="0 0 650 400" preserveAspectRatio="xMaxYMax meet">
```
dennism1997: Can't we put this inside another file and load it into the page?

src/styles/main.css
```diff
@@ -0,0 +1,63 @@
+@font-face {
```
dennism1997: duplicate file

src/index.html
Ishadijcks: Not sure, but to improve loading times we want to load the smallest number of files possible.
So we could do it, and have gulp compile it to a single `html` file.

src/styles/main.css
Ishadijcks: Ah I see, webstorm put the `main.ss` inside the `main.less`

Ishadijcks:
pr outdated because of other merge to develop.

Ishadijcks:
Fixed

Ishadijcks:
Ugh, noticed a bug in alreadyCaughtShiny

src/scripts/Player.ts
```diff
@@ -122,7 +124,17 @@ class Player {
...
+    public static alreadyCaughtPokemonShiny(pokemonName: string) {
+        for (let i: number = 0; i < this.caughtPokemonList.length; i++) {
+            if (this.caughtShinyList()[i] == pokemonName) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public static capturePokemon(id: number, pokemonName: string, shiny: boolean = false) {
```
dennism1997: id is not necessary anymore

dennism1997: and how did I make this yellow lol

Ishadijcks: Fixed

dennism1997:
we should rotate the route numbers too. Isn't as simple as adding rotate(-90, 0,0) unfortunately

Aegyo:
In calculateRouteCssClass, the GameConstants.RouteCssClass parts don't need to be observable because they never change. What is the purpose of using an enum for this? Just returning the string seems simpler, and you have needed to type out the string anyway to get it from GameConstants.

Ishadijcks:
Yeah good point, I guess no need to make them an enum.

PR #38, opened 2017-07-02
Implement evolution

Aegyo:
Pokemon now evolve when they reach the appropriate level, only check when they gain a level, and don't checkever again after they have evolved.

Was the evolved boolean on CaughtPokemon only there to skip evolution checks? If so it is not needed with this method and can be removed.


Ishadijcks:
Can you add a check to not apply the subscription if `level > evoLevel`?
For when we create `CaughtPokémon` from the savefile, don't want to create a new subscription.
Also that dispose is really cool.

Aegyo:
On second thought, it may be better to use the evolved boolean for preventing the evolver being created. This would allow catching of pokemon at higher levels, either from wild or with the perk to hatch pokemon at level 50.

All old pokemon get marked evolved, so It should not be a problem.

Ishadijcks:
Yeah I agree.

PR #40, opened 2017-07-02
Feature/map improvement

Aegyo:
Prevented a lot of unnecessary access checks, and also fixed route 25 thinking it is route 9

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -355,8 +355,8 @@
                         </g>
 
                         <!--Route 25-->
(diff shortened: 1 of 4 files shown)


Ishadijcks:
Good work, totally forgot about gymBadges triggering a re-evaluation!

Aegyo:
Moving to a route was causing every route to check what class it should have, so I've used .peek() on the Player.route observable to prevent the computable subscribing to it.

Of course, this broke the current route being set ever, so it is now handled by the moveToRoute function. The data-route attribute on each \<rect> is needed for the jquery selector.

PR #43, opened 2017-07-03
Feature/paypal

Ishadijcks:
Add a non-obnoxious modal where people can donate if they'd like to support us.
Donations will mostly go towards hosting fees, and thanking contributors (maybe some coffee to keep us going).
If it sounds too beggy, please let me know as that is not the intention.

image

image


dennism1997:
Maybe make mew a pink background? And I'm not sure about the unown on the next row. I think always having them divided in 2 rows looks better

Ishadijcks:
Agreed
image

PR #44, opened 2017-07-03
Implement saving

Aegyo:
Majority of changes here are updating other files to reference the instance of Player (named player) instead of the class.

The only property not moved into the instance is Player.routeKillsNeeded, I'm not sure why this is observable, but as far as I know it will never change. Unless there are plans for this I don't know about yet, it might be best to move this into GameConstants.

I haven't encountered any issues with saving or loading in my limited testing. If you want to test it out, there is a Save.reset() function which will clear out the localStorage and reset your browser, just in case something breaks (or you want to test loading the game when there is no saved player)


Ishadijcks:
`Player.routeKillsNeeded` can be decreased to 7 as a prestige perk

Aegyo:
Ah ok, I'll move that into the constructor also then.

src/scripts/Save.ts
```diff
@@ -0,0 +1,35 @@
...
+    public static filter(object: any, keep: string[]): Object {
+        let filtered = {}, prop;
+        for (prop in object) {
+            if (keep.indexOf(prop) > -1) {
+                filtered[prop] = object[prop]
+            }
+        }
+        return filtered
+    }
+
```
Ishadijcks: Clever, I like it

Aegyo: Should I add a comment explaining the parameters of that function, or is it fairly clear what they should be?

Ishadijcks: Yeah, would be nice!

Ishadijcks:
Getting a typescript compile error:
```src\scripts\Player.ts(36,13): error TS2322: Type 'KnockoutObservableArray<{}>' is not assignable to type 'KnockoutObservableArray<string>'.```

at 
```this._caughtShinyList = ko.observableArray(savedPlayer._caughtShinyList);```

Aegyo:
Strange, I don't get that error. Any idea what is causing that?

Ishadijcks:
Also... 
image

Aegyo:
Yeah I just noticed that too, it's new from that last commit. I'll close this until I fix it and make sure there are no other bugs.

Ishadijcks:
Calling `Save.reset()`
Routekills works,
Let the game save, 
Refresh browser,
NaN Routekills

Aegyo:
I'm still not getting that compile error, even when restarting gulp.

Ishadijcks:
Saving seems to work now, but before I merge I'm going to search for the cause of the error.

Ishadijcks:
`ko.observableArray(savedPlayer._caughtShinyList);`
to
`ko.observableArray<string>(savedPlayer._caughtShinyList);`
fixes the problem.

I guess my WebStorm is somehow stricter than yours or something?

PR #47, opened 2017-07-04
Shiny Pokemon

Aegyo:
- Adds shiny images
- Displays shiny image for enemy wild pokemon and in pokemon list
- caughtShinyList is used to keep track of which pokemon are shiny
- Catching a shiny pokemon will update caughtShinyList, turning any existing non-shiny version into a shiny


src/scripts/Player.ts
```diff
@@ -175,13 +175,35 @@ class Player {
...
+        for (let i=0; i< this._caughtPokemonList().length; i++) {
+            let pokemon = this._caughtPokemonList()[i];
+            if (pokemon.name == pokemonName) {
+                this._caughtPokemonList.splice(i, 1);
+                return true;
+            }
+        }
+        return false
+    }
+
```
Ishadijcks: Hold on
Why are you removing and adding the Pokémon to the list?
This resets the experience and is not needed as with the caughtShinyList a caughtPokemon doesn't need to be marked as shiny.

Aegyo: It should only be removing a pokemon if you just caught a shiny pokemon, and don't already have that pokemon shiny.

So should it instead just be turning the pokemon you already have into a shiny?

Ishadijcks: Yeah seems easier

Aegyo: Ok, I'll change it to that method

Ishadijcks:
Are the shiny pokemon different images? I just thought they changed colors?

image

image

PR #49, opened 2017-07-05
Sorting for pokemon list

Aegyo:
This should be easily extendable to other properties, just add an option to the sort list.


src/scripts/Player.ts
```diff
@@ -18,6 +18,8 @@ class Player {
     private _shinyList: boolean[];
     private _notCaughtBallSelection: GameConstants.Pokeball;
     private _alreadyCaughtBallSelection: GameConstants.Pokeball;
+    private _sortOption: KnockoutObservable<string>;
+    private _sortDirection: KnockoutObservable<string>;
```
Ishadijcks: Isn't this just a boolean `sortAscending'?

src/scripts/Player.ts
```diff
@@ -18,6 +18,8 @@ class Player {
     private _shinyList: boolean[];
     private _notCaughtBallSelection: GameConstants.Pokeball;
     private _alreadyCaughtBallSelection: GameConstants.Pokeball;
+    private _sortOption: KnockoutObservable<string>;
```
Ishadijcks: Could be an enumator for clarity.

src/scripts/Player.ts
```diff
@@ -312,7 +327,7 @@ class Player {
     }
 
     public toJSON() {
-        let keep = ["_money", "_dungeonTokens", "_caughtShinyList", "_route", "_caughtPokemonList", "_routeKills", "_routeKillsNeeded", "_region", "_gymBadges", "_pokeballs", "_shinyList", "_notCaughtBallSelection", "_alreadyCaughtBallSelection"];
+        let keep = ["_money", "_dungeonTokens", "_caughtShinyList", "_route", "_caughtPokemonList", "_routeKills", "_routeKillsNeeded", "_region", "_gymBadges", "_pokeballs", "_shinyList", "_notCaughtBallSelection", "_alreadyCaughtBallSelection", "_sortOption"];
```
Ishadijcks: If you're saving the `sortOption`, you also want to save the `sortDirection`

src/scripts/pokemons/PokemonHelper.ts
```diff
@@ -62,4 +62,39 @@ class PokemonHelper {
         return src;
     }
 
+    public static compareBy(property: string, direction: number): (a: CaughtPokemon, b: CaughtPokemon) => number {
```
Ishadijcks: I like this, much better than a method for each sorting option

src/scripts/pokemons/PokemonHelper.ts
```diff
@@ -62,4 +62,39 @@ class PokemonHelper {
...
+                } else if (_a.id > _b.id) {
+                    return 1;
+                }
+            } else if (_a[property] < _b[property]){
+                res = -1;
+            } else if (_a[property] > _b[property]) {
+                res = 1;
+            } else {
+                res = 0
+            }
```
Ishadijcks: Can't this be simplified to res = _a[property] - _b[property]`?

Aegyo: This way will also sort strings (pokemon name)

Ishadijcks: Good point!

src/scripts/Player.ts
Aegyo: It is essentially just a boolean, but the select element always seems to have a string as it's value

src/scripts/Player.ts
Aegyo: Can you elaborate how that would work? I'm not used to using enums

Ishadijcks: The advantage of using enumerators is that you can make sure you only one of the values that you want to return, and you have those values in a nice concise place. For example with the GameStates you can easily see a list of all the possible states.

In this case you could make 
```
    export enum SortOption{
        "Shiny",
        "Level",
        "Attack",
    }
```

Then giving `_sortDirection` the type `KnockoutObservable<GameConstants.SortOption>`
Since the enum is a string, the conversion between `value` and the variable should work.

Aegyo: Knockout can use an array to populate a select element, maybe that would be a better choice?

Ishadijcks: Perhaps create an array from the enumerator options:

https://stackoverflow.com/questions/18150659/converting-enums-to-array-of-values-putting-all-json-values-in-an-array

https://github.com/Microsoft/TypeScript/issues/12877

Aegyo: I was planning to have an array of objects like `{optionText: Level, optionValue: "levelObservable"}` since sometimes the option text should be different to the property name. Also I think this allows for returning something other than a string.

Is it possible to create an enum which would become an array like that?

Ishadijcks: What does `Level` mean here?

Aegyo: I missed out some "", it should just be the string which is displayed in the options list

Aegyo: Ah, how about making an array somewhere (a Settings class maybe), with the optionValue for each being some GameConstants.SortOption?

Ishadijcks: Sounds good, perhaps in `GameConstants`?

dennism1997:
Is it possible to sort it by just clicking on the titles? Or maybe make ascending/descending a checkbox instead of an input.

Aegyo:
It would be possible to sort by clicking titles, but then there isn't a title for shiny or id, so the dropdown select would probably still be needed.

The ascending/descending thing being a checkbox is something I thought of, but I wasn't sure of a way to make it look good. I can change it over to a checkbox now though if you think that would be better.

PR #55, opened 2017-07-07
Fix/player constructor

Aegyo:
New properties weren't being added to old saves, which could be a problem for future updates. This checks if each property exists on savedPlayer, and uses a default if it does not. A couple needed to be checked for typeof == undefined since they could have falsey values, resulting in the default always being set.

Also removed _shinyList since _caughtShinyList is used instead.


Ishadijcks:
Yep good one.

PR #61, opened 2017-07-10
Add max height to caughtPokémon list.

### LINKED ISSUE #57: The pokemon list needs a max height
# Current Behavior
The pokemon list just keeps expanding below

# Improved Behavior
It should have some sort of max height

Ishadijcks:
All overflow is scrollable

Fixes #57

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -739,7 +739,7 @@
                 <h2>Pokeball management</h2>
             </div>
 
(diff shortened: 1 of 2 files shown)


dennism1997:
Can you make it a percentage of the viewport instead of a hardcoded value?

PR #62, opened 2017-07-10
Feature/pokeballs

Ishadijcks:
Add a Pokéball selection to the main screen.

The user can select which Pokéball to use on new and caught Pokémon.
If that Pokéball is not available, the next highest is used.


dennism1997:
I think having them both initialize at pokeball is better. Also, the player should start with a few pokeballs in the inventory

PR #82, opened 2017-07-12
Feature/shop

dennism1997:
Shops are not populated yet. That is something we still have to think about. Also, items are not implemented yet, but there are some dummy classes


Ishadijcks:
The price under the items doesn't update anymore?

Ishadijcks:
Can you add a -1, -10, -100 and +1 button
Other than that it's great!

PR #87, opened 2017-07-13
Feature/types

dennism1997:
Add type effectiveness to the pokemon attack


Aegyo:
Is this _shards array for keeping track of the shard upgrades? If so it might be best to rename it, since there will need to be a separate array to keep track of collected shards for spending on upgrades, and it would make more sense for that to be named shards rather than the upgrades.

Ishadijcks:
Agreed

dennism1997:
Isha and me discussed about the pokemon attack observable. We thought it was best to show the regular pokemon attack (without types factored in) and do the actual damage with types. That way the pokemon attack does not flicker that much. What do you think? 

We were not sure about upgrades either, as upgrades are pretty boring. You buy them and you get something in return, but there's no real strategy behind it.

Aegyo:
Showing the actual attack your pokemon do to an enemy would allow players to focus on collecting a specific shard type so that they can focus on improving their damage against whatever they are weakest to.

dennism1997:
Maybe then we should also show the type of the pokemon you are battling

Aegyo:
That information could be shown in the pokedex, it might also be helpful to give players a type effectiveness chart.

Ishadijcks:
What needs to be done to make this pr done?

Aegyo:
I think just gaining shards when you defeat a pokemon.

How many shards should be awarded for defeating dungeon and gym pokemon? Also if a pokemon only has a single type, should the player get double for that type?

Ishadijcks:
Nah it's one shard for each type, so single type is 1 shard awarded.
Dungeon can be 3 each, with 20 for the boss (maybe a penalty with auto-dungeons)
Gyms at 5 each, as they have no replayability really.

Aegyo:
Shard modal currently looks like this:
modal

Ishadijcks:
Are you using the old colors? They should be neatly stored in an array somewhere

Aegyo:
Yeah these are the old unlocked colors, I'll move them to an array in TypeHelper.

Ishadijcks:
Can you change the titles to:

Normal 1000<shard image>

I liked how colorful the old modal was.

This looks definitely clean and concise.

Also if it's not too much work, is it possible make the progress bar have 10 steps?

Aegyo:
How is this?
modal

Ishadijcks:
Can you make the not bought portion of the progress bar the old locked color?

src/scripts/Battle.ts
```diff
@@ -50,6 +50,10 @@ class Battle {
     public static defeatPokemon() {
         player.gainMoney(this.enemyPokemon().money);
         player.gainExp(this.enemyPokemon().exp, this.enemyPokemon().level, false);
+        player.gainShards(this.enemyPokemon().type1, 1);
+        if (this.enemyPokemon().type2 != -1) {
```
Ishadijcks: Change -1 to GameConstants.PokemonType.None`

Aegyo:
Yeah, here's an updated image:

modal

Ishadijcks:
I like it, but I'll change the locked to something lighter, right now it's a tad confusing.

Maybe an opacity of 0,6 will do it?

src/scripts/dungeons/DungeonBattle.ts
```diff
@@ -7,6 +7,11 @@ class DungeonBattle extends Battle {
         DungeonRunner.fighting(false);
         player.gainMoney(this.enemyPokemon().money);
         player.gainExp(this.enemyPokemon().exp, this.enemyPokemon().level, false);
+        let shardAmount = DungeonRunner.fightingBoss() ? 20 : 3;
+        player.gainShards(this.enemyPokemon().type1, shardAmount);
+        if (this.enemyPokemon().type2 != -1) {
+            player.gainShards(this.enemyPokemon().type2, shardAmount);
+        }
```
Ishadijcks: Hmm I think it's better to have something like `shardReward` as a property of Battle Pokémon, and move the amount gained to `GameConstants`

src/scripts/gym/GymBattle.ts
```diff
@@ -13,6 +13,10 @@ class GymBattle extends Battle {
     public static defeatPokemon() {
         player.gainMoney(this.enemyPokemon().money);
         player.gainExp(this.enemyPokemon().exp, this.enemyPokemon().level, false);
+        player.gainShards(this.enemyPokemon().type1, 5);
+        if (this.enemyPokemon().type2 != -1) {
+            player.gainShards(this.enemyPokemon().type2, 5);
+        }
```
Ishadijcks: Same as above.

Ishadijcks: Not sure what your opinion on this is, but maybe the `None` type check should be done in the player.gainShards method?

src/scripts/types/TypeHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+        [  1, 0.5, 0.5, 0.5,   1,   2,   1,   1,   1,   1,   1,   1,   2,   1,   1,   1, 0.5,   2], // STEEL
+        [  1, 0.5,   1,   1,   1,   1,   2, 0.5,   1,   1,   1,   1,   1,   1,   2,   2, 0.5,   1], // FAIRY
+    ];
+    //@formatter:on
+
+    public static getAttackModifier(a1: GameConstants.PokemonType, a2: GameConstants.PokemonType, d1: GameConstants.PokemonType, d2: GameConstants.PokemonType): number {
+
+        if (a1 == GameConstants.PokemonType.None || d1 == GameConstants.PokemonType.None) {
+            return 1;
+        }
```
Ishadijcks: This should never happen right?

Perhaps we need a developers warning for this. Something to let the user know to submit a bug report?

src/scripts/types/TypeHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+        let defendType1 = GameConstants.PokemonType[d1];
+        let defendType2 = GameConstants.PokemonType[d2];
+
+        let list = player.shardUpgrades;
+        //TODO factor in shard value
+        let m1 = TypeHelper.typeMatrix[attackType1][defendType1];
+        m1 += (list[attackType1][this.valueToType(m1)]() * GameConstants.SHARD_UPGRADE_STEP);
+
+        let m2 = 1, m3 = 1, m4 = 1;
+        if (d2 != -1) {
```
Ishadijcks: `PokemonType.None` can again be used here

src/scripts/types/TypeHelper.ts
Aegyo: The pokemon attack shown when fighting is calculated against a type of (None, None), this is the only time that d1 has no type. a1 should never be None

Aegyo:
It's probably better to just change the locked hex values to lighter colors

Ishadijcks:
Agreed

PR #89, opened 2017-07-14
Catch time depends on type of ball used

dennism1997:


Aegyo:
Actually it doesn't make sense to have any catch time at all when "None" is selected. The whole throwing a pokeball part is skipped anyway in Battle.defeatPokemon for that case, so it might not make much of a difference.

PR #91, opened 2017-07-15
Fix some errors when fighting the Elite 4

Aegyo:
Indigo Plateau uses town.gymList instead of town.gym, causing the game to crash when you run out of time against one of the elite 4 and it attempts to create a notification with player.town().gym().leaderName


Aegyo:
There were a couple of other places which had basically the same issue, so I've put the commits with this pr

PR #92, opened 2017-07-15
Feature/dungeon

Aegyo:
- randomIntBetween was returning a random int between min and (min + max - 1), now it returns a random int between min and max (previously the last dungeon boss would never be a possibility)
- Pokeball now appears next to caught pokemon in dungeons, and a pokeball next to the dungeon name when all have been caught
- When running out of time while fighting a dungeon pokemon, the fight would continue when starting a new dungeon. Fixed by setting fighting to false when dungeon is lost

I'm not entirely happy about where I've put the dungeonCompleted function. On one hand it makes sense because the logic is identical to routeCompleted, but RouteHelper maybe isn't the best place to have a function for dungeons.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -356,13 +356,13 @@
                             <div>
                                 <span data-bind="text: DungeonRunner.dungeon.name()">Dungeon name</span>
                                 <!--If all Pokémon on the route are caught-->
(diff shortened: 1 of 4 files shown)


dennism1997:
ListCompleted is really nice :)

Issue #60, opened 2017-07-10
Design the Pokéball selection interface


Ishadijcks:
Currently it looks like this:
![image](https://user-images.githubusercontent.com/9715314/28013249-b2739c86-6568-11e7-93b9-6365b3ddfcea.png)

The idea is that the player can select which Pokéball to use for new and Caught Pokémon.

It's really confusing to use with the numbers in the bottom, so I'm looking for something better.

Thoughts?

Aegyo:
How about something like this?
![Pokeball Selection](https://puu.sh/wJKkF/f1da8287a8.png)

Ishadijcks:
Ah yeah, although I would remove the infinity

PR #96, opened 2017-07-15
Pokeball selection interface

Aegyo:
As suggested in #60 
Pokeball selection

Column titles might not be descriptive enough, perhaps a tooltip somewhere would be best for explaining in more detail.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1196,37 +1196,6 @@
                                 </tr>
                             </tbody>
                         </table>
-
-                    <!--
-                        <span>Caught Pokémon</span>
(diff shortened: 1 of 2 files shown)


dennism1997:
Can u make the text center vertically?

dennism1997:
Hmm it still looks a bit cramped. Maybe it could use some padding? And change the "Amt" to a '#'? 
image
This is with:
`.pokeball-table td {
  padding: 0 0 4px;
  width: 33.33%;
  vertical-align: middle;
}`

Aegyo:
Good idea, that looks better.

PR #101, opened 2017-07-26
Feature/fill shops 2

Ishadijcks:
Add items to all the shops.

Currency images will be added on another branch


src/scripts/items/BattleItem.ts
```diff
@@ -4,9 +4,31 @@ class BattleItem extends Item {
...
+                break;
+            case GameConstants.BattleItemType.Token_collector:
+                basePrice = GameConstants.ItemPrice.Token_collector;
+                break;
+            case GameConstants.BattleItemType.Item_magnet:
+                basePrice = GameConstants.ItemPrice.Item_magnet;
+                break;
+            case GameConstants.BattleItemType.Lucky_incense:
+                basePrice = GameConstants.ItemPrice.Lucky_incense;
+                break;
```
Aegyo: Could this be simplified to `GameConstants.ItemPrice[ GameConstants.BattleItemType[type] ]` for every case?
Similarly for the other item types.

src/scripts/shop/Shop.ts
```diff
@@ -4,8 +4,14 @@
 class Shop {
     items: KnockoutObservableArray<Item>;
```
Aegyo: Why an observable array? As far as I know, the items in a shop will always be the same, so I don't see what benefit we get from making the array observable.

Ishadijcks: Ah I think it already was, but you're right

src/scripts/items/BattleItem.ts
Ishadijcks: I remember trying that and it not working, but I'll look into it again

Ishadijcks:
Run `Save.initializeItems` and `Save.initializeMultipliers()`

Aegyo:
Ah, that fixed the numbers. Battle item images are named like Xattack, but the item names in code are like xAttack

Ishadijcks:
Stupid windows and its case insensitiveness

PR #104, opened 2017-07-28
Feature/underground

Aegyo:
You will need the "Explorer kit" to access the underground. Entrance is a dungeon square to the right of Vermillion City.


src/scripts/Notifier.ts
```diff
@@ -22,7 +22,7 @@ class Notifier {
             },
             offset: 20,
             spacing: 10,
-            z_index: 1031,
+            z_index: 1050,
```
Ishadijcks: Why was this needed?

Ishadijcks:
A "You don't have any treasures yet" message would be nice

PR #122, opened 2017-08-27
More achievements

Aegyo:
Continued from tklaxbro's branch,

Adds achievements for:
- Clicks
- Captures
- Defeats
- Tokens
- Diamonds
- Pokemon Attack
- Dig Items
- Dig Deeper
- Gym Clears
- Dungeon Clears
- Hatch Eggs

I couldn't find how much bonus each achievement should have, so I just put all the ones I added to 0.01 bonus.


dennism1997:
Maybe a counter would be nice, which shows the total bonus you get from achievements

Aegyo:
Any suggestions for bonus amounts on dig items, dig deeper, hatch, gym clears, and dungeon clears?

PR #140, opened 2017-09-06
Add filters to pokedex

dennism1997:
For some reason, the filters were not merged into develop, so here they are


Ishadijcks:
When the list is longer than the screen, a scroll bar appears on the right. This moves the entire modal to the left a bit. This looks a bit odd. I think we should avoid the scrollbar and set a max height to the Pokédex, and scroll within the element. (like the caught list)

PR #142, opened 2017-09-07
Obtain mystery egg when the first pokemon reaches lvl 100

### LINKED ISSUE #137: Mystery egg should be awarded when a pokemon turns lvl 100

dennism1997:
Closes #137


Ishadijcks:
```maxLevelPokemonList()``` is used for the breeding interface right?

PR #151, opened 2017-10-02
Added code for town highlighting

BaineGames:
supports showing of locked town, unlocked town, and current town
supports currentown being highlighted
added new player method for getting current town

### DIFF
--- a/src/scripts/worldmap/MapHelper.ts
+++ b/src/scripts/worldmap/MapHelper.ts
@@ -5,6 +5,7 @@ class MapHelper {
             if (this.accessToRoute(route, region)) {
                 $("[data-route='" + player.route() + "']").removeClass('currentRoute').addClass('unlockedRoute');
                 player.route(route);
+                player.currentTown("");
                 $("[data-route='" + route + "']").removeClass('unlockedRoute').addClass('currentRoute');
                 Battle.generateNewEnemy();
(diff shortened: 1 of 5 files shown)


dennism1997:
Nice work! Maybe we want a slightly different color scheme for the towns, but we can change that later

src/scripts/worldmap/MapHelper.ts
```diff
@@ -59,9 +72,10 @@ class MapHelper {
     public static moveToTown(townName: string) {
         if (MapHelper.accessToTown(townName)) {
             Game.gameState(GameConstants.GameState.idle);
-            $("[data-route='" + player.route() + "']").removeClass('currentRoute').addClass('unlockedRoute');
+            $("[data-route='" + player.route() + "']").removeClass('currentRoute').addClass('unlockedRoute'); //pretty sure any jquery in typescript does not work fyi
```
dennism1997: Why do you think it does not work?

BaineGames: when you console log out the jquery selector inside of any location in typescript - nothing gets pulled. I think the css classes are coming from the function calculateRouteCssClass instead of here.

dennism1997: Yeah this line does nothing I think, you can remove it. But everywhere else jquery should work, just tested it out

PR #154, opened 2017-10-02
Enhancement/import html

BaineGames:
NEEDED NEW PACKAGE

`npm install gulp-html-import --save-dev`


Aegyo:
Pretty sure this is all done in #151 and #155

PR #158, opened 2017-10-02
Make dungeon ticket buyable from Pewter City shop

BaineGames:
removed dungeon ticket from brock win
moved dungeon ticket to buyable one time item in pewter city
added data-bind on col-sm-3 in shop index.html area to hide dungeon
ticket
causes a knockoutjs error in console once you buy the ticket and open another shop in the same
session. no idea why. @Ishadijcks  said @dennism1997  can fix that. 💯


BaineGames:
Also duplicated dungeon ticket png from key items to items to allow display on shop as src is hard coded to that directory.

src/scripts/shop/ShopHandler.ts
```diff
@@ -61,6 +66,16 @@ class ShopHandler {
         input.val(newVal > 1 ? newVal : 1).change();
     }
 
+    public static ownKeyItem(name: string): boolean {
+
+        console.log(name);
+        if(name =='Dungeon_ticket' && player.hasKeyItem('Dungeon ticket')){
```
dennism1997: Probably better to just check player.hasKeyItem and then replace the underscore with a space

src/scripts/shop/ShopHandler.ts
```diff
@@ -37,12 +37,17 @@ class ShopHandler {
...
                 Notifier.notify("You don't have enough money to buy " + this.amount() + " " + item.name() + multiple, GameConstants.NotificationOption.danger)
             }
-        } else if (item.currency == GameConstants.Currency.questpoint) {
+        } else if (item.currency == GameConstants.Currency.questpoint) { 
             if (player.hasQuestPoints(item.totalPrice())) {
                 player.payQuestPoints(item.totalPrice());
-                item.buy(this.amount());
-                item.increasePriceMultiplier(this.amount());
-                Notifier.notify("You bought " + this.amount() + " " + item.name() + multiple, GameConstants.NotificationOption.success)
+                if(item.name() =="Dungeon_ticket"){
```
dennism1997: I'd rather have this in the buy method of key item i think

src/scripts/shop/ShopHandler.ts
BaineGames: isha said no issue with a hard coded fix for just dungeon ticket. if we change it to be dynamic - it will return false for non key item items. so I opted to first check and verify then item in question is dungeon_ticket and if so - also check if they own it. Not a permanent deal - but needed to get progress since i currently do not know a way to check if an item is a key item or not.

src/scripts/shop/ShopHandler.ts
BaineGames: did not know you can do that. built the keyitem.ts as a needed modification to allow it to be purchased - was not aware i can use the buy method. can you show me how once the bug is fixed?

Ishadijcks:
Conflicts with index.html, please edit and I can review

src/scripts/items/buyKeyItem.ts
```diff
@@ -0,0 +1,22 @@
+class buyKeyItem extends Item {
```
dennism1997: Maybe we can merge this class with the current key item class

src/scripts/shop/ShopHandler.ts
```diff
@@ -61,6 +62,15 @@ class ShopHandler {
         input.val(newVal > 1 ? newVal : 1).change();
     }
 
+    public static ownKeyItem(name: string): boolean {
+        let keyItem = GameConstants.KeyItemType[name];
+        if( keyItem != undefined && player.hasKeyItem(name.replace("_", " ")) ) {
```
dennism1997: return this value instead of this if statement

src/scripts/items/buyKeyItem.ts
Aegyo: Wouldn't this force all key items to have a price, totalPrice etc, which is going to be useless for most of them?

Ishadijcks:
See #174

PR #160, opened 2017-10-03
Feature/quests

Aegyo:
Has the following quest types:
- DefeatPokemons
- CapturePokemons
- GainMoney
- HatchEggs
- MineLayers
- CatchShinies
- DefeatGym
- DefeatDungeon
- Use Pokeballs
- Gain Shards
- Gain Tokens
- Use Oak Item

"Quest level" (increased by gaining questXP from completing quests) currently does nothing other than provide a different set of quests, but it could be used to give some bonus amount to quest point rewards.

10 quests are decided based on date, player quest level, and number of times the player has "refreshed" the quest list. Refreshing lets players get a different set of 10 quests, but costs money (free when all quests are completed, just over 1 million when none are completed)


dennism1997:
Maybe it's possible to scale the quest difficulty with your progress (eg. the amount of pokemons you have, or the amount of badges). What do you think?

Ishadijcks:
Here is my list of improvements, feel free to disagree with any of these :)

- [ ] Possibility to cancel quests after they have been started.
- [ ] A big reward when all quests are completed. (just the logic behind it, rewards come later)
- [ ] Reset all quests when a new day arrives.
- [ ] Show a minute timer with how much time left for the next quest.
- [ ] I'm gonna give the interface an overhaul ;)
- [ ] Shiny amount is always 1, no need for it to be plural in the description.
- [ ] Use the quest point image instead of the description
- [ ] Add more quests!

Ishadijcks:
image
image

image
image
image

dennism1997:
Maybe switch the claim and questlist buttons around, so questlist is always on the left

PR #166, opened 2017-10-05
Gen 2 pokemon data update.

BaineGames:
Added sprites for gen 2 pokemon.
added all to pokemonlist
started filling in evolution stuff


Aegyo:
"Attack" should be the highest between attack and special attack, a quick glance through (well, just one that I checked so far - Xatu) seems like you have just gone with normal attack for everything.

PR #171, opened 2017-10-10
Other requirement options for dungeons, towns and routes

Aegyo:
Adds the option to restrict towns and routes by dungeons, and dungeons by badges.

Route 4 now requires Mt. Moon, route 20 requires Seafoam Islands, and Indigo Plateau requires Victory Road.


dennism1997:
This is genius, how have we not thought of this before haha

PR #169, opened 2017-10-05
gen 2 johto gym work - needs more info

BaineGames:
added sprite assets for gen2 gyms
added 8 new johto badges to gameconstants
added 8 new gyms to gmy.ts
-still needs hp settings
-still needs on win text

PR #177, opened 2017-10-12
Baine/gen 2/gyms

Aegyo:
Same as #169 but with corrections


src/scripts/GameConstants.ts
```diff
@@ -280,7 +280,19 @@ namespace GameConstants {
         "Bruno" = 10,
         "Agatha" = 11,
         "Lance" = 12,
-        "Champion" = 13
+        "Champion" = 13,
+        "Zephyr" = 14,
```
dennism1997: Wouldn't this be easier when we divide it by region? Like with routes?

Aegyo: Yeah that could deal with a couple of issues (Bruno being in Kanto and Johto, each region having a champion)

Aegyo: How would you keep it as an enum though? Maybe have a KantoBadges enum and a JohtoBadges enum, then something like
```
Badges = {
  0: KantoBadges,
  1: JohtoBadges
}
```

PR #180, opened 2017-10-13
Add button on breeding display to buy the next egg slot

### LINKED ISSUE #145: The player should be able to buy more eggslots using quest points.
In the locked eggslot, a buttons should appear.
When pressed, given that the player has enough quest points, it should give the player one more eggslot.

Aegyo:
Closes #145


dennism1997:
image
On my screen the egg slots are too big for the window

PR #181, opened 2017-10-14
New change log system using Markdown files.

BaineGames:
just make a new file in src/changelog.
the files should appear in asc order 0000 then 0001 then 0002 i think.
edited the gulp task for html import so the import will pull changelog.html into root index.html
and then markdown parse the smaller files to be inserted into the index.html section.

exsiting css will apply so we may want to normalize that modal window
with new css rules.


dennism1997:
Yeah the text should align to the left. For the rest I like it!

PR #185, opened 2017-10-15
Chat System V1

BaineGames:
Still uses CDN. I recommend leaving it that way and build an override stylesheet to include later.
Notes:

- Locked down to only one server domain
- Auto joins default Pokeclicker room on register
- Auto popup of Pokeclicker room on register and sign in while collapsing main bar
- Disables encryption
- Auto uses nickname from login when joining room
- Disables contact requests
- Disables bookmarks
- Disables ability to leave room (can close, but reopen on refresh or through rooms tab)
- Disables clear message command in chat room
- Disables user created chatrooms (Some override style applied to disable chatroom form as it is not a config option)


Ishadijcks:
Closed for later.

PR #187, opened 2017-10-16
Feature/working items

Ishadijcks:
Go to the item tab to use your evolution stones :)


Aegyo:
It would be nice if the pokemon that get pushed onto a second line were centered
overflowing pokemon

I think the cursor should be a pointer when over the stone images, and in the menu it might be better to call it "Items" instead of "Item".

PR #188, opened 2017-10-16
Gen 2

Aegyo:
Does not yet have choosing a new starter, and no popup when you can travel to Johto. Also pokemon health on routes, dungeons, gyms are not balanced in the slightest. Shops also haven't been set up.

After all Kanto pokemon have been caught, a button appears in Pallet town which will send you to Johto. To travel back, you first have to fight your way through Johto to Olivine, where there is a Dock for travel to other regions. Similarly, there is a dock in Kanto at Vermillion City, but the player won't be able to access it until they have been to Johto once.


src/components/shipModal.html
```diff
@@ -0,0 +1,27 @@
...
+            <div class="modal-body">
+                <img src='assets/images/ship.png'>
+                <div style='margin-top:20px'>
+                    <span data-bind='ifnot: player.region == 0'>
+                        <button class='btn btn-primary' onclick='MapHelper.moveToTown("Vermillion City");player.region = 0;$("#ShipModal").modal("hide")'>
+                            Kanto
+                        </button>
+                    </span>
+                    <span data-bind='ifnot: player.region == 1'>
+                        <button class='btn btn-primary' onclick='MapHelper.moveToTown("Olivine City");player.region = 1;$("#ShipModal").modal("hide")'>
```
dennism1997: Maybe it's better to move what's inside the onclick to methods in the maphelper class

Aegyo: I had it like this to plan for when gen 3 gets added, and we have two buttons.

src/components/shipModal.html
```diff
@@ -0,0 +1,27 @@
+<div class="modal noselect fade" id="ShipModal" tabindex="-1" role="dialog" aria-labelledby="ShipModalLabel">
```
dennism1997: What is this file?

src/scripts/GameConstants.ts
```diff
@@ -321,7 +325,20 @@ namespace GameConstants {
...
+        "Fog" = 17,
+        "Storm" = 18,
+        "Mineral" = 19,
+        "Glacier" = 20,
+        "Rising" = 21,
+        "Will" = 22,
+        "Koga" = 23,
+        "Bruno2" = 24,
+        "Karen" = 25,
+        "JohtoChampion" = 26,
```
dennism1997: Isn't Karen the Johto Champion?

src/components/shipModal.html
Aegyo: Modal that opens when you click one of the docks at Olivine or Vermillion. Accessed through the dungeon square next to those cities.

src/scripts/GameConstants.ts
Aegyo: Lance is the Johto Champion

src/scripts/Player.ts
```diff
@@ -851,6 +864,18 @@ class Player {
         this._questPoints(value);
     }
 
+    public ableToTravel() {
```
dennism1997: These two methods may have a better place in the maphelper. At least that's where I would look for these methods

Aegyo: Yeah, I wasn't 100% sure where they should go.

src/scripts/worldmap/MapHelper.ts
```diff
@@ -5,6 +5,7 @@ class MapHelper {
             if (this.accessToRoute(route, region)) {
                 $("[data-route='" + player.route() + "']").removeClass('currentRoute').addClass('unlockedRoute');
                 player.route(route);
+                player.region = region;
```
dennism1997: Isn't player.region an observable now? So shouldn't it be `player.region(region)`

Aegyo: player._region is observable, but we have get and set methods for player.region so this works fine

src/components/shipModal.html
dennism1997: Oh huh I put it at the wrong location. I meant to ask about the test.html file

src/scripts/GameConstants.ts
dennism1997: Oh yeah

src/scripts/worldmap/MapHelper.ts
dennism1997: Cool

src/components/shipModal.html
Aegyo: oh, whoops, that was just so I could see the diff between the map in develop and my KantoSVG. It shouldn't have been in the pr

BaineGames:
and so it begins

dennism1997:
The map feels kinda cramped. Don't really know what to do about it tho. Maybe some space around the edges but not sure if that's possible

Issue #204, opened 2018-01-15
Some issues


Ishadijcks:
Update later.

Tutorial
- [x] Hide Pokéballs during the tutorial.
- [x] Hide Oak items during the tutorial.
- [x] Hide Quests during the tutorial.
- [ ] Explain how to go to another route.
- [ ] Introduce how Pokéballs work.
- [ ] Introduce how Oak items work.
- [x] Explain how to go to another town.

PokemonList
- [x] Sorting input should start at "Sort By..."

Towns
- [x] Gymbutton should have some more padding.
- [x] Shops should have a back to city button.
- [ ] Towns should always have the same height as the main interface.

Shops
- [x] Reset amount input to 1 after buying.
- [x] The buy button for dungeon tickets depends on if you have 1000 money, not 1000 QP.

Dungeons
- [ ] Dungeons should have the same height as towns.

PokéDex
- [x] Only show Pokémon up to the highest ID encountered (defeated).
- [x] Add logic so pokedex refreshes when it is accessed. (NEW FROM NOAH)

Achievements
- [x] Add tooltip explanation of total bonus.
- [x] Add names to all achievements.
- [x] Dungeon tokens are broken.

Donate
- [x] Rewrite the Donate modal, it sounds douchy.
- [x] Remove exclamation mark.

Items
- [ ] Evo stones, block this tab until the first stone is unlocked.
- [x] Key items should be the first active tab.

Save
- [x] Style import save button.
- [x] Give the converter text a bit more padding.
- [x] Make the height consistent.

Stats
- [x] Remove it for now.

Shards
- [x] Remove title

Changelog
- [ ] Center title

- [ ] Key items, center description vertically

dennism1997:
>  Only show Pokémon up to the highest ID caught.

Why?

Ishadijcks:
That's the way it works in the game.

Aegyo:
I think the games will show highest id encountered, no need to be caught.

We could probably just use highest defeated, since I don't think we track encounter numbers

Ishadijcks:
Ah yeah you're right.

Ishadijcks:
Unused issue

PR #205, opened 2018-01-16
removed stats from gamemenu

BaineGames:
per issue #204 - removed stats from gamemenu - commented out for now rather than full deletion. tested locally and does not appear to cause any problems.


Ishadijcks:
"Works on my machine" :)

BaineGames:
lol hardy har har :P

PR #206, opened 2018-01-16
removed shards title - leaving h4 to preserve help location

BaineGames:
per issue #204 - removed shards title. if we remove the h4 element - help img moves to the left - not sure if we want it that way or not - and easiest path is to empty the h4 as opposed to remove and fix help css.


Ishadijcks:
Yeah the tooltip should stay to the right.

PR #207, opened 2018-01-16
updated donation text to something else - removed exclamation mark

BaineGames:
per #204 - request to remove ! and rewrite the modal. Did my own first pass. Feel free to review and update via change if need be if the text is not suitable.


Ishadijcks:
Much less douchy, thanks mate :)

PR #210, opened 2018-01-16
Key items is now active tab by default and ordered first

BaineGames:
per #204 - key items should be the default first active tab.


Ishadijcks:
Duplicate of #209

PR #215, opened 2018-01-17
Add Gulp build to travis

dennism1997:
Everything should work the samen, except the gulpfile.js is now called gulpfile.babel.js. 
I also cleaned up the gulpfile and made all anonymous functions into arrow functions, for consistency


Ishadijcks:
Removed tsc so it doesn't run twice?

dennism1997:
Exactly

dennism1997:
Builds are around 1:20 (20 seconds slower) now, so not a huge impact

PR #217, opened 2018-01-18
Change equality signs to triple ones

dennism1997:


Ishadijcks:
That's surprisingly few

PR #223, opened 2018-02-13
Start gym balancing, gen1 pokemon are weaker in Johto

dennism1997:
probably still too weak, but it's a start


src/scripts/GameHelper.ts
```diff
@@ -65,4 +65,8 @@ class GameHelper {
         }
     }
 
+    public static isGen1(id) {
+        return id <= GameConstants.AMOUNT_OF_POKEMONS_GEN1;
+    }
+
```
Ishadijcks: Change to isGen(), that returns the generation the id belongs to.

src/scripts/Player.ts
```diff
@@ -704,12 +712,16 @@ class Player {
         // TODO Calculate pokemon attack by checking upgrades and multipliers.
         let attack = 0;
         for (let pokemon of this.caughtPokemonList) {
+            let multiplier = 1;
+            if (this.region > 0 && GameHelper.isGen1(pokemon.id)) {
```
Ishadijcks: And here `if this.region != isGen(id)`

dennism1997:
@Ishadijcks

PR #232, opened 2018-02-16
Questlines

Aegyo:
I've tested this a little and it seems to work, but I wouldn't be surprised if I've missed something and it breaks spectacularly.

The tutorial quests might need changing a little, particularly when you have a quest to go kill stuff on route 2, but you can't access it yet.

### DIFF
--- a/src/components/questDisplay.html
+++ b/src/components/questDisplay.html
@@ -40,12 +40,12 @@
             </div>
         </div>
     </div>
-    <div data-bind='if: (player.tutorialProgress() < 6) && (QuestLineHelper.tutorial.totalQuests > 0) && (QuestLineHelper.tutorial.curQuest() < 6)'>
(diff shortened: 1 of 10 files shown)


dennism1997:
Looking good! I think there should also be a quest that tells you to defeat 10 pokemon on Route 1

src/components/questDisplay.html
```diff
@@ -1,11 +1,11 @@
 <div class="page-item" style="background-color: white; min-height: 100px">
     <br>
     <button style="vertical-align: middle;" class='btn btn-primary'
-            data-bind='visible: !player.currentQuest(),
+            data-bind='visible: (player.tutorialProgress() == 6) && !player.currentQuest(),
```
dennism1997: Is 6 the final tutorial quest? A constant is better then, or maybe a function that counts them

Ishadijcks:
Small bug where the progressbar of the tutorial shows 0/1 instead of the real progress.
If you can fix this and the magic number I aprove, it looks good :)

Ishadijcks:
Great

PR #236, opened 2018-02-20
Fix/moneyrounding

### LINKED ISSUE #234: Money is not always rounded
It is rounded when displayed, but the money the player has should always be an integer. Not sure why this is not the case now.

Ishadijcks:
Fixes #234


Ishadijcks:
Ah, must have branched wrong.

PR #251, opened 2018-06-04
hide panels during tutorial

BaineGames:
hides all panels during the tutorial and re-enables them when the tutorial is completed.


Ishadijcks:
Maybe don't hide the questpanel as that explains what you have to do

PR #257, opened 2018-06-05
Refactor quest things

Aegyo:
Let me know if you notice anything else which could be done better in the quest code

### DIFF
--- a/src/scripts/Player.ts
+++ b/src/scripts/Player.ts
@@ -944,6 +944,12 @@ class Player {
         this._questPoints(value);
     }
 
+    public gainQuestPoints(value: number) {
+        player.questPoints += value;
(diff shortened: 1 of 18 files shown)


src/scripts/Player.ts
```diff
@@ -944,6 +944,12 @@ class Player {
         this._questPoints(value);
     }
 
+    public gainQuestPoints(value: number) {
+        player.questPoints += value;
+        GameHelper.incrementObservable(this.statistics.totalQuestPoints, value);
+        Game.animateMoney(value,'playerMoneyQuest');
+    }
+
```
Ishadijcks: Did this method not exist yet?

Aegyo: I expected it to, but couldn't find it

Ishadijcks: Weird

PR #258, opened 2018-06-05
Fix underground

Aegyo:
I don't know how this broke, but it is fixed now.
Also, you actually get evolution stones now instead of it lying to you.


src/scripts/underground/Underground.ts
```diff
@@ -62,7 +62,13 @@ class Underground {
         let item = Underground.getMineItemById(id);
         
         if(item.isStone()){
-            //gainItemByName(item.name);
```
Ishadijcks: Kinda surprised this didn't work tbh

PR #260, opened 2018-06-09
Fix/pokeball price

### LINKED ISSUE #249: Money for multi-buy is not correct for pokeballs (might be intentional)
I realized that this might be an accident so I gave a warning in the title. 
the final lines for the price of an item is as follows: 
```js
if (this.name() == "Pokeball") {
    return this.price() * amount;
}
let res = (this.price() * (1 - Math.pow(GameConstants.ITEM_PRICE_MULTIPLIER, amount))) / (1 - GameConstants.ITEM_PRICE_MULTIPLIER);
return Math.floor(res);
```

I don't know why you'd want to have the pokeballs not follow this patern. 
Sidenote: They still increase in price after you buy, just the math isn't correct for bulk buys. -> 
1st pokeball is 100, second is 101. But 10 cost 1000 flat.

dennism1997:
Closes #249

### DIFF
--- a/src/scripts/items/Item.ts
+++ b/src/scripts/items/Item.ts
@@ -12,19 +12,24 @@ abstract class Item {
         this.basePrice = basePrice;
         this.currency = currency;
         this.price = ko.observable(this.basePrice);
-        this.totalPrice = ko.computed(function () {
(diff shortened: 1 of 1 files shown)


Ishadijcks:
Clever ;)

PR #264, opened 2018-06-25
Make levelType work

tklaxbro:
I saw that levelType wasn't working so that is what the first commit is for. 

Once I got everything working VSC was screaming at me that it could not compare LevelType and a string. I figured that if I parseInt() the type it would fix the issue.

There is this weird behavior where GameConstants.LevelType[] is returning a string. Even weirder GameConstants.LevelType[GameConstants.LevelType[]] works perfectly fine and returns a LevelType.

### DIFF
--- a/src/scripts/pokemons/PokemonHelper.ts
+++ b/src/scripts/pokemons/PokemonHelper.ts
@@ -30,8 +30,8 @@ class PokemonHelper {
 
     public static calculateLevel(pokemon: CaughtPokemon): number {
         let level;
-        switch (PokemonHelper.getPokemonByName(pokemon.name).levelType) {
-
+        let type = GameConstants.LevelType[PokemonHelper.getPokemonByName(pokemon.name).levelType];
+        switch (type) {
             case GameConstants.LevelType.slow:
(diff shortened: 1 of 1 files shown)


src/scripts/pokemons/PokemonHelper.ts
```diff
@@ -30,16 +30,16 @@ class PokemonHelper {
 
     public static calculateLevel(pokemon: CaughtPokemon): number {
         let level;
-        switch (PokemonHelper.getPokemonByName(pokemon.name).levelType) {
-
+        let type = parseInt(GameConstants.LevelType[PokemonHelper.getPokemonByName(pokemon.name).levelType]);
```
Ishadijcks: This should not be needed. What was causing it again?

tklaxbro:
Because GameConstants.leveltype is an enum and the switch was getting the string of the Pokémon type and not the int. We had to get the leveltype of the string to change the string to the int.

tklaxbro:
Now that I look back I have no clue why I made a variable, it could of been done in the switch statement

Issue #254, opened 2018-06-04
Dungeons don't award items yet.


Ishadijcks:
- [ ] Get a notification when an item is gained from a chest.
- [ ] Award the following items for dungeon number `x` .
```

	1: ["X Attack", "X Click", "X Attack", "X Click", "X Attack", "X Click", "Lucky Incense"],
	2: ["X Attack", "X Click", "X Attack", "X Click", "X Attack", "X Attack", "Lucky Incense"],
	3: ["X Attack", "X Click", "X Attack", "X Click", "X Attack", "X Attack", "Token Collector"],
	4: ["X Attack", "X Click", "X Attack", "X Click", "X Attack", "Item Magnet"],
	5: ["X Attack", "X Click", "X Attack", "X Click", "X Click", "Lucky Incense"],
	6: ["X Attack", "X Click", "X Attack", "X Click", "Item Magnet"],
	7: ["X Attack", "X Click", "X Attack", "X Click", "X Exp"],
	8: ["X Attack", "X Click", "X Attack", "X Click", "Token Collector"],
	9: ["X Attack", "X Click", "X Attack", "X Click", "Item Magnet"],
	10: ["X Attack", "X Click", "X Click", "Lucky Incense"],
	11: ["X Attack", "X Click", "X Attack", "Item Magnet"],
	12: ["X Attack", "X Click", "X Click", "Item Magnet"],
	13: ["X Attack", "X Click", "X Attack", "X Exp"],
	14: ["X Attack", "X Click", "X Click", "Token Collector"],
	15: ["X Attack", "X Click", "Lucky Incense"],
	16: ["X Attack", "X Click", "Item Magnet"],
	17: ["X Attack", "X Click", "X Exp"],
	18: ["X Attack", "X Click", "Token Collector"],
	19: ["X Attack", "Lucky Incense"],
	20: ["X Click", "Item Magnet"],
	21: ["X Attack", "X Exp"],
	22: ["X Click", "Token Collector"],
	23: ["Lucky Incense", "Item Magnet", "Token Collector"],
	24: ["X Exp", "Lucky Incense", "Token Collector"],
	25: ["X Exp", "Item Magnet"],
}
```

BaineGames:
closing as it was merged

PR #267, opened 2018-08-04
Make dungeons reward items

tklaxbro:
#254 

Awards one item to the player using player.gainItem()
Then notifies the user that they got 1 (item name)
Follows the chart given in #254 (even though there are only 21 dungeons)

### DIFF
--- a/src/scripts/dungeons/Dungeon.ts
+++ b/src/scripts/dungeons/Dungeon.ts
@@ -7,7 +7,7 @@ class Dungeon {
 
     name: KnockoutObservable<string> = ko.observable("");
     pokemonList: string[];
-    itemList: string[]
+    itemList: GameConstants.BattleItemType[]
     baseHealth: number;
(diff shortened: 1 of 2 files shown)


Ishadijcks:
I like the solution, can you change the `itemList` from a `string[]` to a `BattleItemType[]` and use BattleItemType.xAttack as the itemnames. To prevent typos :)

If you do that I can approve it

src/scripts/dungeons/Dungeon.ts
```diff
@@ -7,6 +7,7 @@ class Dungeon {
 
     name: KnockoutObservable<string> = ko.observable("");
     pokemonList: string[];
+    itemList: string[]
```
BaineGames: @tklaxbro put a semicolon here to make me happy.

Ishadijcks: Why does this still show up as a string[]?

PR #272, opened 2018-08-06
Feature/display items

BaineGames:
alrighty, heres a bit of big one so ill do my best to walk you through how i did this. I built an effectEngine that can be used for a number of things, keeping it generalized so we are not just locked into doing something only for battle items.

first off - we have the effectEngineRunner class. This class is defined to keep track of an effect name and a counter for how many seconds are remaining for the effect. Once the effect counter has hit 0, it will be destroyed from the list. The list is attached to player.effectEngine - similar in the way equipped oak items are stored. this class has a standard tick() method and addEffect(string) method.

The tick() method is used in Main.ts where the counter is incremented. and if the counter exceeds the gameconstant for the effect runner - it will run the tick function. currently this is set to run about every second.

the addEffect(string) method is used currently in battle items use() method.

I also added a small meter to the existing battleitemscontainer that @Ishadijcks started. Inside the effectEngine class it updates those meters with a small green progress bar to illustrate how much time is left. If you activate more than one of the same item, the ticker updates accordingly and begins to decay. You can activate an item manually by using ItemList[itemname].use()

added 2 game constants, one for EFFECT_ENGINE_TICK = 100 (meaning 1 second) and ITEM_USE_TIME = 30 meaning each effect lasts for 30 seconds. Isha said do the same thing but 50%. So i did 30 seconds instead of 60 seconds and 1.5 multipliers (keep reading) instead of flat out doubles.

Next was implementing the actual effects of the effectEngine. You do this by just checking if(player.effectEngine['effectName']) exists.

so for the xAttack - we check if player.effectEngine['xAttack'] is true, and if so - set attack to itself times 1.5 - otherwise keep existing attack calculation.

Same goes for moneytogain for the Lucky_incense.

I did not implement the other items yet as not sure where the logic is to do so quite yet. Also wanted to get some feedback on this before I continued on it. so @Ishadijcks and whoever else wants to take a look, go for it.


src/scripts/Player.ts
```diff
@@ -440,9 +442,13 @@ class Player {
         // TODO add money multipliers
         let oakItemBonus = OakItemRunner.isActive("Amulet Coin") ? (1 + OakItemRunner.calculateBonus("Amulet Coin") / 100) : 1;
         let moneytogain = Math.floor(money * oakItemBonus * (1 + AchievementHandler.achievementBonus()))
+        if(this.effectEngine['Lucky_incense']){
+            moneytogain = Math.floor(moneytogain * 1.5);
```
Ishadijcks: My memory might not be correct but isn't there a battleItem created somewhere with this multiplier?

src/scripts/Player.ts
```diff
@@ -785,7 +799,13 @@ class Player {
             }
         }
 
-        return Math.round(attack);
+        //add item effect bonus here
```
Ishadijcks: Done

src/scripts/effectEngine/effectEngineRunner.ts
```diff
@@ -0,0 +1,25 @@
...
+                $('#'+item+'-meter').css('width',newWidth + "%");
+            }
+            if(player.effectEngine[item] <= 0){
+                delete player.effectEngine[item];
+                Notifier.notify("The "+item+" has worn off!", GameConstants.NotificationOption.warning);
+            }
+        }
+    }
+
+    public static addEffect(itemName: string){
```
Ishadijcks: Change this to the enumerator for BattleItems to prevent typo's

src/scripts/effectEngine/effectEngineRunner.ts
```diff
@@ -0,0 +1,25 @@
+class effectEngineRunner {
+    public static counter: number = 0;
+
+    public static tick() {
+        this.counter = 0;
+        
+        for(let item in player.effectEngine){
+            if(item != 'equals'){
```
Ishadijcks: What is this 'equals'?

Ishadijcks:
@BaineGames Have you checked my review comments?

src/scripts/effectEngine/effectEngineRunner.ts
BaineGames: without it - an extra equals: NaN object is added. no clue why or where - but its there lol. this check removes that item. Could be something related to knockout but no clue.

src/scripts/effectEngine/effectEngineRunner.ts
BaineGames: Done

src/scripts/Player.ts
BaineGames: not sure what you mean, the only enums i see for incense is the price and new declared string i built

src/components/battleItemContainer.html
```diff
@@ -0,0 +1,31 @@
+<div class="page-item" id="battleItemContainer">
+    <div class="card">
+        <div class="row justify-content-sm-center">
+            <div class="battle-item-small">
+                <img src="assets/images/items/xAttack.png">
```
RedSparr0w:
```suggestion
                <img src="assets/images/items/xAttack.png" class="clickable" onclick="ItemList.xAttack.use()">
```

src/scripts/effectEngine/effectEngineRunner.ts
RedSparr0w: its because of [`Array.prototype.equals` here](https://github.com/Ishadijcks/pokeclicker/blob/develop/src/scripts/safari/SafariBody.ts#L350)

RedSparr0w: I've applied a fix here https://github.com/Ishadijcks/pokeclicker/pull/285/commits/50214c7b95076b4cff03c48210670f37ff40575c
By changing the object from `[]` → `{}`,
But this does mean if you already have the `effectEngine` object attached to your save, you will need to apply `player.effectEngine = {}` manually

RedSparr0w:
@Ishadijcks Are there any changes needed so we can get this merged?

Ishadijcks:
I think I have neglected this pr as it has some conflicts. I don't think I have ever ran it, so I'll have to do that first.

Ishadijcks:
If you could fix the conflicts that'd be great

RedSparr0w:
I can't fix the conflicts because it's not my PR and i'm not a maintainer.
There should be a merge conflict UI that you can use before merging though?

src/scripts/items/ItemHandler.ts
```diff
@@ -26,12 +26,14 @@ class ItemHandler {
...
-            return Notifier.notify("No Pokémon selected", GameConstants.NotificationOption.danger);
+            Notifier.notify("No Pokémon selected", GameConstants.NotificationOption.danger);
+            return;
         }
         let amountTotal = Math.min(this.amountSelected(), player.itemList[this.stoneSelected()]());
 
-        if(!amount){
-            return Notifier.notify(`You don't have any ${this.stoneSelected().replace(/_/g, ' ')}s left...`, GameConstants.NotificationOption.danger);
+        if(amountTotal == 0){
+            Notifier.notify("You don't have any stones left...", GameConstants.NotificationOption.danger);
```
RedSparr0w: Just checking if this change was intentional:
```diff
-`You don't have any ${this.stoneSelected().replace(/_/g, ' ')}s left...`
+'You don't have any stones left...'

RedSparr0w:
Is this ready for merge now?
Any other changes needed?
Would be good to get this merged before any other conflicts appear.

Ishadijcks:
Yeah I've been holding off on this one as I'm kind of conflicted (ha) on the current presentation of the time left. It's not clear yet, but we could fix that later of course

RedSparr0w:
I think it would be better to just merge as is, and then update the look of it, as it has quite a lot if changes, it is very susceptible to merge conflicts.

RedSparr0w:
Merged in #382

PR #273, opened 2018-08-18
Translations

tklaxbro:
This creates a backbone for translations using the controversial Language class. This allows for two types of values, functions and normal returns. Ex. oakItemRequirement requires an argument for the amount of captured pokemon so it is a function. 

A bunch of more translations needed to be added and HTML needs to be added. I am just making this PR to see if anyone has a better method of doing it (Aegyo was talking about making it an object and not a class so I'd like your opinion on that)

(you don't see that 1st commit :eyes:)


Ishadijcks:
I'm not a fan of the translations as it slows down development by requiring devs to implement strings in all languages. It would make more sense to add translations once the game is finished.

This is also a new feature, and we're trying to release as soon as possible by cutting features.

I do however believe that this pr functions as a nice Proof of Concept if we want to implement this in the future. A system like this can be completely separated from the main game and would be a great tool to have implemented generically for future projects. I will close this pr for now as I have no plans for merging it any time soon. I hope you'll understand.

PR #275, opened 2018-09-02
added basic hover effects on map. 

BaineGames:
does not work for specialty locations like day care. these do not follow the same css class patterns as the other items. its a start.


Ishadijcks:
Some incorrect behavior with this implementation.

It stops the hovering color when hovering over the number. This makes it flicker weirdly.
Unfortunately the SVG can not be changed. Needs a workaround.
https://stackoverflow.com/questions/6725288/svg-text-inside-rect

Also needs to work for special locations before merging.

PR #279, opened 2019-03-22
Evo stone fix

fuws:
The evolution stone menu had three issues:

1. Major issue: it's showing all base form Pokemon that can evolve using said stone, including those that the player doesn't have.
2. Major issue: Eevee evolutions were not coded correctly.
3. It doesn't have any indicator for if the player already has the evolved form (caught or shiny).

What this change looks like:

I do not have Exeggcute, so it doesn't show up for Grass Stone:

image

Correctly shows which evolved form I have and not have, and whether I have them shiny. Eevee evolutions also appear correctly now.

image

If you look closely, you'll see that the regular pokeball is a bit farther from the right than the shiny pokeball. This is something I was not successful at addressing. I feel it's a minor issue that can be put into backlog.

### DIFF
--- a/src/scripts/items/EvolutionStone.ts
+++ b/src/scripts/items/EvolutionStone.ts
@@ -22,31 +22,10 @@ class EvolutionStone extends Item {
     }
 
     public static computeEvolution(type: GameConstants.StoneType, pokemon: string): string {
-        if (pokemon == "Eevee") {
-            switch (type) {
-                case GameConstants.StoneType.Fire_stone: {
-                    return "Flareon";
-                }
-                case GameConstants.StoneType.Water_stone: {
-                    return "Vaporeon";
-                }
-                case GameConstants.StoneType.Thunder_stone: {
-                    return "Jolteon";
-                }
-            }
-        } else if (pokemon == "Gloom") {
-            switch (type) {
-                case GameConstants.StoneType.Leaf_stone: {
-                    return "Vileplume";
-                }
-                case GameConstants.StoneType.Sun_stone: {
-                    return "Bellossom";
-                }
-            }
-        }
-        else {
-            return PokemonHelper.getPokemonByName(pokemon).evolution;
-        }
+        // Assume stones and evolutions in pokemonList are consistent in ordering
+        let pkmObj = PokemonHelper.getPokemonByName(pokemon);
+        let index = ("" + pkmObj.evoLevel).split(", ").indexOf(GameConstants.StoneType[type]);
+        return pkmObj.evolution.split(", ")[index];
     }
 }
 
--- a/src/scripts/items/EvolutionStone.ts
+++ b/src/scripts/items/EvolutionStone.ts
@@ -14,33 +14,31 @@ class EvolutionStone extends Item {
         player.gainItem(GameConstants.StoneType[this.type], n)
     }
 
-
-
     public use(pokemon?:string) {
         let shiny = PokemonFactory.generateShiny(GameConstants.SHINY_CHANCE_STONE);
+        let evolution = EvolutionStone.computeEvolution(this.type, pokemon);
+        player.capturePokemon(evolution, shiny, false);
+        return shiny;
+    }
+
+    public static computeEvolution(type: GameConstants.StoneType, pokemon: string): string {
         if(pokemon == "Eevee") {
-            switch (this.type) {
+            switch (type) {
                 case GameConstants.StoneType.Fire_stone: {
-                    player.capturePokemon("Flareon", shiny, false);
-                    break;
+                    return "Flareon";
                 }
                 case GameConstants.StoneType.Water_stone: {
-                    player.capturePokemon("Vaporeon", shiny, false);
-                    break;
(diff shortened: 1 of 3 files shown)


Ishadijcks:
Looks great!

Do you know what is causing the pokeball difference? Might be an issue in the sprite?

fuws:
I currently use two different `<span>`s with `data-bind` conditional on caught status. Maybe the span order is causing this. I tried to put `data-bind` at the `` level and it didn't work. I can look into that more.

fuws:
I fixed the pokeball appearance difference. I just moved the padding up one level.

New look:
image

src/scripts/items/EvolutionStone.ts
```diff
@@ -14,33 +14,40 @@ class EvolutionStone extends Item {
...
+                }
+                case GameConstants.StoneType.Sun_stone: {
+                    return "Bellossom";
                 }
             }
         }
         else {
-            let evolution: string = PokemonHelper.getPokemonByName(pokemon).evolution;
-            player.capturePokemon(evolution, shiny, false);
+            return PokemonHelper.getPokemonByName(pokemon).evolution;
```
Ishadijcks: Maybe build a check in case this doesn't return anything (i.e. we forgot to hardcode a special evolution).

src/scripts/items/EvolutionStone.ts
```diff
@@ -14,33 +14,40 @@ class EvolutionStone extends Item {
...
     }
 
-
-
     public use(pokemon?:string) {
         let shiny = PokemonFactory.generateShiny(GameConstants.SHINY_CHANCE_STONE);
-        if(pokemon == "Eevee") {
-            switch (this.type) {
+        let evolution = EvolutionStone.computeEvolution(this.type, pokemon);
+        player.capturePokemon(evolution, shiny, false);
```
Ishadijcks: Or build the check here, see other comment

src/scripts/items/EvolutionStone.ts
```diff
@@ -14,33 +14,40 @@ class EvolutionStone extends Item {
...
+        return shiny;
```
Ishadijcks: Why return the boolean `shiny`?

fuws: It returns `shiny` because `ItemHandler.useStones` needs it to determine when to stop using the stones if player selects "Until Shiny". But the implementation of `ItemHandler.useStones` is a bit flawed, which I can fix in a separate PR.

src/scripts/items/EvolutionStone.ts
fuws: I don't think we ever run into that problem, because if a stone is coded incorrectly, then `PokemonHelper.getPokemonsWithEvolution` would see that a pokemon doesn't evolve with this stone, and wouldn't show this pokemon on the UI. So the player has no way of selecting it. 

Another way of saying this is, players are only able to select the pokemons that we have coded correctly. So an incorrect coding won't result in page rendering failure; it'll result in the base form pokemon not showing.

fuws:
I updated `computeEvolution` so we don't have to hard code the Eevee and Gloom evolutions in code here. It assumes the ordering of stones and evolved forms are consistent (which they are). I checked that the page renders correctly.

src/scripts/pokemons/PokemonList.ts
```diff
@@ -1795,7 +1795,7 @@ const pokemonList = [
         "name": "Eevee",
         "catchRate": 45,
         "evolution": "Vaporeon, Jolteon, Flareon, Espeon, Umbreon",
-        "evoLevel": "Water Stone, Thunder Stone, Fire Stone, Time Stone, Time Stone",
+        "evoLevel": "Water_stone, Thunder_stone, Fire_stone, Time Stone, Time Stone",
```
RedSparr0w: For consistency:
```suggestion
        "evoLevel": "Water_stone, Thunder_stone, Fire_stone, Time_Stone, Time_Stone",
```

RedSparr0w: Also on a side note, Would it be better to store these as an Array:
```javascript
"evoLevel": ["Water_stone", "Thunder_stone", "Fire_stone", "Time_Stone", "Time_Stone"],
```

Or even something like this:
```javascript
"evoLevel": {
  "Water_stone": ["Vaporeon"],
  "Thunder_stone": ["Jolteon"],
  "Fire_stone": ["Flareon"],
  "Time_Stone": ["Espeon", "Umbreon"],
},
```

src/scripts/items/EvolutionStone.ts
Ishadijcks: Good point

RedSparr0w:
> I agree with @RedSparr0w, it would be cleaner to store the 'evoLevel' in a dictionary so there is no need for string parsing. Other than that I believe this is ready to merge!

Probably fine to merge as is, then sort that out in a later PR?

Ishadijcks:
Fair enough, could you create an issue for it?

Issue #74, opened 2017-07-11
The map jumps up and down when switching between routes, towns and dungeons


Ishadijcks:
Not all towns, dungeons and routes have the same height, so the map keeps moving.
This looks very ugly, and all of them should have the same height.

PR #280, opened 2019-03-27
Fix main display height jumping issue

fuws:
The idea is simple: have the currencies bar and attack statistic bar appear at all times. Then, it's a simple matter of regulating the middle chunk to have a fixed height and have the images scale to it.

To this end, I made the following adjustments:
- Reorganized code so all the different views occur between the Currency and Attack bars.
- Made the views a little taller (at 240px) because I have to make room for 6 stacking buttons in the Pokemon League.
- Have all town images have the same height (240px).
- Moved dungeon Pokemon list to the side, so the dungeon image can take the full height. I had to make the image panel a tiny bit narrower to make the Pokemon list render well.

**In town:**
image

**Dungeon (Pokemon list doesn't need scrolling)**
image

**Dungeon (Pokemon list needs scrolling)**
image

One small issue I wasn't able to resolve, is that when fighting gyms, the display grows by a couple of pixels. I think it has to do with the `position:relative` style put into `gymView.html`. Maybe another dev may have ideas on fixing that.

#74

### DIFF
--- a/src/components/gymView.html
+++ b/src/components/gymView.html
@@ -1,6 +1,6 @@
 <div id="gymView" data-bind="if: Game.gameState() === GameConstants.GameState.gym">
 
-    <div style="position:relative; height: 220px;">
+    <div style="position:relative">
         <div id="gymCountdownView">
                 <div class="row" style="display: inline-block; vertical-align: middle;">
                     <div class="col-sm-6 offset-sm-3">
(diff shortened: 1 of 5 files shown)


Ishadijcks:
This is great, you have no idea how much this bugged me to the point I couldn't stand to fix it.

Not a huge fan of the pokemons on the side but I don't have a better idea for now...

fuws:
I thought leaving the dungeon Pokemon list at the bottom just makes the image height hard to control. So it was either on the side, or add an `Info` button that shows the Pokemon list upon clicking. I prefer to have the user do fewer clicks to get information. So on the side it is. 

Maybe something better can be done down the road. It's quite easy to move this block of code around.

Aegyo:
Small issue in Celadon shop: The "buy" bar gets pushed into the attack box
Celadon Shop

fuws:
I grouped shop items tighter together in large screens, and further apart in smaller screens, while introducing a scroll bar. Because of fixed overall height, we pretty much have to use scroll bars for things that don't fit. Hopefully we only need to use it for shops and dungeon Pokemon list.

image

image

PR #298, opened 2019-05-26
Add a button to report a bug

Ishadijcks:
image

Links to [here](https://gitreports.com/issue/Ishadijcks/pokeclicker)

I've also moved the non-game related buttons to the bottom of the menu.

### DIFF
--- a/src/components/gameMenu.html
+++ b/src/components/gameMenu.html
@@ -26,7 +26,7 @@
             <a class="dropdown-item" href="#saveModal" data-toggle="modal">Save</a>
         </li>
         <li>
-            <button class="dropdown-item" href="#changelogModal" data-toggle="modal">Changelog</button>
+            <a class="dropdown-item" href="#changelogModal" data-toggle="modal">Changelog</a>
         </li>
         <li>
             <a class="dropdown-item" href="https://pokeclicker.reddit.com" target="_blank">Subreddit</a>
--- a/src/components/gameMenu.html
+++ b/src/components/gameMenu.html
@@ -14,9 +14,6 @@
             <button class="dropdown-item" href="#">Stats</button>
         </li> -->
         <li>
-            <a class="dropdown-item" href="https://pokeclicker.reddit.com" target="_blank">Subreddit</a>
-        </li>
-        <li>
             <a class="dropdown-item" href="#saveModal" data-toggle="modal">Save</a>
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


src/components/gameMenu.html
```diff
@@ -14,26 +14,29 @@
...
         </li>
         <li>
-            <a class="dropdown-item" href="#showItemsModal" data-toggle="modal">Items</a>
+            <a class="dropdown-item" href="https://pokeclicker.reddit.com" target="_blank">Subreddit</a>
         </li>
         <li>
             <a class="dropdown-item" href="#donateModal" data-toggle="modal">Donate</a>
         </li>
+        <li>
+            <a class="dropdown-item" href="https://gitreports.com/issue/Ishadijcks/pokeclicker" target="_blank">Report a bug</a>
```
RedSparr0w: Do you think it would be better to link to `https://github.com/Ishadijcks/pokeclicker/issues/new` instead and create some [issue templates](https://help.github.com/en/articles/creating-issue-templates-for-your-repository).
I just feel like we will get a lot lower quality bug reports using gitreports?

Ishadijcks: The main advantage is to allow anonymous reports, I don't think that people are willing to create a GitHub account for reporting a bug.

The gitreport contains a link to the issues as sort of a bait for people with GitHub accounts.

I'm not entirely sure what would be best here.

RedSparr0w: Hmm, that is true.
Maybe just link to gitreports for now, and see how it goes, if we end up with a lot of low quality reports, then move to the github issue link, otherwise leave it as is?

Ishadijcks: Sounds good

src/components/gameMenu.html
```diff
@@ -14,26 +14,29 @@
...
-            <a class="dropdown-item" href="#achievementsModal" data-toggle="modal">Achievements</a>
+            <a class="dropdown-item" href="#showItemsModal" data-toggle="modal">Items</a>
         </li>
         <li>
-            <button class="dropdown-item" href="#changelogModal" data-toggle="modal">Changelog</button>
+            <a class="dropdown-item" href="#saveModal" data-toggle="modal">Save</a>
         </li>
         <li>
-            <a class="dropdown-item" href="#shardModal" data-toggle="modal">Shards</a>
+            <button class="dropdown-item" href="#changelogModal" data-toggle="modal">Changelog</button>
```
RedSparr0w: This one seems to be the only `button` element, not really an issue, but may as well change it now.
```suggestion
            <a class="dropdown-item" href="#changelogModal" data-toggle="modal">Changelog</a>
```

Ishadijcks: Huh that is weird

Issue #291, opened 2019-05-25
The dungeon ticket is still selected after buying it


Ishadijcks:
## Current Behavior

Game keeps the DungeonTicket selected instead of selecting an item that is available.

## Possible Solution

Reset the selection back to the first item in the shop whenever a one-off item is bought

## Steps to Reproduce

1. Start a fresh save, complete the tutorial for enough quest points
2. Buy the dungeon ticket
3. Exit the shop
4. Return to the shop
5. Press buy
6. You are

## Screenshot of screen or console (optional)
![image](https://user-images.githubusercontent.com/9715314/58369675-e8daf480-7efd-11e9-8662-3126cd2a4416.png)

PR #305, opened 2019-05-29
Mark key item in shop as sold out after buying

fuws:
**Key items in shops**

**Current behavior**
After buying, the item stays in shop. If you re-enter the shop, the item is invisible but still selected.

**New behavior**
After buying, the item stays in shop, but purchase is no longer possible even if there is enough currency. An error message prints when attempting to buy again. Once you re-enter the shop, the item is marked as "Sold Out".

Attempt to buy when the shop is still open
image

Re-enter the shop: item is still selected but is sold out
image

Marking as "Sold Out" as opposed to making invisible has the benefit that players can revisit the shop and know where the items are bought from. If item becomes invisible, players will forget where they bought it.

**Unresolved problem**
Immediately after buying, knockout's `if` binding doesn't kick in, so the key item isn't marked as Sold Out right away. It only gets marked as Sold Out after one re-enters the shop. I marked this as a TODO in `ShopHandler`.

**Save to test this change with**
[Pokeclicker save - shop key item test.txt](https://github.com/Ishadijcks/pokeclicker/files/3233929/Pokeclicker.save.-.shop.key.item.test.txt)

#291

### DIFF
--- a/src/components/shopView.html
+++ b/src/components/shopView.html
@@ -1,16 +1,21 @@
 <div id="shopView"
      data-bind="if: Game.gameState() === GameConstants.GameState.shop">
 <div class="row justify-content-center" data-bind="foreach: ShopHandler.shopObservable().items">
-   <div class="col-sm-3"  data-bind="visible: ShopHandler.ownKeyItem(name())">
+   <div class="col-sm-3">
        <div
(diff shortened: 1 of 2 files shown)


Ishadijcks:
Replaced by #307

PR #306, opened 2019-05-29
Use ISO date format to name save files

### LINKED ISSUE #304: Save file name format
Current save file, when downloaded, has the following format: `Pokeclicker save - 29_5_2019 @ 9_25`. 

Change this to `PokeClicker Save YYYYMMDD HH-MM-SS` (or something similar), so save files can be sorted in chronological order on one's computer.

fuws:
Current save file names use DDMMYYYY.

ISO date format (YYYY-MM-DD HH-MM-SS) allows chronological sorting of save files on people's computers.

Closes #304

### DIFF
--- a/src/scripts/Save.ts
+++ b/src/scripts/Save.ts
@@ -23,12 +23,7 @@ class Save {
         let element = document.createElement('a');
         element.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(btoa(JSON.stringify(player))));
         let currentdate = new Date();
-        let datetime = "" + currentdate.getDate() + "/"
-            + (currentdate.getMonth() + 1) + "/"
-            + currentdate.getFullYear() + " @ "
-            + currentdate.getHours() + ":"
-            + currentdate.getMinutes();
-        let filename = "Pokeclicker save - " + datetime + '.txt';
+        let filename = "PokeClickerSave_" + currentdate.toISOString() + '.txt';
         element.setAttribute('download', filename);
 
         element.style.display = 'none';
--- a/src/scripts/Save.ts
+++ b/src/scripts/Save.ts
@@ -23,7 +23,8 @@ class Save {
         let element = document.createElement('a');
         element.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(btoa(JSON.stringify(player))));
         let currentdate = new Date();
-        let filename = "PokeClickerSave_" + currentdate.toISOString() + '.txt';
+        let datestr = currentdate.toISOString().replace("T", " ").slice(0, 19);
+        let filename = "PokeClickerSave_" + datestr + '.txt';
         element.setAttribute('download', filename);
 
         element.style.display = 'none';


Aegyo:
`toISOString()` is in the format `YYYY-MM-DDTHH:MM:SSZ`
Do those colons get stripped out when saving? They will probably cause an error at least on windows machines if not removed.

Ishadijcks:
I am not a fan of the unneeded 'T', 'Z' and milliseconds. Also colons might be a problem on Windows I think. Can you please use the old function but reorder the year, month and date?

src/scripts/Save.ts
```diff
@@ -23,12 +23,7 @@ class Save {
         let element = document.createElement('a');
         element.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(btoa(JSON.stringify(player))));
         let currentdate = new Date();
-        let datetime = "" + currentdate.getDate() + "/"
-            + (currentdate.getMonth() + 1) + "/"
-            + currentdate.getFullYear() + " @ "
-            + currentdate.getHours() + ":"
-            + currentdate.getMinutes();
-        let filename = "Pokeclicker save - " + datetime + '.txt';
+        let filename = "PokeClickerSave_" + currentdate.toISOString() + '.txt';
```
RedSparr0w:
```suggestion
        let filename = 'PokeClickerSave_' + currentdate.toISOString().replace(/T|\.\d+Z/g, ' ').replace(/:/g, '-').trim() + '.txt';
```
Output: `2019-05-29 22-49-53`

RedSparr0w:
> Can you please use the old function but reorder the year, month and date?

But the old function has colons and also slashes, which means it basically has the same problem.

Any illegal characters seem to be replaced with `_` on windows.

fuws:
I removed the `T` and anything after seconds. `toISOString` is just much easier to work with. 

The big problem with the `Date.getMonth()` or similar methods is they return a number. So if it's less than 10, you need to pad zero. It becomes quite ugly.

src/scripts/Save.ts
```diff
@@ -23,12 +23,8 @@ class Save {
         let element = document.createElement('a');
         element.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(btoa(JSON.stringify(player))));
         let currentdate = new Date();
-        let datetime = "" + currentdate.getDate() + "/"
-            + (currentdate.getMonth() + 1) + "/"
-            + currentdate.getFullYear() + " @ "
-            + currentdate.getHours() + ":"
-            + currentdate.getMinutes();
-        let filename = "Pokeclicker save - " + datetime + '.txt';
+        let datestr = currentdate.toISOString().replace("T", " ").slice(0, 19);
```
RedSparr0w: This will still leave you with: `2019-05-30 01:48:25` which _should_ show as `2019-05-30 01_48_25` on windows.
I feel like it would be best to have a format that will show the same on all systems?

fuws: I think as long as it's consistent within one's own system, we're good. The goal is for players to be able to sort the files and find the one they want easily.

Ishadijcks:
Is the 0 padding fixed now?

fuws:
I'm still using ISO format, which has no issues with zero padding. I removed the unnecessary characters as you suggested.

Ishadijcks:
Ah alright, I was worried the 0 padding might mess with the slice function.

PR #307, opened 2019-05-29
Fuws/shop keyitem

Ishadijcks:
Replacement of https://github.com/Ishadijcks/pokeclicker/pull/305


RedSparr0w:
You should be able to use something like this:
```html
<!-- ko ifnot: ShopHandler.ownKeyItem(name()) -->
  <div id="the item div in here"></div>
<!-- /ko -->
```

src/scripts/shop/ShopHandler.ts
```diff
@@ -65,9 +69,9 @@ class ShopHandler {
         input.val(newVal > 1 ? newVal : 1).change();
     }
 
-    public static ownKeyItem(name: string): boolean {
-        let keyItem = GameConstants.KeyItemType[name];
-        return !(keyItem != undefined && player.hasKeyItem(name.replace("_", " ")));
+    public static itemAvailable(item: Item): boolean {
+        // Key items, when bought, are no longer available for purchase
+        return !((item instanceof buyKeyItem) && player.hasKeyItem(item.name().replace("_", " ")))
     }
```
Aegyo: This method would make more sense as an instance method of Item. Then buyKeyItem (or any other item type we want to limit availability of) can override the method with its own logic.

Ishadijcks: Agreed, clever

PR #309, opened 2019-05-29
shop keyitem fix

RedSparr0w:
Alternative to #305, #307 & #310
Hides the dungeon token once it has been purchased.
Reversed the logic for `ownKeyItem`, now returns `true` when the player already owns the item.

closes #307 
closes #310


Ishadijcks:
I like it better if it's still in the shop, just not available anymore.

PR #316, opened 2019-05-31
Add dark mode

RedSparr0w:
Add toggle for dark mode in menu. 

How it looks:

image


Ishadijcks:
I think a dark mode would be amazing, but I think we should design something of a 'settings' panel before implementing more of such features. I'll open an issue for it.

I don't think it's fair to reject this pull request due to the lack of another feature, but I'd like someone else's opinion on it.

Aegyo:
`player.darkMode` and the menu button can be replaced by the `theme` setting added in #325, and the knockout binding on `body` can observe that setting's `observableValue`

Ishadijcks:
Yeah already working on it :)

PR #318, opened 2019-05-31
Evolution Stone UI Clarification

### LINKED ISSUE #317: Evolution stone interface
**Two things:**

1. **Issue:** No matter which amount the player selects, the stones usage always stops when a shiny is encountered. This mechanic is great. But this isn't very clear to the player. E.g. player may think the 1000 option would keep using the stone even if shiny is encountered.
image

**Suggested improvement**: Make it clear to the player that stone usage always stops after a shiny is encountered. Change the `Until shiny` to `Infinity` or the infinity symbol.

2. **Issue:** When a shiny is encountered, there is an off-by-one error when displaying how many stones got used. 
image

**Console code to reproduce:**
`player.gainItem("Fire_stone", 1000) // Or any other stone`
`GameConstants.SHINY_CHANCE_STONE = 1`
Then use the stone on a Pokemon to evolve.

fuws:
Closes #317 

**Counts stones used properly**
image

**Clarify that all options stop when shiny**
image

### DIFF
--- a/src/components/itemModal.html
+++ b/src/components/itemModal.html
@@ -118,12 +118,12 @@
                                     <div class="col" style="text-align: center">
                                         <button class="btn btn-secondary amount-button list-group-item-action"
                                                 data-bind="click: function() {ItemHandler.amountSelected(Infinity)}, css:{ 'pokemon-selected': ItemHandler.amountSelected() === Infinity}">
-                                            <span>Until shiny</span>
+                                            <span>All</span>
                                         </button>
                                     </div>
                                 </div>
                             </div>
-
+                            <p><b>Evolution stone usage stops when a shiny evolution is encountered</b></p>
                             <div class="row no-gutters justify-content-sm-center">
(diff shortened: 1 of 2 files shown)


Ishadijcks:
The off-by-one error is a nice catch!

I am not a big fan of the (until shiny) part, it's a bit too vague.
Maybe we can add a checkbox above something like

- [x] Stop using Evolution Stones when you capture a shiny.

I don't think anybody will turn it off, but I believe it conveys the message nicely.

As I was reviewing this, I noticed that the statement `amount = i + 1` is a bit too implicit, while somewhat clever.

I know you didn't introduce this problem, but would you mind refactoring it to something where `amount` would become `max_amount`. And when the `if` is triggered an `actual_amount` would be set to `i`, after which the for loop is broken out of. I believe this makes it much more intuitive.

fuws:
I force pushed to keep the commit history simple. New look:

image

@Ishadijcks

RedSparr0w:
Nice work, Looks good!

PR #320, opened 2019-06-01
Add Explorer kit to shop

fuws:
Added to Cinnabar Island shop. I chose Cinnabar because it's after route 11 and currently has only 3 items in shop. So adding another one won't clutter the shop.

image

**Code to remove your Explorer kit to see it in shop:**
`player._keyItems(player._keyItems().filter(x=>x!="Explorer kit"))`


RedSparr0w:
Nice, do you think it should be a bit more expensive though? as it is quite late game for kanto?

fuws:
I think so. The `TODO` said to add for 100. I think something like 300-500 is appropriate. Let's see what @Ishadijcks thinks.

RedSparr0w:
I think even 5000 wouldn't be too bad, considering stones can be 2500, and eggs 1000.
And you can also obtain stone from the underground.

src/scripts/GameConstants.ts
```diff
@@ -568,6 +568,7 @@ namespace GameConstants {
         "EvolutionStone": 2500,
 
         "Dungeon_ticket": 250,
+        "Explorer_kit": 100,
```
Ishadijcks:
```suggestion
        "Explorer_kit": 5000,
```

PR #321, opened 2019-06-01
Correct plot:hover css selector

### LINKED ISSUE #294: Farm countdown tooltip acts buggy
While hovering over a farmplot the berry and the tooltip start to blink weirdly. Probably an issue with the tooltip but I'm not sure.

Aegyo:
```
.plot {
  :hover {
    thing: value;
  }
}
```
compiles to `.plot :hover`, ie every child of a `.plot` which is being hovered over.
Adding the `&` makes it compile to `.plot:hover`.

I have no idea why this fixes #294 , but it does.

### DIFF
--- a/src/styles/farm.less
+++ b/src/styles/farm.less
@@ -35,7 +35,7 @@
   padding-bottom: 20%;
   overflow: hidden;
 
-  :hover {
+  &:hover {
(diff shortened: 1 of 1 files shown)


Ishadijcks:
The best kind of fix

PR #325, opened 2019-06-01
Add a generic settings menu

Ishadijcks:
With this feature you can easily add settings that the user can tweak.

In the `settingsModal.html` you can create a custom UI grid with all the settings.
Adding a setting is as simple as adding the following line in `Settings.ts`.
```
Settings.add(new BooleanSetting("test", "Do the settings work?", false));
```
And 
```
<div class="col-md-4">
    <div data-bind="template: {
        name: 'MultipleChoiceSettingTemplate',
        data: Settings.getSetting('theme')
    }">
    </div>
</div>	
```

Other setting types can easily be added when needed.

I've refactored the GameConstants.Option a bit as it did nearly exactly what I needed.


src/components/settingsModal.html
```diff
@@ -0,0 +1,38 @@
...
+                <span>Settings</span>
+            </div>
+            <div class="modal-body">
+                <div class="row">
+                    <div class="col-md-4">
+                        <div data-bind="template: { name: 'MultipleChoiceSettingTemplate', data: Settings.getSetting('theme')}"></div>
+                    </div>
+                    <div class="col-md-8">
+                        <div data-bind="template: { name: 'MultipleChoiceSettingTemplate', data: Settings.getSetting('test')}"></div>
+                    </div>
```
Aegyo: Do you think this should be done via a `ko foreach:Settings.list` thing, and the Settings object can have a property for the template it should use?

src/scripts/settings/Settings.ts
```diff
@@ -0,0 +1,59 @@
...
+Settings.add(
+    new MultipleChoiceSetting("theme", "Theme",
+        [
+            new GameConstants.Option("Default", 'default.css'),
+            new GameConstants.Option("Dark Mode", 'dark.css')
+        ],
+        "default.css",
+    )
+);
+Settings.add(new BooleanSetting("test", "Do the settings work?", false));
```
Aegyo: Do we want this setting in the develop branch?

src/scripts/settings/Settings.ts
```diff
@@ -0,0 +1,59 @@
...
+
+class Settings {
+    static list: Setting[] = [];
+
+
+    static add(setting: Setting) {
+        if (this.list.indexOf(setting) === -1) {
+            this.list.push(setting);
+        }
+    }
```
Aegyo: `this.list.indexOf(setting)` will only not be `-1` if the object `setting` has already been put in the list. A different object with identical properties will not match.

I think what we really want to be checking here is whether there is already a setting with the same name, as that would cause a conflict with setting/loading.
eg, `if (!this.getSetting(setting.name))`

```
x = {'a':1}
y = {'a':1}
list = []
list.push(x)
list.indexOf(x)
> 0
list.indexOf(y)
> -1
```

src/components/settingsModal.html
Ishadijcks: I've considered that, but then you can't manually order it however you want.

src/scripts/settings/Settings.ts
Ishadijcks:
```suggestion
```

src/scripts/settings/Settings.ts
```diff
@@ -0,0 +1,58 @@
+/// <reference path="Setting.ts" />
+/// <reference path="BooleanSetting.ts"/>
+/// <reference path="MultipleChoiceSetting.ts"/>
+
+class Settings {
+    static list: Setting[] = [];
+
+
+    static add(setting: Setting) {
+        if (this.list.indexOf(setting) === -1) {
```
Ishadijcks:
```suggestion
        if (!this.getSetting(setting.name))) {
```

src/scripts/settings/Settings.ts
Ishadijcks: Interesting, fixed

src/scripts/settings/Settings.ts
Ishadijcks: Removed

src/components/settingsModal.html
Aegyo: True, I guess it would need some kind of `order` property too. Seems like a bit too much UI work for just the couple of settings we have right now. Might be worth revisiting when there are more settings and that html file is growing out of control.

PR #327, opened 2019-06-01
Implement energy restore items

fuws:
Energy restore costs money and restores a certain percentage of player's maximum mine energy, unaffected by Cell Battery. Cell battery's tool tip was updated to reflect this information.

Usage fails if energy is already full, with an error message. A success message prints if the restore was successful, with the actual restored amount.

image

image

**Balancing thoughts**
This is very preliminary. For small restore (10% max energy), with base mine stats, 5 energy = 100 seconds passive regen. On the routes, 100 seconds ~ 20K money (200 per second). So small restore costs 20K. 

Medium/Large restore prices scale according to their effect. But the benefit here is, player doesn't have to buy as many to achieve the same effect, hence reducing cost.

**Testing**
`player.gainKeyItem("Explorer kit") // Give you access to the mine`
`ItemList['SmallRestore']._increaseCount(100) // Give yourself some restore items`
`ItemList['SmallRestore'].use()`

### DIFF
--- a/src/scripts/GameConstants.ts
+++ b/src/scripts/GameConstants.ts
@@ -547,9 +547,9 @@ namespace GameConstants {
         "Item_magnet": 1500,
         "Lucky_incense": 2000,
 
-        "SmallRestore": 100,
-        "MediumRestore": 100,
-        "LargeRestore": 100,
+        "SmallRestore": 20000,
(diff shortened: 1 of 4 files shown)


RedSparr0w:
Ooo, nice work, i wasn't sure what these were for, now i know :smile:

PR #329, opened 2019-06-01
Dark mode using the settings menu

Ishadijcks:
Closes #316

I have changed to `Notifier.Notify()` functions to console logs. Otherwise loading an incorrect setting would crash as the Notify function wasn't loaded yet.


RedSparr0w:
If you could have a look at #336 before merging this, I've added some more stuff that may be useful.

Aegyo:
Waiting on the dark mode changes from #336 before reviewing this in more depth, but it looks good so far.

PR #335, opened 2019-06-01
Set min height of images

### LINKED ISSUE #333: When catching a Pokémon, the healthbar jumps up for a short time
This is visually very disturbing. It does not happen without catching.

Example:
https://gfycat.com/NextPiercingJaguarundi

RedSparr0w:
Alternative to #334 
Have set the minimum height of the pokeball & pokemon images, this should eliminate the jumping that occours when an image has to be loaded.

closes #333 
closes #334

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -200,7 +200,7 @@
 
                             <div>
                                 <div data-bind="ifnot: Battle.catching">
-                                    <img class="clickable" id="enemy" style="min-height:96px;"
+                                    <img class="clickable" id="enemy"
                                          data-bind="click: function() {Battle.clickAttack()},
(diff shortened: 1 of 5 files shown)


Ishadijcks:
Why is this better?

Ishadijcks:
I think you branched from the wrong commit?

I get the following error.

```
TypeError: OakItemRunner.getOakItemByName(...) is undefined
```

RedSparr0w:
Yeah, shouldn't matter, the changes are still the same and won't be affected when merged.

This fixes the jumping issue for when pokemon images need to load, not just pokeballs.
Also does it without adding extra elements, just using styles.

Ishadijcks:
I know, It makes reviewing more of a hassle.

RedSparr0w:
Thanks,

Yeah, i'll update my develop branch :+1:

PR #336, opened 2019-06-02
Update Dark Mode

RedSparr0w:
**Note:** This is to be merged into the `feature/dark-mode` branch then merged with #329

**Note2:** To review this easier i would suggest hiding whitespace changes:
Screenshot from 2019-06-02 13-34-32


RedSparr0w:
Fixed dark mode tabs:
Screenshot from 2019-06-02 13-17-04

RedSparr0w:
Make tabs more tab like:
Screenshot from 2019-06-02 13-32-20
Screenshot from 2019-06-02 13-32-56

RedSparr0w:
Update save panel tabs:
Screenshot from 2019-06-02 13-45-45

Ishadijcks:
Please split this pr into changes for the dark theme and the change in the settings menu.

I am not a fan of the changed settings as I didn't use a foreach deliberately so it was possible to customize the UI however we might want later.

I do like the dark theme changes :)

RedSparr0w:
Removed settings panel change, is just the dark mode changes now, if you could re-open.

RedSparr0w:
>I am not a fan of the changed settings as I didn't use a foreach deliberately so it was possible to customize the UI however we might want later.

Although this is no longer a part of this PR, I feel like for now it would be better to have it as the list as it looks a little nicer IMO, until more settings are added?

image

Should I re-open that part as a new PR? or are we fully set on keeping it how it is?

PR #338, opened 2019-06-02
Generic upgrade system and Underground refactor

Ishadijcks:
So this got kind of got outta hand.

I've added a generic `Upgrade` class which can be easily extended for more customizable upgrades. This works quite well with the new UndergroundUpgrades.

While I was doing this, I also refactored a ton of Underground logic which was necessary for it to work with the new Upgrade logic.

Nearly all of the logic related to the Mine and Underground has been moved from Player and GameConstants to classes where they are needed. Only mineInventory is remaining but I've added TODO's to fix that in another pr.

Save logic was added to the Underground, which is now responsible for saving itself instead of having the player object do it.

Please properly test the save logic, don't want to screw that up.

This pr will reset your Underground Upgrades, so make sure to get yourself some diamonds:

`player.diamonds= 1000;`

https://gfycat.com/LankyHospitableIberianchiffchaff


src/scripts/Save.ts
```diff
@@ -7,6 +7,13 @@ class Save {
         localStorage.setItem("player", json);
         localStorage.setItem("mine", Mine.serialize());
         localStorage.setItem("settings", Settings.save());
+
+        let saveObject = {};
+
+        saveObject[Underground.saveKey] = Underground.save();
+
+        localStorage.setItem("save", JSON.stringify(saveObject));
```
Ishadijcks: This is how I want to do all saving in the future.

src/scripts/upgrades/Upgrade.ts
```diff
@@ -0,0 +1,76 @@
+///<reference path="../GameConstants.ts"/>
+///<reference path="Cost.ts"/>
+///<reference path="CostFactory.ts"/>
+
+class Upgrade {
+    name: any;
+    displayName: string;
+    maxLevel: number;
+    _level: KnockoutObservable<number> = ko.observable();
+    increasing: boolean;
```
Aegyo: Might be worth adding a comment here for when we forget what this is.

src/scripts/underground/Underground.ts
```diff
@@ -1,50 +1,81 @@
...
- * Created by dennis on 03-07-17.
- */
 class Underground {
+    public static saveKey: string = "underground";
+
     public static itemSelected;
     public static energyTick: KnockoutObservable<number> = ko.observable(60);
     public static counter: number = 0;
 
+    public static energy = ko.observable(50);
```
Aegyo: As much as I love refreshing the page to refill energy, this should probably be saved and loaded :)

src/scripts/underground/Mine.ts
```diff
@@ -1,10 +1,12 @@
 class Mine {
+    public static sizeX: number = 25;
+    public static sizeY: number = 12;
     public static grid: Array<Array<KnockoutObservable<number>>>;
     public static rewardGrid: Array<Array<any>>;
     public static itemsFound: KnockoutObservable<number>;
     public static itemsBuried: number;
     public static rewardNumbers: Array<number>;
-    public static toolSelected: KnockoutObservable<GameConstants.MineTool> = ko.observable(GameConstants.MineTool["Chisel"]);
+    public static toolSelected: KnockoutObservable<Mine.Tool> = ko.observable(0);
```
Aegyo: The `0` here seems a little opaque.
```suggestion
    public static toolSelected: KnockoutObservable<Mine.Tool> = ko.observable(Mine.Tool.Chisel);
```

Bart-van-Ingen: Yeah there was a problem here.
`Mine.Tool.Chisel` can't be used before its declaration in the namespace, and the namespace cannot come before the class it is merged with. I'm not sure of a better solution but I'll add a comment here

Ishadijcks: Yeah there was a problem here.
`Mine.Tool.Chisel` can't be used before its declaration in the namespace, and the namespace cannot come before the class it is merged with. I'm not sure of a better solution but I'll add a comment here

src/scripts/underground/Underground.ts
Ishadijcks: 😅

src/scripts/upgrades/Upgrade.ts
Ishadijcks: Done

src/scripts/upgrades/Upgrade.ts
```diff
@@ -0,0 +1,79 @@
...
+
+class Upgrade {
+    name: any;
+    displayName: string;
+    maxLevel: number;
+    _level: KnockoutObservable<number> = ko.observable();
+
+    // Describes whether this upgrade increases or decreases a number.
+    // (e.g. power is increasing, time is decreasing).
+    increasing: boolean;
```
fuws: `Upgrade.increasing` doesn't appear to be used anywhere. The only place where it might be used `Underground.getEnergyRegenTime`; but these underground upgrades are hard-coded to be either + or - the benefit. Is `increasing` redundant? Even if we have a later upgrade that doesn't hard code the direction of the math, we can still have `Upgrade.bonusList` be negative numbers, right?

src/scripts/Save.ts
fuws: `localStorage` is itself an Object. Is it worth creating an extra layer called `save`, when you eventually plan to move everything here? 

Maybe a simpler structure like `localStorage.setItem(Underground.saveKey, Underground.save());`.

src/scripts/upgrades/Upgrade.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    // Optional array of costs
+    costList: Cost[] = [];
+    // Optional array of benefits
+    bonusList: number[] = [];
```
fuws: Don't have to do this now. But we should think about if there's a need for upgrades with unlimited levels.

src/scripts/upgrades/Upgrade.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    constructor(name: any, displayName: string, maxLevel: number, costList: Cost[], bonusList: number[], increasing = true) {
+        this.name = name;
+        this.displayName = displayName;
+        this.maxLevel = maxLevel;
+        this.level = 0;
+        this.costList = costList;
+        this.bonusList = bonusList;
```
fuws: Is it necessary (or even possible) to have a way to make sure that `costList.length >= maxLevel && bonusList.length >= maxLevel + 1` during compile?

src/scripts/upgrades/Upgrade.ts
Ishadijcks: It is used in the ui, to display a + or - sign. My idea was to have bonuses always be positive for consistency, when a value gets calculated the bonus gets interpret by that function as you see in the energyRegenTime.

src/scripts/upgrades/Upgrade.ts
Ishadijcks: This can be achieved by having the `bonusList` empty and overriding the `calculateBonus()` function to be infinite.

src/scripts/upgrades/Upgrade.ts
Ishadijcks: We could write some tests for it if we ever decide to implement those.

src/scripts/upgrades/Upgrade.ts
fuws: Make sense.

src/scripts/upgrades/Upgrade.ts
fuws: If the function interpretation and the display are consistent, then wouldn't it be more concise to have `current_value + bonus = new_value` at all times, even if bonus is negative? If the word bonus gives the connotation of something positive, can change it to something else, like "effect".

PR #339, opened 2019-06-04
Allow for multiple concurrent quests

fuws:
Number of quest slots is hard-coded to 1. But this can be easily modified with the `QuestHelper.questSlots` function.

These screenshots are taken with the case of 2 quest slots allowed:

**Main page display**
image

**Quest modal (can start more quests when one is already in progress)**
image

**Summary of code design changes:**
- Have `player.currentQuests` be a observableArray that tracks active quests
- Move logic to check whether quests are available into `QuestHelper.ts`. 
- In `Quest.ts`, combine `quit` and `claimReward` functions into a single `exit` function, that either quits or claims reward based on quest completion status.

### DIFF
--- a/src/components/questDisplay.html
+++ b/src/components/questDisplay.html
@@ -19,7 +19,7 @@ style="background-color: white; min-height: 100px">
                 <div class="col-sm-11 col-md-8">
                     <div class="row" style="margin-bottom: 10px">
                         <div class="col-9" style="padding:0px; margin:0px">
-                            <div class="progress" style="padding:0px; text-align:center" 
+                            <div class="progress" style="margin-top:6px; padding:0px; text-align:center" 
                                 data-bind='tooltip: {title: QuestHelper.questList()[$data.index].progressText, trigger: "hover"}'>
                                 <div class="progress-bar bg-success" role="progressbar" 
                                     data-bind="attr:{style: 'width:' + (QuestHelper.questList()[$data.index].progress() * 100) + '%'}"
@@ -28,7 +28,7 @@ style="background-color: white; min-height: 100px">
                             </div>
                         </div>
                         <div class="col-3">
-                            <button style="margin:0px; padding:0px; min-width:60px"
+                            <button style="margin:0px; padding:0px; height:20px; min-width:60px"
                                     data-bind='
                                         click: function(){QuestHelper.questList()[$data.index].exit();}, 
                                         text: QuestHelper.questList()[$data.index].isCompleted() ? "Claim" : "Quit", 
--- a/src/components/questDisplay.html
+++ b/src/components/questDisplay.html
@@ -1,42 +1,41 @@
 <div class="page-item" id="questDisplayContainer"
 style="background-color: white; min-height: 100px">
-    <br>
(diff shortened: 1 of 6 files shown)


src/scripts/quests/Quest.ts
```diff
@@ -25,11 +25,10 @@ abstract class Quest {
         this.notified = false;
     }
 
-    claimReward() {
+    exit() {
```
RedSparr0w: i feel like `exit` isn't very descriptive of the function,
Do you think something like `endQuest` would work better?
```suggestion
    endQuest() {
```

src/components/questDisplay.html
```diff
@@ -1,42 +1,41 @@
...
+            <p style="margin-bottom: 0px" data-bind="text: QuestHelper.questList()[$data.index].description"></p>
+            <div class="row justify-content-sm-center">
+                <div class="col-sm-11 col-md-8">
+                    <div class="row" style="margin-bottom: 10px">
+                        <div class="col-9" style="padding:0px; margin:0px">
+                            <div class="progress" style="margin-top:6px; padding:0px; text-align:center" 
+                                data-bind='tooltip: {title: QuestHelper.questList()[$data.index].progressText, trigger: "hover"}'>
+                                <div class="progress-bar bg-success" role="progressbar" 
+                                    data-bind="attr:{style: 'width:' + (QuestHelper.questList()[$data.index].progress() * 100) + '%'}"
+                                    aria-valuemin="0" aria-valuemax="100">
```
RedSparr0w: If you wanted to have the text inside the progress bar similar to how the breeding progress bar is shown, you should be able to do something like this:
```suggestion
                                    aria-valuemin="0" aria-valuemax="100">
                                    <span data-bind="text: Math.floor((QuestHelper.questList()[$data.index].progress() * 100)) + '%'">0%</span>
```
**Note:** The span element will need to be the same width as the full progress bar (80% for the breeding progress bar)
image

RedSparr0w:
This is looking awesome!
Can't wait to test it out :+1:

src/scripts/quests/QuestHelper.ts
```diff
@@ -135,16 +135,28 @@ class QuestHelper{
...
+        // Two conditions for starting new quest:
+        // 1. Current quests not exceed maximum slots
+        // 2. At least one quest is neither completed nor in-progress
+        if (player.currentQuests().length >= this.questSlots()()) {
+            return false;
+        }
+        for (let i = 0; i < QuestHelper.questList().length; i++) {
+            if (!(player.completedQuestList[i]() || QuestHelper.questList()[i].isCompleted() 
+                    || QuestHelper.questList()[i].inProgress()())) {
+                return true;
```
RedSparr0w: I _think_ you could replace the whole `for` block with this?
```javascript
return QuestHelper.questList().some(q=>!q.isCompleted() && !q.inProgress()());
```

fuws: You do need to check against `player.completedQuestList[i]()`, because this stores completed quests from previous game sessions, while `Quest.isCompleted()` only gets set to true when you complete the quest in the current session.

PR #346, opened 2019-06-06
Number of pokeballs reset when player revisits safari.

### LINKED ISSUE #343: Number of Pokeballs not reset when revisiting safari zone
When the safari zone loads (statically, with the web page) the value ```Safari.balls``` is set to 30. This number is never reset to 30 until the page is refreshed.

This allows for the player to lose all of their pokeballs and revisit the safari zone with 0 pokeballs. They can then continue to use pokeballs as ```Safari.balls``` reaches negative values.

There are two possible expected behaviors for entering the safari zone. It's up for discussion regarding which one we should use:
1. Whenever the player enters the safari zone they enter with 30 pokeballs.
2. The player starts with 30 pokeballs. Whenever the player enters the safari zone, they enter with the number of pokeballs they currently have (ex: a player enters, throws 5, leaves, and re-enters with 25 left). If the player enters the safari zone with 0 pokeballs, the number is reset to 30.

jameswasson:
resolves #343

### DIFF
--- a/src/scripts/GameConstants.ts
+++ b/src/scripts/GameConstants.ts
@@ -163,6 +163,8 @@ namespace GameConstants {
     
     export const SAFARI_BATTLE_CHANCE = 5;
 
+    export const SAFARI_BASE_POKEBALL_COUNT = 30;
+
     export const LEGAL_WALK_BLOCKS = [
         0,
         10,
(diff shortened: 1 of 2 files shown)


Ishadijcks:
While this definitely solves it, it will probably be refactored once we make this dynamic.
Can you add a static public attribute to the Safari class, and a method to calculate the amount of starting pokeballs?

Safari.ts
```
BASE_POKEBALL_COUNT = 30;

...

calculateStartPokeballs(){
	return BASE_POKEBALL_COUNT;
}

...

Safari.balls(calculateStartPokeballs())

```

Now later we can simply change the `calculateStartPokeballs` to `return BASE_POKEBALL_COUNT + some_upgrade_value`. :)

PR #352, opened 2019-06-07
Add linter (eslint)

RedSparr0w:
tslint is being replaced by eslint, It hasn't fully been released yet so unsure if you want to merge this.
_See https://github.com/palantir/tslint/issues/4534 for more info_

I've just left the rules as standard.

There is still a ton of errors, i've fixed some, but we can slowly remove more and more errors as we work on the project.

The main errors/warnings that pop up are when the return type hasn't been declared, usually:
```typescript
function nothingToReturn(): void {
```

I think merging this would mostly be good for keeping consistency in the code in terms of spacing etc..

Usage:
`npm test` - Shows all errors and warnings
`npm run test-error` - Only shows errors, ignores warnings
`npm run test-fix` - Will show all warnings and errors, and attempt to fix any that can be fixed automatically such as spacing issues

### DIFF
--- a/src/scripts/pokemons/PokemonFactory.ts
+++ b/src/scripts/pokemons/PokemonFactory.ts
@@ -86,7 +86,7 @@ class PokemonFactory {
         let maxHealth: number = Math.floor(baseHealth * (1 + (chestsOpened / 5)));
         let catchRate: number = this.catchRateHelper(basePokemon.catchRate);
         let exp: number = basePokemon.exp;
-        let money: number = 0;
+        let money = 0;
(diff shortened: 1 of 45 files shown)


src/scripts/pokemons/PokemonFactory.ts
```diff
@@ -100,7 +100,7 @@ class PokemonFactory {
         let maxHealth: number = Math.floor(bossPokemon.baseHealth * (1 + (chestsOpened / 5)));
         let catchRate: number = this.catchRateHelper(basePokemon.catchRate);
         let exp: number = basePokemon.exp;
-        let money: number = 0;
```
Ishadijcks: Why remove it here?

PR #355, opened 2019-06-08
Lower memory usage

### LINKED ISSUE #332: Repeated breeding crashes chrome
Repeated breeding crashes google chrome.

After breeding some pokemon, the breeding modal begins to lag after selecting a pkm, as well as the map in selecting a rout. 
When then continuing to breed, battle ticks slow down until one tick (pkm dmg is applied) takes 3-5 seconds. 
Shortly afterwards chrome crashes. According to chrome debugging, out-of-memory error while in knockout-latest.js (if I understand it correctly) 

Console

Sources

Savefile: 
https://pastebin.com/wHX7ihbp

Chrome Version: Version 74.0.3729.169 (Official Build) (64-bit)
last sync with github: 01.06.2019 1pm (ish)

RedSparr0w:
closes #332 it should reduce the amount of memory used to around ~25-50%

It may still need some more investigating, but this should help a ton.

I've removed the knockout bindings, the map seems to update properly without them though:

memory usage over 20 seconds using the following script:
```javascript
setInterval(()=>{
	MapHelper.moveToRoute(Math.floor(Math.random() * 6) + 1, 0);
}, 10);
```

Before | After
:--: | :--:
`88MB`→ `562MB` | `85.9MB`→`224MB`
memory-usage-withko | memory-usage-noko

### DIFF
--- a/src/components/JohtoSVG.html
+++ b/src/components/JohtoSVG.html
@@ -14,18 +14,18 @@
     <path class="city" data-town="TOWN"
           data-bind="
           click:function(){MapHelper.moveToTown('TOWN')},
-          attr: { class: MapHelper.calculateTownCssClass('TOWN')() }"
+          attr: { class: MapHelper.calculateTownCssClass('TOWN') }"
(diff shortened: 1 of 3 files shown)


Ishadijcks:
Seem pretty high still. Any idea what's causing it?

RedSparr0w:
Not a clue, but definitely better than before.

PR #357, opened 2019-06-10
Fix pokedex, add defeated dungeon pokemon

### LINKED ISSUE #356: Pokedex ends at Nr.244 instead of 251
Pokedex only shows Pokemon up to Nr.244,Nr.245-251 are missing

image

RedSparr0w:
closes #356 

Dungeons didn't count Pokemon defeated,
Also updated the Pokedex to show up to either the highest caught Pokemon id or the highest defeated Pokemon id _(whichever is greater)_.

### DIFF
--- a/src/scripts/dungeons/DungeonBattle.ts
+++ b/src/scripts/dungeons/DungeonBattle.ts
@@ -9,6 +9,7 @@ class DungeonBattle extends Battle {
         player.gainExp(this.enemyPokemon().exp, this.enemyPokemon().level, false);
         player.gainShards(this.enemyPokemon());
         player.addRouteKill();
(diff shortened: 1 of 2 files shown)


src/scripts/dungeons/DungeonBattle.ts
```diff
@@ -9,6 +9,7 @@ class DungeonBattle extends Battle {
         player.gainExp(this.enemyPokemon().exp, this.enemyPokemon().level, false);
         player.gainShards(this.enemyPokemon());
         player.addRouteKill();
```
Ishadijcks: Should this be removed then?

RedSparr0w: Yeah, looks like `player.route()` is always `0` when you're in a dungeon.
So it's just tracking total dungeon kills.

Issue #363, opened 2019-06-11
Unobtainable Gen2 Pokemon


fuws:
Aside from regular level-ups, these are all currently unobtainable Gen2 Pokemon, and the way to get them in the original games (source: serebii.net, Gen2 Dex) 

ID | Name | Method
-- | -- | --
152 | Chikorita | Starter
155 | Cyndaquil | Starter
158 | Totodile | Starter
169 | Crobat | Happiness
172 | Pichu | Breed from Gen1
173 | Cleffa | Breed from Gen1
174 | Igglybuff | Breed from Gen1
175 | Togepi | Egg given in Violet City
176 | Togetic | Happiness
177 | Natu | Ruins of Alph
185 | Sudowoodo | Route 36 unique
186 | Politoed | Trade Poliwhirl with King's Rock
190 | Aipom | Headbutt in various locations
191 | Sunkern | National park, route 24
192 | Sunflora | Sun stone
196 | Espeon | Happiness
197 | Umbreon | Happiness
198 | Murkrow | Route 7, 16, 18
199 | Slowking | Trade Slowpoke with King's Rock
201 | Unown | Ruins of Alph
204 | Pineco | Headbutt in various locations
208 | Steelix | Trade Onix with Metal Coat
211 | Qwilfish | Route 32, 12, 13
212 | Scizor | Trade Scyther with Metal Coat
213 | Shuckle | Rock smash at various locations
214 | Heracross | Headbutt in various locations
218 | Slugma | Route 16, 17, 18
223 | Remoraid | Route 44
228 | Houndour | Route 7
230 | Kingdra | Trade Seadra with Dragon Scale
233 | Porygon2 | Trade Porygon with Up-Grade
235 | Smeargle | Ruins of Alph
237 | Hitmontop | Evolve Tyrogue (A=D)
238 | Smoochum | Breed from Gen1
239 | Elekid | Breed from Gen1
240 | Magby | Breed from Gen1
242 | Blissey | Happiness

These are the main categories, and suggestions on what to do with them:

- **Starters**: Have the grass/fire/water eggs be modified when `player.highestRegion == 1` to include them.
- **Happiness**: Have `Time_stone` be either purchasable in some shop or be mined (or both). For the Eevee ones, have it depend on time of day.
- **Trade**: Implement using existing `Trade_stone`. Ignore all the items that trading requires (e.g. Metal Coat).
[...]

PR #368, opened 2019-06-12
Allow pokemon to evolve again after breeding, apply custom evolutions, give tyrogue, eevee cutom evolution

RedSparr0w:
Refer #363 

This allows Tyrougue to evolve correctly, also allows you to obtain espeon/umbreon.
And can also be used for other pokemon in the future (Happiness based etc..)

Current problems that will need to be addressed at some point:
-  ~~The pokemon can only evolve once as the `evolved` flag is not reset on when breeding~~
- ~~Even if you reset the `evolved` flag once they hatch, the evolver observer isn't started again~~


fuws:
So with this, the player may still fail to get Hitmontop during the prestige?

RedSparr0w:
Yes, but in a seperate PR I will update the breeding functions to reset the caught pokemon `evolved` state to `false`, and re attach the evolver observer.

So you may have to breed Tyrogue a bunch of times before getting Hitmontop

RedSparr0w:
Eh, decided to fix it in this PR.

Now any pokemon can evolve again after hatching.

On a side note, it looks like it is impossible to obtain a shiny pokemon from evolving.

src/scripts/pokemons/CaughtPokemon.ts
```diff
@@ -49,4 +38,38 @@ class CaughtPokemon {
         plainJS = ko.toJS(this);
         return Save.filter(plainJS, keep);
     }
+
+    public evoTrack(reset = false){
```
RedSparr0w: I don't know what to call this, taking any and all suggestions!

RedSparr0w:
I've added another custom function for eevee,
But it might work better as an array?

```javascript
        "evolution": ["Vaporeon",    "Jolteon",       "Flareon",    ["Espeon", "Umbreon"]],
        "evoLevel":  ["Water_stone", "Thunder_stone", "Fire_stone", "Time_stone"],
```

RedSparr0w:
>I think I'm missing something, how is it decided if you get Hitmontop?

It's decided randomly

https://github.com/Ishadijcks/pokeclicker/pull/368/files#diff-3e703bba7e316c46a04035db1cfbba4eR3174

src/scripts/pokemons/CaughtPokemon.ts
Ishadijcks: `checkForEvolution()`?

Ishadijcks:
Can you fix the merge conflicts please?

RedSparr0w:
Conflicts resolved 👍

PR #371, opened 2019-06-15
Gen2 evolution stones

fuws:
Add Dragon scale, Metal coat, King's rock, and Upgrade to Johto shops (costs 2500 QP each). Using these from the Item menu will directly evolve Pokemon (in the actual Pokemon series game, need to trade them while holding these items). 

In the Items menu, Pokemon will only appear to be available to evolve using these items if player has unlocked Johto. 

**Shop in Azalea Town (shops for other evolution items look similar)**
image

**In Items menu (Johto unlocked)**
image

**In items menu (Johto not accessible yet)**
image

This makes the following Pokemon obtainable in #363 : Politoed, Steelix, Scizor, Kingdra, and Porygon2. 

Slowking is still unavailable, because Slowpoke needs to be coded to evolve both using leveling (to Slowbro) and Kings rock (to Slowking). Current code doesn't easily support this. We'll need to change up some of the leveling code to handle this. It is probably good to have its own PR for this.


RedSparr0w:
If #368 is merged it will make Slowking easy to implement,
Should only need the following:
```javascript
        "evolution": ["Slowbro", "Slowking"],
        "evoLevel": [37, "Kings_rock"],
```

fuws:
I was thinking your work might make it easy so I didn't spend time trying to get Slowking to work, @RedSparr0w

PR #373, opened 2019-06-16
(Alt #368) Allow pokemon to evolve again after breeding, handle multiple evolutions, time stone

RedSparr0w:
This PR is an alternative version to #368 

Just use arrays of pokemon for multiple evolutions instead of a custom function
Also hides pokemon you shouldn't be able to evolve to yet (not upto that region)
And will show a red or yellow pokeball based on if you have caught all the evolutions (for eevee using the timestone)

Still need an image for the timestone and a away to obtain them.(assuming this is how eevee will be handled as currently coded)


fuws:
Do you plan to add `Time_stone` to some shops? Is that going to be the same as evolving by happiness?

RedSparr0w:
I just used `Time_stone` as that is what was already in the code for `Eevee`, not sure if that is the best way or not, but i guess it should be added to a store in Johto if we are going to use that.

fuws:
I think your comment elsewhere about lv100 being the evolution criterion for happiness is a reasonable one. We may not even need the `Time_stone`. But let's see what @Ishadijcks thinks.

RedSparr0w:
True, i can always remove the `Time_stone` references and just change `Eevee` to this:
```javascript
        "evolution": ["Vaporeon",    "Jolteon",       "Flareon",    ["Espeon", "Umbreon"]],
        "evoLevel":  ["Water_stone", "Thunder_stone", "Fire_stone", 100],
```
Which will work exactly the same way as `Tyrogue` but at level 100, will randomly evolve into either form.

PR #374, opened 2019-06-19
Typed egg in Johto

fuws:
Have typed eggs' Pokemon list depend on player's highest region. This makes all gen2 starters obtainable. #363 

Also moved some Pokemon from Route 28 to typed eggs. In the Pokemon Gold/Silver, these Pokemon were obtainable from routes in the Kanto map. 

Closes #376

### DIFF
--- a/src/scripts/breeding/BreedingHelper.ts
+++ b/src/scripts/breeding/BreedingHelper.ts
@@ -73,7 +73,8 @@ class BreedingHelper {
     }
 
     public static createTypedEgg(type: GameConstants.EggType): Egg {
-        let name = HatchList[type][Math.floor(Math.random() * HatchList[type].length)];
+        let region = player.highestRegion;
+        let name = HatchList[type][region][Math.floor(Math.random() * HatchList[type][region].length)];
         return BreedingHelper.createEgg(name, type);
     }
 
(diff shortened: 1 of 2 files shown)


src/scripts/breeding/BreedingHelper.ts
```diff
@@ -73,7 +73,8 @@ class BreedingHelper {
     }
 
     public static createTypedEgg(type: GameConstants.EggType): Egg {
-        let name = HatchList[type][Math.floor(Math.random() * HatchList[type].length)];
+        let region = player.highestRegion;
+        let name = HatchList[type][region][Math.floor(Math.random() * HatchList[type][region].length)];
```
RedSparr0w: or what about something like this, so you aren't locked to 1 region:
```javascript
const hatchable = HatchList[type].filter(p=>PokemonHelper.calcNativeRegion(p) <= player.highestRegion);
return hatchable[Math.floor(Math.random() * hatchable.length)];
```

fuws: I thought about this approach. It makes `HatchList` easier to structure. But issue here is that when player gets to higher regions, their chance of getting their desired Pokemon get much smaller. If we want to maintain the 4 Gen1 Pokemon in Fire egg, and at least add the Gen2 starter to the egg, then the Gen2 starter will have 20% probability, compared to 25% probability for the Gen1 starter. The probability gets worse the more higher-region Pokemon or more higher regions we add. This could be one of the ways that higher prestiges get more difficult - but it's a strange way I feel.

RedSparr0w: The reasons I think it would be okay are:
- By that point you will likely have an absolute ton of QP
- By this point in the game you have probably hatched a ton of eggs and are used to it.
- Eggs will be hatching slightly faster because of the routes adding more steps the further you are
- Kind of adds another reason to continue collecting QP
  - it currently feels pointless increasing the QP once you complete the gen1 dex and have run out of things to buy.

fuws: You still do need QP for these eggs. It's really a matter of whether players will feel fine with only a 12.5% chance of getting a starter from an egg. We'll have to look at how quest point gains work in Gen2 (because quests now only require actions done in Gen1 map).

Another solution could be to add 2 extra Pokemon per new region, and have 50% chance of getting a random Pokemon from player's current region, and 50% chance of getting a random Pokemon from previous regions. This logic will be a little convoluted, but will make defining `HatchList` easier.

Ishadijcks: Maybe the cleaner solution is to implement the egg as a different item for each region, that way we can also tweak the prices better.

RedSparr0w: We could use a function like below, which will weight newer generations with a higher percentage than older generations (if we keep it how you've coded it now, but only have that regions Pokemon in each sub array)
```javascript
const hatch_list = [['first gen'], ['second gen'], ['third gen']];
function getRandomHatchWeighted(){
    const hatchable = hatch_list.slice(0, 2 /* player highest gen + 1 */);
    const test = [];
    hatchable.forEach((possible_hatches, index)=>{
        for (i = 0;i < index + 1; i++){
            test.push(possible_hatches);
        }
    });
    return test[Math.floor(Math.random() * test.length)];
}
getRandomHatchWeighted();
```
This way when we reach Johto, Johto Pokemon will have a 66% chance of hatching, 33% Kanto
Then in Hoenn, Hoenn Pokemon will have a 50% chance of hatching, 33% Johto 16% Kanto
etc..

fuws: @RedSparr0w Your function would weight regions 1 - N according to their region number, correct? With this, the chance to roll the newest region is `2/(N+1)`, which may be too low for higher regions.

RedSparr0w: I don't believe it should be too much of an issue making the later regions slightly harder, you should have a ton of QP by that point to be able to afford the extra eggs.

Or maybe even just change it to be:
```javascript
    hatchable.forEach((possible_hatches, index)=>{
        const toAdd = test.length;
        for (i = 0; i < toAdd; i++){
            test.push(possible_hatches);
        }
    });
```

Which would make it so the newest generation always has a 50% chance of hatching

Kanto | Johto | Hoenn | Sinnoh
:--: | :--: | :--: | :--:
`100%` | `-` | `-` | `-`
`50%` | `50%` | `-` | `-`
`25%` | `25%` | `50%` | `-`
`12.5%` | `12.5%` | `25%` | `50%`

PR #376, opened 2019-06-27
Make separate typed eggs for Gen2

fuws:
Alternative to #374 . Closes #374 

Implement @Ishadijcks 's suggestion of making separate egg items for Gen2 eggs. Now use `{Type}_egg_{GenNumberInRomanNumeral}` naming convention for these eggs.

**Shop in New Bark Town:**
image

**Shop in Cerulean City:**
image


RedSparr0w:
I prefer #374 compared to this approach, Just feels cleaner IMO.

fuws:
@RedSparr0w Yes. This approach requires quite a few places to change. And these places need to be changed every time new generations get added.

PR #387, opened 2019-12-23
Fix semantic errors, Update TypeScript

RedSparr0w:
Fixup some semantic errors,
Update typescript to latest version.


Ishadijcks:
Seems to work fine

PR #407, opened 2019-12-27
Change entry logic to start an app class.

Ishadijcks:
**Please review carefully, breaking potential**

## Introduction
I have introduced an App class which contains just the game object for now. Later it will get some methods to manipulate games (`startPrestigeGame()`, `startChallenge()`, etc..).

This will be very useful later when we compile with a single entry point as well. 

We can now update the gamestate with `App.game.gameState = GameConstants.GameState.paused`

## Changes
- Move some UI methods from Game to GameController
- Refactor game from global to app property
- Change gamestate to proper knockout variable


src/scripts/App.ts
```diff
@@ -0,0 +1,31 @@
+class App {
+
+    static readonly debug = false;
+    static game: Game;
+
+    static start() {
+        if (!App.debug)
+            Object.freeze(GameConstants);
```
Aegyo: I know this works as is, but would be nice to have some { } around the !debug code

src/scripts/Game.ts
```diff
@@ -113,7 +46,7 @@ class Game {
         Underground.counter += GameConstants.TICK_TIME;
 
         GameHelper.counter += GameConstants.TICK_TIME;
-        switch (Game.gameState()) {
+        switch (App.game.gameState) {
```
Aegyo: Should this switch be `this.gameState` instead? I know it would have the same effect (unless we have multiple Games running...), but it seems strange that Game knows about App.

src/scripts/KeyBinds.ts
```diff
@@ -1,7 +1,7 @@
-$(document).ready(function() {
-    $(document).on("keydown", function(e) {
+$(document).ready(function () {
+    $(document).on("keydown", function (e) {
         let keyCode = e.keyCode;
-        if (Game.gameState() == GameConstants.GameState.safari) {
+        if (App.game.gameState == GameConstants.GameState.safari) {
```
Aegyo: Other gameState checks use `===`, may as well make it consistent while we are here. Same "issue" in the keyup handler.

src/scripts/KeyListeners.ts
```diff
@@ -1,7 +1,7 @@
 $(document).on("keydown", function (e) {
     let keyCode = e.keyCode;
 
-    if (Game.gameState() == GameConstants.GameState.dungeon) {
+    if (App.game.gameState == GameConstants.GameState.dungeon) {
```
Aegyo: Oh, there is a KeyListeners file. What is the difference between this and KeyBinds? (Other than doing the same thing for a different feature)
Maybe something to deal with Another Time.

src/scripts/App.ts
Ishadijcks: 100% agree, must have missed it while copy-pasting

src/scripts/Game.ts
Ishadijcks: Ah yes, a remainder from that I did the refactor in 2 steps, it was static before

src/scripts/KeyListeners.ts
Ishadijcks: No idea, seems both perfect to go into the GameController class

Ishadijcks:
Thanks! I have implemented all your feedback.

I'm just waiting for someone to run it and then we can merge it

PR #475, opened 2019-12-31
Refactor Badges into a separate feature

### LINKED ISSUE #476: Johto Badges gained not show in Badge Case [BUG]
All badges in Kanto are working OK.

Once you reach Johto, the first 2 gyms badges are OK.

From Goldenrod City, the badges gained after defeating the gym leaders not showing up in Badge Case.  From Plain Badge onwards......

Ishadijcks:
Closes #476 

A small refactor of the badge logic. Move it from the player to its own class

You can try it out here:
https://pokeclicker-development.herokuapp.com/

- Move Badges from GameConstants to BadgeCase namespace
- Change badgeList from observable array of strings to array of observable booleans


src/scripts/App.ts
```diff
@@ -11,8 +11,13 @@ class App {
         Preload.load(App.debug).then(function () {
             OakItemRunner.initialize();
             UndergroundItem.initialize();
-            App.game = new Game(new Breeding(), new Pokeballs(), new Wallet());
+            App.game = new Game(new Breeding(), new Pokeballs(), new Wallet(), new BadgeCase(16));
```
Aegyo: This 16 is something of a magic number, and happens to be incorrect as we count each  elite4 and champion as badges (see the numbers in BadgeCase.Badge enum, and issue #476)

src/scripts/App.ts
```diff
@@ -11,8 +11,13 @@ class App {
...
 
+            KeyItemHandler.initialize();
+            AchievementHandler.initialize();
+            player.gainKeyItem("Coin case", true);
+            player.gainKeyItem("Teachy tv", true);
+            player.gainKeyItem("Pokeball bag", true);
```
Aegyo: These three lines seem more suited for KeyItemHandler.initialize()

src/scripts/Player.ts
```diff
@@ -363,7 +347,7 @@ class Player {
         }
 
         for (let pokemon of this._caughtPokemonList()) {
-            if (pokemon.levelObservable() < (this.gymBadges.length + 2) * 10) {
+            if (pokemon.levelObservable() < (App.game.badgeCase.badgeCount() + 2) * 10) {
```
Aegyo: Was this intended to limit pokemon to level 20 before the first badge? I noticed a while ago that you could reach level 30 but never bothered to find out why. It is because we are now counting BadgeCase.Badge.None as a badge, so badgeCount before any real badges is already 1. Not sure if we want to change this line, or change badgeCount to make more sense.

src/scripts/App.ts
Ishadijcks: Agree, was more of a quickfix, I think they shouldn't exist anyway and just be part of the defaults in `KeyItems`. I'll fix it when refactoring KeyItems

src/scripts/Player.ts
Ishadijcks: `BadgeCase.Badge.None` is kind of counterintuitive anyway, I'll try to change it so it's no longer needed.

src/scripts/App.ts
Ishadijcks: Yeah excellent point. Later I was thinking of creating some sort of a `GameFactory` with methods like `createEasyPrestige()`, `createMediumPrestige()`, `createChallenge1()`, etc...

The reason for this parameter was so that the badge case doesn't display badges you can't get during this game, which I realize I still have to implement.

Maybe we should find a different way to store the elite four, adding them as badges is kind of iffy anyway.

Ishadijcks:
I decided to keep the E4 as it is, but the BadgeCase now takes a parameter for the highest badge in the current game

PR #479, opened 2020-01-01
Have Item class determine max number purchasable with player wallet

### LINKED ISSUE #477: [BUG] Selecting MAX when buying key items freezes the game
**Steps to Reproduce:**

1. Go to Pewter City shop
2. Select Dungeon Ticket
3. Click Max
4. Game freezes

**Expected Behavior:**
Max on a key item should not do anything, as only one can be bought.

fujnw:
Closes #477 

Closes #491 _(alternative PR)_


fujnw:
Good point. With the geometric series math, we need to build another special case for Pokeballs.

src/scripts/items/Item.ts
```diff
@@ -14,23 +14,19 @@ abstract class Item {
...
 
     maxAmount(cost: number): number {
         // Returns the maximum amount of items that can be bought
-        let amt = 0;
-
-        // TODO: Change to binary search if this becomes slow
-        while (this.totalPrice(amt) <= cost) {
-            amt++;
+        if (this.price() <= 0) {
+            return 1;
```
Ishadijcks: Why would this.price() be smaller than 0?

fujnw: It shouldn't. But just in case for whatever reason it is, the game won't crash. Feel free to remove this part.

PR #480, opened 2020-01-01
Proxy for ObservableArray

Aegyo:
Similar to the ArrayOfObservables, here is a curtain to hide ObservableArrays behind.

`new ObservableArrayProxy(["things","and","stuff])`

I have converted player._mineInventory to use this, and tested that the treasure tab updates when mining, selling, and daily dealing.

If you wish, you can compose this with ArrayOfObservables to get an observable array of observables. This is a consequence of careful planning, and not just dumb luck, I swear :smile: 

```
let test = new ObservableArrayProxy( new ArrayOfObservables( ["first","second","third"] ) )
test.push("egg")
test.join(", ")
-> "first, second, third, egg"
test.subscribe(console.log)
test.getSubscriptionsCount()
-> 1
ko.toJS(test)
-> (4) ["first", "second", "third", "egg"]
```


Ishadijcks:
Seems to work great!

PR #488, opened 2020-01-03
Refactor Party into its own feature

Ishadijcks:
Major refactor of the caughtPokemon system.
Play at https://pokeclicker-development.herokuapp.com/
- Move caughtPokemon and shinyPokemon to new `Party` feature.
- Gaining xp is no longer a responsibility for the player
- Trim the pokemon list to gen3 max
- Rewrite evolutions to be more scalable.
- Add `caughtStatus` for display of pokeball indicators
- Partydisplay now updates instantly.

Other
- Fix a bug in the settings template


src/scripts/Save.ts
```diff
@@ -203,19 +199,20 @@ class Save {
...
+        //     let shiny = parseInt(pokemon['shiny']);
+        //     let name = pokemon['name'];
+        //     if (shiny == 1 && player.caughtShinyList.indexOf(name) == -1) {
+        //         player.caughtShinyList().push(pokemon['name']);
+        //         converted.push(pokemon['name']);
+        //     }
+        // }
+        // if (converted.length > 0) {
+        //     Notifier.notify("You have gained the following shinies: " + converted, GameConstants.NotificationOption.success)
+        // }
```
RedSparr0w: Can we just comment this out as a code block?

Ishadijcks: Yeah I will be fixing this in a big save refactor which is up next

RedSparr0w:
Importing of saves seems to no longer work.
[PokeClickerSave_2020-01-06 19_56_07.txt](https://github.com/pokeclicker/pokeclicker/files/4027641/PokeClickerSave_2020-01-06.19_56_07.txt)

Ishadijcks:
All refactored features don't import correctly, I'm changing saves a bit later so that should make it work again

PR #490, opened 2020-01-04
Safari updates

Aegyo:
Some safari changes and code cleanup.

There is still more to be done here, switching some code to knockout and making this into a Feature, but that is mostly just code quality.
This PR can be merged now if we want safari fixes as soon as possible, or we can wait a day or two until I finish up the other things.


Ishadijcks:
I'm not in a hurry to get it merged into develop, but I'l deploy it so people can try it out :)

PR #492, opened 2020-01-04
Use eslint

RedSparr0w:
Use eslint.

Should make for tidier and more consistent code.

Some warnings will need to be fixed up later, unless we don't mind those rules being off.


src/scripts/Battle.ts
```diff
@@ -6,7 +6,7 @@
 class Battle {
     static enemyPokemon: KnockoutObservable<BattlePokemon> = ko.observable(null);
 
-    static counter: number = 0;
+    static counter = 0;
```
Ishadijcks: Weird that this is a fix

RedSparr0w: Its automatically assigned by typescript because we are assigning a number

Ishadijcks: Ah makes sense

RedSparr0w:
@Ishadijcks 
Any chance we can get this merged before we end up with a ton more conflicts?

Ishadijcks:
Is the of the errors automated?
I'd rather not create a ton of conflicts with the `refactor/party` branch

RedSparr0w:
You would just need to run `npm run eslint-fix` which should fix most common problems, there may be a couple that can’t automatically be fixed though.

package.json
```diff
@@ -6,7 +6,10 @@
   "scripts": {
     "start": "gulp",
     "website": "gulp website",
-    "test": "gulp scripts"
+    "test": "npm run ts-test && npm run eslint",
+    "ts-test": "gulp scripts",
+    "eslint": "eslint --config ./.eslintrc.js --ext ts ./src/scripts",
+    "eslint-fix": "eslint --config ./.eslintrc.js --ext ts --fix ./src/scripts"
```
Ishadijcks: Can you remove `test` and `ts-test` as their names are somewhat misleading, in travis we already run `tsc` so I don't think running `gulp-scripts` has any additional benefit

.travis.yml
```diff
@@ -9,4 +9,5 @@ install:
 script:
 #  Compile, fail on any error
 - tsc
-- npm test
+- npm run ts-test
```
Ishadijcks: Remove this line, see other comment

RedSparr0w:
```suggestion
```

package.json
RedSparr0w: I think it's probably worth keeping them for testing locally?

RedSparr0w: I would say it would be better to use `gulp scripts` overall because the output is nicer,
But it doesn't throw on semantic errors right?

RedSparr0w:
Unfortunately no way around it being this big 😅 
The main reason is because of all the `let` -> `const`.

package.json
Ishadijcks: What if we change `test` to something like `validate-all` so that is nice to set as a pre-commit hook.
```
    "validate-all": "gulp scripts && tsc && npm run eslint",
```

Ishadijcks:
Is there more stuff like the `travis.yml` and the `package.json` that you manually updated? I have scrolled through and couldn't find any but you never know

RedSparr0w:
Only thing manually updated was the `travis.yml`, `package.json`, `.eslintrc.js` and these 2 commits here: 
https://github.com/pokeclicker/pokeclicker/pull/492/commits/f66624cf1e2aca824917d146809f46d8983042ff
https://github.com/pokeclicker/pokeclicker/pull/492/commits/c810cbfb2097c08c7411421ce1331c82c6026409

Which both had to be done manually as those ~57 errors couldn't be fixed automatically.

RedSparr0w:
And a couple things when i merged develop branch last time too, as it didn't quite automatically merge, the conflicts were a bit strange, but should all be just eslinting in those files now, as i reset the files and relinted them.

package.json
RedSparr0w: `tsc` doesn't seem to be something that is automatically installed? doesn't work on my system.

PR #493, opened 2020-01-04
Log oak item use statistic even at max level

### LINKED ISSUE #489: [BUG] OakItem quests don't progress is the oakItem is at max level
Quests that require you 'get the benefit' from an Oak Item don't tick up if the Oak Item is already at max level. 

Check the `OakItem.use()` method to see what goes wrong and how you can fix it :)

fujnw:
Closes #489

### DIFF
--- a/src/scripts/oakItems/OakItem.ts
+++ b/src/scripts/oakItems/OakItem.ts
@@ -23,12 +23,13 @@ class OakItem extends ExpUpgrade {
     }
 
     use(exp: number = this.expGain) {
-        if (this.isMaxLevel() || !this.isActive) {
+        if (!this.isActive) {
             return;
         }
-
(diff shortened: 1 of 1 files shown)


src/scripts/oakItems/OakItem.ts
```diff
@@ -23,12 +23,13 @@ class OakItem extends ExpUpgrade {
     }
 
     use(exp: number = this.expGain) {
-        if (this.isMaxLevel() || !this.isActive) {
+        if (!this.isActive) {
             return;
         }
-
+        if (!this.isMaxLevel()) {
+            this.gainExp(exp);;
```
Ishadijcks:
```suggestion
            this.gainExp(exp);
```

PR #486, opened 2020-01-03
Refactor shards into a Feature

fujnw:
Shard wallet and upgrades are both housed inside `Shards` feature for now. The fact that shards need a separate set of currencies makes it hard to just extend the existing Wallet and Upgrade features. I didn't want the added complexity of creating 18 new currencies and 72 upgrade objects just for this.

PR #497, opened 2020-01-05
Refactor shards into its own Feature

fujnw:
Shard wallet and upgrades are both housed inside `Shards` feature for now. The fact that shards need a separate set of currencies makes it hard to just extend the existing Wallet and Upgrade features. I didn't want the added complexity of creating 18 new currencies and 72 upgrade objects just for this.

**Moved from #486** 

**Note**: The shard feature requires Shard Case to unlock. There is currently no way of obtaining Shard Case in the game.


Ishadijcks:
You can run all the commands travis is running locally as well. That way your builds don't have to fail

RedSparr0w:
You should just be able to run `npm run eslint-fix` and it should pass.
You may need to run `npm install` first though.

PR #499, opened 2020-01-05
Fix closing the breeding modal putting you back to route 5

Ishadijcks:


RedSparr0w:
I thought this was by design rather than an issue?
Same thing happens with the underground and farm.

Ishadijcks:
Yeah it was but it is somewhat undeserved as your eggs are there as well. So it makes sense to access it from the breeding spot that has its own ui on the main screen.

Underground and Farm are fine as their feature only lives in their modal

PR #500, opened 2020-01-05
Normalize OakItem progress by xp gained per action

Ishadijcks:


src/scripts/oakItems/OakItem.ts
```diff
@@ -56,6 +56,11 @@ class OakItem extends ExpUpgrade {
     }
 
     // Knockout getters/setters
+    get progressString(): string {
+        let nextLevelExp = this.level === 0 ? this.expList[this.level] : this.expList[this.level] - this.expList[this.level - 1];
```
RedSparr0w: Can we change this to a `const`

Ishadijcks: Vietnam flashbacks to eslint

PR #501, opened 2020-01-05
Update changelog to reflect previous PRs

RedSparr0w:
Update the changelog with previous commits


RedSparr0w:
Note that I have also added a comment between version to easily see where a new version starts/begins.
I think adding a date next to a version "release" will also help anyone when deciding which version to add stuff to in the changelog.
Although we aren't currently "releasing" it is still probably worth bumping it very occasionally.

PR #507, opened 2020-01-06
Fix saves importing/exporting on refactor/party branch

RedSparr0w:
Fix saves importing/exporting on refactor/party branch


RedSparr0w:
Previously downloaded saves will not import correctly, but any new saves should.

PR #518, opened 2020-01-08
Reduce eslint warnings [27]

RedSparr0w:
reduce the number of warnings in eslint `37 → 27`

### DIFF
--- a/src/scripts/Notifier.ts
+++ b/src/scripts/Notifier.ts
@@ -1,3 +1,4 @@
+/* eslint @typescript-eslint/camelcase: off */
 class Notifier {
 
     public static notify(message: string, type: GameConstants.NotificationOption){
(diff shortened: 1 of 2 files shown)


src/scripts/Notifier.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint @typescript-eslint/camelcase: off */
```
Ishadijcks: Why turn this off?

RedSparr0w: I don't think we can control the names of those parameters for notifications,

```javascript
            allow_dismiss: true,
            newest_on_top: true,
            z_index: 1050,
            url_target: '_blank',
            mouse_over: null,
            icon_type: 'class',
```

It's still turned on for every other file though, just not this one.

Ishadijcks: Ah fair point, but does it actually trigger there? You'd expect for it to only trip at the definitions

RedSparr0w: You would think so, but unfortunately it does detect it.

PR #519, opened 2020-01-08
Fix a bug where only the first stone evolution would be triggered

Ishadijcks:


fujnw:
The random evolution works as intended. 

There're some UI/notification oddities though. The Item panel will show the pokeball beside Eevee only when both its Time_stone evolutions are obtained, which is fine, as long as we're aware of this behavior. 

What's a little more strange, is if you have a ton of Time_stones and start hunting for shiny. Using many stones as once will stop when encountering a shiny. So the first time one of the evolution is shiny, you stop using stones and get a notification saying you caught a shiny - all is good. But the second time you catch a shiny, if it happens to be the one you already caught, you will still stop using stones. In this case, you'll get notified that you used X stones, but won't get notified on why it stopped. This will repeat until you caught the shiny of the other evolution. 

This is not a big deal. And seeing that Time stone isn't available in Underground, players likely won't have a ton of them. But we should be aware of this.

RedSparr0w:
>There're some UI/notification oddities though. The Item panel will show the pokeball beside Eevee only when both its Time_stone evolutions are obtained, which is fine, as long as we're aware of this behavior.

It was set up this way so the user _should_ be more aware that they still haven't obtained all the evolutions, although I think this should only affect Eevee anyway.

RedSparr0w:
Does Tyrogue evolve correctly?
He has 3 possible evolutions via leveling.

PR #520, opened 2020-01-08
Fix obtaining base forms when breeding

### LINKED ISSUE #465: When hatching Dragonite or Dragonair for first time, you obtain a Dratini
I have bought a single Dragon egg.  After I breed it Dragonite pops out.  At the same time a message stating that you have found a Dratini nearby.

### LINKED ISSUE #515: [BUG] Breeding Pokemon no longer gives you their baby form
**Description:**

When breeding Pokemon such as Pikachu, you will no longer "find" the baby form once hatched.

**Steps to Reproduce:**

1. Reach Johto region
2. Breed Pikachu
3. Hatch Pikachu
4. No Pichu

**Expected Behavior:**

1. Reach Johto region
2. Breed Pikachu
3. Hatch Pikachu
4. Get a Pichu

RedSparr0w:
closes #515 
closes #465 

- Fixes obtaining babies when breeding Pokemon
- Fixes obtaining next gen Pokemon via breeding and leveling back up.
- Fixes obtaining evolved forms when you already have them.

### DIFF
--- a/src/scripts/breeding/Breeding.ts
+++ b/src/scripts/breeding/Breeding.ts
@@ -237,16 +237,15 @@ class Breeding implements Feature {
     }
 
     public calculateBaseForm(pokemonName: string): string {
+        const devolution = pokemonDevolutionMap[pokemonName];
         // Base form of Pokemon depends on which regions players unlocked
-        if (!(pokemonName in pokemonDevolutionMap)) {
+        if (!devolution || PokemonHelper.calcNativeRegion(devolution) > player.highestRegion()) {
             // No devolutions at all
-            return pokemonName;
-        } else if (PokemonHelper.calcNativeRegion(pokemonDevolutionMap[pokemonName]) > player.highestRegion()) {
             // No further devolutions in current unlocked regions
             return pokemonName;
         } else {
             // Recurse onto its devolution
-            return this.calculateBaseForm(pokemonDevolutionMap[pokemonName]);
+            return this.calculateBaseForm(devolution);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/scripts/breeding/Breeding.ts
```diff
@@ -237,16 +237,15 @@ class Breeding implements Feature {
...
+        if (!devolution || PokemonHelper.calcNativeRegion(devolution) > player.highestRegion()) {
             // No devolutions at all
-            return pokemonName;
-        } else if (PokemonHelper.calcNativeRegion(pokemonDevolutionMap[pokemonName]) > player.highestRegion()) {
             // No further devolutions in current unlocked regions
             return pokemonName;
         } else {
             // Recurse onto its devolution
-            return this.calculateBaseForm(pokemonDevolutionMap[pokemonName]);
+            return this.calculateBaseForm(devolution);
```
Ishadijcks: Why is this recursively?

if `A -> B -> C` it would make sense to only gain `A` while breeding `B` and not while breeding `C`. Although then you would also get `B` while breeding `C` and that doesn't make sense

Ishadijcks: Wait what is preventing this from triggering on all evolved pokemons in its current form?

RedSparr0w: I think it makes more sense that you would only get `A` while breeding, due to being the base form.
At least that’s how it works in the main series games from memory.

RedSparr0w: The breeding map was empty due to the evolutions refactoring.

Ishadijcks: What if we add a flag to pokémon that can only be acquired this way, and only put those in the devolutionmap

RedSparr0w: Done, can now only receive babies via breeding/devolution map.

PR #550, opened 2020-01-13
Fixed the Pokedex Filtering for Type Search

Ash031:


Ishadijcks:
The second filter doesn't seem to work correctly.

If I select `Normal/Flying` I only want pokemons with that exact type

src/scripts/wildBattle/PokemonsPerRoute.ts
```diff
@@ -26,7 +26,7 @@ const pokemonsPerRoute = {
             headbutt: [],
         },
         5: {
-            land: ['Pidgey', 'Pidgeotto', 'Rattata', 'Jigglypuff', 'Oddish', 'Meowth', 'Mankey', 'Abra', 'Bellsprout'],
+            land: ['Pidgey', 'Pidgeotto', 'Rattata', 'Jigglypuff', 'Oddish', '!(((owth', 'Mankey', 'Abra', 'Bellsprout'],
```
Ishadijcks: ?

Ash031: Probably hit the wrong stuff when editing the parenthesis as I was browsing throught the repo, will fix.

Ishadijcks: Good :)

Ash031: It's fixed

Ishadijcks:
Why are there things changed in the `package-lock`? Did you run `npm update`? Please remove those

Ishadijcks:
Otherwise the logic seems to work correctly :)

Ash031:
> Why are there things changed in the `package-lock`? Did you run `npm update`? Please remove those

I haven't run anything other than npm install and npm start, probably some minor automatic changes from the package manager?

Ishadijcks:
Hmm odd, that shouldn't happen I think

PR #553, opened 2020-01-14
Fix dungeon double notification

### LINKED ISSUE #509: [BUG] It is possible to complete and fail a dungeon at the same time
**Description:**
When the timer of a dungeon runs out while you are catching the boss, you can fail and complete a dungeon at the same time.

**Expected Behavior:**

You should either win or lose the dungeon. I think that if you manage to defeat it before the dungeon, you should still win, even if capturing takes longer.

**Steps to Reproduce:**
Make sure you will throw a ball at the boss
Open dungeon
Do the thing
Time the boss fight perfectly to have the capture animation appear, but not finish before running out of time

**Screenshots:**

https://media.discordapp.net/attachments/553647470224343040/664184737857798145/unknown.png

fujnw:
I add two new observables to track dungeon status: `defeatedBoss` and `dungeonFinished`. The former ensures that when timer goes out, the correct notification is sent. The latter ensures that only one such notification is sent.

Closes #509 

Updated dungeon behavior: If the boss catching animation goes overtime, you complete the dungeon successfully but the catch fails (you move to town, where `enemyPokemon` is set to null).


RedSparr0w:
Would this be the expected behavior?

I would assume you would still have the chance to capture the Pokemon, rather than instant fail?

PR #566, opened 2020-02-27
Update master to current dev testing version

RedSparr0w:
Push latest changes to master/live version, host from GitHub so we don't have to worry about using up the free hours on Heroku.


RedSparr0w:
Ton of conflicts would probably be easier to do a `git reset --hard redsparr0w/testing-live`

RedSparr0w:
Close as hosting dev version via Heroku.

Still may be worth looking at using GitHub pages _(or something else)_ for hosting so it won't run out at the end of each month.

PR #568, opened 2020-03-03
Fix importing old save shiny Pokemon

RedSparr0w:
Attached save below for testing:
[pokeclicker_2020-03-03T23_28_14.819Z.txt](https://github.com/pokeclicker/pokeclicker/files/4284275/pokeclicker_2020-03-03T23_28_14.819Z.txt)

### DIFF
--- a/src/scripts/Save.ts
+++ b/src/scripts/Save.ts
@@ -183,7 +183,7 @@ class Save {
             const p = JSON.parse(json);
             Save.convertShinies(p.caughtPokemonList);
             $('#saveModal').modal('hide');
-        } catch(e) {
+        } catch (e) {
             Notifier.notify('Invalid save data.', GameConstants.NotificationOption.danger);
         }
     }
--- a/src/scripts/Save.ts
+++ b/src/scripts/Save.ts
@@ -184,20 +184,21 @@ class Save {
         $('#saveModal').modal('hide');
     }
 
-    // TODO (@Isha) reimplement
-    public static convertShinies(list: Array<string>) {
-        // let converted = [];
-        // for (let pokemon of list) {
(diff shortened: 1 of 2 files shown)


src/scripts/Save.ts
```diff
@@ -184,20 +184,21 @@ class Save {
...
-        //     }
-        // }
-        // if (converted.length > 0) {
-        //     Notifier.notify("You have gained the following shinies: " + converted, GameConstants.NotificationOption.success)
-        // }
+    public static convertShinies(list: Array<any>) {
+        const converted = [];
+        list = list.filter(p => p.shiny);
+        for (const pokemon of list) {
+            const id = +pokemon.id;
```
Ishadijcks: `+`?

src/scripts/Save.ts
```diff
@@ -184,20 +184,21 @@ class Save {
...
-        //         converted.push(pokemon['name']);
-        //     }
-        // }
-        // if (converted.length > 0) {
-        //     Notifier.notify("You have gained the following shinies: " + converted, GameConstants.NotificationOption.success)
-        // }
+    public static convertShinies(list: Array<any>) {
+        const converted = [];
+        list = list.filter(p => p.shiny);
+        for (const pokemon of list) {
```
Ishadijcks: Does this work with the any type?

src/scripts/Save.ts
RedSparr0w: Just insuring its a number

src/scripts/Save.ts
RedSparr0w: Sure does, otherwise typescript throws an error expecting that everything in the array is a string.

PR #571, opened 2020-03-10
Add Logbook to the game

### LINKED ISSUE #447: logbook?
ok so, it is annoying that you don't know what happens when you're afk, so why not make a logbook?
it may look like a spiral textbook, and be accessible from start menu 
and a simple way of doing it is to make it display the 20 last not berry notifications !
so we will be able to know what shiny we caught !

RedSparr0w:
Closes #447 

This PR adds a log book to the game, which currently only tracks Pokemon encounters.
- Uncaught Pokemon encounter.
- Capture/Escape previously uncaught Pokemon.
- Shiny Pokemon encounter.
- Capture/Escape shiny Pokemon.

The log book will only keep up to 1000 logs, Only 100 of which will be kept with the save file or on refresh.

image


Ishadijcks:
I like it!

Maybe we should refine the purpose a bit. Is it for everything that has happened or do we only want to log important events (e.g. Oak items unlocks/upgrades, achievements, shinies, badges gained, first completions of dungons/roads)?

RedSparr0w:
At the moment it's only tracking new/shiny Pokemon events.

We could always add more stuff later on:
- Oak item unlocked/upgradable/upgraded
- Key item obtained
- Achievement completed
- Quest level increased
- Quest completed
- Farm ready/harvested/planted
- Underground completed/energy full
- Underground trades
- Shop purchases

Issue #199, opened 2017-11-12
Quests finishing


dennism1997:
- [ ] balance the rewards
- [ ] Big reward for completing them all
- [x] Beginner quests that work as a tutorial

PR #574, opened 2020-03-11
Balance Dungeon Quest Rewards

Aegyo:
Fixes a feature where dungeon quests were estimating that you only take 0.0034 attacks per pokemon, therefore your quest reward should be similarly low.

Also accounts for time spent farming the dungeon tokens to actually enter and complete the dungeons, rewarded at afk levels.


Aegyo:
Referencing #199 so we can see status of this pr from that issue

PR #583, opened 2020-03-15
[CAN MERGE NOW]Rebalance Pokemon attack

anapple96:
The rework is here.....for the Kanto and Johto pokemons, all based around this calculator : https://jscalc.io/calc/ErtO3M6wsibdGOvW

Thanks Mega Rayquaza, Eru, Neon Toxel and Beespoon on Discord for helping me filling in the datas.(Especially Eru, this man pulled some serious weighs)
We have fully reworked the whole Johto dex, only a few Kanto mon away!

PR #586, opened 2020-03-17
WIP: Update Pokemon attack formula

anapple96:
Well, this's basically the same as #583 and #585 .But, it also has Mega...at least a few of them, more will come in the future.(Megas are unobtainable, just the stats are here)


anapple96:
well, technically, a few mega's stat is in the pr so...

anapple96:
Not just an attack formula change

src/scripts/pokemons/PokemonList.ts
```diff
@@ -13370,7 +13370,7 @@ const pokemonList: {
             'type': [PokemonType.Grass],
             'eggCycles': 20,
             'levelType': LevelType.mediumslow,
-            'exp': 230,
+            'exp': 236,
```
RedSparr0w: Could you please revert this change

src/scripts/pokemons/PokemonList.ts
```diff
@@ -14172,12 +14172,49 @@ const pokemonList: {
...
+                'hitpoints': 78,
+                'attack': 130,
+                'specialAttack': 130,
+                'defense': 111,
+                'specialDefense': 85,
+                'speed': 100,
+            },
+
+
+        },
```
RedSparr0w: Can we also remove the mega evolutions from this PR

src/scripts/pokemons/PokemonList.ts
```diff
@@ -14172,12 +14172,49 @@ const pokemonList: {
...
     ];
 
 
 // TODO move to its own initialize method that gets called on game start.
 pokemonList.forEach(p => {
-    p.attack = Math.max(...Object.values(p.base));
+
+    p.attack = Math.round(Math.max(Math.round(Math.max(1, Math.floor(Math.sqrt((2 * p.base.hitpoints)) * (2 * Math.round(Math.sqrt(p.base.attack * p.base.specialAttack) + Math.sqrt(p.base.speed)) * Math.sqrt(2 * Math.round(Math.sqrt(p.base.defense * p.base.specialDefense) + Math.sqrt(p.base.speed))))))) / 300));
```
RedSparr0w: Could you please reduce the complexity of this formula by splitting it up into a few extra variables.
And there are a couple redundant rounding/max/floors in here too.
I think we should also make the minimum attack be 10 as 1 is just far too low of an attack amount.

If you need any help with these just @ me on the discord chat.

src/scripts/pokemons/PokemonList.ts
anapple96: fixed!

src/scripts/pokemons/PokemonList.ts
anapple96: done!

src/scripts/pokemons/PokemonList.ts
anapple96: gone, reduced to atoms

anapple96:
This is merge-ready!

anapple96:
Yeah, i think /250 is a good idea, as early game is slightly too hard with /300

anapple96:
Hmm, /225 seems like the best one

RedSparr0w:
Thanks for the PR 🎉

PR #587, opened 2020-03-17
Add keyboard shortcuts

RedSparr0w:
Fixup space to start Gym/Dungeons.
Added `-` and `+` to change current route.
Added check to make sure route exist.

### DIFF
--- a/src/scripts/GameController.ts
+++ b/src/scripts/GameController.ts
@@ -103,9 +103,9 @@ class GameController {
                 e.preventDefault();
             } else if (App.game.gameState === GameConstants.GameState.town) {
                 if (keyCode == 32) { // space
-                    if (player.town().gym) {
+                    if (player.town().gym()) {
                         GymRunner.startGym(player.town().gym());
-                    } else if (player.town() instanceof DungeonTown) {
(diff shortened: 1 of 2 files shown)


Ishadijcks:
Nice, good idea.
Maybe we can think of something to collect these shortcuts somewhere to avoid being unable to find them later. I've had multiple bugs where something didn't work because an eventlistener in a file somewhere was handling the reguest.

Maybe some mapping where Keycombination + GameState => callback.

RedSparr0w:
> I've had multiple bugs where something didn't work because an eventlistener in a file somewhere was handling the reguest.

I've Fixed the dungeons which were capturing **every** key.

I've also changed the larger ones to a switch statement to make it easier to spot which keys are being captured.

I'm not sure if you wan't to move it to be like the Safari Zone which has the movement handled in the class:
```js

            const keyCode = e.keyCode;
            if (App.game.gameState === GameConstants.GameState.safari) {
                const dir = GameConstants.KeyToDirection[keyCode];
                if (dir) {
                    e.preventDefault();
                    Safari.move(dir);
                }
                if (keyCode == 32) { // space
                    e.preventDefault();
                }
            }
```

PR #599, opened 2020-03-20
Dual type egg images

Aegyo:
Most colors were chosen by a pixel color in the old egg images. Some of these combinations didn't look great when scaled up, so I have replaced a few with the light and dark options from https://bulbapedia.bulbagarden.net/wiki/Category:Type_color_templates

Dragon looks ok by itself, but combinations don't seem great (eg, Dragonite - Dragon/Flying). Could use some input on what people think a good alternative for this one could be.

Also, there is a slight corner on each side of the egg. I'll attempt to fix this before merging, but it doesn't seem too noticeable even when you know it is there.


Aegyo:
This may be useful for testing different colors - https://jsfiddle.net/jessica_scoltock/913hg28u/169/

src/scripts/breeding/Egg.ts
```diff
@@ -8,6 +8,8 @@ class Egg implements Saveable {
     shinySteps: number;
     pokemon: string;
     type: GameConstants.EggType;
```
Ishadijcks: Should this regular type still be used?

Aegyo: Unless we stop showing fossils as their fossil image, we need this type to differentiate between normal breeding and EggType.Fossil

This enum is also used when choosing and setting up pokemon from shop bought eggs.

It seems like this enum is two things at once: shopEggType and eggDisplayType. 
We should probably separate it into these two things and just pass the displayType into the constructor (Pokemon or Fossil right now, but potentially something else in the future), and we can rename the property to reflect this.

src/scripts/breeding/Egg.ts
```diff
@@ -47,9 +63,7 @@ class Egg implements Saveable {
         this.pokemon = json['pokemon'];
         this.type = json['type'];
         this.notified = json['boolean'];
```
RedSparr0w: I know this isn't a part of your PR, but shouldn't this be:
```suggestion
        this.notified = json['notified'];
```

PR #601, opened 2020-03-20
Move pokemonMap, pokemonMapId to a proxy object

RedSparr0w:
Not sure if this will provide any performance improvements or make it worse, haven't had the chance to test it out.

But I think this approach would save on memory usage overall?


Ishadijcks:
Not sure, why would it be better?

RedSparr0w:
So I did a quick test, and it seems to take a little longer to access the objects.
From ID 1 → 1000 overall _(note: only ~800 Pokemon exist in the object)_
```
pokemonMapId: 0.9ms
pomeonMap (proxy): 15ms
```
It did save on some memory, as we aren't creating 2 extra objects, although it is kind of negligible.
It does make it slightly easier for accessing Pokemon _(not having to worry about accessing by id or name)_, but that's about it.

PR #612, opened 2020-03-21
Fix log book new encounters

### LINKED ISSUE #610: [BUG] Log Book adds first encounter to log as NEW encounter
**Info**
 - OS: `Windows 10`
 - Browser: `Chrome`
 - Version: `Latest Develop Branch`

**Description:**
Log Book adds first encounter to log as NEW encounter.

**Steps to Reproduce:**
1. Open Log Book
2. Note current logs
3. Refresh page
4. New log of first encountered Pokemon as NEW encounter

**Expected Behavior:**
Only previously unencountered Pokemon should be added as NEW encounter

**Screenshots:**
image

RedSparr0w:
closes #610

I've also moved all the Log Book stuff into it's own folder to keep it more organised.


src/scripts/Game.ts
```diff
@@ -84,6 +84,7 @@ class Game {
         this.keyItems.initialize();
         this.oakItems.initialize();
         this.farming.initialize();
+        this.load();
 
         // TODO refactor to proper initialization methods
         Battle.generateNewEnemy();
```
RedSparr0w: This was the line causing the problem for the log book.

Ishadijcks: That's weird, do you have any idea why?

RedSparr0w: It’s because the party hasn’t loaded yet, so every Pokémon is counted as uncaught.

PR #613, opened 2020-03-21
Fix background not scaling correctly

### LINKED ISSUE #524: [BUG] Background Stretching
**Info:** Windows 10, Chrome (suspected _all_ browsers)

**Description:** When any of the Drop-Down UI elements on the left (Pokeballs, Battle Items, Pokemon List) are collectively low enough that they enable vertical scrolling, the background splash image is forcibly resized both downwards _and_ horizontally (both left and right), rather than just adding more of the background's bottom color (Lime Green-ish) to prevent stretching.

This is briefly magnified when using a battle item in addition to the expanded Left UI, as the Battle Item timers slightly increase the height of the Battle Items UI element.

This issue _also_ occurs if the Middle UI (Location, Map) is stretched beyond a certain point, such as the Viridian Forest or Ruins Of Alph pictures. However, Locations that stretch the Middle UI by lesser amounts, such as Power Plant and Victory Road, do _not_ cause this to happen, so... 

It seems this is only caused when a sufficient amount of vertical space is taken up, suggesting that the Background Image is taller than is typically viewable and slightly more vertical scrolling space is available, but when the UI reaches further downwards than expected, the image stretches downwards and away to encompass the increased size of the UI, rather than filling the additional space with the Lime Green-ish colour on the bottom of the background image.

Additionally, if the image is already stretched to a certain point, throwing a Rock in the Safari Zone briefly causes further expansion (out of the three animations that appear above the pokemon, the yellow first hitmarker causes this, but the red angry tics do not, so the additional stretching reverts as soon as the yellow hitmarker fades).

I will include several image caps with text to better explain this issue below

This issue does _not_ occur if the right (Oak Items, Quest, Hatchery) group of UI elements enables vertical scrolling. I am assuming that this is because, while they do take up a greater amount of vertical space, they aren't collectively tall enough at full height to stretch the background. _However_, this might change if the user has multiple quests active at once. I am unable to test this at present.

Worth noting, the introduction of the vertical scrollbar _does_ compress the image diagonally upwards and to the left very slightly. It is barely visible, but jarring if it happens repeatedly (such as Location images that _just_ push the Map below the center of the screen, but not enough to cause the Background Stretching issue), and rather ruins the immersion.

**Steps to Reproduce:** See above.

**Expected Behavior:** When additional vertical space is taken up, the Lime Green-ish colour that takes up the bottom of the background picture should expand downwards or already be present below the Background Picture/a certain height.

**Possible Quick & Easy Solution:** Set the Page Background to the same hue of Lime Green-ish as the bottom of the Background Image and disable Background Image Resizing/limit Background Image Resizing to not take into account the Vertical Scrollbar/Scrolling. This way, vertical height is irrelevant, as it will lead to an infinite expanse of the Lime Green-ish colour (which won't look all that good if it goes too far down, but is fine at this level of UI element height).

**Screenshots:** This is about as clear as I can make it. Hope this helps visualise the problem:
Normal Background
Hatchery Extended
Pokemon List Extended
Battle Item Timer
Location Extended
Safari Right
Safari Left

### LINKED ISSUE #577: [BUG] Background Y-axis repeat issue.
**Info**
 - OS: Windows 10
 - Browser Chrome
 - Version 0.9.2

**Description:**
When you have a High res monitor, for example a 4K screen at 100% text size, and zoom to 150% in chrome, the background repeats on the Y-axis at it looks very bad.

**Steps to Reproduce:**

1. Have a high resolution screen for example 4K and have 100% windows text size.
2. Go to Chrome -> PokeClicker
3. Ctrl+Scroll to zoom to 150%
4. See error in the bottom where the sky of the background repeats.

**Expected Behavior:**
The background should probably stretch to fit any resolution.

**Screenshots:**
ISSUE:
PokeclickerIssueBGrepeat

With css FIXED:
PokeclickerIssueBGrepeatFIXED

**Additional Information:**
Here is the css code I fixed it with locally (can't commit it because I'm not good enough at github):

body {
    background-color: #7dad71;
    background-repeat: repeat-x;
    background-size: cover;
    overflow-x: hidden ;
}

Ishadijcks:
Closes #577
Closes #524 

Had to add it to the preloader

### DIFF
--- a/src/scripts/utilities/Preload.ts
+++ b/src/scripts/utilities/Preload.ts
@@ -63,6 +63,9 @@ class Preload {
                 const loader = $('#loader');
                 loader.css('background', 'url(assets/images/background.png) top');
                 loader.css('background-size', 'cover');
+                loader.css('overflow-x', 'hidden');
+                loader.css('background-color', '#7dad71');
+                loader.css('background-repeat', 'repeat-x');
                 resolve();
             };
             img.src = 'assets/images/background.png';
@@ -77,6 +80,9 @@ class Preload {
                 const body = $('body');
                 body.css('background', 'url(assets/images/background.png) top');
                 body.css('background-size', 'cover');
+                body.css('overflow-x', 'hidden');
+                body.css('background-color', '#7dad71');
+                body.css('background-repeat', 'repeat-x');
                 resolve();
             };
             img.src = 'assets/images/background.png';
--- a/src/styles/main.less
+++ b/src/styles/main.less
@@ -88,10 +88,6 @@ button {
   cursor: pointer;
 }
 
-body {
-    overflow-x: hidden;
-}
-
 body:hover {
   cursor: default;
 }


src/scripts/utilities/Preload.ts
```diff
@@ -63,6 +63,10 @@ class Preload {
                 const loader = $('#loader');
                 loader.css('background', 'url(assets/images/background.png) top');
                 loader.css('background-size', 'cover');
+                loader.css('overflow-x', 'hidden');
+                loader.css('background-color', '#7dad71');
+                loader.css('background-repeat', 'repeat-x');
+
```
Ishadijcks:
```suggestion
```

PR #614, opened 2020-03-21
Egg/Breeding refactor

RedSparr0w:
Moved the `hatch()` method to `Egg`.
Moved the `addSteps()` method to `Egg`.
Added a check for if the egg can hatch yet. _(can currently just call hatch on any egg instantly)_
Moved `EggType` out of GameConstants.
Added `EggType.None` for empty eggs.

_Note: This should work seamlessly with peoples current saves._


src/scripts/breeding/Breeding.ts
```diff
@@ -120,24 +120,8 @@ class Breeding implements Feature {
         amount *= App.game.oakItems.calculateBonus(OakItems.OakItem.Blaze_Cassette);
 
         amount = Math.round(amount);
-        for (const obj of this._eggList) {
-            // TODO(@Isha) fix this properly.
```
Ishadijcks: :sweat_smile:

src/scripts/breeding/Breeding.ts
```diff
@@ -20,27 +20,27 @@ class Breeding implements Feature {
     }
 
     initialize(): void {
-        this.hatchList[GameConstants.EggType.Fire] = [
+        this.hatchList[EggType.Fire] = [
             ['Charmander', 'Vulpix', 'Growlithe', 'Ponyta'],
             ['Cyndaquil', 'Slugma', 'Houndour', 'Magby'],
```
Ishadijcks: With the `refactor/routes`, do you have any idea how we can move this multiple region stuff to world? It seems to make way more sense here but that would leave region-specific data fractured.

RedSparr0w: We could change `hatchList = {}`
And when we initialise the region, just add it to `Breeding.hatchList[region] = this.hatchList`?

Feel like there is probably a better way, but at least that would keep all region data in 1 central location.

PR #618, opened 2020-03-24
Move eggs to lowest available egg slot.

### LINKED ISSUE #549: Eggs move to open slot when egg is hatched [Request]
Not sure what else to add that doesn't just repeat the title but, it would be nice for eggs to move to the newly opened slot when you hatch an egg because the hatch button is slightly small.

RedSparr0w:
When you hatch an egg, any eggs in "higher" slots will be moved down.

Closes #549

### DIFF
--- a/src/scripts/breeding/Breeding.ts
+++ b/src/scripts/breeding/Breeding.ts
@@ -125,7 +125,7 @@ class Breeding implements Feature {
         }
     }
 
-    public gainPokemonEgg(pokemon: PartyPokemon) {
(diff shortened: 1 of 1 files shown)


Ishadijcks:
I still don't understand why, but the implementation seems correct

anapple96:
Based on my understanding, if the hatchery has 4 hatchable eggs, the player can just spam click the hatch button of the first slot to hatch all of them, which reduces the time

PR #619, opened 2020-03-24
Capping egg slot to 4

anapple96:
Wonky stuffs happened.


src/scripts/breeding/Breeding.ts
```diff
@@ -223,7 +223,11 @@ class Breeding implements Feature {
     }
 
     public gainEggSlot() {
-        this.eggSlots += 1;
+        if (this.eggSlots < 4){
+            this.eggSlots += 1;
+        }
+    } else {
+        this.eggSlots += 0;
```
Ishadijcks: Redundant

anapple96: redundant?

anapple96:
This is more of a safety measure, as getting more than 4 egg slots is impossible without actually messing with `App.game.breeding.gainEggSlot()`

src/scripts/breeding/Breeding.ts
Ishadijcks: Why increase a variable y 0, it is not needed

src/scripts/breeding/Breeding.ts
```diff
@@ -223,7 +223,14 @@ class Breeding implements Feature {
     }
 
     public gainEggSlot() {
-        this.eggSlots += 1;
+        if (this.eggSlots < 4){
```
Ishadijcks: Don't compare with the magic number `4`. Use `this.eggslots.length` instead.

anapple96: I don't actually see how will this change anything??

Ishadijcks: If we leave the 4 there, and decide to allow for more eggslots in the future, it will introduce a bug. This is not the case if you use the eggslots length variable

anapple96: so basically `this.eggSlots < this._eggList.length`?

anapple96: ok, fixed

src/scripts/breeding/Breeding.ts
anapple96: fixed

anapple96:
I think it's a Travis problem

src/scripts/breeding/Breeding.ts
```diff
@@ -223,7 +223,13 @@ class Breeding implements Feature {
     }
 
     public gainEggSlot() {
-        this.eggSlots += 1;
+        if (this.eggSlots < this.eggList.length){
+            this.eggSlots += 1;
+        }
```
Ishadijcks: Too many brackets

anapple96: fixed

src/scripts/breeding/Breeding.ts
```diff
@@ -223,7 +223,11 @@ class Breeding implements Feature {
     }
 
     public gainEggSlot() {
-        this.eggSlots += 1;
+        if (this.eggSlots < this.eggList.length){
+            this.eggSlots += 1;
+        } else {
+            console.error("eggSlots is larger than 4.");
```
Ishadijcks: Magic number, same reasoning as above

anapple96: `eggSlots is larger than eggList.length.`  ok?

src/scripts/breeding/Breeding.ts
```diff
@@ -223,7 +223,11 @@ class Breeding implements Feature {
     }
 
     public gainEggSlot() {
-        this.eggSlots += 1;
+        if (this.eggSlots < this.eggList.length){
+            this.eggSlots += 1;
+        } else {
+            console.error("eggSlots is larger than eggList.length.");
```
Ishadijcks: Please refer to the docs to see how string templates work
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

anapple96: I've been using C# too much.

anapple96:
*technically* fixed all the problems

PR #625, opened 2020-04-04
Fix routeKill quests giving inaccessible routes

anapple96:
Some Quests requires you to go to route 47 and 48, which don't exist


Ishadijcks:
Please make sure you branch from the latest commit on develop

Ishadijcks:
Not needed as routes 47 and 48 will be added.

PR #628, opened 2020-04-06
Show currencies in shops

Ishadijcks:
Screenshot from 2020-04-06 21-11-06


src/components/shopModal.html
```diff
@@ -4,6 +4,22 @@
         <div class="modal-content">
             <div class="modal-header text-center">
                 <h4 data-bind="text: player.town().name() + ' Shop'">Shop</h4>
+                <h4 style="display: inline;">
```
greeny, line 7: Bootstrap has `d-inline` class, which may be better over inline styles :)

Ishadijcks: Turns out it wasn't even needed

PR #630, opened 2020-04-06
Don't allow the player to gain infinite eggslots

Ishadijcks:
Had to make a new one because Travis broke.

Closes #619


anapple96:
Travis hates me

PR #634, opened 2020-04-06
Disable shards until ShardCase is unlocked

### LINKED ISSUE #293: Lock shards behind the shard case key item
The shard case will be unlocked after the first prestige. Currently you can access the shard menu at any time, even without the KeyItem.

Ishadijcks:
Closes #293 

- Disable the Start Menu entry if the shardcase is not yet unlocked.
- Don't allow the player to gain shards until they unlock the case

### DIFF
--- a/src/components/gameMenu.html
+++ b/src/components/gameMenu.html
@@ -19,8 +19,9 @@
         <li>
             <a class="dropdown-item" href="#logBookModal" data-toggle="modal">Log Book</a>
         </li>
-        <li>
-            <a class="dropdown-item" href="#" data-bind="click:function(){App.game.shards.openShardModal()}">Shards</a>
+        <li data-bind="attr: {  title: App.game.shards.canAccess() ? '' : 'Progress further to unlock...'}">
+            <a class="dropdown-item" href="#" data-bind="click:function(){App.game.shards.openShardModal()},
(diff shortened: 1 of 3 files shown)


src/components/gameMenu.html
```diff
@@ -41,7 +42,9 @@
             <a class="dropdown-item" href="https://discordapp.com/invite/rHF3V54" target="_blank">Discord</a>
         </li>
         <li>
-            <a class="dropdown-item" href="https://gitreports.com/issue/pokeclicker/pokeclicker?name=Anonymous&issue_title=[BUG%20REPORT]%20" target="_blank">Report a bug</a>
+            <a class="dropdown-item"
+               href="https://gitreports.com/issue/pokeclicker/pokeclicker?name=Anonymous&issue_title=[BUG%20REPORT]%20"
+               target="_blank">Report a bug</a>
```
Ishadijcks, line 47: Formatting

RedSparr0w:
Do you think it would be better to just hide the shards entry until it is unlocked?
Or update the `openShardModal` to just show a error notification like the daycare/routes etc?

Ishadijcks:
I like that it is teasing the player in the menu. However I noticed that the bootstrap `.disabled` also disable any pointer events so the denied notification doesn't show up anymore. Easy fix would be to make our own disabled class that only changes the style.

PR #636, opened 2020-04-07
fixed money position logic on gym battle win #522

### LINKED ISSUE #522: [BUG]  Gym Leader 'Money Earned' Positioning
When a Gym Leader is defeated (in a rematch, have not tested first victories but don't expect the behaviour to change), the '+##,###' pop-up appears in the top-right of the screen, rather than over the PokeDollars counter.

IshaD: Yeah this happens with all gym money, probably because the modal messes with the position.

BaineGames:
Closes #522 
- Added ID to the h1 title as a default reference to use in the animate money function.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -148,6 +148,7 @@
 
 <div class="container">
     <div class="row justify-content-center">
+        <!--The id title is needed for a default marker for money animation to display when not in a wild battle where the playerMoney span is not visible-->
         <h1 id="gameTitle">Pokéclicker</h1>
     </div>
     <div class="row justify-content-lg-center" style="text-align: center">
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -148,7 +148,7 @@
 
 <div class="container">
     <div class="row justify-content-center">
-        <h1>Pokéclicker</h1>
+        <h1 id="gameTitle">Pokéclicker</h1>
```
Ishadijcks, line 151: Can you add a comment here explaining that the money is showed here. Good chance we will remove this title in the future.

BaineGames: sure

PR #649, opened 2020-04-09
Add a check to see if any KeyItems should be gained on load

Ishadijcks:
Only needed in case requirements change between saves.


Ishadijcks:
I suspected this could cause some issues with the initialization order. So far no issues have been reported by Sentry so I think it works

PR #661, opened 2020-04-16
Update gameMenu.html

InQuognito:
Suggestion: Rename to 'Menu' for clarity and to save space


RedSparr0w:
I personally prefer `Start Menu` over `Menu` but i'll let @Ishadijcks decide if this should be changed or not.

Ishadijcks:
Same, that's what it's called in the games

PR #667, opened 2020-04-21
Always generate the max number of daily deals in Underground

### LINKED ISSUE #666: Always generate at least one daily deal
On this day, April 21 2020, the daily deal generator can be empty. There should always be at least one daily deal.

fujnw:
Not **always**, but with very very high probability.
Closes #666

Deals on 2020-04-21:
image


Ishadijcks:
Nice, there are some slight issues however. Because of the for loop it always does the 10 * max tries, even if it is done after max tries. Can you rewrite it to a while loop like
```
int i = 0;
while(i < maxTries && deals.length < maxDeals) {
i++
...
}
```
This doesn't waste performance without the risk of an inifinite loop.

fujnw:
Done @Ishadijcks

Issue #573, opened 2020-03-11
[BUG REPORT] Lag when sorting Pokemon by Attack


Ishadijcks:
Submitter: Joshua Myers
When I have the pokemon list in the bottom left sorted to "Attack and reverse checked" the game behaves oddly. It's like a stutter or lag. Not really sure how to describe it. More like buffering, like when you are loading a video. I've cleared my cookies and cache. This also didn't start happening until today or yesterday.

RedSparr0w:
Strange, nothings changed on the live developement version recently

Ishadijcks:
Odd indeed, let's hope someone can reproduce it

danfroese:
I only notice lag when using xAttack items, but it's pretty extreme. Could be a related issue with attack calculation?

Ishadijcks:
I think so, it only seems to happen whenever the game has been running for a while though.

RedSparr0w:
Lag is also more extreme when sorting by level iirc

PR #684, opened 2020-06-05
Fix Lurking Computed Observers

Aegyo:
This pull request fixes at least one cause of pokeclicker accumulating memory as you play.
The main source of memory leaking it fixes is from clickAttacks

All of the changes are changing ko.computed into ko.pureComputed wherever possible.

A ko.computed will subscribe to any observables that it sees, which adds this ko.computed to an update list on each observable. When we lose any normal references to this computed (eg, a variable or DOM node), we expect the garbage collector to free up the memory. Since the computed is still referenced in this hidden update list of one or more observables, the garbage collector thinks we still want the function and does not free its memory.

This happens every time we calculate click attack (eg, every time we do a clickAttack), because we are checking if the x attack item is active by creating a computed, evaluating it, and then forgetting about. Except the effectList remembers every single one of those computed functions.

We can solve this in two ways:
1. Make sure to call dispose() on the computed when we are done with it
2. Use a pureComputed instead

A pureComputed will unsubscribe from all observables when it no longer has any subscribers to itself. This method is only a problem for computeds that we want to stay subscribed even if nothing is subscribed to itself, because we want the computed to perform some side effect when an observable changes - for example, creating a notification.

I have changed as many computed functions as I can to pureComputed, and eyeballed the remaining ones to see if they are disposed properly. The minesquare one in Underground looks possibly leaky.

### DIFF
--- a/src/scripts/GameHelper.ts
+++ b/src/scripts/GameHelper.ts
@@ -5,10 +5,10 @@ class GameHelper {
     public static counter = 0;
     public static currentTime: KnockoutObservable<Date> = ko.observable(new Date());
     public static tomorrow: Date = GameHelper.getTomorrow();
-    public static msUntilTomorrow: KnockoutComputed<number> = ko.computed(function () {
+    public static msUntilTomorrow: KnockoutComputed<number> = ko.pureComputed(function () {
         return Number(GameHelper.tomorrow) - Number(GameHelper.currentTime());
     });
-    public static formattedTimeUntilTomorrow: KnockoutComputed<string> = ko.computed(function () {
+    public static formattedTimeUntilTomorrow: KnockoutComputed<string> = ko.pureComputed(function () {
         let milliseconds = GameHelper.msUntilTomorrow();
         const hours = Math.floor(milliseconds / GameHelper.MS_IN_HOUR);
         milliseconds -= hours * GameHelper.MS_IN_HOUR;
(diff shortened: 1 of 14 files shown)


Aegyo:
I ran into this memory leak when I was trying to create an autoclicker that didn't lag the game. I used clickAttack instead of the more powerful pokemon attack for no reason whatsoever, and watching my memory get consumed rather rapidly.

You can observe this yourself on `develop` branch, and then this branch to compare the difference, by running this code below:

```
class Idler {
    constructor(idleFunction) {
        this._function = idleFunction;
        this.loops = 0;
        this.functionCalls = 0;
        this.looping = false
    }

    start() {
        this.loops++;
        requestIdleCallback(this.loop.bind(this));
        this.looping = true;
    }

    stop() {
        if (this.looping) {
            this.shouldStop = true
        }
    }

    loop(deadline) {
        if (this.shouldStop) {
            this.looping = false
            this.shouldStop = false
            return
        }
        
        while (deadline.timeRemaining() > 0) {
            let finished = this._function();
            this.functionCalls++;
            if (finished) {
                return;
            }
        }

        this.start();
    }
}

autoclicker = new Idler(function(){App.game.gameState === GameConstants.GameState.gym ? GymBattle.clickAttack() : Battle.clickAttack()})
```

To start autoclicking, run `autoclicker.start()`, and to stop, `autoclicker.stop()`

Or to make your comparisons between code versions easier:
```
// Take a memory snapshot via your browser dev tools
// Make sure you are on a route

// Start auto clicking and schedule a stop after 30 seconds
autoclicker.start(); setTimeout(function(){autoclicker.stop()},30 * 1000)

// Take another memory snapshot after the money numbers slow back down to normal

// Take another one after a minute or two to convince yourself the garbage collector has done all it can
```

Ishadijcks:
Nice find, I'll deploy it to the test version

Aegyo:
This could be the cause of issue #573, but I don't have a good way to verify that.

PR #702, opened 2020-07-07
Add option to sort pokemon list by base attack

### LINKED ISSUE #682: [Suggestion] Add an option to sort Pokemon by base attack
It would be nice if players could sort Pokemon by base attack rather than their boosted attack, for breeding purposes.

Aegyo:
There have been many requests for sorting by base attack, so people know which is the best pokemon to breed.

I didn't like having to add to a large switch statement to have a new sort option, and also have to add code in Settings.ts to register it, so I have refactored the sort options to have all the data in one place.

The Record type annotation on sortOptionConfigs is a nice feature, TypeScript will complain if we don't add an entry for all possible enum values

Fixes #682

### DIFF
--- a/src/scripts/settings/SortOptions.ts
+++ b/src/scripts/settings/SortOptions.ts
@@ -3,7 +3,8 @@ enum SortOptions {
     'name' = 1,
     'attack' = 2,
     'level' = 3,
-    'shiny' = 4
+    'shiny' = 4,
+    'baseAttack' = 5,
 }
 
 type SortOptionConfig = {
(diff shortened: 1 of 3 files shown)


src/scripts/settings/SortOptions.ts
```diff
@@ -3,5 +3,53 @@ enum SortOptions {
...
+    text: string;
+
+    // How to get the comparison value from a PartyPokemon
+    getValue: (p: PartyPokemon) => any;
+
+    // true if the default sort direction should be descending
+    invert?: boolean;
+}
+
+const SortOptionConfigs: Record<SortOptions, SortOptionConfig> = {
```
Ishadijcks, line 21: I did not know this was a thing

PR #717, opened 2020-08-01
Implement timed evolution

Ishadijcks:
As well as some cherry-picked commits from before.

Add a new Timed Evolution class which can be used like this:
```
new TimedStoneEvolution('Eevee', [new TimedStoneData(6, 18, 'Espeon')], 'Umbreon', GameConstants.StoneType.Time_stone)
```


RedSparr0w:
👍 Looks good to me,

But I think it could be simplified slightly as there are only 2 different time based evolutions (day/night) according to [bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Category:Pok%C3%A9mon_that_evolve_based_on_time).

Maybe just use it like this:
```js
// new TimedStoneEvolution(base_pokemon, day_pokemon, night_pokemon),
new TimedStoneEvolution('Eevee', 'Espeon', 'Umbreon'),
```
It should also probably have `GameConstants.StoneType.Time_stone` as the default stone type so it can be optionally included,
As it's likely to be the only stone used for it, due to _(as far as I could see)_ all the timed evolutions either being from friendship or leveling up during the time of day.

src/scripts/pokemons/PokemonList.ts
```diff
@@ -14177,7 +14177,7 @@ pokemonList.forEach(p => {
 
     p.attack = Math.max(10, Math.floor(Math.sqrt(baseDefense * baseStamina) * baseOffense / 250));
     if (p.baby) {
-        p.evolutions?.forEach(evo => pokemonDevolutionMap[evo.evolvedPokemon] = evo.basePokemon);
+        p.evolutions?.forEach(evo => pokemonDevolutionMap[evo.getEvolvedPokemon()] = evo.basePokemon);
```
Aegyo, line 14180: This makes `pokemonDevolutionMap` vary depending on what time of day you loaded the game. It isn't going to cause a problem with Eevee, but it could with some hypothetical future pokemon.

src/scripts/party/evolutions/Evolution.ts
```diff
@@ -1,30 +1,30 @@
...
         }
 
         if (notification) {
-            Notifier.notify(`Your ${this.basePokemon} evolved into a ${this.evolvedPokemon}`, GameConstants.NotificationOption.success);
+            Notifier.notify(`Your ${this.basePokemon} evolved into a ${this.getEvolvedPokemon()}`, GameConstants.NotificationOption.success);
         }
 
         const shiny = PokemonFactory.generateShiny(GameConstants.SHINY_CHANCE_STONE);
-        App.game.party.gainPokemonById(PokemonHelper.getPokemonByName(this.evolvedPokemon).id, shiny, true);
+        App.game.party.gainPokemonById(PokemonHelper.getPokemonByName(this.getEvolvedPokemon()).id, shiny, true);
```
Aegyo, line 27: Super unlikely to happen, but it might be safer to only calculate `this.getEvolvedPokemon()` once, in case it changes between throwing the notification and gaining the pokemon.

Ishadijcks: Very good point, I've stored the result.

src/scripts/pokemons/PokemonList.ts
Ishadijcks: I've added a comment regarding this. I _think/hope_ this will ever become an issue

RedSparr0w: It will affect chingling and budew IIRC

PR #739, opened 2020-09-14
Add Pikachu starter

RedSparr0w:
Add Pikachu as a possible starter if the player doesn't select a starter 5 times (by clicking off the modal)


Aegyo (approved):
:+1:

PR #740, opened 2020-09-14
Unlock requirements for roamers

RedSparr0w:
Apply requirements to the roamers.
They will not appear on the routes until these requirements have been met.

### DIFF
--- a/src/scripts/pokemons/RoamingPokemonList.ts
+++ b/src/scripts/pokemons/RoamingPokemonList.ts
@@ -1,0 +1,31 @@
+class RoamingPokemonList {
+    public static list: Array<Array<RoamingPokemon>> = new Array(GameConstants.MAX_AVAILABLE_REGION).fill([]);
+
+    constructor() {
+
+    }
+
+    public static add(region: GameConstants.Region, roamer: RoamingPokemon): void {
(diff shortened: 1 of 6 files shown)


src/scripts/pokemons/RoamingPokemonList.ts
```diff
@@ -0,0 +1,44 @@
+class RoamingPokemonList {
+    public static list: Array<Array<RoamingPokemon>> = [];
```
Aegyo, line 2:
```suggestion
    public static list: Partial<Record<GameConstants.Region, Array<RoamingPokemon>>> = {};
```
This suggested change isn't necessary, but is more specific about what can be used as object keys for this list.
The `Record<GameConstants.Region` ensures we are using a Region as the key, and the `Partial<..>` is just there so it doesn't complain when we haven't added something for every single region.

Feel free to ignore if you want to keep this as an Array :+1:

Aegyo (approved):

src/scripts/pokemons/RoamingPokemonList.ts
RedSparr0w: That makes sense, and this way it also won't break if we happen to add any Pokemon to the `none` region for whatever reason.

PR #742, opened 2020-09-16
Add more breeding filters, Display text

RedSparr0w:
Add breeding filter by text (allows for regex)

Add option for display value (attack, base attack, attack bonus, egg steps)

image

### DIFF
--- a/src/components/breeding.html
+++ b/src/components/breeding.html
@@ -19,6 +19,10 @@ aria-labelledby="breedingModalLabel" aria-hidden="true">
                 <div class="tab-pane active" id="breeding-pokemon">
                     <div class="text-left">
                         <div class="form-row">
+                            <div class="form-group col-md-6 col-6">
+                                <label>Search</label>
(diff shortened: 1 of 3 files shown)


Aegyo (approved):
:+1:

Aegyo:
I am just noticing that we don't get the rounded corners on hatchery pokemon like we do in the pokedex. The `btn` class seems to be overriding the border-radius from `pokedexEntry`. It also looks like the type colors for dual types bleed out the other end somehow, which isn't happening in the pokedex... any idea what is causing that one?

RedSparr0w:
> I am just noticing that we don't get the rounded corners on hatchery pokemon like we do in the pokedex. The `btn` class seems to be overriding the border-radius from `pokedexEntry`. It also looks like the type colors for dual types bleed out the other end somehow, which isn't happening in the pokedex... any idea what is causing that one?

I remember messing around with it ages ago to try fix it to be the same as the pokedex, don't remember what wasn't working for me though 🤔

Just adding the border-radius still leaves it with some overflow though:
image

PR #743, opened 2020-09-17
Fix up events checking

RedSparr0w:
setTimeout is run instantly if the number is too large:
The delay (in milliseconds) for these functions is a 32 bit signed quantity, which limits it to 2147483647 ms or 24.855 days.

The main change is here:
```diff
-                this.initialize();
-            }, timeTillEventStart - 7 * GameConstants.DAY);
+                this.checkStart();
+            }, Math.min(7 * GameConstants.DAY, timeTillEventStart - 7 * GameConstants.DAY));
```
It was running the `initialize` function constantly over and over if the event start time was more than ~25 days away, as that's the limit for setTImeout.
But it should now run a set timeout for 7 days or until the event start, whichever is less.

🤞 Should reduce some lag.

### DIFF
--- a/src/scripts/specialEvents/SpecialEvent.ts
+++ b/src/scripts/specialEvents/SpecialEvent.ts
@@ -129,7 +129,7 @@ class SpecialEvent {
 
         // If more than 1 day, don't notify the player yet
         if (timeTillEventEnd > 1 * GameConstants.DAY) {
-            // Check again when 7 days left until event start, or in 7 days (whichever is less)
+            // Check again when 1 day left until event start, or in 1 day (whichever is less)
             setTimeout(() => {
                 this.checkEnd();
             }, Math.min(1 * GameConstants.DAY, timeTillEventEnd - 1 * GameConstants.DAY));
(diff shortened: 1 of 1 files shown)


src/scripts/specialEvents/SpecialEvent.ts
```diff
@@ -78,39 +111,33 @@ class SpecialEvent {
...
-    start() {
-        this.status = SpecialEventStatus.started;
+    checkEnd() {
         const timeTillEventEnd = this.timeTillEnd();
 
-        this.notify('on now!', Math.min(1 * GameConstants.HOUR, timeTillEventEnd), GameConstants.NotificationOption.success);
-        this.startFunction();
+        // If more than 1 day, don't notify the player yet
+        if (timeTillEventEnd > 1 * GameConstants.DAY) {
+            // Check again when 7 days left until event start, or in 7 days (whichever is less)
```
Aegyo, line 132: This comment doesn't match the code below :eyes:

Aegyo (approved):
:+1:

src/scripts/specialEvents/SpecialEvent.ts
RedSparr0w:
```suggestion
            // Check again when 1 day left until event start, or in 1 day (whichever is less)
```

RedSparr0w: Whoops 😅

PR #749, opened 2020-09-20
[WIP] Unova

RedSparr0w:
Creating this PR so it's easier to keep it up to date with develop branch, and keep track of overall progress.

There is a board at https://github.com/pokeclicker/pokeclicker/projects/2 and a milestone at https://github.com/pokeclicker/pokeclicker/milestone/7 to track the individual tasks that are remaining before we can release Unova.


SpenserJ (approved):
Looking good to me. As long as it doesn't crash on load and doesn't enable the region yet, I think we're at a good point to merge :+1:

src/scripts/GameConstants.ts
```diff
@@ -461,6 +461,8 @@ namespace GameConstants {
         'Razor_claw',
         'Razor_fang',
         'Prism_scale',
+        'Sachet',
+        'Whipped_dream'
```
RedSparr0w, line 465: comma dangle
```suggestion
        'Whipped_dream',
```

src/scripts/breeding/Breeding.ts
```diff
@@ -223,7 +247,8 @@ class Breeding implements Feature {
     }
 
     public getEggSlotCost(slot: number): number {
-        return 500 * slot;
+        return 500
+          * slot;
```
RedSparr0w, line 251:
```suggestion
        return 500 * slot;
```

PR #752, opened 2020-09-21
Move `ko: if seen` conditional in pokedex image html

Aegyo:
The silhouette image for unseen pokemon is not being shown since the fix to the hatchery display.
This moves the `ko if` to surround just the things we don't want for the unseen pokemon.

Before hatchery changes:
Before hatchery changes

After hatchery changes:
After hatchery changes

With this change:
With this change


RedSparr0w (approved):
Looks good 👍

PR #755, opened 2020-09-21
fixed burmy icon

LuqDragon:
imgonline-com-ua-twotoone-iEa6uFigF5mPlQRT


RedSparr0w:
Just a note for future PRs:
We don't normally update the docs folder in PRs as it just makes it harder to review and removes any config settings from the files.

src/components/shopModal.html
```diff
@@ -23,7 +23,7 @@ <h4 class="col-6 col-sm-4 col-lg-2">
                                      data-bind="template: { name: 'caughtStatusTemplate', data: {'status': $data.getCaughtStatus()}}">
                                </span>
                            </knockout>
-                           <img src="" height="36px" data-bind="attr:{ src: 'assets/images/items/' + name().replace(/\W/g, '_') + '.png' }">
+                           <img src="" height="36px" data-bind="attr:{ src: 'assets/images/items/' + name().replace(/[. ]/g, '_') + '.png' }">
```
RedSparr0w, line 26: I'm not sure about this change as it will allow disallowed filenames to be tried still:
`Unown (?).png`
I think we would be better off maybe using allowed chars instead of a list of disallowed characters.
Maybe like:
```
/[^\w.-]/g
```
Which would allow any A-Z chars, 0-9 digits, `_`, `.` and `-` in the filename.
Or `/[^\w.-\\(\\)]/g` if we also want to allow `()`.

LuqDragon: The other files have a `_` instead spaces, so we need to standardize the file names. Like the `Unown (?).png` the burmy have spaces instead `_`, so if we allow it, we dont need to change to `_`,  but Mr. Mime file is `Mr__Mime.png` cause of dot and space, if we apply the `/[^\w.-\\(\\)]/g` the file need to be `Mr._Mime.png`.

RedSparr0w: I think the `.` looks okay in the `Mr._Mime.png` filename, we could dissallow dots aswell though if we want?
`/[^\w-\\(\\)]/g`

RedSparr0w: Which will basically keep all the file names the same (expect burmy)

LuqDragon: Ok, just spaces will be replaced so

LuqDragon: So it was like this:
regex: `/[^\w.-\\(\\)]/g`
Mr. Mime: `Mr._Mime.png`
Burmy (plant): Burmy_(plant)
OK?

RedSparr0w: Sounds good 👍

RedSparr0w (approved):
👍 Thanks for the PR

PR #760, opened 2020-09-22
Nerf click attack speed

anapple96:
The game has fallen into the Cookie Clicker trap and it's not a good thing.
Also buffed attack speed by a tiny bit to compensate


src/scripts/Battle.ts
```diff
@@ -26,9 +26,9 @@ class Battle {
      */
     public static pokemonAttack() {
         // TODO: figure out a better way of handling this
-        // Limit pokemon attack speed, Only allow 1 attack per 900ms
+        // Limit pokemon attack speed, Only allow 1 attack per 750ms
         const now = Date.now();
-        if (this.lastPokemonAttack > now - 900) {
+        if (this.lastPokemonAttack > now - 750) {
```
RedSparr0w, line 31: Why are we decreasing this?

anapple96: "Also buffed attack speed by a tiny bit to compensate"

Ishadijcks:
This seems like a knee jerk.
Also nerfing how often you can click is not nerfing click attack. You're nerfing autoclickers, and boosting attack for legit players.

anapple96:
Yeah, that's the main purpose, the title is misleading

src/scripts/Battle.ts
RedSparr0w: This doesn't actually increase the pokemon attack speed though, It just allows people to cheat pokemon attack faster

anapple96: I always forget the -900 in Battle.ts doesn't actually control the speed

anapple96: Reverted

PR #761, opened 2020-09-22
Add eggs and stones to all region shops.ts

Ultima1990:
The idea is to have these items function as hints for where one might find the Pokémon they need to complete the local dex. As such only the items one will need in that region were added in this push. A.I. no fire stone in Hoenn, no dragon egg in Johto.
An attempt was made to have grass/water/fire eggs available fairly early, especially in a region with late docks. Starters available before late docks makes sense to me. Stones didn't need to be quite so early, functioning more as hints than anything else.
I hope I put stuff in locations that makes sense. I tried. Maybe too hard sometimes (electric egg in Sunyshore, very late).

Final touches: added ultra and great ball in places where they should become available in those regions, and the Veilstone dept. store.

IMO this is better than having everything available in the shortcut shop, but it should alleviate the amount of dock travel people are having to do right now.

### DIFF
--- a/src/scripts/towns/Town.ts
+++ b/src/scripts/towns/Town.ts
@@ -313,10 +313,12 @@ const VeilstoneCityShop = new Shop([
     ItemList['Token_collector'],
     ItemList['Item_magnet'],
     ItemList['Lucky_incense'],
-    ItemList['Fighting_egg'],
 ]);
 const CelesticTownShop = new Shop([
     ItemList['Dragon_egg'],
+]);
+const CanalaveCityShop = new Shop ([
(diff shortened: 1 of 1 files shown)


Ultima1990:
Late edit to move Sinnoh's fighting egg from Veilstone to Canalave. Veilstone made sense because it's the fighting gym, but it's also the shopping mall. Rather not have non-shopping mall stuff in that shop. Canalave also makes sense because it's access to Iron Island, and the only new Sinnoh mon in the egg is Riolu.

RedSparr0w (approved):
Awesome, thanks for the PR

PR #762, opened 2020-09-23
Filter to achievements

LuqDragon:
achievement


src/components/achievementsModal.html
```diff
@@ -10,8 +10,45 @@ <h5 data-bind="tooltip: { title: 'As you earn achievements, you gain up to 100%
...
+                                    <select class="custom-select" onchange="AchievementHandler.getAchievementOption(this.value, AchievementHandler.filter.type())">
+                                        <option value="all" selected>All</option>
+                                        <option value="false">Uncompleted</option>
+                                        <option value="true">Completed</option>
+                                    </select>
+                                </div>
+                                <div class="form-group col-md-6 col-6">
+                                    <label>Type</label>
+                                    <select class="custom-select" onchange="AchievementHandler.getAchievementOption(AchievementHandler.filter.status(), this.value)">
+                                        <option value="all" selected>All</option>
```
Aegyo, line 28: We could use a knockout foreach binding here to save us having to update in the future with new achievement types.

There is an example of using this with an enum in `src/components/shipModal.html`
```html
<div style='margin-top:20px' data-bind="foreach: GameHelper.enumStrings(GameConstants.Region)">
```
and in the same place, a nice utility function to turn the enum key into something better to display
```js
text: GameConstants.camelCaseToString($data)
```

```js
>> GameConstants.camelCaseToString("itemDiggedUp")
<- "Item Digged Up"
```

src/scripts/achievements/AchievementTypes.ts
```diff
@@ -0,0 +1,17 @@
+enum achievementType {
```
Aegyo, line 1: All of our enums so far have begun with a capital letter, it would be nice to keep that consistent

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -19,12 +21,26 @@ class AchievementHandler {
...
+
+    public static getAchievementListWithIndex() {
+        this.achievementListFiltered = this.achievementList.filter(a => a.region <= player.highestRegion() &&
+                                        (this.filter.status() == 'all' ? true : a.unlocked == JSON.parse(this.filter.status())) &&
+                                        (this.filter.type() == 'all' ? true : a.type == achievementType[this.filter.type()]));
+        this.getNumberOfTabs();
+        return this.achievementListFiltered.slice(this.navigateIndex() * 10, (this.navigateIndex() * 10) + 10);
+    }
+
+    public static getAchievementOption(status: string, type: string) {
```
Aegyo, line 40: `get` implies that this function is for retrieving a value, but this doesn't return anything.
I think it makes more sense to call this `updateAchievementOption` or even `updateAchievementFilter`

LuqDragon: you are right, this function is weird, I will dismember it to each filter and change its name, what do you know?

src/scripts/achievements/AchievementTypes.ts
LuqDragon: ok

src/components/achievementsModal.html
LuqDragon: ok, I will do that

src/scripts/achievements/AchievementHandler.ts
Aegyo: That sounds like a good idea to me

RedSparr0w:
I feel like we may be better off just filtering by the achievement `Requirement` type, rather than creating a new type to filter by.

As far as I can tell, they are all pretty much the same.

That way when new Achievements/Requirements are added, everything is all in one area without having to also update the filter types.

I'm not sure if this is the best approach or not though.

As it would likely also require adding a `friendlyName/displayName` or similar to each of the requirements for use in the select element.

LuqDragon:
true, i didnt realize that, i will work on it

src/scripts/achievements/AchievementType.ts
```diff
@@ -1,17 +1,33 @@
...
+    'capture': CapturedRequirement.name,
+    'defeat': DefeatedRequirement.name,
+    'shiny': ShinyPokemonRequirement.name,
+    'egg': HatchRequirement.name,
+    'ball': PokeballRequirement.name,
+    'route': RouteKillRequirement.name,
+    'gym': ClearGymRequirement.name,
+    'dungeon': ClearDungeonRequirement.name,
+    'click': ClickRequirement.name,
+};
```
LuqDragon, line 33: i dont know if this is best, but i don't know a way to get it automatically from files, so I just change to names of classes and use this to filter

RedSparr0w (approved):
PR looks good!

Thank you 👍

PR #763, opened 2020-09-23
Add ability to collapse shard types

RedSparr0w:
This might not be the best solution.

Due to the way the shards stuff is all in a table, it doesn't actually collapse when hiding it is just delayed slightly before hiding, but when showing again, is shown instantly.


RedSparr0w:
Example:
shards

Aegyo:
If you add a `transition: none` style to the table rows, the collapsing is instant

RedSparr0w:
I had to move each shard to it's own thead & tbody within the table, so that it was only collapsing 1 element by id to completely remove the delay.
Otherwise it was still delayed with `transition: none`

src/styles/shards.less
```diff
@@ -10,6 +10,12 @@
     .progress-bar {
         z-index: 1;
     }
+
+    .collapsing {
+        -webkit-transition: height 0s;
+        transition: height 0s;
```
Aegyo, line 16: Any particular reason to specify the `-webkit-` browser prefix only? It looks like all major browsers have supported the unprefixed version since ~2013.

Speaking of browser prefixes, I learnt recently that [autoprefixer](https://github.com/postcss/autoprefixer) is a thing that exists. Might be useful when we eventually try to make the game work nicely with older browser versions.

Aegyo (approved):
Works nicely :+1:

src/styles/shards.less
RedSparr0w: That looks pretty good, although not sure how far back we want to go in supported versions.
But at least if that was implemented it's one less thing to worry about.

PR #764, opened 2020-09-23
Nerf Explorer Kit price

anapple96:
5k too dang expaenseeve


RedSparr0w (approved):
Looks good 👍 
Thanks

Ishadijcks:
Do you want to give a restitution to people who bought it already?

anapple96:
Maybe, but I don't think we need it

anapple96:
This game is in beta after all, they know full and well the game can change completely overnight. Like the attack formula change for example

RedSparr0w:
The extra 3k QP back would probably help a lot of the newer players though.

anapple96:
Hmmm, maybe, maybe

KrushGames:
Yeah i think 3 eggs or a new stone would definitely help some of the new players, I see a bunch of people complaining about egg price

RedSparr0w (changes requested):
Would you also be able to add something to the update function to refund the players 3k QP please.

src/scripts/Update.ts
```diff
@@ -236,6 +236,18 @@ class Update implements Saveable {
             }
         }
 
+        if (this.isOlderVersion(this.saveVersion, '0.5.4')) {
+            try {
+                //Correct statistics
+                App.game.wallet.currencies[GameConstants.Currency.questPoint] += 3000;
```
Ishadijcks, line 242: Now you're giving it to everyone instead of people who have bought it.

Ishadijcks (changes requested):

src/scripts/Update.ts
```diff
@@ -236,6 +236,18 @@ class Update implements Saveable {
...
+
+                // Update save data
+                this.setSaveData(saveData);
+            } catch (ಠ_ಠ) {
+                console.error('[update] v0.5.4 - Couldn\'t update player statistics..', ಠ_ಠ);
```
RedSparr0w, line 247:
```suggestion
                console.error('[update] v0.5.4 - Couldn\'t refund players Quest Points..', ಠ_ಠ);
```

src/scripts/Update.ts
RedSparr0w: This also won't add to the players Quest Points, as it's not quite initialized yet.
`App.game.update` is the first thing initialized so that all the other data can update correctly based on what we change.
You will need to modify the `saveData.wallet` object.

PR #765, opened 2020-09-23
Nerf egg prices

anapple96:
Egg RNG is brutal in Kanto


RedSparr0w:
I think this should be left until at least #757 is resolved.

anapple96:
Yeah, will just leave it here for now. If we say no later, I'll close it

PR #771, opened 2020-09-23
Add use all battle items button under each item

D0rako:
Clicking hundreds of times to keep battle items active for hours gets kinda old after a few regions. This should add buttons similar to the shop screens that'll use all of a battle item in a single click under each item. I've set it to only show after arriving in Johto (similar to the shortcuts menu), as this probably shouldn't be a feature that's available early in the game.

### DIFF
--- a/src/components/battleItemContainer.html
+++ b/src/components/battleItemContainer.html
@@ -30,6 +30,26 @@
 
             </td>
           </tr>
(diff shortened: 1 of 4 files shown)


RedSparr0w:
I like the idea, but maybe instead of adding a use all button, we could use something like a [bootstrap popover](https://getbootstrap.com/docs/4.1/components/popovers/#four-directions) with multiple options similar to this:
image

PR #772, opened 2020-09-23
Fix/farm notification

LuqDragon:
I started thinking this was broken, but was not implemented yet, so i think found a good solution to it (thanks for the code in discord, RedSparr0w)

### DIFF
--- a/src/scripts/farming/Farming.ts
+++ b/src/scripts/farming/Farming.ts
@@ -36,11 +36,12 @@ class Farming implements Feature {
     update(delta: number): void {
         const timeToReduce = delta * App.game.oakItems.calculateBonus(OakItems.OakItem.Sprayduck);
         this.plotList.forEach(plot => {
-            if(!plot.isEmpty() && plot.timeLeft == 0 && !plot.notified) 
+            if (!plot.isEmpty() && plot.timeLeft == 0 && !plot.notified) {
                 plot.notified = true;
+            }
             plot.reduceTime(timeToReduce);
         });
-        if(this.plotList.some(plot => !plot.isEmpty() && plot.timeLeft == 0 && !plot.notified)) {
+        if (this.plotList.some(plot => !plot.isEmpty() && plot.timeLeft == 0 && !plot.notified)) {
             Notifier.notify({ message: 'A berry is ready to harvest!', type: GameConstants.NotificationOption.success, sound: GameConstants.NotificationSound.ready_to_harvest, setting: GameConstants.NotificationSetting.ready_to_harvest });
         }
     }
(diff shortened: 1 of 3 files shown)


src/scripts/farming/Farming.ts
```diff
@@ -36,8 +36,14 @@ class Farming implements Feature {
     update(delta: number): void {
         const timeToReduce = delta * App.game.oakItems.calculateBonus(OakItems.OakItem.Sprayduck);
         this.plotList.forEach(plot => {
+            if (!plot.isEmpty() && plot.timeLeft == 0 && !plot.notified) {
+                plot.notified = true;
```
RedSparr0w, line 40: I feel like this should be done after the notification is sent or on the same cycle that the time is set to 0,
Otherwise there is a whole cycle where the notification has been sent, but `plot.notified == false` before setting it to `true`.

Or another way that we could go about this is something like:

Add a flag for whether we should notify or not (default false)
Loop the plots
-- Reduce the time for each plot
-- If any plot reaches 0 and update the plot to `notified = true` then flip the flag to `true` as well
Check the flag, and send out a notification if true

Which would also allow us to remove the extra check at the end for the notification.

RedSparr0w (changes requested):
👍 Looking good, thanks for the PR,
Just left 1 comment for a potential modification.

src/scripts/farming/Farming.ts
LuqDragon: The whole cycle with `plot.notified == false` is to detect a non notified in the `some`, notify, and in the next cycle, when they have notified, make `plot.notified = true`, just to avoid another `foreach`, cause I don't know who was notified in the `some`. To notify in the same `foreach` they will notify all the plots, I'm avoiding that.

A flag is a good option too, but I don't know where we gonna reset it, if we reset after notification, the flag will be turned true again in the next cycle, cause how can we detect the update to `notified`?

RedSparr0w: In terms of the flag, something like this:
```

        const timeToReduce = delta * App.game.oakItems.calculateBonus(OakItems.OakItem.Sprayduck);
        let notify = false;
        this.plotList.forEach(plot => {
            plot.reduceTime(timeToReduce);
            if (!plot.isEmpty() && plot.timeLeft == 0 && !plot.notified) {
                plot.notified = true;
                notify = true;
            }
        });
        if (notify) {
            Notifier.notify({ message: 'A berry is ready to harvest!', type: GameConstants.NotificationOption.success, sound: GameConstants.NotificationSound.ready_to_harvest, setting: GameConstants.NotificationSetting.ready_to_harvest });
        }
    }
```
It's set to false at the start of the cycle, then set to true if we should notify.

LuqDragon: Ok, that's good. I forgot the `notified` will avoid to notify again xD, thanks

RedSparr0w (approved):
Looks Great 👍
Thanks for the PR!

PR #773, opened 2020-09-24
Add sorting by breeding efficiency

nls0:
Breeding efficiency is the base attack to egg cycle ratio.

### DIFF
--- a/src/scripts/settings/SortOptions.ts
+++ b/src/scripts/settings/SortOptions.ts
@@ -6,6 +6,7 @@ enum SortOptions {
     'shiny' = 4,
     'baseAttack' = 5,
     'breedingEffiency' = 6,
+    'highestRegionBreedingEfficiency' = 7,
 }
 
 type SortOptionConfig = {
@@ -57,6 +58,12 @@ const SortOptionConfigs: Record<SortOptions, SortOptionConfig> = {
     [SortOptions.breedingEffiency]: {
         'text': 'Breeding Efficiency',
         'getValue': p => (p.baseAttack / pokemonMap[p.name].eggCycles),
+        'invert': true,
+    },
+
+    [SortOptions.highestRegionBreedingEfficiency]: {
+        'text': 'HRBE',
+        'getValue': p => (p.baseAttack / pokemonMap[p.name].eggCycles * (PokemonHelper.calcNativeRegion(p.name) < player.highestRegion() ? (player.highestRegion() + 1) * 0.1 : 1)),
         'invert': true,
     },
 };
--- a/src/scripts/settings/SortOptions.ts
+++ b/src/scripts/settings/SortOptions.ts
@@ -5,8 +5,8 @@ enum SortOptions {
     'level' = 3,
     'shiny' = 4,
     'baseAttack' = 5,
-    'breedingEffiency' = 6,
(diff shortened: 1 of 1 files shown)


src/scripts/settings/SortOptions.ts
```diff
@@ -5,6 +5,8 @@ enum SortOptions {
     'level' = 3,
     'shiny' = 4,
     'baseAttack' = 5,
+    'breedingEffiency' = 6,
```
Aegyo, line 8: Should be `breedingEfficiency`

src/scripts/settings/SortOptions.ts
```diff
@@ -52,4 +54,16 @@ const SortOptionConfigs: Record<SortOptions, SortOptionConfig> = {
...
     },
+
+    [SortOptions.breedingEffiency]: {
+        'text': 'Breeding Efficiency',
+        'getValue': p => (p.baseAttack / pokemonMap[p.name].eggCycles),
+        'invert': true,
+    },
+
+    [SortOptions.highestRegionBreedingEfficiency]: {
+        'text': 'HRBE',
```
Aegyo, line 65: `HRBE` is all players are going to see in game, it would be ideal if they could know what this is without having to go to discord/wiki/reddit

Maybe just abbreviating some of the words? Something like `Max reg. breeding eff.`?

Aegyo (changes requested):

src/scripts/settings/SortOptions.ts
nls0: What about `Max Region Breed Eff.` ?

I think the discord uses the term `Breeding Attack`, is that better than `Breeding Efficiency`?

Aegyo: Yes, `reg.` doesn't look very clear, might be better to spell that one out entirely

I don't think `Breeding Attack` is a good description, it makes it sound like the pokemon highest on this list will gain the most attack per breed, which is a bit different to what we are actually sorting by.

Might be best to ask people on the discord to see what would make sense without being too long.

RedSparr0w: I might just be misunderstanding this option, But I'm not entirely sure what the use case would be for this one?
Would it not just make more sense to use the `Breeding Efficiency` sort and then filter by the region you want to breed?

nls0: Well, I'm only in Hoenn so far, and you're right that it's not that important, but even in Hoenn it puts Gyarados second in the list if I want to raise my Hoenn attack. Clefable is next highest out-of-region at around 60, which is kind of low, but still higher than a lot of Hoenn-native pokemon. I think in higher regions, as the penalty for out-of-region pokemon reduces, other pokemon can go higher in the list than region-native pokemon.

I just started playing a few days ago, though, so I can't say for certain how useful it will be.

RedSparr0w: I think it might be best to just do the standard one for now, and we can always re add this further down the line if needed.

RedSparr0w (approved):
Changes look good to me 👍 
Thanks for the PR!

jaahay:
Literally just came here to propose this. Thank you :thumbsup:

PR #774, opened 2020-09-24
temporary fix to safari bug

LuqDragon:
just a temporary fix until we don't discover what is causing the real issue


RedSparr0w (approved):

Hopefully this solves the issue.

PR #775, opened 2020-09-24
Add Sinnoh Roamers

Ultima1990:
Cresselia and Mesprit unlocked as roamers in Sinnoh after you clear their respective dungeons. Also removed them from those dungeons.

Make roamers roam again.

I swear there's some other file that needs changing, but I entirely forget which one... maybe my mind's playing tricks.


RedSparr0w (approved):
Looks good to me!

Not sure why these weren't roaming already 🤔

Ultima1990:
Because we didn't have a way to unlock roamers, and we did have the lake and island.

PR #767, opened 2020-09-23
Fix game not working on Firefox 70

Ishadijcks:
Closes #766 

Probably caused by the use of `?.`
There are no such statements left in the HTML files

PR #776, opened 2020-09-24
Replace null coalescing with logical OR to improve older browser support

Ishadijcks:
Could be related to #767


RedSparr0w (approved):
Changes look good!

The prospect result should never be falsy anyway _(as far as I can tell)_,
So this change shouldn't affect anything _(other than compatibility)_.

PR #777, opened 2020-09-24
Increase Shoal Cave DT cost.

Ultima1990:
Twas too cheap for when it's unlocked. This seems more reasonable to me, between Mt. Pyre and Seafloor Cavern.


RedSparr0w (approved):
Thanks,
That seems more inline with the other dungeons 👍

Issue #747, opened 2020-09-18
Move from keyCode → KeyboardEvent.code


RedSparr0w:
[`.keyCode` is depreciated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), we should move to using [`KeyboardEvent.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) in the future.

![image](https://user-images.githubusercontent.com/7288322/93546837-d2d92900-f9b7-11ea-935d-e58d8d5e0f43.png)

PR #780, opened 2020-09-26
Migrating to KeyboardEvent.code

Mephistic:
This updates all uses of `event.keyCode` to use the newer, un-deprecated `KeyboardEvent` and should resolve https://github.com/pokeclicker/pokeclicker/issues/747.

Testing:
* I tested these changes in the most recent Chrome/Firefox/Edge and all seems fine.

One nit: It seems like we need to use `e.originalEvent.code` instead of just `e.code` because of the way TypeScript plays with https://github.com/jquery/jquery/issues/3978 - the `code` property will be available on the event object when debugging, but it doesn't seem to register as available when compiling.


src/scripts/GameController.ts
```diff
@@ -115,33 +108,33 @@ class GameController {
     }
 
     static addKeyListeners() {
-        $(document).on('keydown', function (e) {
+        $(document).keydown(function (e) {
```
Aegyo, line 111: Event aliases are deprecated, we should be using `.on(eventName`

Aegyo:
Seems like the type definitions we are using for jquery are missing the the `code` property for events. It was added to the jquery event object over 2 years ago.

I wonder how long it would take to get a pr approved over at DefinitelyTyped...

Mephistic:
Let's find out: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48201

Mephistic:
Looks like the DefinitelyTyped PR was merged so we can just use `e.code`

Aegyo (approved):

PR #782, opened 2020-09-26
Make Kyurem forms available

Ultima1990:
Some opinions needed. Some help also. Consider this a prototype for how it could be done. Basically: Black DNA and White DNA drop as items from Zekrom and Reshriram in Dragonspiral Tower. Can be used to evolve Kyurem, like any old evolution stone item. Thinking ~10% drop, which is why I need a separate dungeon boss held item chance constant. Or separate drop chances for these specific mons, or these specific items. I think dungeon boss might be easier?
Images are very placeholder of course.
Would like to add DNA Splicer key item for flavor. Haven't found what file contains key item unlock conditions.
If this method of obtaining these forms is approved, it could be re used for Necrozma.
Haven't been able to get Kyurem to show up in the evolution stones interface. The stones show up, the mon doesn't. I did give myself Kyurem, it's in my list. I'm sure I forgot something somewhere or messed something up, please point me in the right direction.
NoKyurem


anapple96:
The art for the dna is interesting

PR #784, opened 2020-09-26
Make Rotom forms available

Ultima1990:
This adds the Rotom forms to the Old Chateau boss pool after the base form is obtained.
In the main game Platinum the Rotom forms are available in the Team Galactic Eterna Building, which is accessible after obtaining the Eterna City gym badge, same as the Old Chateau. Both require the same HM (cut) to access.
The room that houses the Rotom forms does require a Secret Key, which was obtained from events so meh who cares.


RedSparr0w (approved):
👍 Thanks for the PR

PR #785, opened 2020-09-26
added a use multiplier to battle items

LuqDragon:
I thought that the other PR would implement that, but nothing changed, so I resolved to add.

I think this button fits the model, but not discard the idea of a dropdown to choose the multiplier.

I was afraid to change these functions thinking they were used by all the items, but are used just to battle items, so I will try to make more generic as possible in the next updates.

The 'all' option is a little controversial too.
mult2

closes #771


src/components/battleItemContainer.html
```diff
@@ -2,6 +2,17 @@
     <div class="card-header p-0" data-toggle="collapse" href="#battleItemContainerBody">
       <span>Battle Items</span>
     </div>
+    <button type="button" class="btn btn-sm btn-primary" 
+            style="position: absolute; right: 0px; top: 0px; width: auto; height: 41px;" 
+            data-bind="text: ItemHandler.multipliers[ItemHandler.multIndex()]"
+            onclick="ItemHandler.incrementMultiplier()"
+            oncontextmenu="ItemHandler.decrementMultiplier(); return false"
+            title="Controls"
+            data-toggle="popover"
```
RedSparr0w, line 11: I think we should use a `tooltip` here to match the style of everything else.

LuqDragon: Ah, ok.

First, I thought that, but I found this popover and decided to try it.

LuqDragon (commented):
nice update, this simplifies so much xD

src/components/battleItemContainer.html
RedSparr0w: Actually, maybe we should leave it, and move the other things over to popovers instead.
It does look nicer IMO.

LuqDragon: maybe I just change the title to 'Use Multiplier' instead 'Controls', the problem is that this popover is relatively big than a tooltip

RedSparr0w (approved):
👍 Thanks for the PR!

PR #788, opened 2020-09-27
Move suggestion template into correct folder for GitHub issues

SpenserJ:
There is currently an [issue template](https://github.com/pokeclicker/pokeclicker/issues/new/choose) for bug reports but not one for suggestions, even though there is a [file in the project root](https://github.com/pokeclicker/pokeclicker/blob/2721ea31722b1635b66afda905e14ba1b3d1c43a/issue_improvement_template.md). This adds a suggestion template, and cleans up the old templates that aren't being used


RedSparr0w (approved):
Looks good, thanks 👍

PR #789, opened 2020-09-27
"fix" the complete kanto pokedex achievement

LuqDragon:
I "fixed" the complete Kanto Pokedex achievement, cause many people to come to discord saying this is a bug, so I made a new requirement to unique pokemons by region if afterward implement new achievements by region if want.


RedSparr0w:
I think it would be better to use something similar to what the `MapHelper.ableToTravel` method uses, As that also filters out different forms which also probably shouldn't count toward the achievement.

https://github.com/pokeclicker/pokeclicker/blob/325e0722ff8983e0f88784d9aa31862fc84be442/src/scripts/worldmap/MapHelper.ts#L204

And maybe at the same time update that to use a shared function/filter.

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -137,7 +137,7 @@ class AchievementHandler {
         AchievementHandler.addAchievement('Keeping Oak Really Busy', 'Capture 75 unique Pokémons', new CaughtPokemonRequirement(75), 0.15);
         AchievementHandler.addAchievement('Surpassing Ash', 'Capture 88 unique Pokémons', new CaughtPokemonRequirement(88), 0.05);
         AchievementHandler.addAchievement('I Wanna be The Very Best', 'Capture 100 unique Pokémons', new CaughtPokemonRequirement(100), 0.20);
-        AchievementHandler.addAchievement('I Should Probably Take a Break', 'Complete the Kanto Pokédex!', new CaughtPokemonRequirement(151), 0.50);
+        AchievementHandler.addAchievement('I Should Probably Take a Break', 'Complete the Kanto Pokédex!', new CaughtUniquePokemonsByRegionRequirement(GameConstants.Region.kanto), 0.50);
```
RedSparr0w, line 140: PR looks good to me,
Do you think we should make this bonus apply the Kanto region percentage instead of the global percentage?

_not really sure which makes more sense_

LuqDragon: i dont know about that, i dont know how the bonus really works, I have in my save more than 100%, its not global?
Dont make much sense, for me, just for Kanto cause when you complete Kanto you travel to other regions, just come back for quests or shinies.

RedSparr0w (approved):
Changes look good 👍 
Thanks for the PR

Issue #730, opened 2020-09-06
[BUG] Ability to capture gym leaders Pokemon


RedSparr0w:
**Info**
 - OS: `Windows 10`
 - Browser: `Chrome`

**Description:**

You are able to catch team leaders pokemon

**Steps to Reproduce:**

1. Go to a route enable all catch options
2. Defeat Pokemon
3. As soon as the catch animation starts switch to a gym and start the battle
4. Might catch team leaders Pokemon

**Expected Behavior:**

Should catch the Pokemon you originally were going to catch.

PR #793, opened 2020-09-27
Prevent Player From Catching Gym Pokemon

Mephistic:
This change snapshots the defeated enemy pokemon at the start of the catching process so we try to catch that pokemon instead of the last pokemon battled.

This should resolve https://github.com/pokeclicker/pokeclicker/issues/730


RedSparr0w:
Thanks for the PR!

This does seem to solve the problem of being able to capture Gym Pokemon, but it also introduces a new bug.
When the real Pokemon is captured, the Gym Pokemon is replaced with a 10k HP Rattata.

Gif Below:
Gym_Capture

LuqDragon:
```typescript
this.attemptCatch(enemyPokemon);
if (player.route() != 0) {
    this.generateNewEnemy();
}
```
this solve, but impact on other things?

RedSparr0w (approved):
👍 Change look good thanks.

PR #804, opened 2020-09-30
Add introduction to certain features using intro.js

### LINKED ISSUE #786: [Feature] Intro/Guide for things as they are unlocked
Implement something like https://introjs.com/

Would be good to add one for most of the modules (pokeball selector, quest list after tutorial, oak items)

Example:
image

RedSparr0w:
Closes #786 
_Note that before testing this branch you will need to run `npm i` to install the new dependency._

Add introduction text for some elements using intro.js
Probably more that needs to be added, but this should cover the basics.

Images of currently implemented intros:
image
image
image
_note that it covers notifications too (which is for the best IMO)_
image
image


RedSparr0w:
Note that before testing this branch you will need to run `npm i` to install the new dependency.

Aegyo (approved):
Looks good to me.

If you felt like making a small addition, I think the introduction for the pokeball selection could let people know about hovering over each table header for more information about what that column controls. It doesn't seem hugely clear without the information in those tooltips.

RedSparr0w:
> If you felt like making a small addition, I think the introduction for the pokeball selection could let people know about hovering over each table header for more information about what that column controls. It doesn't seem hugely clear without the information in those tooltips.

👍 Added some info in the initial pokeball intro tooltip, as it doesn't seem to play nicely with tables for some reason.

PR #805, opened 2020-09-30
Update shiny chance

RedSparr0w:
Make dungeon Pokémon use the `SHINY_CHANCE_DUNGEON` constant.
Increase shiny chance for Evo stones.
_(stones are similar price to shop pokemon, and this is the value that was in the original & we have in the faq 😅)_


Aegyo (approved):
:+1:

PR #806, opened 2020-09-30
fix "lag" caused by filtering breedable list

LuqDragon:
Fix to change breedable list just when open or when change the filter


RedSparr0w (approved):
Nice work!

Changes look good to me 👍

PR #811, opened 2020-10-01
Added sorting UI to breeding filter area

### LINKED ISSUE #798: [SUGGESTION] Hatchery List Sort Options
It would be nice to have hatchery sorting options alongside the filtering options. I could sort the filtered Pokemon by current attack value, or step count, etc.

**Current Behavior**
The current hatchery list allows us to filter our available Pokemon but no sorting options are present.

**Improved Behavior**
Additional options that allow us to set a primary (and possibly secondary) sort value. A dropdown could be added to sort by the desired criteria. A toggle to chose sort direction as well would be useful.

**Reason for change**
This allows users to more easily pick which Pokemon they want to get an egg for. Maybe they want to keep training their strongest Pokemon or maybe they want the ones quickest to hatch; this would make it easier to find them.

**Detailed Description**
A dropdown could be added to sort the filtered Pokemon by Pokemon Number, Attack, Attack Bonus, Base Attack, or Egg Steps.  A second dropdown could be added with the same values as a secondary sort in case of a tie in the first value. Additional there should be another toggle of some sort to chose the sort direction.

walkerboh:
Added a version of the sorting UI from the Pokemon List into the Day Care filtering section. Using the same fields so the values stay consistent.
closes #798


RedSparr0w (changes requested):
Thanks for the PR!

Just one change needed, as it doesn't seem to update the hatchery list when changed due to #806,

On a side note:
It takes up a lot of space on mobile, I think we probably will also need some sort of way to show/hide the filters, but that can be done in a separate PR.
image

walkerboh:
I don't love the solution so please let me know if that was a bad variable to subscribe against. I tried subscribing to the observable values of the two Settings but that would update the breeding list before caughtPokemon was updated so the order wouldn't change.

RedSparr0w:
Just commenting before I forget to note that the `sort direction` select doesn't change when you change the `sort direction` on the Pokemon list.

And there is also a console.log left over.

Otherwise seems to be fine, the update doesn't seem to happen often without expecting it which is good.

walkerboh:
Added toggles to show and hide the breeding filters and sorting. Default is hidden.
image

RedSparr0w (approved):
Looks good,
Thanks for the PR!

PR #812, opened 2020-10-01
Allow Mystery Egg to hatch from Dragon Eggs pool

RedSparr0w:
Mystery Egg currently ignores the Dragon egg pool.
This PR allows it to pull from there too.


Aegyo (approved):
:+1:

PR #813, opened 2020-10-01
Fix egg hatch shiny chance

RedSparr0w:
Fix egg hatch shiny chance.

It is currently being double calculated:
Shiny chance is being lowered based on shiny steps,
and then also lowered if shiny charm active upon hatching.

### DIFF
--- a/src/scripts/oakItems/OakItems.ts
+++ b/src/scripts/oakItems/OakItems.ts
@@ -38,6 +38,15 @@ class OakItems implements Feature {
         return oakItem.calculateBonus();
     }
 
+    calculateBonusIfActive(item: OakItems.OakItem) {
+        const oakItem = this.itemList[item];
+        if (oakItem == undefined) {
+            console.error('Could not find oakItem', item, 'This could have unintended consequences');
+            return 1;
+        }
+        return oakItem.calculateBonusIfActive();
+    }
+
     isUnlocked(item: OakItems.OakItem) {
         if (this.itemList[item] == undefined) {
             return false;
--- a/src/scripts/breeding/Egg.ts
+++ b/src/scripts/breeding/Egg.ts
@@ -78,9 +78,9 @@ class Egg implements Saveable {
             return;
         }
         let shinyChance = GameConstants.SHINY_CHANCE_BREEDING;
-        const shinyCharmBonus = shinyChance / App.game.oakItems.calculateBonusIfActive(OakItems.OakItem.Shiny_Charm);
-        const bonusStepsChance = shinyCharmBonus * Math.min(1, this.shinySteps / this.steps());
-        shinyChance -= bonusStepsChance;
+        const shinyCharmReduction = shinyChance - shinyChance / App.game.oakItems.calculateBonusIfActive(OakItems.OakItem.Shiny_Charm);
+        const shinyChanceReduction = shinyCharmReduction * Math.min(1, this.shinySteps / this.steps());
+        shinyChance -= shinyChanceReduction;
 
         const shiny = PokemonFactory.generateShiny(Math.ceil(shinyChance), true);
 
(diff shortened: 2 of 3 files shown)


nls0:
`App.game.oakItems.calculateBonusIfActive(OakItems.OakItem.Shiny_Charm) * Math.min(1, this.shinySteps / this.steps());` will evaluate to 0 if `shinySteps` is 0. Need to take the weighted average of 1 and oakbonus based on the percentage of shiny steps.
Divisor should be
(1 * non-shiny% + oakbonus * shiny%)
(1 - shiny%) + oakbonus * shiny%
1 + oakbonus * shiny% - shiny%
1 + shiny% * (oakbonus - 1)

RedSparr0w:
Thanks @nls0, missed that 😅 

I've updated it a little more,
Think it should be correct now.

I've also made some browser test of the same calculations:
```js
eggShinyChance = (shinySteps, steps, shinyCharmBoost = 2) => {
        let shinyChance = 1000; // 1000 for simplicity sake
        const shinyCharmReduction = shinyChance - shinyChance / shinyCharmBoost;
        const shinyChanceReduction = shinyCharmReduction * Math.min(1, shinySteps / steps);
        shinyChance -= shinyChanceReduction;

        return Math.ceil(shinyChance);
}

console.log(eggShinyChance(0, 100, 2)     == 1000);
console.log(eggShinyChance(100, 100, 2)   == 500);
console.log(eggShinyChance(50, 100, 2)    == 750);
console.log(eggShinyChance(100, 100, 1.5) == 667);
console.log(eggShinyChance(10, 100, 1.5)  == 967);
console.log(eggShinyChance(100, 100, 1.7) == 589);
console.log(eggShinyChance(100, 100, 1.9) == 527);
```

nls0:
eggShinyChance(0, 100, 2)     == 1000 gives 0.10% chance
eggShinyChance(100, 100, 2)   == 500 gives 0.20%
eggShinyChance(50, 100, 2)    == 750 gives 0.13%

I would expect the chance to be halfway between if I have the charm on half the time, since I had 1/1000 chance for half the steps and 1/500 chance for the other half, so halfway between would be 1/667.
If you want to do that, you can do
`shinyChance /=`
`1 + Math.min(1, this.shinySteps / this.steps()) * (App.game.oakItems.calculateBonus(OakItems.OakItem.Shiny_Charm) - 1)`

src/scripts/oakItems/OakItems.ts
```diff
@@ -38,6 +38,15 @@ class OakItems implements Feature {
         return oakItem.calculateBonus();
     }
 
+    calculateBonusIfActive(item: OakItems.OakItem) {
```
Aegyo, line 41: This is the same as calculateBonus, but just calls something different at the end. Feels like we could just be passing some `pretendActive` parameter to calculateBonus instead.

Aegyo: On further thought, I think also passing some number `bonusModifier` would be good, to have the oak item do this extra math for us. (with the intention to be used as a 0-1 value to scale the amount of bonus we actually get)

Aegyo (commented):
I don't like the location of these changes too much, this kind of low level shiny calculation in the middle of egg hatching code feels out of place.

I would rather see generateShiny take an extra parameter to scale the bonus we get from the shiny charm, possibly using the presence of that extra parameternor taking another one to calculate shiny bonus as if the charm is active right now.

Additionally, this kind of thing does not seem exclusive to shiny calculations, we might want something similar for instance with underground energy recovery, or maybe we want to add some feature where an oak item "charges up" to get increasing multipliers the longer you have it equipped. I think it would be better to move the oak item chance modification into the oak item calculateBonus method (or the one higher up in Upgrade?), and make necessary changes to pass the numbers all the way through to there.

I'm open to disagreements though, if someone feels this way is better.

RedSparr0w:
closing this for now, will pick it back up later, but if anyone else wants to fix it before that, feel free 😅

PR #816, opened 2020-10-01
Add more eggs everywhere

Ultima1990:
This adds Mystery Egg to the first gym town of every region.

Also adds Electric and Fighting eggs to shops in regions where they do add new mons to the pool, but no unique egg-only mons.


src/scripts/towns/Town.ts
```diff
@@ -192,10 +197,19 @@ TownList['Mt Silver'] = new DungeonTown('Mt Silver', [new RouteKillRequirement(1
...
+const DewfordTownShop = new Shop([
+    ItemList['Fighting_egg'],
+]);
 const SlateportCityShop = new Shop([
     ItemList['Greatball'],
     ItemList['Water_egg'],
 ]);
+const MauvilleCityShop = new Shop([
+    ItemList['Electric_egg'],
+]);
```
RedSparr0w, line 212: I think you may have forgotten to add this to the town 😅

Ultima1990: Gotta love that Travis is still fine with it. Other way around and Travis throws a fit. Can't have a shop added to a town only, but not added below //region shops.

RedSparr0w (approved):
Looks good to me!
Thanks for the PR

PR #817, opened 2020-10-01
Add mons to various eggs

Ultima1990:
This adds
Numel to Hoenn fire egg pool
Buizel to Sinnoh water egg pool
Budew to Sinnoh grass egg pool
Shinx to Sinnoh electric egg pool

Also it removes Electrike and Manectric from Hoenn routes, making the electric egg useful in that region.


RedSparr0w (approved):
👍 Looks good

PR #818, opened 2020-10-02
Prefer higher selected pokeball when encountering shinies

nls0:
When encountering a shiny, the game should use the higher selected pokeball if the shiny is also already caught and the caught setting is higher. Same logic for uncaught. Wouldn't want someone to encounter a new shiny roamer and throw an ultraball instead of masterball if they have new pokemon set to masterball.

A potential issue would be with the new pokeballs and the order they fit in the list.


RedSparr0w (approved):
Changes look good to me!

Thanks for the PR 👍

RedSparr0w:
> A potential issue would be with the new pokeballs and the order they fit in the list.

I think we will need to calculate which ball gives the higher chance based on the current encounter, and use that once/if we add new balls.

PR #819, opened 2020-10-02
Add notification settings for egg hatching and gyms won

nls0:
I didn't add settings for "additional finds" when hatching or for new captures or shiny captures since they are one time events, but maybe they can have settings, too. I don't know where to place the gym notification setting either, so I just put it at the bottom.


RedSparr0w (approved):
Awesome, thanks!
Changes look good 👍

RedSparr0w:
> I didn't add settings for "additional finds" when hatching or for new captures or shiny captures since they are one time events

I don't think we need to add options for the one time events, or things that we would likely want the player to see.

PR #820, opened 2020-10-02
Fix Manaphy => Phione evo

anapple96:


RedSparr0w (approved):
Changes look good,

Thanks for the PR 👍

PR #822, opened 2020-10-02
Change effect timers to show up to 99 hours

nls0:
Originally shows up to 60 minutes, while it's common to have many hours running at endgame. Knowing how many hours is useful for people who want to run overnight and how many more they need to use.

### DIFF
--- a/src/scripts/effectEngine/effectEngineRunner.ts
+++ b/src/scripts/effectEngine/effectEngineRunner.ts
@@ -30,8 +30,10 @@ class EffectEngineRunner {
 
     public static updateFormattedTimeLeft(itemName: string) {
         const times = GameConstants.formatTime(player.effectList[itemName]()).split(':');
-        if (+times[0] > 0) {
-            return player.effectTimer[itemName]('60:00+');
+        if (+times[0] > 99) {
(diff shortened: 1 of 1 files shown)


src/scripts/effectEngine/effectEngineRunner.ts
```diff
@@ -30,8 +30,10 @@ class EffectEngineRunner {
 
     public static updateFormattedTimeLeft(itemName: string) {
         const times = GameConstants.formatTime(player.effectList[itemName]()).split(':');
-        if (+times[0] > 0) {
-            return player.effectTimer[itemName]('60:00+');
+        if (+times[0] > 99) {
+            return player.effectTimer[itemName]('99h+');
+        } else if (+times[0] > 0) {
+            return player.effectTimer[itemName](`${times[0]}h`);
```
Aegyo, line 36:
```suggestion
            return player.effectTimer[itemName](`${+times[0]}h`);
```
Maybe do this so it says `1h` instead of `01h`. Might just be me, but I think `01h` looks odd.

Aegyo (changes requested):

Aegyo (approved):

PR #823, opened 2020-10-02
Big Kalos PR

Ultima1990:
Added requirements to Kalos routes.
Added Kalos towns, and their requirements.
Added Kalos shops, and their items.
Added Kalos Dungeons, and their Pokémon and requirements.
Added 2 new evolution items, and changed the Pokémon that use them.
Added images for the new items.

Notes:
Couldn't find 120x120 images for the items, so they are ugly upscaled 80x80s.
Gym.ts line 1067. There must be an elegant solution for boy/girl.
There's commented stuff for Kalos Power Plant and Pokéball Factory. They are Team Flare dungeons. Just trainers, no wild mons.
Team Flare Secret HQ is a Team Flare dungeon, but it contains the box legends to I thought that'd be very required.
There is another Team Flare dungeon I haven't mentioned. It's inside Lumiose City, and would open up just before and be a requirement for Team Flare Secret HQ.
Tower of Mastery contains no wild mons and only one trainer with only one mon. It does give Lucarionite/Mega Lucario. Could be important.
Everything needs a balance pass for HP and rewards.


src/scripts/badgeCase/BadgeCase.ts
```diff
@@ -128,5 +128,19 @@ namespace BadgeCase {
...
         'Elite_UnovaChampion',
+        //Kalos
+        'Bug',
+        'Cliff',
+        'Rumble',
+        'Plant',
+        'Voltage',
+        'Fairy',
+        'Psychic',
+        'Iceberg',
```
RedSparr0w, line 139: Hmm, very creative badge names aren't they 🤔

src/scripts/towns/Town.ts
```diff
@@ -517,3 +517,89 @@ TownList['Pinwheel Forest'] = new DungeonTown('Pinwheel Forest', [new GymBadgeRe
...
+    new OneFromManyRequirement([
+        new MultiRequirement([
+            new GymBadgeRequirement(BadgeCase.Badge.Iceberg),
+            new RouteKillRequirement(10, 21),
+        ]),
+        new MultiRequirement([
+            new GymBadgeRequirement(BadgeCase.Badge.Iceberg),
+            new RouteKillRequirement(10, 22),
+        ]),
+    ]),
```
RedSparr0w, line 603: I think this should be doing the same thing, as by default the array is basically a `MultiRequirement`
```suggestion
    new GymBadgeRequirement(BadgeCase.Badge.Iceberg),
    new OneFromManyRequirement([
        new RouteKillRequirement(10, 21),
        new RouteKillRequirement(10, 22),
    ]),
```

RedSparr0w (changes requested):
Looking good,

I think we should probably just use the original 80x80 images for now, until/if we can find something better.

PR #824, opened 2020-10-03
Damage Calculator

Aegyo:
Adds a damage calculator to the start menu.
The damage calculator has settings for
- Include Breeding: includes pokemon that are currently inside eggs when calculating damage
- Base Attack Only: ignores the extra attack from breeding
- Region: selects the region, for nativeRegion effects. Region None means ignore the region debuffs
- Type1 and Type2: select the type of the enemy pokemon we are calculating damage against

It tells you your total damage, and also provides a bar breaking down that damage to the type level. This type-breakdown bar has each section colored with the TypeLocked color from the shard upgrades modal, and a tooltip telling you what type this section is for and how much damage it is providing.

You can also select a type, and you will be given a list of all pokemon that have that type and the damage each will do against an enemy pokemon with the chosen settings.

For dual pokemon, the type-breakdown bar takes 50% of the damage for each type, the detail table of pokemon with a type tells you the full damage the pokemon will do however.

Damage Calculator Image

There might be bugs, I haven't fully tested it yet. This PR is mostly to get feedback from people to see if it is missing something we would like to have (and hope someone else does the testing for me :eyes:)


RedSparr0w (commented):
Code all looks good to me!

A few minor appearance changes that I think could help:
- Have the table go from edge to edge inside the modal (especially for mobile)
- Align the Pokémon names/images to the left of the first column
- Vertically align the text in the `Attack` and `Type` columns with the Pokémon name


Also i'm not sure if this is feasable, would the Pokémon List and this table be able to share a template?
And just pass through some variables for which columns should be displayed.

Aegyo:
I've put the images into their own column so they all line up, and we have the space to do that anyway. Also made the table text vertically centered, and the table extends to the edges now. I didn't like how the bar with different types looked when it extended to the edges, so everything else is wrapped in a div with a little padding.

New image

RedSparr0w:
Awesome, just one more change needed.

If you have any Pokemon currently breeding in the hatchery and have `Include Breeding` unticked, it shows NaN for total damage, and hides the types that are in the hatchery:
_(image with Grass/Poison in hatchery)_
image

RedSparr0w (approved):
Changes look good 👍 

Thanks for the PR!

Issue #803, opened 2020-09-29
[BUG] Other items being added to player.itemList


RedSparr0w:
**Description:**
Extra items not in game are being added to the `player._itemList`, a couple of examples:
`<<BS_START>>{\"bs\":{\"hardReload\":true,\"scroll\":{\"x\":0,\"y\":0}}}<<BS_START>>` (added on my own save)
`中文(简体)`
`MEBBİS`

**Steps to Reproduce:**
- UNKNOWN

**Expected Behavior:**
No items not defined in the script should be added to the itemList

**Additional Information:**
This bug can also stop peoples saves from being downloaded - ref #801 

Edit: the `<<BS_START>>` seems to be a [browser sync issue](https://github.com/BrowserSync/browser-sync/issues/914) so shouldn't affect the live site.

PR #827, opened 2020-10-03
Strip non-ascii characters from discord usernames prior to saving.

### LINKED ISSUE #801: [BUG] non-ASCII characters in discord username breaks your save
**Description:**
Game stops loading on update because we try to download a save, which hits this line:
https://github.com/pokeclicker/pokeclicker/blob/develop/src/scripts/Save.ts#L51
`btoa` throws an error at this point with:
```
Uncaught (in promise) DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range
```

Probably also causes the game to crash if you try downloading your save normally.

**Steps to Reproduce:**

1. Link to a discord with emoji or other non-ascii characters in the username
2. Download save

**Expected Behavior:**
Game doesn't crash

**Additional Information:**
[StackOverflow question with multiple solutions](https://stackoverflow.com/questions/23223718/failed-to-execute-btoa-on-window-the-string-to-be-encoded-contains-characte)
Some people are also getting foreign characters into the itemlist, which also breaks their save. I have no idea how or why that is happening, but fixing this bug will fix those saves from crashing too.

davmillar:
Resolves #801

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -28,7 +28,7 @@
     <script src="libs/Sortable.min.js"></script>
 
     <!--Minified scripts-->
-    <script src="scripts/script.min.js?v=0.5.5"></script>
+    <script src="scripts/script.min.js?v=0.5.5-patch2"></script>
 
     <!--Popper.js-->
     <script src="libs/popper.min.js"></script>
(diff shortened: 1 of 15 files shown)


RedSparr0w:
Do you think it would be worth removing non-ascii characters from the entire save string, as we've also had strange instances of other strings being added to the itemList (#803)

davmillar:
I think it would be a good safety measure, but it might make tracking the root of that particular problem a bit harder.

RedSparr0w:
Good point!

I think we should probably handle the username escaping in the `Discord.toJson()` method so everything is kept together in terms of Discord data.

https://github.com/pokeclicker/pokeclicker/blob/fe32b12f9f062e220b8325129d96c15f653ae91e/src/scripts/discord/Discord.ts#L148-L154

RedSparr0w (approved):
Looks good thanks 👍

PR #828, opened 2020-10-03
Add an egg step for each step taken in Safari Zone.

davmillar:


RedSparr0w (approved):
👍 Looks good, Thank you.

PR #829, opened 2020-10-03
Squeeze Pokemon name in list if it's more than 15 chars long.

davmillar:


SpenserJ (approved):
Looks great :+1: Thanks for fixing this

Old (horizontal scrollbars):
image

New:
image

RedSparr0w:
Do you think it's worth just doing it for all the names?
As i'm still getting a horizontal scroll bar with `Armored Mewtwo` which is 14 characters long.
Or even just lower it to 13/14?

SpenserJ:
Oh the joys of variable width fonts. `iiii` is much skinnier than `wwww`, and `Armored Mewtwo` has wider characters than the average pokemon.

I think we need to decide whether we want to be overly cautious (13+) but compress things that don't need to be compressed, be explicit (have a constant defining which pokemon should be compressed), or look into a more advanced solution like SVG text (can scale without the aspect ratio) or some JS to resize. We should also do some testing on different zoom levels and device pixel densities, since that may eliminate options.

SVG seems like it would be the simplest to implement, and the most responsive, but we may need to pick some magic numbers first. https://codepen.io/chriscoyier/pen/zaRxdm

SpenserJ:
Going to close this now that #961 is merged as a full rework of the layout

PR #831, opened 2020-10-03
Allow users to view their eggs and fossils even when they cannot use …

davmillar:
…them in an egg slot.


src/components/breeding.html
```diff
@@ -158,10 +160,6 @@ <h5 class="modal-title" id="breedingModalLabel" data-bind="tooltip: { title: `Yo
                         </ul>
                         <!-- /ko -->
                     </div>
```
SpenserJ, line 162: Looks like you missed a `</div>` (pair of the one from the old line 146), and it is causing the close button to wrap to the right of the modal.

Old:
image

Now:
image

While you're fixing that, please also unindent that block by one level (4 spaces) so that we have:
```
                <div class="tab-pane" id="breeding-fossils">
                    <p class="text-danger" data-bind="visible: !App.game.breeding.hasFreeEggSlot()">
```

instead of
```
                <div class="tab-pane" id="breeding-fossils">
                        <p class="text-danger" data-bind="visible: !App.game.breeding.hasFreeEggSlot()">
```

SpenserJ (commented):
Looking really good, and I'm so excited to finally be able to see my egg and fossil list :+1:. Just the one comment about the broken close button and then we should be good to merge

SpenserJ (approved):
Looks great! Thanks for improving this :+1:

PR #832, opened 2020-10-03
Added Old Timey Pikachu special pokemon and created a daylight saving…

davmillar:
…s time event to spawn it.

Feel free to reject if it's a dumb idea. Also not sure if it should happen in spring as well, but it seemed like 'fall back' was more appropriate for the 'back in time' aspect.


RedSparr0w:
I like the event idea,

But I would prefer if we could keep all the sprites in the similar style,
Maybe we could find a newer style "fat pikachu" sprite for the event.
image

anapplealt:
We can instead use the Gmax Pikachu sprite, which is basically this but modern (yes, I'm using this account because I forgot my 2Auth device on my main)

PR #834, opened 2020-10-03
Prevent energy usage in the seconds between finishing a dig level and…

davmillar:
… the new level starting.


Ishadijcks:
Should probably also remove the `setTimeout` here if it's still active

RedSparr0w (approved):
Nice!

Looks good to me,
Thanks for the PR!

PR #835, opened 2020-10-03
chore(circular): moves notification constants into separate file

SpenserJ:
This is needed in order to begin reducing the number of circular dependencies within the code, and support the eventual move to proper TypeScript dependencies and type checking. An example of a circular dependency for `GameConstants.ts` (there are ~136) is `GameConstants.ts > utilities/Sound.ts > settings/Settings.ts > GameConstants.ts`, and the only reason these haven't cause a noticeable issue is because everything is pushed into `window` and most of the code isn't set up to import the types that it uses.

Despite being such a large number of lines changed, the changes can be summarized as:
* Move the three `Notification*` enums into a `NotificationConstants.ts` file
* Split most of the `Notifier.notify` calls onto multiple lines to improve readability (some lines were ~300 characters long)
* Replace `GameConstants.Notification` with `NotificationConstants.Notification` (case by case, instead of using replace all, just to double check everything)

### DIFF
--- a/src/components/settingsModal.html
+++ b/src/components/settingsModal.html
@@ -38,7 +38,7 @@
                     <div class="tab-pane" id="settings-sound">
                         <div data-bind="template: {name: 'RangeSettingTemplate', data: Settings.getSetting('sound.volume')}"></div>
                         <table class="table table-striped table-hover m-0">
-                            <tbody data-bind="foreach: Object.values(GameConstants.NotificationSound)">
+                            <tbody data-bind="foreach: Object.values(NotificationConstants.NotificationSound)">
                               <tr data-bind="template: { name: 'BooleanSettingTemplate', data: Settings.getSetting(`sound.${$data.name}`)}"></tr>
                             </tbody>
                         </table>
@@ -46,7 +46,7 @@
 
                     <div class="tab-pane" id="settings-notifications">
                         <table class="table table-striped table-hover m-0">
-                            <tbody data-bind="foreach: Object.values(GameConstants.NotificationSetting)">
+                            <tbody data-bind="foreach: Object.values(NotificationConstants.NotificationSetting)">
                                 <tr data-bind="template: { name: 'BooleanSettingTemplate', data: $data}"></tr>
                             </tbody>
                         </table>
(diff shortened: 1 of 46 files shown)


RedSparr0w:
Thanks for the change, this definitely makes more sense.

Though do you think instead of having the `NotificationConstants` these could just be a static property of `Notifier`?

```js
class Notifier {
    static setting = [];
    static sound = [];
    static type = [];
}
```

SpenserJ:
Potentially, however the driving force behind a few of these changes (there are quite a few coming) is reducing the circular dependencies. As long as I can set it up that Notifier doesn't import other code (or at least something that eventually imports Notifier), it shouldn't be a problem. Sound and Setting is a bit of a challenge with how it is currently written, since they have a circular dependency between them, so it might be better as a future improvement, but I'll take a quick look when I have some time if it is a blocker

RedSparr0w:
> there are quite a few coming

It might be worth us merging the [Unova branch](https://github.com/pokeclicker/pokeclicker/tree/unova) sooner than later then, otherwise these are going to create a bunch of conflicts there.

SpenserJ:
Yeah, I'm sure I'll be causing many conflicts as I move things around. Is there a list of outstanding tasks for that branch? I would be happy to help finish it up before I cause more chaos

RedSparr0w:
The PR is #749, there's still quite a lot to do before Unova is playable,

We can merge the PR into the develop branch once we have the town images though,
_(it won't make Unova playable, but will allow us to avoid merge conflicts)_
Or we can just disable trying to preload the images for unreleased regions (which is probably a good idea anyway).

SpenserJ:
Fantastic. I'm a fan of merging things in more regularly and having it disabled or hidden behind feature flags. Do you have any interest in using [project boards](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/creating-a-project-board) to track some of the work at a higher level (probably the top level tasks from the PR), or would you prefer to keep it in code?

SpenserJ:
@RedSparr0w I've updated this to target the `unova` branch now

RedSparr0w (approved):
Changes look good!

Thanks for the PR 👍

PR #836, opened 2020-10-03
chore(circular): move BadgeTypes into separate file

SpenserJ:
Similar to #835, this is needed in order to begin reducing the number of circular dependencies within the code, and support the eventual move to proper TypeScript dependencies and type checking. An example of a circular dependency for `BadgeCase.Badges` is `gym/Gym.ts > achievements/GymBadgeRequirement.ts > badgeCase/BadgeCase.ts -> gym/Gym.ts`, and the only reason these haven't cause a noticeable issue is because everything is pushed into `window` and most of the code isn't set up to import the types that it uses.

This PR is relatively straight forward, and just changes reference for `BadgeCase.Badge` to `BadgeTypes`. We can't use the `BadgeCase` namespace from a separate file, and if it was possible, doing so would result in the same circular dependency.


SpenserJ:
Whoops. I see a trailing comma snuck in there. I'll fix that up when I'm back on the computer.

Is the reason for avoiding trailing commas just that we want to ensure browser compatibility, and is that something that TypeScript is currently handling for us? I find that trailing commas tend to make development and PRs easier since you tend to have fewer lines changing and more consistency when coding.

SpenserJ:
@RedSparr0w I've updated this to target the unova branch now

RedSparr0w (approved):
Changes look good, thanks 👍

PR #839, opened 2020-10-04
Add animation for eggs that are imminently ready or ready to hatch.

davmillar:


src/styles/breeding.less
```diff
@@ -1,3 +1,75 @@
...
+  }
+  60% {
+    -webkit-transform: translateX(3px) rotate(2.4deg);
+            transform: translateX(3px) rotate(2.4deg);
+  }
+  75% {
+    -webkit-transform: translateX(-2px) rotate(-1.2deg);
+            transform: translateX(-2px) rotate(-1.2deg);
+  }
+}
```
RedSparr0w, line 41: I don't believe there's any benefit to having the `@-webkit-keyframes` anymore as none of our supported browser don't support `@keyframes` as far as I am aware.

<https://caniuse.com/mdn-css_at-rules_keyframes>

Also the same thing goes for `-webkit-transform` & `-webkit-animation` used below.

<https://developer.mozilla.org/en-US/docs/Web/CSS/WebKit_Extensions#Formerly_proprietary_properties_that_are_now_standard>

RedSparr0w:
Personally I think the animations might be happening too often and a little too fast/distracting,
Maybe if they were slowed down a bit and not constantly happening it would be nicer.

But I'll leave it here for now in case anyone else want's to weigh in.

_Example gif of current animations:_
eggs

SpenserJ:
I'd agree that the animations happen a bit too often. The speed of the animation seems good to me, but a delay of 3-5 seconds between cycles would be a bit less overwhelming

SpenserJ (approved):
This feels really good to me :+1: 
Peek 2020-10-04 08-51

RedSparr0w:
Nice!
Looks good a bit slower.

Could you also add an `animation-delay` to each of them with a slight variance,
So if they are all put in the hatchery at the same time, they will be out of sync when wobbling.

Something similar to this:
_(not sure classes & what sort of delays would be good values)_
```css
.egg-slot:nth-child(5n) {
  animation-delay: 0.48s;
}
.egg-slot:nth-child(5n+1) {
  animation-delay: 0.22s;
}
.egg-slot:nth-child(5n+2) {
  animation-delay: 0.35s;
}
.egg-slot:nth-child(5n+3) {
  animation-delay: 0.45s;
}
.egg-slot:nth-child(5n+4) {
  animation-delay: 0.15s;
}
```

RedSparr0w (approved):
Awesome, Thanks for the PR!
Changes look good 👍

PR #840, opened 2020-10-04
Skip preloading disabled features like towns

SpenserJ:
This is based off a discussion in #835, and will disable the preloading of unreleased towns (avoiding pointless requests and console warnings) unless the feature flag has been enabled and a development build is running. It is a quick and dirty feature flag system, but should do the trick for now.

Changes:
* `FEATURE_FLAGS` from `config.js` is injected into `window.featureFlags`
* If `window.featureFlags && window.featureFlags.preloadUnreleasedTowns`, the preloader will try (and currently fail) to load images for towns in Unova and Kalos, otherwise it will skip them (no warnings in a prod build)
* `Town` has been reworked to accept 3 arguments instead of the previous 6. This avoids the need for passing `null` or `undefined` to get to the argument that you want, and makes the instantiation of the class a bit more clear:
  * `name: string` - Required
  * `region: GameConstants.Region` - Required
  * `optional: {}`
    * `requirements?: (Requirement | OneFromManyRequirement)[]`
    *`shop?: Shop`
    * `dungeon?: Dungeon`
* `DungeonTown` and `PokemonLeague` have been updated to match
    * `npcs?: NPC[]`

### DIFF
--- a/_config.js
+++ b/_config.js
@@ -5,4 +5,7 @@ module.exports = {
     DEV_BANNER: (() => process.env.HEROKU)(),                               // optional (false) - If the development banner should be shown
     DISCORD_CLIENT_ID: false,                                               // optional (false) - Discord bot ID
     DISCORD_LOGIN_URI: false,                                               // optional (false) - Discord login proxy link (example: https://gist.github.com/RedSparr0w/46000c523b21e846adbc912d9800e5b8)
+    FEATURE_FLAGS: {                                                        // optional - Toggle feature flags for development
+        preloadUnreleasedTowns: false,                                      // optional (false) - Enable preloading images for unreleased towns
+    },
 };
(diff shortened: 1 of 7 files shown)


RedSparr0w:
I'm not too sure how I feel about this, it adds another place to update/miss when regions/towns are added/updated.
It also seems quite pointless to add a region to the towns that isn't used anywhere though, but I feel like that might be a better option?

SpenserJ:
The initial approach that I had been taking was another argument in the `Town` constructor, but that didn't feel like a great solution because I had to pass `undefined` for multiple arguments just to set it. Down the road I think having a `Region` class that references it's towns may simplify a large amount of logic, similar to how a `Town` lists shops, but that is a much larger undertaking.

I'm open to however you feel this would be best implemented. I picked the easiest to implement solution since this is hopefully temporary while a new region is being added, but it also isn't integrated deeply enough to really fill out the variable name that I used (`unreleasedTowns`)

SpenserJ:
@RedSparr0w Let me know if this is more along the lines of what you were thinking. I've also updated the description to represent the changes

RedSparr0w (approved):
👍 Looks good thanks!

PR #841, opened 2020-10-04
Make route level based on order and region

nls0:
Level of wild pokemon spawned is currently just route number * 2. Level is only used as a flat multiplier for exp gained. Egg steps, money, hp, and tokens are based on the route order instead of route number. The sixth route in Kanto (25) has almost the same level (and therefore exp) as the second to last route in Johto (26). Also, Unova starts at route 1, which would be a big hit to exp gained under the current system.

The main intent is to fix the order issues while keeping the level per region approximately the same. With this formula, overall Kanto route levels are the same, Johto around 20% higher, Hoenn 2% lower, and Sinnoh 3% higher. Unova would also be higher than Sinnoh instead of the same as Kanto.

A simpler method would just be route order * 2, though that would be a big hit to current exp gains.

exp

### DIFF
--- a/src/scripts/pokemons/PokemonFactory.ts
+++ b/src/scripts/pokemons/PokemonFactory.ts
@@ -42,7 +42,7 @@ class PokemonFactory {
     }
 
     public static routeLevel(route: number, region: GameConstants.Region): number {
-        return route * 2;
+        return Math.floor(MapHelper.normalizeRoute(route, region) * 2 + 20 * Math.pow(region, 2.3));
     }
 
     public static routeHealth(route: number, region: GameConstants.Region): number {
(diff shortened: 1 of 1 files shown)


RedSparr0w:
Changes to level look good (thanks for the graph too),

Would you be able to show the changes to Dungeon Tokens would be affected too?

nls0:
The dungeon token calculation previously normalized the route to be route order and then called routeLevel, but all that did was multiply it by 2. I replaced the function call by just a multiplication by 2, so nothing has changed for dungeon tokens.

RedSparr0w (approved):
Yeah that makes sense, Thanks 😅 

Changes look good!

nls0:
Yeah, another option would be to base the token gains on the new level calc (and remove the redundant call to normalize route). That would multiply the token gain by the same factor as between red and yellow in the graph. I didn't do that because I figure that's better left for a balance patch.

PR #842, opened 2020-10-04
Add duplicate deal checking to daily deal generator.

davmillar:


src/scripts/underground/DailyDeal.ts
```diff
@@ -39,7 +39,27 @@ class DailyDeal {
...
+
+        if (item1Name !== item2Name) {
+            return false;
+        }
+
+        if (DailyDeal.sameDealExists(item1Name, item2Name, dealList)) {
+            return false;
+        }
+
+        return ( !DailyDeal.reverseDealExists(item1Name, item2Name, dealList) && !this.item1.isStone() );
```
RedSparr0w, line 53: For the sake of consistency we should separate these 2 as you've done with the above 2 checks.

RedSparr0w (changes requested):
Thanks for the PR, just 1 small change.

Definitely makes sense to check for duplicates 👍

src/scripts/underground/DailyDeal.ts
```diff
@@ -39,7 +39,35 @@ class DailyDeal {
...
     private isValid(dealList: Array<DailyDeal>): boolean {
-        return ( (this.item1.name !== this.item2.name) && !DailyDeal.reverseDealExists(this.item1.name, this.item2.name, dealList) && !this.item1.isStone() );
+        const item1Name = this.item1.name;
+        const item2Name = this.item2.name;
+
+        if (item1Name !== item2Name) {
+            return false;
+        }
+
+        if (!this.item1.isStone()) {
```
RedSparr0w, line 49: I _think_ this might be the wrong way around?
Although I am a bit tired, so I may be wrong 😓
```suggestion
        if (this.item1.isStone()) {
```

RedSparr0w (approved):
Awesome!

Thanks for the PR 👍

src/scripts/underground/DailyDeal.ts
```diff
@@ -39,7 +39,35 @@ class DailyDeal {
     }
 
     private isValid(dealList: Array<DailyDeal>): boolean {
-        return ( (this.item1.name !== this.item2.name) && !DailyDeal.reverseDealExists(this.item1.name, this.item2.name, dealList) && !this.item1.isStone() );
+        const item1Name = this.item1.name;
+        const item2Name = this.item2.name;
+
+        if (item1Name !== item2Name) {
```
RedSparr0w, line 45: Whoops, needed to flip this one too 😅 
```suggestion
        if (item1Name == item2Name) {
```

PR #843, opened 2020-10-04
Fix Discord code logged in check, Add info about linking Discord

### LINKED ISSUE #833: [SUGGESTION] More info about linking discord
The first button in the Save menu is to Link Discord, which is making some people think that their save will be stored "in the cloud" and linked to their discord account.
There is also no mention of this being required for the discord bot codes to work (and I don't think the code verifier lets you know if you are trying to enter a discord code without linking a discord first)

**Improved Behavior**

A tooltip on Code and Link Discord, or a Help tab in the Save menu explaining what everything does and does not do

**Reason for change**

People are misunderstanding the purpose, and thinking their save files are stored safely.

RedSparr0w:
closes #833 

Added a note about not having cloud saves.

image

Fixed check for if user has linked Discord.

image

Also note the Discord bot already lets users know to link their account, but the extra Notification should help some confusion.

image


SpenserJ (approved):
:+1: Looks good to me

PR #847, opened 2020-10-04
Add Achievement Tracker

Mephistic:
Saw this was requested a few times in the Discord - this PR adds an Achievement Tracker bar to the top. Players can click on an achievement in the Achievements list to track it on the main screen.

Unselected:
Screen Shot 2020-10-04 at 11 18 29 AM
Achievement in Progress:
Screen Shot 2020-10-04 at 11 17 07 AM
Completed Achievement:
Screen Shot 2020-10-04 at 11 18 03 AM

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -124,8 +124,10 @@
 
         <!--Middle column-->
         <div id="middle-column" class="col-lg-6 col-md-12 order-md-first">
-            <!-- Achievement Tracker -->
-            @import "achievementTrackerContainer.html"
+            <div id="middle-top-sort-column">
+                <!-- Achievement Tracker -->
+                @import "achievementTrackerContainer.html"
+            </div>
 
             <div id="battleContainer" class="card sortable-disabled border-secondary mb-3">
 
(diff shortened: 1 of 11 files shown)


src/scripts/achievements/AchievementTracker.ts
```diff
@@ -0,0 +1,51 @@
...
+    constructor() {
+        this.trackedAchievement = ko.observable(this.defaults.trackedAchievement)
+    }
+
+    initialize(): void {
+
+    }
+
+    canAccess(): boolean {
+        return App.game.party.caughtPokemon.length >= 110;
```
Mephistic, line 19: I picked this threshold because:
* I didn't think we should dump this on a player right at the start
* I wouldn't imagine players would seriously start going for achievements before they hit the wall near the end of Kanto anyway, which is around ~120/130 caught Pokemon.

Happy to adjust/remove this if people feel it's too restrictive.

src/index.html
```diff
@@ -124,6 +124,9 @@ <h1 id="gameTitle">Pokéclicker</h1>
 
         <!--Middle column-->
         <div id="middle-column" class="col-lg-6 col-md-12 order-md-first">
+            <!-- Achievement Tracker -->
+            @import "achievementTrackerContainer.html"
```
Mephistic, line 128: I tried this as both the current "bar at the top of the screen" and as a box in the right sidebar similar to the Quests list - I thought the approach here looked better, but happy to adjust if it's too intrusive.

FWIW, here was my pass at the sidebar box: 
Screen Shot 2020-09-30 at 11 54 18 PM

SpenserJ (approved):
I love this. I wasn't sure about it being in the middle column at first, but I think it fits in nicely and doesn't increase the height of the page at all :+1:

RedSparr0w (changes requested):
Awesome work, Thank you.

I do like how it looks above the "main area",
But if you could move it to underneath the "main area" into the sortable columns would be better though,
as then players can drag and drop it to wherever they like (a less known feature).

Mephistic:
Oh neat, I didn't realize you could sort these!

Rather than moving the tracker down, I split the middle column into two sort areas - one above the battle box and one below it so we can sort around it.

RedSparr0w (approved):

PR #849, opened 2020-10-04
Merge develop into unova to prevent crash with missing Discord username

SpenserJ:
It would be good to get `unova` back in sync with `develop` because it is currently crashing for saves that are missing the Discord setting (which my development save doesn't have), and quite a few things have merged in the past week


RedSparr0w:
🤔 There shouldn't have been this many missing commits as develop was merged into unova yesterday: https://github.com/pokeclicker/pokeclicker/pull/749/commits/6a9df1b79ee02da2afa513d0d1801499f924fad4

PR #862, opened 2020-10-05
Save Mine as part of Underground

### LINKED ISSUE #353: Mine state doesn't change between saves
## Tested Version:
`develop` @ https://github.com/Ishadijcks/pokeclicker/commit/4033d2f582f9d3946f86da62d033e3b35bc377f0

## Description:
The mines state is saved seperately to the player data, causing the mine to be in the same state as you left it, even when reloading your save files or starting a new game.

## Reproduction:
- Load a save file
- Mine out part of the mine
- Reload the same save file
- The mine will still be partially mined

fujnw:
Closes #353 

On first load of the new code, the Mine should regenerate itself. Afterwards, it should save properly.
This doesn't make `Underground` a `Feature` yet. It simply brings `Mine` under the umbrella of `Underground` when saving.


RedSparr0w (changes requested):
Thanks for the PR,

Working good on current saves, but crashes on new game (after catching starter),
Here's some images for the error and code where the problem occurs.
image
image

fujnw:
I made a fix @RedSparr0w . Not the greatest but works.

SpenserJ:
I don't think the mine grid is instantiated until it is unlocked, so a safer fix might be:
```diff
- grid: this.grid.map(row => row.map(val => val())),
+ grid: this.grid?.map(row => row.map(val => val())),
```

That will only run the map function if `this.grid` has a value, instead of instantiating the grid when we're trying to save, which should match the previous behavior. You'll want to make the same change in `loadSavedMine` so that it doesn't crash if there is no mine in the save file:
```diff
- this.grid = mine.grid.map(row => row.map(val => ko.observable(val))),
+ this.grid = mine.grid?.map(row => row.map(val => ko.observable(val))),
```

RedSparr0w:
This fixes the game from stopping at the beginning, but for some reason it also doesn't show the mine if you reach it before reloading:
image

I'm not too sure what this is caused by as calling `Underground.showMine()` doesn't seem to work either.

RedSparr0w (approved):

RedSparr0w:
Awesome work!

Thanks for the PR 🎉

PR #867, opened 2020-10-05
Added Badges and Gym Lead/E4

unclebanks:
Added badges for Unova/Kalos
Added GL/E4 for Unova


RedSparr0w (approved):
Looks good 👍 

Thanks for the PR!

PR #868, opened 2020-10-05
allow Unown to breed other Unown to unlock the letters not yet access…

davmillar:
…ible in the game.


RedSparr0w (commented):
I like the idea,
Though I'm not sure I want them to be available so easily so early as we also have them on the Discord bot `!shop` (`DISCOR`).
And the other letters can be used in other events for easy filler.

RedSparr0w:
Closing as per above.

PR #869, opened 2020-10-05
fix(lint): adds Radix argument to all calls to parseInt

SpenserJ:
`parseInt` should always be called with the radix argument to ensure it parses the number as a base-10 number. Certain values may trigger it to assume a different radix if one is not provided. This is less of an issue with modern browsers, but should help to prevent an obscure bug from appearing down the road

More info:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description
- https://eslint.org/docs/rules/radix


RedSparr0w (approved):
Looks good thanks 👍

PR #871, opened 2020-10-06
TypeScript modules

SpenserJ:
This is a draft of some of the work that I'm doing to remove circular references and add full TypeScript support to the project. The changes I am making will cause many conflicts, so I'll try to keep it as up to date as possible. It may be worth merging this in smaller pieces so that the conflicts are easier to deal with, however it includes some temporary code that is required to support the migration to TypeScript modules, and may make the codebase more confusing until the migration is complete.

## High level overview

TypeScript modules are written using `import` and `export` statements, instead of the legacy `///<reference ... />`, and they bring a few key improvements over our current architecture:
* Full type checking throughout the code base
* Ability to bundle the code in a cleaner way (it isn't obvious in this PR due to the extra code required to support both modules and scripts simultaneously, but I promise it will be cleaner than our Gulp setup)
* Improved compilation and reloading in development (also currently limited due to supporting old and new)

## Technical details

The code in the `src/scripts` folder has been written to rely on all variables/classes/types/etc existing in the global `window`, and very few files are using the `///<reference ... />` directive to load in the types that it depends on. Instead of trying to migrate all of the code at once (I already tried only to run into 180+ circular references and a mess of merge conflicts with develop), I am slowly moving code into the `src/modules` folder which will use `import` and `export` almost exclusively. There are a few types that will be loaded through `///<reference ... />` to `src/scripts`, and all of the transpiled module code will be exposed back into the `window` and will have type declarations that are placed into `src/declarations` so that the old `src/scripts` code can still rely on the types.

All code from `src/modules` is bundled with webpack into `build/scripts/modules.min.js` and loaded before `scripts.min.js` in the browser (so that it can inject itself into the window global), and is loaded with `ts-loader` and the `./tsconfigWithModules.json` configuration. In the future, webpack will also allow us to transpile, minify, and/or bundle assets like styles (including LESS or SCSS) and images, and generate the html required to serve the game.

## Files moved from `./src/scripts` to `./src/modules`

| Old path (`scripts`)  | New path (`modules`) | Changes |
| ------------- | ------------- | ------------- |
| `common/Saveable.ts`  | `DataStore/common/Saveable.ts`  | None |
| `common/Feature.ts`  | `DataStore/common/Feature.ts`  | None |
| `badgeCase/BadgeCase.ts` | `DataStore/BadgeStore.ts` | Logic from `highestAvailableBadge` function moved into `BadgeCaseController`, as it was the only file calling the function |
| `achievements/Statistics.ts` | `DataStore/StatisticStore.ts` | `getGymIndex` and `getDungeonIndex` moved into `GameConstants.ts`, following pattern of similar GameConstant helpers |

## Note about testing

When you run this branch, it will generate type declarations in `./src/declarations`. Due to how the current codebase is set up for loading TypeScript files, you will not be able to run any other branches until you delete `./src/declarations`. Since it is autogenerated and .gitignored, it is safe to delete, and the only reason it lives in `./src/` is to make it easier to reference from the existing scripts. Once everything has been moved over, these declarations won't be required and can be removed from the build process.

***When switching away from this branch, delete `./src/declarations` or your code will fail to compile***

## Summary

* Modules will help us write cleaner code and eliminate the possibility of bugs caused by certain race conditions and circular dependencies
* Having both modules and scripts at the same time will be messy
* Merging this in smaller pieces will make migration significantly easier and will begin to show the benefits much earlier
* There will be some headaches as we find and fix the circular dependencies


SpenserJ:
This should be ready for review now that #835 and #836 are merged :+1:

RedSparr0w (approved):
👍 

Issue #790, opened 2020-09-27
[SUGGESTION] Add new Pokeball types, Update Pokeball selector


RedSparr0w:
Update Pokeball selector to be just 1 row:
![image](https://user-images.githubusercontent.com/7288322/94357655-2ace0900-00f7-11eb-803c-7d2f28bd3ccd.png)

When you click the Pokeball/cross it will open a modal with the different Pokeballs available for selection and a small description about each Pokeball (similar to the oak items)

Add new Pokeball types:
_(we don't need all of them straight away, can always just add the ones with a use, then the rest later on)_
```prolog
Lure Ball: water pokemon higher catch rate
Moon Ball: moon stone pokemon higher catch rate
Friend Ball: ???
Love Ball: ???
Heavy Ball: we dont have weight in our stats, could base it on hp or something instead
Fast Ball: high speed pokemon higher catch rate
Premier Ball: premium pokeball?...
Repeat Ball: already caught pokemon increased chance
Timer Ball: less time remaining in a dungeon higher chance
Nest Ball: lower route higher chance
Net Ball: water/bug higher chance
Dive Ball: water higher chance
Cherish Ball: ???
Heal Ball: ???
Quick Ball: faster catches (500ms)?
Dusk Ball: higher catch rate at night or in route/dungeons that use the cave background
```

I think it would be worth not having them all available straight from the start, maybe introduce a couple each new region.

HLXII:
Expanding on this idea.

We can split what poke balls are available by region (which gives some additional progression and uniqueness to regions).

NOTE: Numbers are speculative and will probably change for balance.

# Johto

## Apricorn Farm

Implement an Apricorn farm (much like the Berry farm in Kanto). Apricorn trees will take some hours to grow, but will remain permanent in the farm unless the player decides to remove it. Apricorn trees can only be removed once they're fully grown.

We can decide how automated this farm can be. I had the idea that once fully grown, they would produce Apricorns regularly (maybe they have a certain number of harvests until they die?). However, they can also behave like Berry farms where it's a one-time harvest.

Kurt will be an NPC in Azalea town and you can trade in Apricorns for special Poke balls.

Poke ball | Description
------------ | -------------
Fast Ball | Created from a White Apricorn. 15 catching power for Electric/Flying. 5 otherwise.
Level Ball | Created from a Red Apricorn. Improved catch rate for lower routes. Lowest route in region will have 15 catching power. Highest route will have 5.
[...]

PR #872, opened 2020-10-06
Pokeball Selector Modal

walkerboh:
Started looking into updating the Pokeball selector to reclaim some space and prepare for potential new Pokeball types (as referenced in #790). Also added a Pokeball class to make the creation of future Pokeball types hopefully easier.

Looking for feedback on my implementation, not sure if if was desired to just move the whole selector into a modal or select each category separately.
image
image

Currently having an issue where the tooltip is not updating on the Pokeball selector display, but it is working fine on the modal.


SpenserJ:
I haven't had a chance to try the code, but everything is looking good from what I've read. The modal has a lot of empty space right now, and if we're going to be adding the additional balls that were mentioned in the the original suggestion at some point down the road, maybe we should add another column (or 3) to give a little description of the benefit of the ball and show the catch rate bonus and catch time.

walkerboh:
I wasn't sure if it was desired to make those details known but I can look
at using some of that empty space more effectively and display those.

RedSparr0w:
I was thinking the modal could look a little something like this:
image

So it just shows all the different Pokeball types once, not a column for each selection,
The modal just shows for 1 at a time (caught, caught shiny, new, new shiny)

SpenserJ:
I really like the concept of a smaller selector modal, especially if we place it below (or beside, as long as it is near) the Pokeball status component

walkerboh:
That design looks pretty good. I'll start changing the modal up.

walkerboh:
Pretty much what Red had in mind:
image

Right now it opens up more in the center of page, would be nice if I could get to open closer to the pokeball selector but not sure if it is possible to dynamical locate the modal based on where the user has the pokeball selector on the main page.

src/components/pokeballSelector.html
```diff
@@ -10,47 +10,34 @@
...
+                    <td data-bind="template: {name: 'selectedPokeballDisplayTemplate', data: { 'value': App.game.pokeballs.alreadyCaughtSelection, 'field': App.game.pokeballs._alreadyCaughtSelection, 'title': 'Already Caught Pokemon'}}"></td>
+                    <td data-bind="template: {name: 'selectedPokeballDisplayTemplate', data: { 'value': App.game.pokeballs.alreadyCaughtShinySelection, 'field': App.game.pokeballs._alreadyCaughtShinySelection, 'title': 'Already Caught Shiny Pokemon'}}"></td>
+                    <td data-bind="template: {name: 'selectedPokeballDisplayTemplate', data: { 'value': App.game.pokeballs.notCaughtSelection, 'field': App.game.pokeballs._notCaughtSelection, 'title': 'New Pokemon'}}"></td>
+                    <td data-bind="template: {name: 'selectedPokeballDisplayTemplate', data: { 'value': App.game.pokeballs.notCaughtShinySelection, 'field': App.game.pokeballs._notCaughtShinySelection, 'title': 'New Shiny Pokemon'}}"></td>
+                </tr>
+                <tr>
+                    <td data-bind="template: {name: 'selectedPokeballCountTemplate', data: App.game.pokeballs.pokeballs[App.game.pokeballs.alreadyCaughtSelection]}"></td>
```
SpenserJ, line 26: I think we should make the entire row clickable, since the image takes up so little space in comparison.

src/components/pokeballSelectorModal.html
```diff
@@ -0,0 +1,46 @@
...
+                        <thead>
+                            <tr>
+                                <th width="20%">Ball</th>
+                                <th width="80%">Description</th>
+                            </tr>
+                        </thead>
+                        <tbody>
+                            <tr data-bind="template: { name: 'pokeballSelectionTemplate', data: { 'image': 'None', 'type': -1, 'description': 'Do not catch' }}"></tr>
+                            <!-- ko foreach: App.game.pokeballs.pokeballs -->
+                            <tr data-bind="template: { name: 'pokeballSelectionTemplate', data: $data}"></tr>
```
SpenserJ, line 24: I think we should make the entire row clickable, since the image takes up so little space in comparison.

SpenserJ (commented):
If we wanted to position it near the pokeball selector, we can do so with something like this:
```js
const $pokeballSelector =  $('#pokeballSelector');
const selectorOffset = $pokeballSelector.offset();
const selectorHeight = $pokeballSelector.height();
const selectorWidth = $pokeballSelector.width();
$('#pokeballSelectorModal .modal-dialog').css({
  left: selectorOffset.left + (selectorWidth / 2),
  top: selectorOffset.top + selectorHeight,
  transform: 'translateX(-50%)',
  margin: 0,
});
```

image

The challenge will be making sure that it chooses a good position regardless of where you put the selector (e.g., if its the last item in the column, we want to go above), and that it doesn't go off screen on smaller resolutions. We'd also need to disable that logic if its on mobile, since we'd want it centered.

After running through all of the potential edge cases, I no longer think it is worth trying to position relative since it will be more of a hassle than it is worth IMO.

RedSparr0w (approved):
Changes look good to me!

Thanks for the PR 👍

PR #873, opened 2020-10-06
Bulk buys -- configurable shop buttons

davmillar:


src/scripts/settings/Settings.ts
```diff
@@ -89,6 +89,14 @@ Settings.add(new MultipleChoiceSetting('breedingDisplay', 'Breeding progress dis
     ],
     'percentage'
 ));
+Settings.add(new MultipleChoiceSetting('shopButtons', 'Shop amount buttons:',
+    [
+        new SettingOption('+10, +100', 'original'),
+        new SettingOption('+100, +1000', 'bigplus'),
+        new SettingOption('\u00d710, \u00f710', 'multiplication'),
```
RedSparr0w, line 96: I'm not sure the divider option makes sense here, would a ×100 option not be more useful?
```suggestion
        new SettingOption('×10, ×100', 'multiplication'),
```
If not, I think we should just use the actual icons here, so we can tell what they are at a glance.
```suggestion
        new SettingOption('×10, ÷10', 'multiplication'),
```

src/scripts/towns/Town.ts
```diff
@@ -114,14 +114,18 @@ const CinnabarIslandResearcher = new NPC('Researcher', [
     'They were trying to clone an ancient Pokémon in the mansion, I wonder if they succeeded.',
     'Apparently the ancient Pokémon escaped, And can be found roaming around Kanto!',
 ]);
+const BigSpender = new NPC('Big Spender', [
+    'I love shopping! When I come in, the cashiers know I want tons of items.',
+    'You can use the Shop Amount Button settings to make it easy for big purchases, too!',
+]);
```
RedSparr0w, line 120: 👍

RedSparr0w (approved):
Changes look good,

Thanks for the PR 👍

PR #874, opened 2020-10-06
Added Some Unova Dungeons in GameConstants

unclebanks:
Added 1 image of Aspertia for initial testing and added some of the dungeons in Unova for checking.


RedSparr0w (approved):
Looks good thanks

PR #880, opened 2020-10-06
Makes Oak Item popover display levels consistently

AbstractBeliefs:
The move to show both current and max level in the Oak Item popovers is
a good one, but it would be more consistent and more clear if the
Level: x / y format was retained at all levels.

Currently, the format changes when you reach max level (indicated most
clearly by the MAX LEVEL label), but there doesn't see to be any benefit
in doing so. On the other hand, if Oak Items ever have varying max
levels (as factored for in the OakItem.maxLevel property), having the
varied format may be confusing (why does this one change format at 5,
and THIS one change format at 10?).

### DIFF
--- a/src/scripts/oakItems/OakItem.ts
+++ b/src/scripts/oakItems/OakItem.ts
@@ -70,6 +70,6 @@ class OakItem extends ExpUpgrade {
     }
 
     tooltip = ko.pureComputed(() => {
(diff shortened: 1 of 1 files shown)


SpenserJ (approved):
I think this makes a lot of sense. Thanks for adding it :+1:

Issue #881, opened 2020-10-06
[BUG] Achievement Tracker not listing entire achievement name


nikwasik:
**Info**
 - OS: Mac Catalina 10.15.6
 - Browser Chrome
 - Version v85.0.4183.83

**Description:**
When selecting an achievement in the achievement tracker, it will list the action that needs to be taken (for example, "Clear 10 Times"), but won't show what the player actively needs to clear.

**Steps to Reproduce:**
1. Select a Gym Battle or Dungeon Completion Achievement
2. See error

**Expected Behavior:**
The tracker should show what the player has to complete for the achievement (i.e., "Clear Azalea Gym 10 Times", "Clear Mt. Moon 10 Times", "Clear Mt. Coronet South 100 Times")

**Screenshots:**
![image](https://user-images.githubusercontent.com/56930858/95218984-7731dc80-07ba-11eb-9450-a2eadf796de8.png)

**Additional Information:**
It appears that the only achievements that have issue with this are the Gym and Dungeon ones, and I can see why; the title of the achievement ("Viridian Forest Expert", "Viridian Forest Explorer", "Viridian Forest Hermit") lists the location, while the subtitle lists what needs to happen for completion. While the subtitle is being pulled for the tracker, the title is not.

PR #883, opened 2020-10-06
Update Achievement Tracker Display

Mephistic:
This PR addresses two pieces of feedback we've received on the Achievement Tracker:
* Updates the achievement descriptions for gym/dungeons to include the name of the gym/dungeon (should fix https://github.com/pokeclicker/pokeclicker/issues/881)
* Adds the Achievement Name to the tracker display

Screenshot: 
Screen Shot 2020-10-06 at 1 23 15 PM


SpenserJ (approved):
Looks great! :+1: I was a bit worried that this might affect the saved statistics (wasn't sure if they were index or name based), but everything works well. Thanks for submitting this

RedSparr0w:
I feel like the title makes the module almost a little too big, maybe we could make it optional somehow?

Other than that adding Dungeon + Gym to description is good 👍

Mephistic:
Yeah, I left out the achievement name in the original PR because of the extra screen real estate it took up (but I made that decision before I realized we could freely move these modules around, so maybe that mitigates the UI pain here?).

The real issue here was just that it wasn't clear which gym/dungeon was being tracked - the name is probably less important now that we've resolved that. I'll revert this to only include the description updates.

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -233,15 +233,16 @@ class AchievementHandler {
...
             GameConstants.RegionGyms[region]?.forEach(gym => {
-                AchievementHandler.addAchievement(`${gym} Gym regular`, 'Clear 10 times', new ClearGymRequirement(10, Statistics.getGymIndex(gym)), 1, region);
-                AchievementHandler.addAchievement(`${gym} Gym ruler`, 'Clear 100 times', new ClearGymRequirement( 100, Statistics.getGymIndex(gym)), 2, region);
-                AchievementHandler.addAchievement(`${gym} Gym owner`, 'Clear 1,000 times', new ClearGymRequirement(1000, Statistics.getGymIndex(gym)), 3, region);
+                const gymTitle: string = gym.includes('Elite') || gym.includes('Champion') ? gym : `${gym} Gym`;
+                AchievementHandler.addAchievement(`${gymTitle} Regular`, `Clear ${gymTitle} 10 times`, new ClearGymRequirement(10, Statistics.getGymIndex(gym)), 1, region);
+                AchievementHandler.addAchievement(`${gymTitle} Ruler`, `Clear ${gymTitle} 100 times`, new ClearGymRequirement( 100, Statistics.getGymIndex(gym)), 2, region);
+                AchievementHandler.addAchievement(`${gymTitle} Owner`, `Clear ${gymTitle} 1,000 times`, new ClearGymRequirement(1000, Statistics.getGymIndex(gym)), 3, region);
```
RedSparr0w, line 239: Do we need these to be capitalized?
```suggestion
                AchievementHandler.addAchievement(`${gymTitle} regular`, `Clear ${gymTitle} 10 times`, new ClearGymRequirement(10, Statistics.getGymIndex(gym)), 1, region);
                AchievementHandler.addAchievement(`${gymTitle} ruler`, `Clear ${gymTitle} 100 times`, new ClearGymRequirement( 100, Statistics.getGymIndex(gym)), 2, region);
                AchievementHandler.addAchievement(`${gymTitle} owner`, `Clear ${gymTitle} 1,000 times`, new ClearGymRequirement(1000, Statistics.getGymIndex(gym)), 3, region);
```
As we are already also adding to old save file sizes with each of the Elite, Champions no longer including `Gym` in their title.

This isn't so much a problem for newer saves, but for saves that have a lot of achievements this could be quite a bit bigger.

image

_side note: it also looks like we went from `tourist` → `regular` at some point too 🤔, we should probably only be saving achievements that currently exist_

Mephistic: Fair point - I thought it read better capitalized, but if it means we can put off switching the logic to use ids instead of names, I'll switch the spelling back.

RedSparr0w (approved):
Changes look good, thanks for the PR 👍

PR #885, opened 2020-10-06
Add 1px space between type sections of damage calc bar

Aegyo:
To help with identifying where one section ends and another begins. Useful when the sections are very thin, or the player has difficulty distinguishing the colours.


SpenserJ (approved):
Great improvement :+1: Thanks!

Old:
image

New:
image

PR #887, opened 2020-10-06
Make eggs all fit on one line when in middle column

Aegyo:
Eggs are huge when the module is dropped into the middle column, this PR makes them squeeze into a single line like mobile view.

Images of how it looks with this addition:

In the middle column
Middle Column

In a side column
Side Column

On mobile
Mobile View

Using images I uploaded to discord... hopefully they don't disappear

### DIFF
--- a/src/styles/breeding.less
+++ b/src/styles/breeding.less
@@ -139,6 +139,11 @@
   background-color: #aaa;
 }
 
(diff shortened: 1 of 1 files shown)


SpenserJ (approved):
Looking great! :+1: :tada:

PR #888, opened 2020-10-06
Fixup the link to images for Pokémon on the dynamic background

RedSparr0w:
Currently the images are linking to `/assets/` which will be looking from the root of the site.

This is only really a problem for pages that aren't hosting it on the base domain such as GitHub pages.


SpenserJ (approved):
Looks good. We will need to remember this down the road if we replace our gulp scripts with a single webpack config for transpilation and bundling

RedSparr0w:
_just for future reference_
This is also the reason we need to use `../assets/` in the css instead of `/assets/` or `assets/`

Issue #877, opened 2020-10-06
[BUG] Breeding filter order is backwards


SpenserJ:
**Description:**
When using sort direction in the day care screen, "Ascending" puts the highest attack/most efficient pokemon at the top and "Descending" puts the lowest at the top. Usually, ascending lists increase as you go through the list and descending decrease.

nls0:
Applies to any sort option where invert is true.

walkerboh:
It looks like the issue is only for certain fields.
Pokemon #, Name, and Egg steps are correct.
Base Attack and Attack are out of order. 
There may need to be a case by case basis for how the sorting works rather than just "reversing" the sorting.

PR #889, opened 2020-10-06
Removed sort option invert to fix sorting

walkerboh:
In reference to #877 
I'm not sure the original purpose of the invert flag, but it was causing Attack, Base Attack, and Breeding Efficiency to be sorted in incorrect order when looking at ascending vs descending. Now they all have the same behavior.

Feel free to close this PR if the invert is still necessary for something I didn't see.

### DIFF
--- a/src/scripts/party/PartyController.ts
+++ b/src/scripts/party/PartyController.ts
@@ -45,15 +45,12 @@ class PartyController {
 
     public static compareBy(option: SortOptions, direction: boolean): (a: PartyPokemon, b: PartyPokemon) => number {
         return function (a, b) {
-            let res, dir = (direction) ? -1 : 1;
+            let res: number;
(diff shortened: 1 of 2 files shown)


RedSparr0w:
I believe the original intent here was that the default be by "best",
other than ID and name which you would expect them to be ASC by default as they have no "best".

We could just replace the ASC/DESC select with a toggle similar to the Pokémon list which is a bit more ambiguous, as it probably doesn't need an entire select dedicated to it either.

walkerboh:
If we want to go that route, we should and something that tells you it is sorted that way, or then we should make all the values as possible display values for the hatchery. Otherwise I could switch the orders, forget, and not know if it is in the "best first" mode or not. If we can display the values then we can at least see that it isn't in the order we want.

PR #891, opened 2020-10-07
Add bomb upgrades to Underground

RedSparr0w:
Added Bomb upgrades to the Underground.

Level | Effeciency | Cost to reach
--- | --- | ---
0 | 10 | `none`
1 | 12 | 50 💎
2 | 14 | 100 💎
3 | 16 | 150 💎
4 | 18 | 200 💎
5 | 20 | 250 💎


src/scripts/underground/UndergroundUpgrade.ts
```diff
@@ -53,3 +53,10 @@ Underground.upgradeList.push(
         GameHelper.createArray(0, 2, 1)
     )
 );
+
+Underground.upgradeList.push(
+    new UndergroundUpgrade(Underground.Upgrades.Bomb_Efficiency, 'Bomb Efficiency', 5,
+        AmountFactory.createArray(GameHelper.createArray(50, 500, 50), GameConstants.Currency.diamond),
+        GameHelper.createArray(0, 10, 2)
```
Aegyo, line 60: There are only 5 steps in the bonus array, but the cost array has 10 steps (max cost is only 250, not 500). Not really a bug, but makes me think one of these numbers maybe isn't what you intended?

Aegyo (approved):

src/scripts/underground/UndergroundUpgrade.ts
RedSparr0w: Ah that makes sense, I wasn't really sure what the array did, as it's not too self explanatory 😅 
What is the first number used for?
Also what are the first/last numbers in the Amount Factory?

Aegyo: Yeah, it is `start, end, increment`, so you either want your bonus array to be built with `(0, 10, 1)` for 10 steps, or the amount array built from `(50, 250, 50)` to just make the 5 steps we actually have a bonus for.

RedSparr0w, line 60: I'm honestly not sure if the bonuses are too cheap/expensive as these are just the first numbers that I tried.
Do you think increments of 1 would work out better?

But the arrays make sense to me now thanks.
_(I probably should have just run the array constructor individually to figure it out)_
```suggestion
        AmountFactory.createArray(GameHelper.createArray(50, 250, 50), GameConstants.Currency.diamond),
        GameHelper.createArray(0, 10, 2)
```

Aegyo: The total cost to upgrade does seem a little cheaper than the others, except daily deals, but I think it is ok like this.

Aegyo (approved):

PR #892, opened 2020-10-07
Apricorn

HLXII:
This isn't fully implemented yet, I just wanted to check whether this feature might be added to the game. Should also be pulled into its own feature branch since there's a lot of development here.

A couple of notes for myself on what's currently not implemented:

* Get Images for Apricorn tree growth from Seed > Sprout > Sapling > Tree > Tree with Berry
  * Update ApricornController.ts to point to the new image paths
  * Also images for Apricorns themselves.
* ApricornFarming.ts
  * Set properties of each Apricorn type (growing time from Seed > Tree, harvesting time from Tree -> Berry, and number of harvests until the tree dies).
  * Find multiplying effects (maybe Sprayduck?). More on the Mulch idea below.
  * Setup notifications. One of Fully grown tree, one on apricorn ripe, one on apricorn overripe and falling.
 * ApricornPlot.ts
   * Setup Mulches - Using the first 4 [Mulches](https://bulbapedia.bulbagarden.net/wiki/Mulch) from Gen IV. Growth will speed up growing/harvesting time. Damp will give multiple apricorns per harvest (usually 1), Stable will always return apricorns on overripe ones dropping. Gooey will replant automatically when the tree dies.
     * Not sure yet whether the mulches can stack on each other, or only one at a time. Can also maybe use the mulches (or the other 4 versions) on Berries as well.
     * Also need to figure out who will sell these.
   * Update this.stage. Not sure how that'll be calculated yet.
 * GameConstants.ts - Update Apricorn distribution to be better (currently just copied berry distribution)
 * Location on Map - Will be diagonally across CherryGrove. Unlocked by Apricorn box (bought at Cherrygrove).
 * Kurt - Will be NPC in Azalea Town. Talking to him opens a shop modal to trade Apricorns for new pokeballs. Will be unimplemented until the new pokeball situation is figured out.
 * Achievements for Apricorns
 * Quests for Apricorns

I can do a full writeup of the details of this mechanic if that's necessary, instead of people parsing through what I've already implemented. I only covered what I haven't done in the comments above.


RedSparr0w:
I may just be missing it myself, but I'm not quite seeing the reasoning behind adding this compared with the current farm situation?

This also seems to have a ton of similarities to the current farm and looks like it could be sharing a lot of the current classes.

HLXII:
Apricorns will be used for the new pokeballs (hopefully). If we're not adding those any time soon, then we don't need this feature I guess. This mechanic is different from the Berry farm also in that it has more depth to it. The Berry Farm mechanic is fairly plain in that you plant stuff, wait, then harvest and repeat.

Apricorn trees will grow, last for multiple harvest cycles (which the player can speed up by manually harvesting), and then die for replanting. I'm also adding Mulches to add for some additional choices for the player.

As for sharing classes, I'd have to create a parent Farming class to handle that. Wasn't really thinking about that at the time, so I can do that next. 

I also had some ideas for adding more details to the Berry Farm mechanic to make it different from Apricorns, but this PR was more for checking if my Apricorn mechanic could be added to the game. If this idea is approved, I'll have an idea on what ideas for the Berry farm would be approved as well.

RedSparr0w:
closing as #973 seems to be a replacement?

PR #894, opened 2020-10-07
feat(breeding): adds setting to control egg hatching animation

### LINKED ISSUE #875: [SUGGESTION] Add setting to disable hatch animations
Hatch animations should be bound to a setting so that they can be disabled if the user does not want them. The [WCAG](https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html) guidelines for accessible sites recommends allowing automatic animations to be stopped to assist users that find them distracting.

**Current Behavior**

Animations play automatically when an egg is 70% ready to hatch

SpenserJ:
I defaulted it to `full` instead of `almost` because there were quite a few questions about whether it was a bug that it was wiggling before it could hatch. This closes #875


src/scripts/settings/Settings.ts
```diff
@@ -106,6 +106,14 @@ Settings.add(new MultipleChoiceSetting('backgroundImage', 'Background image:',
     ],
     'background-day'
 ));
+Settings.add(new MultipleChoiceSetting('eggAnimation', 'Egg Hatching Animation:',
+    [
+        new SettingOption('None', 'none'),
+        new SettingOption('Fully ready', 'full'),
+        new SettingOption('Start wiggling at 50 steps left', 'almost'),
```
RedSparr0w, line 113: Could we make this a little bit shorter, maybe something similar to:
```suggestion
        new SettingOption('Almost & Fully ready', 'almost'),
```
```suggestion
        new SettingOption('Almost ready', 'almost'),
```

RedSparr0w (changes requested):
Looks good, but I think we should shorten the wording for the `almost` option.

src/scripts/settings/Settings.ts
SpenserJ: Done! I went with `Almost & fully ready` (lowercase f for readability) just to make it clear that you get both animations instead of just when they're almost ready

RedSparr0w (approved):

PR #895, opened 2020-10-07
Fix route 35 skip

Ultima1990:
If Plain badge is obtained first, route 35 can currently be skipped. Not intended.
Changed route 36 to require route 35 clear in addition to Plain badge.


Mephistic:
In-game, Route 36 doesn't require Route 35 to be accessed - you can also get to Route 36 by backtracking to Violet City once you have the Plain Badge. Route 36 is a three-way intersection between Route 37 (to Ecruteak), Route 35 (to Goldenrod), and Violet City that's blocked by Sudowoodo - you just need the Plain Badge in order to fight and move past Sudowoodo.

See https://bulbapedia.bulbagarden.net/wiki/Johto_Route_36

RedSparr0w (dismissed):
👍 

_**Edit:** Didn't see the above message_
So should it be staying as is or?

Ultima1990:
Yup, forgot the backtracking option.

PR #897, opened 2020-10-07
chore: require linting to pass before you can commit

SpenserJ:
This will use [husky](https://www.npmjs.com/package/husky) to run `npm run lint` prior to allowing you to commit your code. I'm not sure if everyone will like this workflow, so we can always change it to `pre-push` or even skip the PR entirely if we want. If you want to bypass the check (if you know there are linting issues but you're doing a bigger change and will fix it later, or you've already linted) you can use `--no-verify` (e.g., `git commit --no-verify -m 'Did something'`) to skip the check temporarily, or set the environment variable `HUSKY_SKIP_HOOKS=1` to bypass it more permanently


Aegyo:
Sorry I didn't get to this one sooner. Looks good when I test it, but could you fix the conflict github is complaining about now (and maybe rebase/merge develop for me) please?

SpenserJ:
Rebased and should be good to merge now :+1:

RedSparr0w (approved):
Awesome 👍 

Thanks for the PR.

PR #903, opened 2020-10-08
fix achievement dont refiltering after complete

LuqDragon:
achv
this causes another problem because reset the pages after that, so if you seeing achievement on another page instead of page 1, you will be redirected to page 1, but the tracker is for tracking, so I don't see a real problem in this.

another better way to do that will be welcomed


SpenserJ (commented):
I think this should probably keep the user on the page that they were on. It isn't perfect, because if you're looking at the first achievement on page 10 and one before it is completed, the one you were looking at will now be on the bottom of page 9, but it'll work for any other achievement on the page.

You can implement this by accepting an argument in the `filterAchievementList` function that skips the call to `this.resetPages()`, and passing it a non-default value when the `Achievement` class calls `AchievementHandler.filterAchievementList`:

**AchievementHandler.ts**
```diff
-     public static filterAchievementList() {
+     public static filterAchievementList(retainPage = false) {
          this.achievementListFiltered(this.achievementList.filter(
              a => a.region <= player.highestRegion() &&
              (this.filter.status() == 'all' ? true : a.unlocked == JSON.parse(this.filter.status())) &&
              (this.filter.type()   == 'all' ? true : a.property.constructor.name == this.filter.type()) &&
              (this.filter.region() == 'all' ? true : a.region == +this.filter.region())
          ));
-         this.resetPages();
+         if (!retainPage) {
+             this.resetPages();
+         }
      }
```

**Achievement.ts**
```diff
              player.achievementsCompleted[this.name] = true;
              this.unlocked = true;
-             AchievementHandler.filterAchievementList();
+             AchievementHandler.filterAchievementList(true);
          }
```

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -33,12 +33,19 @@ class AchievementHandler {
...
                                     (this.filter.type()   == 'all' ? true : a.property.constructor.name == this.filter.type()) &&
                                     (this.filter.region() == 'all' ? true : a.region == +this.filter.region())));
-        this.resetPages();
+        if (!retainPage) {
+            this.resetPages();
+        } else if (this.getAchievementListWithIndex().length == 0) {
+            if (this.navigateIndex() > 0) {
+                this.calculateNumberOfTabs();
+                this.navigateIndex(this.navigateIndex() - 1);
+            }
```
SpenserJ, line 47: This is great! We can make it just a bit cleaner by combining the if statement and reusing the `numberOfTabs()` value that you updated
```suggestion
        } else if (this.getAchievementListWithIndex().length == 0 && this.navigateIndex() > 0) {
            this.calculateNumberOfTabs();
            this.navigateIndex(this.numberOfTabs())
```

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -41,14 +41,19 @@ class AchievementHandler {
...
-    public static filterAchievementList() {
+    public static filterAchievementList(retainPage?: boolean) {
         this.achievementListFiltered(this.achievementList.filter((a) => (
             a.region <= player.highestRegion() &&
-            (this.filter.status() == 'all' || a.unlocked == JSON.parse(this.filter.status())) &&
-            (this.filter.type()   == 'all' || a.property.constructor.name == this.filter.type()) &&
-            (this.filter.region() == 'all' || a.region == +this.filter.region())
+              (this.filter.status() == 'all' ? true : a.unlocked == JSON.parse(this.filter.status())) &&
+              (this.filter.type()   == 'all' ? true : a.property.constructor.name == this.filter.type()) &&
+              (this.filter.region() == 'all' ? true : a.region == +this.filter.region())
```
RedSparr0w, line 49: Is there any reason for this change?
As the logic still seems to be the same.

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -41,14 +41,19 @@ class AchievementHandler {
         return this.achievementListFiltered().slice(this.navigateIndex() * 10, (this.navigateIndex() * 10) + 10);
     }
 
-    public static filterAchievementList() {
+    public static filterAchievementList(retainPage?: boolean) {
```
RedSparr0w, line 44: Can we make this explicitly false, instead of falsey
```suggestion
    public static filterAchievementList(retainPage = false) {
```

src/scripts/achievements/AchievementHandler.ts
LuqDragon: maybe some of merges did that, i dont edited this part

LuqDragon: solved

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -41,14 +41,19 @@ class AchievementHandler {
...
             (this.filter.status() == 'all' || a.unlocked == JSON.parse(this.filter.status())) &&
             (this.filter.type()   == 'all' || a.property.constructor.name == this.filter.type()) &&
             (this.filter.region() == 'all' || a.region == +this.filter.region())
         )));
-        this.resetPages();
+        if (!retainPage) {
+            this.resetPages();
+        } else if (this.getAchievementListWithIndex().length === 0 && this.navigateIndex() > 0) {
+            this.calculateNumberOfTabs();
+            this.navigateIndex(this.numberOfTabs());
```
RedSparr0w, line 55: Should this be 1 less due to `this.navigateIndex()` being 0 indexed?
```suggestion
            this.navigateIndex(this.numberOfTabs()  - 1);
```

LuqDragon: true

RedSparr0w (approved):

PR #906, opened 2020-10-08
feat(achievements): adds buttons to jump to first and last pages

SpenserJ:
There was a request for a way to move between achievement pages faster. Implementing a dropdown list didn't look very nice (mainly because the total count would either be beside the list or duplicated in each row), so I went with a start/end button instead. I replaced the arrows with new SVG images because the previous ones had artifacts in dark themes, and these seem to work well with all but one theme (where the background of the button is also dark, which is no change from the previous images).

image
image
image

This also fixes two bugs with the page count:
* If the achievement filter returned an empty array, the modal would show `0/1` pages
* If there were a number of items ending in 1 (11, 21, 31), the last item would be on an inaccessible page


RedSparr0w (changes requested):
Looks good, definitely should help a bit with navigation.

Can we just use the 2 different arrow styles then flip (or rotate) them for the opposite direction?
Also can you please run the `svg` through https://jakearchibald.github.io/svgomg/ to remove any things that aren't needed and reduce the file size.

SpenserJ:
> Can we just use the 2 different arrow styles then flip (or rotate) them for the opposite direction?
Done

> Also can you please run the `svg` through https://jakearchibald.github.io/svgomg/ to remove any things that aren't needed and reduce the file size (prettify is fine).

I had run them through [Scour](https://github.com/scour-project/scour), but I must have missed a few options. Managed to shave off another 130 bytes from each though :+1:

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -20,7 +20,7 @@ class AchievementHandler {
     }
 
     public static calculateNumberOfTabs() {
-        this.numberOfTabs(Math.floor((this.achievementListFiltered().filter(a => a.region <= player.highestRegion()).length - 1) / 10));
+        this.numberOfTabs(Math.min(1, Math.floor(this.achievementListFiltered().filter(a => a.region <= player.highestRegion()).length / 10)));
```
RedSparr0w, line 23: I think you may have reversed the logic here, as only 2 pages ever show up:
```suggestion
        this.numberOfTabs(Math.max(1, Math.floor(this.achievementListFiltered().filter(a => a.region <= player.highestRegion()).length / 10)));
```
image

SpenserJ: Oh goodness. I can't believe I did that xD Should be committed now

RedSparr0w:
One more problem I just noticed, when you have it filtered to show nothing, it shows 2 pages,
I guess this is due to no longer doing `-1` from the total length?:
image

SpenserJ:
That is so strange, because it showed `1/0` when the `- 1` was in there. I'll go back to the drawing board because my math and testing clearly had some flaws

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -20,7 +20,7 @@ class AchievementHandler {
     }
 
     public static calculateNumberOfTabs() {
-        this.numberOfTabs(Math.floor((this.achievementListFiltered().filter(a => a.region <= player.highestRegion()).length - 1) / 10));
+        this.numberOfTabs(Math.max(1, Math.floor(this.achievementListFiltered().filter(a => a.region <= player.highestRegion()).length / 10)));
```
RedSparr0w, line 23: I _think_ this solves the 1/2 pages when filtered for no achievements.
```suggestion
        this.numberOfTabs(Math.max(0, Math.floor(this.achievementListFiltered().filter(a => a.region <= player.highestRegion()).length / 10)));
```

SpenserJ:
Alright, I've done some more thorough testing and found a few more cases where it was incorrect. I've fixed up a few different bugs with the paging, cleaned up some of the logic to make it easier to verify, and tested it with multiple cases:
* 0-10 achievements after filtering: `1/1` pages and you can't use any buttons
* 11 achievements - `1/2` pages and you can go right, or `2/2` and you can go left
* 21+ achievements - Page count is correct, and you can go backward as long as you're on page 2+ (index greater than 0) and you can go forward as long as you're not on the last page (page index + 1 !== number of pages)

There was a subtle bug where the `numberOfTabs` property was zero-based (0 means there is one page), which is why the code originally had a `- 1` in it. This switches it to be 1-based, so we always have a minimum of 1 page, and instead of adding `+ 1` in multiple places (specifically rendering), we only add it when checking if we're on the last page

RedSparr0w (approved):

PR #908, opened 2020-10-08
Swapped modal sort direction to pokemon list style

### LINKED ISSUE #877: [BUG] Breeding filter order is backwards

walkerboh:
closes #877 
Replaced the sort direction dropdown with the icon from the Pokémon List to remove the ascending/descending confusion.


RedSparr0w (approved):
Looks good 👍 

Thanks for the PR

PR #909, opened 2020-10-09
Update Pokemon shop images

RedSparr0w:
Update the Pokemon shop item images:
image

This uses the images from [Pokemon Shuffle](https://www.spriters-resource.com/3ds/pokemonshuffle/),
With the current images it can be hard to make them fit nicely in the window after resizing and trimming etc.

I've shrunk them from `256x256` to `120x120` (same size as the other items).
And also have been run through tinypng.


SpenserJ (approved):
These look great! :+1:

Issue #854, opened 2020-10-04
[TASK] Unova - Images


SpenserJ:
We're missing some images for the Unova region. This should be created as a PR to merge into the [`unova`](https://github.com/pokeclicker/pokeclicker/pull/749) branch.

Some of these assets can be downloaded from [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Main_Page) and then run through a tool such as [TinyPNG](https://tinypng.com) to compress them.
Environment backgrounds just need to be associated in [src/scripts/GameConstants.ts](https://github.com/pokeclicker/pokeclicker/blob/6a9df1b79ee02da2afa513d0d1801499f924fad4/src/scripts/GameConstants.ts#L373-L427) 

- [x] Towns
- [x] Gyms _(Environment)_
- [x] Gym Leaders/Elite 4 _(Environment)_ (Thank you @unclebanks)
- [x] Dungeons
- [x] Dungeons _(Environment)_
- [x] Badges (Thank you @unclebanks)
- [x] Pokemon
  - [x] Sprites
  - [x] Alternative forms
- [x] Underground Fossils

PR #910, opened 2020-10-09
Town images

HLXII:
For https://github.com/pokeclicker/pokeclicker/issues/854#issue-714427516

Found most of the images for towns and dungeons. A couple of notes:
Castelia City is huge and also 3D, so I just used the image they had on Bulbapedia
Relic Castle was a bunch of small maps, so I just used the final floor.
Celestial Tower also had a bunch of small maps. I chose an inner room rather than the top floor.
Dragonspiral Tower was also all 3D. I found a render of the top floor on DeviantArt, not sure if it should be used.
Also I have no idea what "Black and White Park" is, so that town doesn't have an image.


SpenserJ (changes requested):
Please retarget this to `develop`. Unova has been merged and is only still a branch while we port two PRs over from it.

PR #912, opened 2020-10-09
feat(battle): pokemon attack tooltip shows actual attack against enemy

SpenserJ:
There was a request to see the actual attack against the pokemon you're currently battling, without drilling into the damage calculator. Since we already have this information readily available when damaging the enemy pokemon, it was quite simple to add. I used a tooltip instead of having it displayed continuously both for space and because it may be confusing for new players when their damage output changes by a substantial amount.

Route
image

Dungeon
image

Gym
image

Battle Frontier
image


RedSparr0w:
Looks good to me!

What do you think about using a bootstrap tooltip instead though?
Or having it just display the calculated attack via a settings option? (default normal attack without type calculations)

SpenserJ:
I like the look of the bootstrap tooltip, since it has better positioning and the placement arrow is nice, but it doesn't appear to update as the KO bindings change (always shows the first pokemon you battled), and triggering it to update would be a huge performance hit when you're one-shotting. I don't know if anyone would want to have these values displayed continuously via a setting either, since they will change so dramatically based on what you're battling (the screenshots were all from the same save, with only 3 shard types maxed out).

Since bootstrap tooltips can't handle this without a huge perf hit, I'm fine if we want to close this unless you feel the setting would be useful to have.

RedSparr0w:
Strange, not sure why the bootstrap tooltip won't update on this, but it works on the Pokeball selector:
image

RedSparr0w (approved):
Changes look good to me!

Thanks for the PR 👍

PR #913, opened 2020-10-09
Update PokedexHelper.ts

mog-kupo:
Small change to show event Pokemon in the pokedex.


LuqDragon (commented):
now need to change the way to show them 😉 
image

anapple96 (commented):
Can't see an use for this, as we're moving the rest of the eventmons to `x.1` soon:tm:;

mog-kupo:
> Can't see an use for this, as we're moving the rest of the eventmons to `x.1` soon:tm:;

If that change is coming in the next update that's fine... Until then it's a quick fix to see the event mons.

RedSparr0w:
Updated to show `-` IDs as `???`

RedSparr0w (approved):
Thanks for the PR! 👍

PR #914, opened 2020-10-09
unova(images): Added town/dungeon images

HLXII:
I'm too lazy to figure out how to merge into develop, so I just made a new branch.

Notes copied from old PR:

For https://github.com/pokeclicker/pokeclicker/issues/854#issue-714427516

Found most of the images for towns and dungeons. A couple of notes:
Castelia City is huge and also 3D, so I just used the image they had on Bulbapedia
Relic Castle was a bunch of small maps, so I just used the final floor.
Celestial Tower also had a bunch of small maps. I chose an inner room rather than the top floor.
Dragonspiral Tower was also all 3D. I found a render of the top floor on DeviantArt, not sure if it should be used.
Also I have no idea what "Black and White Park" is, so that town doesn't have an image.


SpenserJ (approved):
This is a great set of images. It is a shame some of them are so small, since we'll have to figure out a better way of displaying them as the background or find alternatives, but I think it is a great start and should work until we're able to run through the new regions and test it all :+1:

PR #916, opened 2020-10-10
shortcut '1' to '9' to league

LuqDragon:
add numbers key to enter in league gyms, I put 1 to 9 cause we can apply to others if have more than 5.
trying to make more generic now.


src/scripts/GameController.ts
```diff
@@ -157,6 +157,11 @@ class GameController {
                         DungeonRunner.initializeDungeon(player.town().dungeon());
                     }
                     e.preventDefault();
+                } else if ('gymList' in player.town()) {
+                    const cap = keyCode > 48 && keyCode < 58 ? 48 : keyCode > 96 && keyCode < 106 ? 96 : -1; //1 to 9
+                    if (keyCode > cap && keyCode - cap <= player.town().gymList().length) {
+                        GymRunner.startGym(player.town().gymList()[keyCode - cap - 1]());
```
RedSparr0w, line 163: I'm not sure if it's just me, but the logic seems a little bit confusing,
Maybe we could simplify it a little bit with something like:
```suggestion
                    const number = keyCode > 48 && keyCode < 58 ? keyCode - 48 : keyCode > 96 && keyCode < 106 ? keyCode - 96 : 0; //1 to 9
                    if (number && number <= player.town().gymList().length) {
                        GymRunner.startGym(player.town().gymList()[number - 1]());
```
or something like this:
```suggestion
                    const numberKey = keyCode > 48 && keyCode < 58 ? keyCode - 48 : 0;
                    const numpadKey = keyCode > 96 && keyCode < 106 ? keyCode - 96 : 0;
                    const number = numberKey || numpadKey;
                    if (number && number <= player.town().gymList().length) {
                        GymRunner.startGym(player.town().gymList()[number - 1]());
```

LuqDragon: i liked the first one, but the second is more understandable

RedSparr0w: I also just noticed #780 was merged recently, so this will probably need to be modified.

SpenserJ: My vote is on the second one. I've never been a big fan of nested ternaries, especially when they're all on one line.

RedSparr0w (approved):
Working great 👍 

Thanks for the PR

PR #918, opened 2020-10-11
refactor: Use math to save us a bit of memory when picking shop egg pokemon

Aegyo:
Adds a function to pick a random element from an array, but exponentially distributed, and uses it in the typed egg selection

Should have the same egg rates as before, but we don't need to add many copies of each egglist when picking a random element.

A large test of using this function:
```js
testme = (n,r,t) => {
  const res = {};
  const arr = Array(n).fill().map((x,i)=>i);
  for (i=0;i<t;i++) {
    const x = GameConstants.expRandomElement(arr,r)
    res[x] = res[x] + 1 || 1
  }
  return res
}
```
```js
>> testme(10,2,100000)
<- Object { 0: 180, 1: 188, 2: 372, 3: 764, 4: 1561, 5: 3132, 6: 6229, 7: 12786, 8: 25064, 9: 49724 }
```


src/scripts/GameConstants.ts
```diff
@@ -296,10 +296,22 @@ namespace GameConstants {
...
         return array[GameConstants.randomIntBetween(0, array.length - 1)];
     }
 
+    // Return a random element from the array, with an exponential distribution
+    // The last element has a 1/ratio chance of being chosen, one before last is 1/(ratio^2), etc
+    // The logarithm is clipped up to 0, so the first two elements will have equal chance
+    export function expRandomElement<T>(array: T[], ratio: number): T {
+        const r = Math.random();
+        // Math.random() can be 0, and log(0) is NaN, so we turn it back to 0 if this happens.
+        const logr = (Math.log(r) || 0) / Math.log(ratio);
```
RedSparr0w, line 309: `Math.log(0)` seems to return `-Infinity` so I don't think this will work as expected?
```suggestion
        // Math.random() can be 0, and log(0) is -Infinity, so we turn it back to 0 if this happens.
        const logr = (r ? Math.log(r) : 0) / Math.log(ratio);
```

Aegyo: Huh, so it is...
In that case, we can just rely on the clipNumber taking care of it for us?

RedSparr0w: Yeah, just becomes the last element of the array for `-Infinity`

RedSparr0w (approved):
Changes look good to me 👍 

Thanks for the PR

PR #919, opened 2020-10-11
TypeScript Modules linting

SpenserJ:
This sets up the AirBnB ESLint Config (TypeScript edition) for the modules folder. This is a more substantial config than we are using in scripts, which is needed in order to help provide an earlier indication that the modules are being used incorrectly or will not compile properly.

I noticed that my previous PR created a new `GameConstants.ts` in the modules folder, but did not migrate everything away from the file in scripts, so this will also make that change. In order to avoid some quirks with namespaced declarations in the way that scripts are currently transpiled/minified, I had to take the autogenerated declaration and tweak it, or it would be extremely fragile and hard to automate. Moving forward, any changes to GameConstants will require updating the `GameConstants` file in both the module (the actual values used) and the scripts (a declaration of structure that the scripts can rely on).

This renames `BadgeStore` to `BadgeCase`, as I am not following through with my original plan to refactor the logic just yet. It is more important to get all of the code moved to modules, than it is to clean up some of the state management. I also renamed `BadgeTypes` to `BadgeEnums`, which more accurately represents the values and avoids confusion with the type of badge (gym, elite, etc).

Finally I refactored the `ArrayOfObservables` class into a function, as it really didn't need to be a class (it was returning the value from the constructor), and was just creating memory allocations that weren't used, and moved that into the modules folder.

## Testing

As with the previous PR, no real functionality has changed, so we only need surface-level testing. Please run `npm ci` when checking out this branch, and then run through a quick test of the common game features. There should be no compilation errors, no (new) console errors, and no linting errors.


src/modules/DataStore/StatisticStore.ts
```diff
@@ -130,12 +130,16 @@ export default class Statistics implements Saveable {
 
                     // If it's not an int or less than zero, we do not want to set it
                     const id: number = Math.floor(+prop);
-                    if (isNaN(id) || id < 0 || id != +prop) {
-                        if (isNaN(id)) {
+                    if (Number.isNaN(id) || id < 0 || id !== +prop) {
+                        if (Number.isNaN(id)) {
+                            // eslint-disable-next-line no-console
                             console.trace(`[Statistics] [${array}] Invalid property requested:`, prop);
```
RedSparr0w, line 136: I don't think this does that same thing as it used to.
Fairly certain we need `isNaN` here not `Number.isNaN`

SpenserJ: I'll double check all of this, but the `isNaN` global [should be avoided](https://github.com/airbnb/javascript#standard-library--isnan) in most cases. I will probably replace some of the places that were coercing numbers with `+prop` just so it's clear and reliable though

package.json
```diff
@@ -7,7 +7,7 @@
     "start": "cross-env NODE_ENV=development gulp",
     "test": "npm run ts-test && npm run eslint",
     "ts-test": "gulp scripts",
-    "eslint": "eslint --config ./.eslintrc.js --ext ts ./src/scripts",
+    "eslint": "eslint --ext ts ./src/scripts ./src/modules",
     "eslint-fix": "eslint --config ./.eslintrc.js --ext ts --fix ./src/scripts",
```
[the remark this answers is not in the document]
SpenserJ: This should be fixed now :+1: Good catch

src/modules/DataStore/StatisticStore.ts
SpenserJ: I've switched some of the places that we coerced numbers with `+prop` to `Number(prop)`. There shouldn't be any real difference with them (although `Number(...)` is more strict than `parseInt(..., 10)` would be), and it will continue to work with casting booleans to their `0` and `1` values

src/modules/DataStore/StatisticStore.ts
```diff
@@ -114,108 +117,122 @@ export default class Statistics implements Saveable {
         'shinyPokemonDefeated',
         'shinyPokemonEncountered',
         'shinyPokemonHatched',
-    ]
+    ];
 
     constructor() {
-        for (const prop of this.observables) {
-            this[prop] = ko.observable(0).extend({ numeric: 0 });
```
RedSparr0w, line 121: Is there a reason we are removing all the extensions?

As with just `<number>` we can still end up with decimal values which has happened in the past.

src/modules/enums/Badges.ts
```diff
@@ -1,4 +1,4 @@
-enum BadgeTypes {
+enum BadgeEnums {
```
RedSparr0w, line 1: 👍

RedSparr0w (approved):
👍 Looks good to me,

Just the one thing about the `extends` that I am unsure of.

src/modules/DataStore/StatisticStore.ts
SpenserJ: I had looked through everything using these values and didn't see anything that could generate decimal numbers. Knockout extenders don't play nicely with TS (a lot of errors and complexity around `extend does not accept object property "numeric"`), and I was hoping it was some dead code, since `numeric` also had support for the level of decimal precision but everything used it to just strip all decimals. I'll see if I can find a clean way of restricting the observable and push a fix :+1:

gulpfile.babel.js
```diff
@@ -174,7 +174,7 @@ gulp.task('scripts', () => {
         // Remove default exports
         .pipe(replace(/(^|\n)export default \w+;/g, ''))
         // Replace imports with references
-        .pipe(replace(/(^|\n)import .* from '(.*)((.d)?.ts)?';/g, '$1/// <reference path="$2.d.ts"/>'))
+        .pipe(replace(/(^|\n)import (.* from )?'(.*)((.d)?.ts)?';/g, '$1/// <reference path="$2.d.ts"/>'))
```
RedSparr0w, line 177: Should this be using the 3rd match now?
```suggestion
        .pipe(replace(/(^|\n)import (.* from )?'(.*)((.d)?.ts)?';/g, '$1/// <reference path="$3.d.ts"/>'))
```

SpenserJ: Good catch. Not sure how it didn't crash out on me. Just goes to show how loose the `scripts` code is when it comes to checking type safety I guess

PR #921, opened 2020-10-11
Boosts attack% per breed for Hoenn+

anapple96:
5 extra % from all regions, starting from Hoenn


Heppity:
Wait what does this mean? SO if you breed in Sinnoh it would be 35% of pokemon attack?

PR #922, opened 2020-10-11
Initial SVG implementation for Unova

HLXII:
Notes:
I've pulled the map image from anapple's branch. Not sure if that's the final version though.

Not entirely sure what town exists where Black City/White Forest is. I've left it as "???" for now.
There doesn't seem to be a dock in Unova right now. Not sure if that just wasn't implemented yet, so I commented out the dock for now.
Castelia city has a dungeon, and what seems to be some type of machine/gate thing near it. I'm not sure what both of these are referring to, but I added elements for both.
There are comments for Desert Resort and Undella Bay to become routes. I've still added elements for both dungeons for now.


RedSparr0w:
I haven't had a chance to look at all the changes yet,
But if you are updating the map image `.png` you also need to update the tiled map so it can be reproduced/edited.

HLXII:
I got the updated map from anapple. Not sure if they commited the tiled map, since the .tmx file on the branch wasn't updated either. https://github.com/anapple96/pokeclicker/tree/UnovaMap Here's the branch I got it from.

anapple96:
Delay this PR. There's some changes going up that'll require a remake of the base and SVG map

HLXII:
Closed due to major refactoring of routes by Apple. They'll handle the SVG implementation.

PR #923, opened 2020-10-11
fix(notification): level the gain of sounds to match

SpenserJ:
The notification sounds all played at the same volume percentage, but had different peak gains in the mp3 files. I've run them through `mp3gain -r -k` to balance them against each other and to avoid clipping. They all appear to be about equal, however they may be slightly quieter than desired due to avoiding clipping on some tracks.

If you'd like to compare the new versions to the current, please copy the `./src/assets/sounds` folder prior to checking out. There is a noticeable difference on the quest sounds (quieter), and a more subtle increase in volume on the harvest sound.


RedSparr0w:
To test out sounds from this PR:
```js
Object.values(NotificationConstants.NotificationSound).forEach((sound, index) =>
setTimeout(() => sound.sound.play(), index * 2e3));
```
https://pokeclicker.com:
```js
Object.values(GameConstants.NotificationSound).forEach((sound, index) =>
setTimeout(() => sound.sound.play(), index * 2e3));
```

RedSparr0w (approved):
👍 Sounds good to me!

PR #925, opened 2020-10-11
Fix/berry display

HLXII:
This was how it looked before:
image

This is how it looks now:

image

Sprout stage also appears now:

image

Was also considering updating the sprites to newer ones, but that might be for later as that's a lot of images.


RedSparr0w:
Just in case anyone was curious about the full growth cycle:
grow

RedSparr0w (approved):
Changes look good to me 👍 

Thanks for the PR.

PR #926, opened 2020-10-12
Fix player._itemList having strange text

### LINKED ISSUE #803: [BUG] Other items being added to player.itemList

RedSparr0w:
This PR should hopefully fix #803 

No name was being sent to the PokeBlocks constructor, so it would use `window.name` which would sometimes add other strange characters.

Example when using `window.name = 'test'`
image

Not sure why it just prepends `test` but it was just the other text beforehand.


SpenserJ (approved):
Fixes the issue for me :+1:

PR #928, opened 2020-10-12
Remove GameConstants.RegionRoute

RedSparr0w:
No longer use `GameConstants.RegionRoute`

Simplify some logic.


SpenserJ (approved):
:+1: Looks like a good cleanup

PR #931, opened 2020-10-12
Check if egg hatched before replacing the egg

### LINKED ISSUE #930: [BUG] The hatchPokemonEgg() function can erase said egg and pokemon
When clicking on an egg in the hatchery, the _hatchPokemonEgg()_ function is called.

/src/script/breeding/Breeding.ts

```
public hatchPokemonEgg(index: number): void {
        const egg: Egg = this._eggList[index]();
        egg.hatch();
        this._eggList[index](new Egg());
        this.moveEggs();
    }
```
The _egg.hatch()_ function checks if the egg is ready to hatch and does nothing if it is not.
However the egg is always overriden by a new empty egg right after without checking if it actually hatched. If it did not hatch, the pokemon is not added back into the pokemon list.

This should not be an issue at the moment as the egg buttons in the hatchery are checking if the egg is ready to hatch.
But it could become an issue if these buttons script gets changed or if another way to hatch eggs gets added later.

By the way, the egg class has a function called _canHatch()_, but the egg button does not use this to check if an egg is ready to hatch.

RedSparr0w:
closes #930 

Added a check to see if the egg was actually hatched before replacing it with a new empty egg.


SpenserJ (approved):
:+1: Looks great!

PR #927, opened 2020-10-12
Minor fixes for themes

RedSparr0w:
Fixed a few minor issues in some themes,
Also rearranged the `themes.less` file to be alphabetical.

### Slate:
Fixup the Achievement bonus text:
_before_
image
_after_
image

### Minty:
Fixup the Underground item count text/button
_before_
image
_(also button wasn't full size)_
image

_after_
image

closes #879

PR #932, opened 2020-10-13
Fix minor button UI styling for Underground

HLXII:
Before: (the button actually extends into the skip layer button).
image

After:
image

We also might want to create a CSS class for this, since the info button styling is used in a lot of places.


RedSparr0w:
This should also already be fixed in #927

PR #935, opened 2020-10-13
Allow user to advance route with '=' key

nls0:
...so they don't have to press 'shift' + '=' when not using numpad.


RedSparr0w (approved):
Looks good 👍 

Thanks for the PR

PR #939, opened 2020-10-14
Rework route, add support for named routes

anapple96:
Basically for Unova+ regions


src/scripts/wildBattle/Routes.ts
```diff
@@ -1,5 +1,7 @@
 /// <reference path="../../declarations/enums/Badges.d.ts"/>
 ///<reference path="RegionRoute.ts"/>
+///<reference path="../../declarations/enums/Badges.d.ts"/>
```
SpenserJ, line 3: Badge enums are referenced twice, although the scripts are (frighteningly) forgiving with references

SpenserJ (changes requested):
Looks like all of my route kill statistics and achievements have been reset, but everything else seems to be working well :+1:

anapple96:
Should be fixed now

SpenserJ (changes requested):
Works when I import a save prior to checking out the branch (although I'm notified of achievements I had already completed), but still doesn't work if I import a 0.5.6 save after checking out the branch. This picture is after importing a save that had all of Kanto completed, without running any of the routes post-import
image

src/scripts/Update.ts
```diff
@@ -271,6 +271,42 @@ class Update implements Saveable {
...
+
+                    // Ensure the region has been prepared
+                    acc[region] = (acc[region] || {});
+                    // Track the route with its number in the statistics
+                    acc[region][nextIndex] = nextValue;
+                    return acc;
+                }, {});
+                saveData.statistics.routeKills = result;
+                // Update save data
+                this.setSaveData(saveData);
```
SpenserJ, line 304: I've worked out a simple way of migrating the achievements, so we don't spam the player with a hundred achievement completed notifications while leaving the old achievement names in the state. This will match any `Route * (traveler|explorer|conqueror)` achievements, find the region, and prefix it onto the achievement name.

```suggestion
                this.setSaveData(saveData);

                // Migrate the achievements so we don't spam players with notifications
                const renamedAchievements = Object.entries(playerData.achievementsCompleted)
                    .map(([name, isCompleted]) => {
                        const matchRoute = name.match(/^Route (\d+) (?:traveler|explorer|conqueror)/);
                        // If the name doesn't match a route, return the old key-value pair
                        if (matchRoute === null) {
                            return [name, isCompleted];
                        }

                        const routeNumber = matchRoute ? Number(matchRoute[1]) : null;
                        if (Number.isNaN(routeNumber)) {
                            console.trace('[Update] Could not map region into achievement name:', name);
                            return [name, isCompleted];
                        }

                        // Look up the region for the route, and rename the achievement
                        const [region] = Object.entries(regionRoutes).find(([, check]) => (
                            // Find the region that contains this index
                            check[0] <= routeNumber && routeNumber <= check[1]
                        )) || ['none'];

                        if (region === 'none') {
                            console.trace('[Update] Could not map region into achievement name:', name);
                            return [name, isCompleted];
                        }

                        return [`${GameConstants.camelCaseToString(region)} ${name}`, isCompleted];
                    });
                playerData.achievementsCompleted = Object.fromEntries(renamedAchievements);
                this.setPlayerData(playerData);
```

SpenserJ (changes requested):
Other than my comment for migrating achievements, everything looks great! :+1:

SpenserJ (approved):
Looks great! :+1:

PR #940, opened 2020-10-14
Add breeding queue functionality

RedSparr0w:
Added Breeding/Hatchery Queue functionality.
Added new setting to hide the hatchery screen (never, when egg slots full, when queue full).
Changed the "egg slots full" warning to take up less space.

_Currently there is no way to obtain any queue slots without the console, not sure on the best way._

Every Pokémon added to the queue their attack is taken away from total attack straight away (as if they were breeding).

If there is Pokémon in the queue, eggs will auto hatch and be replaced by the first Pokémon in the queue.

If the queue is empty, the eggs will not auto hatch.

Click the Pokémon in the queue to remove them from the queue.

Example images:
image
image
imageimage

### DIFF
--- a/src/components/breeding.html
+++ b/src/components/breeding.html
@@ -23,6 +23,13 @@ aria-labelledby="breedingModalLabel" aria-hidden="true">
                 <li class="nav-item"><a class="nav-link" href="#breeding-fossils" data-toggle="tab">Fossils</a></li>
             </ul>
 
+            <p class="bg-danger my-0" data-bind="visible: !App.game.breeding.hasFreeEggSlot()">
+                You don't have any free eggslots.
+            </p>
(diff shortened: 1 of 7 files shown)


SpenserJ (changes requested):
I haven't tested so I could be wrong, but this doesn't appear to handle adding/removing eggs and fossils in the queue. We're going to need some additional logic so that cancelling an egg/fossil returns the item to you as well

RedSparr0w:
>this doesn't appear to handle adding/removing eggs and fossils in the queue.

The Eggs and Fossils aren't supported in the queue, just Pokémon,
They get disabled when the hatchery is full.

SpenserJ (approved):
:+1: Good to merge once you fix the bug and tweak the setting name like we discussed in Discord. This is a great QoL improvement :grin:

PR #943, opened 2020-10-15
feature(battle): use bootstrap tooltip for actual attack against

SpenserJ:
Our KO binding handler for tooltips wasn't responding to the tooltip value being updated. This will dig into the bootstrap tooltip data and update the content of the tooltip when it changes. This may be a slight performance hit when updating certain tooltips, so I tried to guard it by verifying that the tooltip was visible and doing direct DOM manipulation instead of relying on jQuery


Aegyo (approved):
:+1:

PR #944, opened 2020-10-15
Check if mine is completed on load

nls0:
When the mine is completed, there is a 1500ms timeout created for the function that generates a new mine. If the game autosaves during this 1500ms interval and the game is closed before the next autosave occurs, the game will be stuck with the old, completed mine.

This intends to fix that issue by checking for mine completion on load.


RedSparr0w:
I think it might be better to just check the mine completion state each time the "energy" meter is checked,
As this has also been a problem for people that don't close their game window.

nls0:
Okay, I will make `checkCompleted()` public, and by  
>each time the "energy" meter is checked  
do you mean `get energy()` (~2 calls per second) or `gainEnergy()` (~1 call per minute)?

SpenserJ:
> do you mean `get energy()` (~2 calls per second) or `gainEnergy()` (~1 call per minute)?

`gainEnergy()` should be sufficient

src/scripts/underground/Mine.ts
```diff
@@ -315,6 +315,8 @@ class Mine {
         this.skipsRemaining(mine.skipsRemaining ?? this.maxSkips);
 
         Underground.showMine();
+        // Check if completed in case the mine was saved after completion and before creating a new mine
+        setTimeout(Mine.checkCompleted);
```
SpenserJ, line 319: Although it isn't required, can you please provide the delay so it is more clear to anyone looking at the code? Can you also please add a comment that the `setTimeout` is required to push it back one event cycle so that `App.game` is available? It won't be required when we're on TypeScript modules, and would be good to clean it up
```suggestion
        // TODO: Remove setTimeout after TypeScript module migration is complete. Needed to that `App.game` is available
        setTimeout(Mine.checkCompleted, 0);
```

SpenserJ (changes requested):

nls0:
Okay, I moved the checkCompleted to a better spot, I think. Checking loadingNewLayer is there in case someone manually completes a mine at the same time the checkCompleted call is made elsewhere, which would have caused it to give multiple completions.

RedSparr0w (approved):
Changes look good to me 👍 

Thanks for the PR!

PR #945, opened 2020-10-15
Fix sorting of new pokemon into list

### LINKED ISSUE #837: [BUG] newly caught pokemon always go to the bottom of the pokemon list
**Info**
 - OS:  Windows 10
 - Browser Firefox
 - Version 0.5.5

**Description:**
When you catch a new pokemon that isn't in your pokedex it appears at the bottom of your pokemon list instead of being being placed in its appropriate condition. 

**Steps to Reproduce:**
Start a new game. Catch a few pokemon. Sort your list by Breeding efficiency (this happens on other sorting methods too). Catch a pokemon you don't have yet. 

**Expected Behavior:**
The newly caught pokemon appears at the correct position in the pokemon list

**Additional Information:**
This also effects where the pokemon appears in the hatchery list. Changing the sorting method of the pokemon list will fix the issue until more new pokemon are caught.

nls0:
Intended to fix #837. It works, though I'm not sure if this is the best way to do it since I'm not that familiar with knockout.


RedSparr0w (approved):
Looks good to me, thanks for the PR 👍

PR #948, opened 2020-10-16
Fix: Remove getArrayOfObservables

SpenserJ:
After #919 was merged, @HLXII and @Aegyo let me know that Knockout wasn't updating when values had changed. After some discussion, we agreed that the proxy wasn't necessary anymore.

This will fix the following observables:
* BadgeCase - Unlocked badges
* Farming - Berry list wasn't updating as berries were harvested or planted
* Shards - Shard wallet and upgrades weren't being displayed or tracked properly
* Wallet - Currencies weren't updating as they were earned

This removes the `getArrayOfObservables` function, as the proxy was allowing us to treat it as an array of (unobserved) values, and was likely to cause some confusion moving forward.


RedSparr0w (approved):
Changes look good to me, values all seem to be updating as expected 👍

src/modules/DataStore/BadgeCase.ts
```diff
@@ -40,15 +40,16 @@ export default class BadgeCase implements Feature {
...
         });
     }
 
     toJSON(): Record<string, any> {
         let shouldReturn = false;
         // We only want to save upto the highest badge we have obtained,
         // everything else is assumed to be false
-        return [...this.badgeList]
+        return this.badgeList
+            .map((v) => v())
```
Aegyo, line 52: alternative that doesn't require making an anonymous function:
```suggestion
            .map(ko.unwrap)
````
I am not sure I like this suggestion any more though. Such a simple operation as function application I had hoped would have a native javascript thing we could map, but I couldn't find anything that did what I wanted on Function or Function.prototype

**Review by Aegyo — COMMENTED** [review]
A couple of opinionated suggestions, feel free to ignore if you disagree

File: src/modules/DataStore/BadgeCase.ts [hunk 2]
```diff
@@ -12,19 +12,19 @@ export default class BadgeCase implements Feature {
 
     defaults: Record<string, any> = {};
 
-    badgeList: Array<boolean> = getArrayOfObservables(emptyBadgeList);
+    badgeList: Array<Observable<boolean>> = emptyBadgeList.map((v) => ko.observable(v));
```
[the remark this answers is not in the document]
SpenserJ: I had the same thought initially, but TS doesn't like it:
```js
badgeList: Array<Observable<boolean>> = emptyBadgeList.map(ko.observable);
// TS2322: Type 'Observable<unknown>[]' is not assignable to type 'Observable<boolean>[]'

badgeList: Array<Observable<boolean>> = emptyBadgeList.map(ko.observable<boolean>);
// TS1005: '(' expected.
```

We also don't need `ko.observable<boolean>(v)` here because TS will prevent me from passing any other type in as the observable value

src/modules/DataStore/BadgeCase.ts
SpenserJ: I've made this change for BadgeCase, Farming, Shards, and Wallet. Its not a huge improvement for characters or readability, but `ko.unwrap` will guard against unwrapping an observable or primitive, so its definitely safer

src/modules/DataStore/BadgeCase.ts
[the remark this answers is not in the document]
Aegyo: This works, but it looks ugly
```ts
badgeList = emptyBadgeList.map<Observable<boolean>>(ko.observable);
```
So lets just leave it alone

src/scripts/wallet/Wallet.ts
```diff
@@ -1,16 +1,14 @@
-/// <reference path="../../declarations/utilities/getArrayOfObservables.d.ts"/>
-
 class Wallet implements Feature {
     name = 'Wallet';
     saveKey = 'wallet';
-    currencies: Array<number>;
+    currencies: Array<KnockoutObservable<number>>;
```
Aegyo, line 4: I feel much better about these type signatures now, it felt like we were lying before

Aegyo (approved):
:+1:

PR #949, opened 2020-10-16
Implement #920 - refund and hide invalid shard types

### LINKED ISSUE #920: [SUGGESTION] Remove Useless Shard Upgrades
**Current Behavior**

It's currently possible to buy shard upgrades that will never have any effect.

This list currently includes the "Immune" upgrades for:
* Fire
* Water
* Grass
* Ice
* Flying 
* Bug
* Rock
* Dark
* Steel
* Fairy

and the "Super Effective" upgrades for:
* Normal

**Improved Behavior**

Players should only be able to buy shard upgrades that can actually have an impact on damage.

**Reason for change**

These useless upgrades are effectively traps for players who don't fully remember the type chart and mis-clickers. This isn't painful in the long run given how easy it is to max out most shard types by the end of Hoenn / early Sinnoh, but is an unnecessary pitfall for Johto / late Kanto players.

**Detailed Description**

* Remove the useless shard upgrades listed above
* Refund shards to players who purchased those upgrades

**Screenshots:**

Current Behavior:
Screen Shot 2020-10-11 at 2 45 59 AM

Desired Behavior:

Screen Shot 2020-10-11 at 2 46 49 AM

adapap:
Closes #920.

Hides invalid shard types (immune for a handful of types, and super effective for normal), and refunds the shards back to the player.
I assumed that the best place to refund is when the save is loaded from JSON.

This is my first time contributing here, hope it can help :)


anapple96 (changes requested):
Looks great, but you'll need to refund the players who bought them before too.

src/scripts/shards/Shards.ts
```diff
@@ -70,6 +88,23 @@ class Shards implements Feature {
...
+    
+    private refundShardUpgrade(
+        typeNum: PokemonType,
+        effectNum: GameConstants.TypeEffectiveness
+    ) {
+        while (this.shardUpgrades[typeNum * Shards.nEffects + effectNum] > 0) {
+            this.shardUpgrades[typeNum * Shards.nEffects + effectNum]--;
+            this.gainShards(this.getShardUpgradeCost(typeNum, effectNum), typeNum);
+        }
+    }
```
RedSparr0w, line 100: I don't think we need to have this as a method of Shards, as it won't be used other than for updating a game save.

src/scripts/shards/Shards.ts
```diff
@@ -8,6 +8,24 @@ class Shards implements Feature {
...
+            [PokemonType.Ice, new Set([GameConstants.TypeEffectiveness.Immune])],
+            [PokemonType.Flying, new Set([GameConstants.TypeEffectiveness.Immune])],
+            [PokemonType.Bug, new Set([GameConstants.TypeEffectiveness.Immune])],
+            [PokemonType.Rock, new Set([GameConstants.TypeEffectiveness.Immune])],
+            [PokemonType.Dark, new Set([GameConstants.TypeEffectiveness.Immune])],
+            [PokemonType.Steel, new Set([GameConstants.TypeEffectiveness.Immune])],
+            [PokemonType.Fairy, new Set([GameConstants.TypeEffectiveness.Immune])],
+            // Super Effective
+            [PokemonType.Normal, new Set([GameConstants.TypeEffectiveness.Very])],
+        ]);
```
RedSparr0w, line 28: Are we able to just automatically calculate this from the type effectiveness array?

In the unlikely event a new type is added, or effectiveness changed, it will then automatically just work.

src/scripts/shards/Shards.ts
```diff
@@ -99,6 +134,14 @@ class Shards implements Feature {
...
             this.shardWallet = getArrayOfObservables(json['shardWallet']);
             this.shardUpgrades = getArrayOfObservables(json['shardUpgrades']);
+            // Refund invalid upgrades
+            for (let t = 0; t < Shards.nTypes; t += 1) {
+                for (let effect = 0; effect < Shards.nEffects; effect += 1) {
+                    if (this.isInvalidUpgrade(t, effect)) {
+                        this.refundShardUpgrade(t, effect);
+                    }
+                }
+            }
```
RedSparr0w, line 144: Any updating of save data should be done in the [`Update.ts` file](https://github.com/pokeclicker/pokeclicker/blob/develop/src/scripts/Update.ts).

docs/index.html
```diff
@@ -7487,6 +7487,7 @@ <h3>
                                         </button>
                                     </td>
                                 </tr>
+                                <!-- /ko -->
```
RedSparr0w, line 7490: Any files in the `docs` folder should be ignored, as we update these automatically when the website is updated.
```suggestion
```

RedSparr0w (changes requested):
👍 Sounds like a good idea to me, just a few changes needed.

adapap:
Added the changes requested. Just not sure if I can access the app instance from `Update.ts`, so I reused the implementation of methods from `Shards.ts` within the update.

RedSparr0w (approved):
Looks good 👍 

Thanks for the PR!

PR #953, opened 2020-10-16
Fix/typescript modules vscode config

SpenserJ:
This will set up an [EditorConfig](https://editorconfig.org/) and `.gitattributes` file that should allow for consistent line endings (and indentation) within the project. It will work best if everyone installs the [EditorConfig](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) plugin in VSCode (or their IDE of choice), since it will ensure that new files have the correct line endings prior to being committed (otherwise ESLint may complain), but it is not required (`npm run eslint-fix` will resolve the issue as well)

LF endings were chosen over CLRF as most/all Windows editors can handle LF endings, but CLRF often shows additional characters on some Mac/Linux editors. This follows the [recommendation from Git](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration) to use LF line endings when committing code on projects that are not exclusive to Windows.

This PR will also fix the VSCode warnings for TypeScript issues inside of the `modules` folder, as well as errors within the `.eslintrc.js` files (`Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser`)


.gitattributes
```diff
@@ -0,0 +1,7 @@
+*.ts		text eol=lf
+*.tsx		text eol=lf
+*.js		text eol=lf
+*.html		text eol=lf
+*.css		text eol=lf
+*.scss		text eol=lf
+*.less		text eol=lf
```
Aegyo, line 7: Tab characters? What heresy is this?

.editorconfig
```diff
@@ -0,0 +1,23 @@
...
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# Matches multiple files with brace expansion notation
+# Set default charset and indentation style
+[*.{ts,tsx,js,html,css,scss,less}]
```
Aegyo, line 14: Want to include `ejs` in this list too (`src/templates`)

Aegyo (approved):
Thanks, looks good to me :+1:

PR #955, opened 2020-10-17
close modal if hotkey is pressed

LuqDragon:
hotkey


RedSparr0w:
I think we might have some functions bound to the `close` button on the gym leader modals.
So we might also need to trigger those if the modal is/was visible.

RedSparr0w:
Or instead of triggering the close function, we could just ignore this input if any modals are currently visible?

LuqDragon:
dont have a close function, just `data-dismiss="modal"` that make the same thing then JQuery right?

the idea is not close the modal?

RedSparr0w:
Ah, I misremembered, it looks like the function is run when we show the modal, not on close.

LuqDragon:
want to implement this as a function to all modals? checking if they a showing?

RedSparr0w:
Currently getting this error:
image

May be easier just to check if any modals are visible, as we likely wouldn't want to allow the battle to restart when covered.

LuqDragon:
i tested three methods:
$('#myModal').is(':visible');   ->  dont work with auto clicker
$('#myModal').data('bs.modal').isShown ;    -> work but we get the error
$('#myModal').hasClass('in');    ->  never works

maybe join them?

RedSparr0w:
> i tested three methods:
> `$('#myModal').is(':visible');` -> dont work with auto clicker
> `$('#myModal').data('bs.modal').isShown;` -> work but we get the error
> `$('#myModal').hasClass('in');` -> never works
> 
> maybe join them?

These ones might also work (haven't tried them out yet)
```js
$('#receiveBadgeModal').hasClass('show')
// or just for any modal
$('.modal.show').length
$('.modal').hasClass('show')
```

LuqDragon:
the problem with `hasClass` is that you spam the number or keep the button pressed, the modal get the `show` class after the modal is fully opened, change to another champion before the modal

src/scripts/GameController.ts
```diff
@@ -149,10 +149,16 @@ class GameController {
                     }
                     e.preventDefault();
                 } else if ('gymList' in player.town()) {
-                    const number = Number(e.key);
-                    // Check if a number higher than 0 and less than total Gyms was pressed
-                    if (number && number <= player.town().gymList().length) {
-                        GymRunner.startGym(player.town().gymList()[number - 1]());
+                    // Close receiveBadgeModal before start another champion
+                    //$('#receiveBadgeModal').removeClass('fade').modal('hide').addClass('fade');
+
```
RedSparr0w, line 154:
```suggestion
```

RedSparr0w (approved):
Changes working great for me!

Thanks for the PR 👍

PR #958, opened 2020-10-17
fix(shop): Don't charge more than x100 per item when bulk buying

### LINKED ISSUE #956: [BUG] Bulk buying shop items is more expensive than buying individually, for large purchases
**Info**
 - OS: Win7
 - Browser: FireFox
 - Version: v0.5.7

**Description:**
 - Shop prices increase for most items the more you buy them, with all except Master Balls returning to their original price with time. There is a cap for single-purchase of x100 times the base price. However, when you buy in bulk, it can calculate the total price as more than x100 the base price. The Discord Server is also in disagreement whether the x100 cap is only meant to apply to Master Balls or not, but it applies to all items regardless with the above behavior.

**Steps to Reproduce:**
 - Buy an item(that grows in price with amount) in bulk (About x20,000 will do) and then keep buying x1 items. It'll never go over x100. However as soon as you try to buy in bulk (like another x1000) it'll continue to scale the item price over x100.

Aegyo:
Item prices were limited to x100 in commit https://github.com/pokeclicker/pokeclicker/commit/fa71be336aff66325fcdd93d9ff229b63a2da1c5, but this wasn't being respected when calculating the price for a bulk buy.

Fixes #956

### DIFF
--- a/src/scripts/items/Item.ts
+++ b/src/scripts/items/Item.ts
@@ -52,8 +52,20 @@ abstract class Item {
         if (this.name() == GameConstants.Pokeball[GameConstants.Pokeball.Pokeball]) {
             return Math.max(0, this.basePrice * amount);
         } else {
-            const res = (this.price() * (1 - Math.pow(this.multiplier, amount))) / (1 - this.multiplier);
-            return Math.max(0, Math.floor(res));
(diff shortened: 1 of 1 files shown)


SpenserJ (approved):
Looks great! Thanks for the fix :+1:

PR #959, opened 2020-10-17
feature(changelog): Add date to UPDATE changelog entries

### LINKED ISSUE #957: [SUGGESTION] Add date to changelog updates
Add the date to the changelog. It's nice to know how often the game is updated

Aegyo:
Dates I have added here were determined from a mixture of commit messages in the format `[v*.*.*]` and the commit history of `Changelog.ts`

Closes #957

### DIFF
--- a/src/scripts/Changelog.ts
+++ b/src/scripts/Changelog.ts
@@ -35,9 +35,9 @@ class ChangelogUpdate extends Changelog {
         version: string,
         date: Date
     ) {
-        super(changelogType.UPDATE, version);
         const dateFormat: Intl.DateTimeFormatOptions = {year: 'numeric', month: 'long', day: 'numeric'};
-        this.description = `<code>${this.description}  -  ${date.toLocaleDateString(undefined, dateFormat)}</code>`;
+        const description = `<code>${version}  -  ${date.toLocaleDateString(undefined, dateFormat)}</code>`;
+        super(changelogType.UPDATE, description);
     }
 }
 
--- a/src/scripts/Changelog.ts
+++ b/src/scripts/Changelog.ts
@@ -1,4 +1,19 @@
-const changelogType = {
+enum ChangeLogType {
+    DEFAULT,
+    UPDATE,
+    NEW,
+    CHANGE,
+    FIXED,
+    REMOVED,
+    EVENT,
+}
+
+type ChangelogConfig = {
+    display: string,
+    label: string,
+}
+
+const changelogType: Record<keyof typeof ChangeLogType, ChangelogConfig> = {
     UPDATE: { display: 'dark', label: 'UPDATE' },
     NEW: { display: 'success', label: 'NEW' },
     CHANGE: { display: 'primary', label: 'CHANGE' },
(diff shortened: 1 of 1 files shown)


src/scripts/Changelog.ts
```diff
@@ -136,7 +156,7 @@ const changelogItems = [
     new Changelog(changelogType.FIXED, 'Fixed breeding filter displayed options being reset after hatchery filled up'),
 
     // v0.5.1
-    new Changelog(changelogType.UPDATE, 'v0.5.1'),
+    new ChangelogUpdate('v0.5.1', new Date(2020, 7, 31)),
```
SpenserJ, line 159: When I backported the tags (https://github.com/pokeclicker/pokeclicker/releases), I found that the date for the commit or version bump often didn't align with the date it was released. The tags might not be any more accurate, but because some of the changelogs were retroactively updated to add missing changes, I think they might be a bit closer to the last time that version was deployed. `0.5.1` seems to have had a commit on September 4th, even though though the version bump and changelog were originally from July 31st. Not sure which ones we want to use, but I figured I'd mention it since I had the same struggle determining dates, and just ended up automating some detection for it

Aegyo: Some of the versions seemed to have a patch or two afterwards, with additions to the changelog group. In these cases I went with the earlier date since that is what we would have used if we were entering dates from the start. I have made an assumption that fda8f15fb8baec3f183b5d0f39e4a1558cdeab53 and 85975e9ad7d3a5cf3d27d8b24bc7cca9d7385485 both had a release to the website though.

src/scripts/Changelog.ts
```diff
@@ -9,15 +24,20 @@ const changelogType = {
...
+}
+
+class ChangelogUpdate extends Changelog {
+    constructor(
+        version: string,
+        date: Date
+    ) {
+        super(changelogType.UPDATE, version);
+        const dateFormat: Intl.DateTimeFormatOptions = {year: 'numeric', month: 'long', day: 'numeric'};
+        this.description = `<code>${this.description}  -  ${date.toLocaleDateString(undefined, dateFormat)}</code>`;
```
SpenserJ, line 40: Minor, but it might be cleaner to push the description through the constructor, instead of initially setting it to `version`
```suggestion
        const dateFormat: Intl.DateTimeFormatOptions = {year: 'numeric', month: 'long', day: 'numeric'};
        const description = `<code>${version}  -  ${date.toLocaleDateString(undefined, dateFormat)}</code>`;
        super(changelogType.UPDATE, description);
```

SpenserJ (approved):

src/scripts/Changelog.ts
Aegyo: Definitely better. I thought I had tried it this way first, but had typescript shout at me for daring to do anything before calling super, but it doesn't happen when I try it now...

SpenserJ: It will scream if anything touches `this` before `super`, but as long as we're only using `const` with no helper methods, it'll usually be okay

SpenserJ (approved):
Looks great! Thanks :+1:

PR #961, opened 2020-10-18
fix(party): rework party list layout to fit in small layouts

SpenserJ:
Taking an alternative approach to #829 to avoid the challenge of dynamically sizing text. This will stack the pokemon's name above its attack and level in small columns, and keep it on a single line in the center column

| Left/Right column | Center column |
| --- | --- |
| image | image |

It may be worth doing a bit more work so that the attack/level "column" line up consistently, but I want to see how we feel about this layout first

### DIFF
--- a/src/components/pokemonListContainer.html
+++ b/src/components/pokemonListContainer.html
@@ -11,17 +11,11 @@
         <td>
             <div>
                 <div class="pokemonName">
-                    <span data-bind="text: name, class: name.length > 15 ? 'squeeze' : ''">Name</span>
+                    <span data-bind="text: name">Name</span>
                     <sup data-bind="visible: App.game.party.alreadyCaughtPokemon($data.id, true)">✨</sup>
                 </div>
-                <div class="pokemonAttack">
-                    Atk: 
-                    <span data-bind="text: attack"></span>
-                </div>
-                <div class="pokemonLevel">
-                    Lvl: 
-                    <span data-bind="text: level"></span>
-                </div>
+                <div class="pokemonAttack" data-bind="text: attack.toLocaleString()"></div>
+                <div class="pokemonLevel" data-bind="text: level"></div>
             </div>
         </td>
     </tr>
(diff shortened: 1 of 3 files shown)


RedSparr0w:
The only thing I'm not too sure about is the `Atk:` and `Lvl:` text.

Maybe we could have the table titles fixed to the top of the container?

I know it's not this way currently, but should we also change the `attack` amount to use `toLocaleString`?

SpenserJ:
> The only thing I'm not too sure about is the `Atk:` and `Lvl:` text. Maybe we could have the table titles fixed to the top of the container?

I had been thinking about that, and wasn't sure if it would be confusing since we'd probably want a similar format in the header. I've set a minimum width of `60px` for both the attack and level sections to help with alignment:

| Left/Right column | Center column |
| --- | --- |
| image | image |

> I know it's not this way currently, but should we also change the `attack` amount to use `toLocaleString`?

Yes. A hundred times yes.

Something I've noticed (and you can see in the left/right column screenshot) is that we still get a horizontal scrollbar sometimes, but it doesn't seem to be related to the width of the list from what I can tell. It seems like the dropdown is taking up too much room, and the scrollbar goes away as soon as I hide it. It doesn't seem to be consistent though and often goes away by itself.

RedSparr0w:
> Something I've noticed (and you can see in the left/right column screenshot) is that we still get a horizontal scrollbar sometimes, but it doesn't seem to be related to the width of the list from what I can tell. It seems like the dropdown is taking up too much room, and the scrollbar goes away as soon as I hide it. It doesn't seem to be consistent though and often goes away by itself.

Maybe we could change the sort select to be the same as the Hatchery, which should take up less space overall:
image


Something still looks slightly off to me, not exactly sure why though.
Maybe with the side column view, we could just show `Attack` and `Level` for the headers, I think the `Name` is mostly implied.

SpenserJ:
> Maybe we could change the sort select to be the same as the Hatchery, which should take up less space overall:
> Maybe with the side column view, we could just show `Attack` and `Level` for the headers, I think the `Name` is mostly implied.

| Left/Right Column | Center column |
| --- | --- |
| image) | image |

The size of the select varies depending on theme, and it looks a bit cleaner in Yeti (default). I'm having trouble getting the arrow to align better in slate _and_ yeti at the same time, so we may have something wacky happening in the theme styles
image

RedSparr0w (approved):
Looks good to me 👍 
Thanks for the PR!

Not sure if you wanted to do anything about the alignment of the arrows in this PR or leave it for another, so will leave this for you to merge.

SpenserJ:
> Not sure if you wanted to do anything about the alignment of the arrows in this PR or leave it for another, so will leave this for you to merge.

I've been unable to figure it out, so I figure it would be better to merge now as its an improvement to most things, and fix that up when we have a chance

PR #964, opened 2020-10-19
Unova SVG +Base Map

anapple96:
Base map + SVG


other/Map tools/Hoenn.tmx
```diff
@@ -59,12 +59,7 @@
...
-  </data>
- </layer>
- <layer id="19" name="Dungeon Extension" width="100" height="60">
-  <data encoding="base64" compression="zstd">
-   KLUv/WDAXKUAACgAABAGEgQAYQZwAO7uMHx0eYAF
+   KLUv/WDAXB0NAMLOKCdQd0UBH5GI3AUgZiEhZgHP5HBW+F/lDLZLROBkASEAAKAEI1iumA/38rl8LafynT5TSekjfaOTy+SiuDxCH+j7iC6RK+Q6Po7P87nNt/FpauKCuCSuiAeHq4Fh4VLwdK7NxZGJ2WY2GI3GldFyYXFVsGwrppQNxUabaCPZRuwDsn2Yp7NxEraBDWE7MCjYCFQujOt0mZpAwAZQIl2jFQJdnwNmqCALCBVnAeCCi0FhFRJAEChSBBFG8BDBHxFAEIGCURRNwLFh8ZVsdRWO11vVaJx0FZMqGXf8JLWRNV6ITaSBkvuUN/BU0kvwst46CvhT1gg+0C9yvHXSZLaske3icnhySJRtif0sctHkWJTL2c8SdPiBeTcd+5qhPjAJOD8zE48a4FVaOKjk8ymPhT4vrzgXDioJ+RbzPsgc7KHQZwDfJejdkJOJLgeafGnWyEk6L5wMDHo++oaal1EWJOBpd0fA9HNhU43rkbI4QNS1BmzefZj2I5WD8xqwMfdhys8pDOggfjIXXJxJLd80TF63gfgwY1LPo/kBesqevqV+
```
RedSparr0w, line 62: You've made changes to the Hoenn map file, but haven't updated the image.

other/Map tools/Unova.tmx
```diff
@@ -14,37 +14,45 @@
...
   </data>
  </layer>
+ <layer id="12" name="Decor" width="100" height="60">
+  <data encoding="base64" compression="zstd">
+   KLUv/WDAXOUAAGgAAM0GAADOBgAAzwYAAwCZmFw+EASqf5x2gAU=
+  </data>
+ </layer>
+ <imagelayer id="13" name="base" visible="0">
+  <image source="../../../../../Downloads/unovabase.png" width="1600" height="960"/>
+ </imagelayer>
```
RedSparr0w, line 57: Could you remove this layer, as it will be a broken reference for anyone else working on the maps.

RedSparr0w (commented):
I'm not sure about removing the "bridge" as it looks odd without anything there, and it also makes the route next to it extremely small to be able to tap/click on.
image

_I haven't had a chance to test out the svg/interactive map yet, just a couple of small changes that I noticed though._

Edit: 
There also seems to be an extra image file in the map tools directory:
image

anapple96:
The small bit is an extra part of the bigger route 3. But eh, extended it anyway

other/Map tools/Unova.tmx
anapple96: Removed

other/Map tools/Hoenn.tmx
anapple96: Gimme a bit, dunno why did I change it, probably some water stuffs

anapple96: Removed

src/components/UnovaSVG.html
```diff
@@ -0,0 +1,705 @@
...
+        { route: 21, region: 4
+        , width: 6, height: 3
+        , x: 91, y: 10
+        , rotate: true
+        })
+    %>
+
+    <!--Undella Bay-->
+    <%- include('templates/mapRoute',
+        { route: Undella Bay, region: 4
```
SpenserJ, line 274: The map crashes right now because this is an invalid object
```suggestion
        { route: "Undella Bay", region: 4
```

src/components/UnovaSVG.html
```diff
@@ -0,0 +1,705 @@
...
+        { route: 4, region: 4
+        , width: 10, height: 3
+        , x: 57, y: 49
+        , noText: true
+        })
+    %>
+
+    <!--Desert Resort-->
+    <%- include('templates/mapRoute',
+        { route: Desert Resort, region: 4
```
SpenserJ, line 72: The map crashes right now because this is an invalid object
```suggestion
        { route: "Desert Resort", region: 4
```

SpenserJ (changes requested):

SpenserJ (commented):
These steps are needed before the map will render:
* Add `@import "UnovaSVG.html"` to `src/index.html`
* Set `GameConstants.MAX_AVAILABLE_REGION` to `4`
* Remove requirement on `Castelia City`
* Add `4: Set(),` to `GameConstants.IceAreas` and the type declaration

You will now be able to access go to the new region, but will not be able to do anything. My previous suggestion for adding `"` around the named routes fixes the EJS rendering, but `src/templates/mapRoute.ejs` will generate invalid KO bindings because `MapHelper.moveToRoute` expects a numeric route and will either need to accept routes by `number | string`, or we need to have the SVG only know about the number. A quick fix to render more is to add `<% const routeValue = typeof locals.route === 'string' ? `"${locals.route}"` : locals.route; %>` to the mapRoute template, and use `routeValue` in place of `locals.route` for the click and class (but not text) bindings.

At this point you should be able to refresh a saved game in Unova (the previous crash did not prevent you from going there, but would crash the game when starting up), and you may be able to play through routes 19 and 20, but cannot proceed any further. Your map will look similar to this, with some issues for `Desert Resort` and `Undella Bay` (named routes that overlap with others) and the greyed out extensions of Route 3:
image

anapple96:
Should be ready for review

src/modules/GameConstants.ts
```diff
@@ -16,7 +16,7 @@ export const MAX_LOAD_TIME = 20000; // 20 Seconds
 export const MUTATION_TICK = 1000;
 export const WANDER_TICK = 1500;
 
-export const MAX_AVAILABLE_REGION = 3; // Sinnoh
+export const MAX_AVAILABLE_REGION = 4; // Sinnoh
```
RedSparr0w, line 19:
```suggestion
export const MAX_AVAILABLE_REGION = 3; // Sinnoh
```

RedSparr0w (approved):

PR #968, opened 2020-10-20
Update how Discord login works, allow any domain.

RedSparr0w:
Allow Discord login from any domain, even if not in "approved list".

This fixes not being able to login to Discord from the desktop version.

No longer storing username as it's not used anywhere.

config setting used for production:
```js
    DISCORD_LOGIN_PROXY: 'https://discord.pokeclicker.com/proxy',
```


SpenserJ (approved):
Looks good, but hard for me to test it. I don't see anything that might be an issue

RedSparr0w:
> Looks good, but hard for me to test it. I don't see anything that might be an issue

You should be able to test it by just adding this to your config and then trying to login:
```
    DISCORD_LOGIN_PROXY: 'https://discord.pokeclicker.com/proxy',
```

PR #970, opened 2020-10-20
Add Region to Gym quests and achievements. Remove numbers from names

anapple96:
Exactly what the title says


anapple96:
Should work fine now.

SpenserJ (changes requested):
We'll want to do another update for this so we don't spam achievements and leave the old achievement names in the save, but we should probably wait for #974 just in case we have another issue with older saves. Let me know if you want me to write the update function. Otherwise this looks like a really great change! :+1:

src/scripts/gym/Champion.ts
```diff
@@ -47,7 +47,7 @@ class Champion extends Gym {
 
 gymList['Champion Blue'] = new Champion(
```
RedSparr0w, line 48: Should this not also be updated to reflect the change?
Otherwise it seems like each of the elites are being referred to 3 different ways:
`Blue`, `Champion Blue`, `Kanto Champion Blue`
```suggestion
gymList['Kanto Champion Blue'] = new Champion(
```

src/scripts/quests/questTypes/DefeatGymQuest.ts
```diff
@@ -1,16 +1,17 @@
...
         this.focus = App.game.statistics.gymsDefeated[GameConstants.getGymIndex(gymTown)];
     }
 
-    private static getDescription(gymTown: string, amount: number): string {
-        let desc = `Defeat ${gymTown} `;
+    private static getDescription(gymTown: string, amount: number, region: number): string {
+        let desc = `Defeat ${gymList[gymTown].town} `;
         if (!gymTown.includes('Elite') && !gymTown.includes('Champion')) {
-            desc += 'gym ';
+            desc += `Gym in ${GameConstants.Region[region].charAt(0).toUpperCase() + GameConstants.Region[region].slice(1)} `;
```
RedSparr0w, line 14: This will uppercase the first letter.
```suggestion
            desc += `Gym in ${GameConstants.camelCaseToString(GameConstants.Region[region])} `;
```

RedSparr0w: We should probably also include the region for Elite/Champion too.

anapple96: Done

src/scripts/gym/Champion.ts
anapple96: Done

PR #971, opened 2020-10-20
fix(statistics): fix bug with porting old or early-region saves

SpenserJ:


RedSparr0w (approved):

PR #972, opened 2020-10-20
Add a way to gain breeding queue slots

RedSparr0w:
Not sure if this is the right way to be going about this or not tbh,
But here it is anyway, we can also add other methods of obtaining extra later on.

Gain queue slots based on highest region.
Highest Region | Slots Gained | Total Slots
:-- | :--: | :--:
Kanto | 0 | 0
Johto | 4 | 4
Hoenn | 4 | 8
Sinnoh | 8 | 16
Unova | 16 | 32


Ultima1990:
I'm not sure I like this way of obtaining breeding slots. IMO they should be obtained upon clearing the route/location that would contain the day care in main series games. Makes sense to me, nice little connection to the main titles, and those day care locations currently have no other function.

Those would be:
Route 34 in Johto
Route 117 in Hoenn
Solaceon town (so route 209) in Sinnoh
Route 3 in Unova

RedSparr0w:
> IMO they should be obtained upon clearing the route/location that would contain the day care in main series games.

It's pretty similar to just unlocking it when you reach the next region, with a bit of extra time after,
We can always change it to be that way later on, but it will require extra code as we don't currently have a reward system for beating routes/unlocking areas.

Claygame:
i think other than making more slots available when you reach new area, they should also be purchasable. Maybe can make them decent price item in battle frontier shop as a limited stock item so people cant buy 100+ slots easily. I have a feeling people will want more than just 8 slots when they have 300+ poke. Can reduce the slots gained os new region unlock to balance it out if you feel that shop purchase can make it too easy.

RedSparr0w:
As mentioned above we can also add other methods of obtaining extra later on,
but this PR is just for an initial method of obtaining them, the amounts can also be adjusted later if need be.

RedSparr0w:
Can be modified in the future if needed, merged for testing and can be adjusted from here.

Issue #902, opened 2020-10-08
[FEATURE PROPOSAL] Berry Farm Overhaul 


HLXII:
# Abstract
The current berry system is fairly removed from the rest of the game. The only connect it has to other mechanics is getting Farm points to then spend on Master balls, or completing quests for QP. This process is also quite tedious, since the Farm mechanic requires constant attention since there is no passive gain implemented in the mechanic.  What I mean by this is that you can passively gain Gold and DP by sitting in a Route with Poke balls. However, to get Farm points, you have to be active every 1-6 minutes.

My goal for this proposal is to integrate the Farm mechanic into other systems, so that the player has more incentive to use the mechanic, and to make the experience a bit richer for the player, rather than clicking Plant All, waiting for a notification, then Harvest All and repeat.

#  Berry Farm

## [Berry Types](https://bulbapedia.bulbagarden.net/wiki/Berry#Generation_III_onwards)

The number of berry types will expand to all canon berries (possibly excluding Gen VI ones, as those don't have [artwork](https://bulbapedia.bulbagarden.net/wiki/Berry#Artwork)). We'll keep the same BerryDistribution constant and continue to drop the same 8 berries implemented. The rest of the berry types will be gained by either:

1. Mutations (see Mutations section)
2. Achievements (?)
   *  I was thinking one berry type would be gained by getting all other berry types, but not so sure about this one.

### Berry Generations

There are a total of 67 berries in canon pokemon. I've separated them into five different generations.

First Gen | Second Gen | Third Gen | Fourth Gen (Typed Berries) | Fifth Gen
------------ | ------------- | ------------- | ------------- | -------------
Cheri |  Persim | Pomeg | Occa | Liechi
Chesto | Razz | Kelpsy | Passho | Ganlon
Pecha | Bluk | Qualot | Wacan | Salac
Rawst | Nanab | Hondew | Rindo | Petaya
Aspear | Wepear | Grepa | Yache | Apicot
Leppa | Pinap | Tamato | Chople | Lansat
Oran | Figy | Cornn | Kebia | Starf
Sitrus | Wiki | Magost | Shuca | Micle
_ | Mago | Rabuta | Coba | Custap
_ | Aquav | Nomel | Payapa | Jaboca
_ | Iapapa | Spelon | Tanga | Rowap
_ | Lum | Pamtre | Charti | Kee
_ | _ | Watmel | Kasib | Maranga
_ | _ | Durin | Haban | Enigma
[...]

PR #973, opened 2020-10-20
Berry Farm Overhaul

HLXII:
See https://github.com/pokeclicker/pokeclicker/issues/902#issue-716952657

This is the first PR mostly for code review, as the last development hasn't been fully implemented yet (see above issue for current progress).

Functionality to check: 
- [ ] Mechanics
   - [ ] New Growth Implementation
   - [ ] Unlocking Plots
   - [ ] Mulch
   - [ ] Mutations
- [ ] UI
   - [ ] Farm UI
   - [ ] Berry/Mulch List
   - [ ] Berrydex Tab
   - [ ] Berrydex Modal
   - [ ] Berry Master UI
- [ ] NPCs
   - [ ] Kanto Berry Master Messages
   - [ ] Berry Master Trades
- [ ] Balance
   - [ ] Berry Growth Times
   - [ ] Berry Harvest Amounts
   - [ ] Berry Farm Tokens
   - [ ] Replant Rates
   - [ ] Mutation Rates
   - [ ] Trade Amounts


Aegyo (commented):
Partial review, I didn't get all the way through it tonight.

I'll take another look tomorrow

src/scripts/farming/Farming.ts
```diff
@@ -5,63 +5,894 @@ class Farming implements Feature {
...
+            [10, 0, 10, 10, 10], BerryColor.Red,
+            ['It takes longer to grow than Berries such as Cheri. The smaller Berries taste better.']);
+        this.berryData[BerryType.Oran]      = new Berry(BerryType.Oran,     [120,180,240,300,600],
+            6, 0.2, 20,
+            [10, 10, 0, 10, 10], BerryColor.Blue,
+            ['Nature\'s gifts came together as one in this Berry. It has a wondrous mix of flavors that spread in the mouth.']);
+        this.berryData[BerryType.Sitrus]    = new Berry(BerryType.Sitrus,   [150,300,450,600,1200],
+            8, 0.2, 30,
+            [0, 10, 10, 10, 10], BerryColor.Yellow,
+            ['Sitrus came from the same family as Oran. It is larget and smoother tasting than Oran.']);
```
anapple96, line 78: Larget

src/scripts/farming/Farming.ts
```diff
@@ -5,63 +5,894 @@ class Farming implements Feature {
...
+        this.mutations.push(new EvolutionMutation(.00001, BerryType.Shuca, BerryType.Watmel, function() {
+            return App.game.oakItems.isActive(OakItems.OakItem.Sprinklotad);
+        }, 'I\'ve heard that using the Sprinklotad will cause Watmel Berries to change!', function() {
+            return App.game.oakItems.isUnlocked(OakItems.OakItem.Sprinklotad) &&
+            App.game.farming.unlockedBerries[BerryType.Watmel]();
+        }));
+        // Coba
+        this.mutations.push(new FlavorMutation(.00001, BerryType.Coba,
+            [0, 10, 0, 15, 0], true,
+            'I\'ve heard that a berry will appear if its surroundings match its flavor profile! If I recall, it tasted a little dry and fairly bitter at the same time.',
```
Aegyo, line 671:
```suggestion
            'I\'ve heard that a Berry will appear if its surroundings match its flavor profile! If I recall, it tasted a little dry and fairly bitter at the same time.',
```

src/scripts/farming/Farming.ts
```diff
@@ -5,63 +5,894 @@ class Farming implements Feature {
...
         this.plotList.forEach(plot => {
-            plot.reduceTime(timeToReduce);
-            if (!plot.isEmpty() && plot.timeLeft == 0 && !plot.notified) {
-                plot.notified = true;
-                notify = true;
+            if (plot.update(timeToReduce)) {
+                change = true;
+            }
+            if (plot.notifications) {
+                plot.notifications.forEach(n => notifications.add(n));
```
Aegyo, line 802:
```suggestion
                plot.notifications.forEach(notifications.add);
```

src/scripts/farming/Farming.ts
```diff
@@ -5,63 +5,894 @@ class Farming implements Feature {
...
+        }
+
+        if (change) {
+            this.resetAuras();
+        }
+
+        if (notifications.size) {
+            notifications.forEach(function(n) {
+                this.handleNotification(n);
+            }, this);
```
Aegyo, line 825: I haven't tried it, but this can probably be changed to
```js
notifications.forEach(this.handleNotification)
```

src/scripts/farming/Farming.ts
```diff
@@ -5,63 +5,894 @@ class Farming implements Feature {
...
+        }
+    }
+
+    // TODO: HLXII Change details of notifier for different notifications
+    handleNotification(type: FarmNotificationType): void {
+        switch (type) {
+            case FarmNotificationType.Ripe:
+                Notifier.notify({
+                    message: 'A berry is ready to harvest!',
```
Aegyo, line 834: `berry` -> `Berry` on all of the notification messages

src/scripts/farming/Farming.ts
```diff
@@ -70,86 +901,127 @@ class Farming implements Feature {
...
-            });
-        }
-
-        plot.berry = BerryType.None;
         App.game.oakItems.use(OakItems.OakItem.Sprayduck);
-        return money;
+
+        plot.die(true);
+
+        this.resetAuras();
```
Aegyo, line 978: Does this need to be wrapped in a `if (!suppressResetAura)`

src/scripts/farming/Plot.ts
```diff
@@ -2,43 +2,434 @@ class Plot implements Saveable {
...
+        this._auras[AuraType.Mutation] = ko.observable(1);
+        this._auras[AuraType.Replant] = ko.observable(1);
 
         this.formattedTimeLeft = ko.pureComputed(function () {
-            return GameConstants.formatTime(Math.ceil(this.timeLeft) / App.game.oakItems.calculateBonus(OakItems.OakItem.Sprayduck));
+            if (this.berry === BerryType.None) {
+                return '';
+            }
+            for (let i = 0;i < 5;i++) {
+                if (this.age < this.berryData.growthTime[i]) {
```
Aegyo, line 53: I think it would be clearer to do
```js
const growthTime = this.berryData.growthTime.find(t => this.age < t)
```
and then replace `this.berryData.growthTime[i]` with `growthTime` in the rest of the calculation, to avoid returning from the middle of a for loop.

src/scripts/farming/Plot.ts
```diff
@@ -2,43 +2,434 @@ class Plot implements Saveable {
...
+            if (this.auraMutation() !== 1) {
+                if (auraStr) {
+                    auraStr += '<br/>';
+                }
+                auraStr += `Mutation: ${this.auraMutation().toFixed(2)}x`;
+            }
+
+            if (this.auraReplant() !== 1) {
+                if (auraStr) {
+                    auraStr += '<br/>';
```
Aegyo, line 104: Maybe just personal preference, but in cases like this, I like to collect all of the strings into an array and then
```js
return strings.join('<br/>')
```
at the end, instead of having to check if we already have stuff at each stage.

src/scripts/farming/Plot.ts
```diff
@@ -2,43 +2,434 @@ class Plot implements Saveable {
...
+            return this.berry === BerryType.None;
         }, this);
+
         this.stage = ko.pureComputed(function () {
             if (this.berry === BerryType.None) {
-                return 1;
+                return PlotStage.Seed;
+            }
+            for (let i = 0;i < 5;i++) {
+                if (this.age < this.berryData.growthTime[i]) {
```
Aegyo, line 120: Another
```js
return this.berryData.growthTime.find(t => this.age < t)

File: src/scripts/farming/Plot.ts [hunk 10]
```diff
@@ -2,43 +2,434 @@ class Plot implements Saveable {
...
+                        break;
+                    case PlotStage.Berry:
+                        tooltip += `${formattedTime} until death`;
+                        break;
+                }
+            }
+
+            if (this.stage() >= PlotStage.Taller && this.berryData.aura) {
+                if (tooltip) {
+                    tooltip += '<br/>';
```
Aegyo, line 156: Another collection of strings and join at the end, if you liked the suggestion.

src/scripts/farming/Plot.ts
```diff
@@ -2,43 +2,434 @@ class Plot implements Saveable {
...
+        let oldStage = PlotStage.Seed;
+        for (let i = 0;i < 5;i++) {
+            if (oldAge <= growthStages[i]) {
+                oldStage = i;
+                break;
+            }
+        }
+        let newStage = PlotStage.Seed;
+        for (let i = 0;i < 5;i++) {
+            if (newAge <= growthStages[i]) {
```
Aegyo, line 239: A couple more loops that can be replaced with `.find`

src/scripts/farming/Plot.ts
```diff
@@ -47,35 +438,63 @@ class Plot implements Saveable {
...
+    /**
+     * Finds the plot indices that are around the plot in a 3x3 square
+     * @param index The plot index
+     * @param filter An optional filter callback for filtering out indices
+     */
+    public static findNearPlots(index: number, filter?: (n: number) => boolean): number[] {
+        const plots = [];
+
+        const colIdx = index % 5;
+        const rowIdx = (index - colIdx) / 5;
```
Aegyo, line 466: These 5s could do with being constants somewhere, and also replaced wherever else is assuming that the farm is a 5x5

src/scripts/farming/aura/Aura.ts
```diff
@@ -0,0 +1,68 @@
+class Aura {
+
+    auraType: AuraType;
+    auraMultipliers: number[];
+
+    constructor(auraType: AuraType, auraMultipliers: number[]) {
```
Aegyo, line 6: You can write this line as
```js
constructor(
    public auraType: AuraType,
    public auraMultipliers: number[],
) { }
```
then you will not need the declarations above the constructor or the setting inside of the constructor, typescript will Just Know.

src/scripts/farming/aura/Aura.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    getLabel(stage: PlotStage): string {
+        if (stage < PlotStage.Taller) {
+            return '';
+        }
+        let label = '';
+        switch (this.auraType) {
+            case AuraType.Attract:
+                label = 'Attract: ';
```
Aegyo, line 45: This entire switch statement looks like it can be replaced with
```js
let label = `${AuraType[this.auraType]}: `;
```
Maybe a `GameHelper.humanifyString` if there might be a multiword AuraType added in the future.

Aegyo (commented):
Another partial review, only 21 more files to go...

Aegyo:
Also, thank you for showing me the word "drupelet" :+1:

src/scripts/farming/Farming.ts
HLXII: Tried this and didn't seem to work

src/scripts/farming/Farming.ts
HLXII: same here

RedSparr0w:
Haven't had a chance to look through any of this yet, but just checking that all the images have been minified with something like https://tinypng.com?

HLXII:
> Haven't had a chance to look through any of this yet, but just checking that all the images have been minified with something like https://tinypng.com?

Yep, though I didn't do that with the plant images since those are fairly small already. Not sure if it was worth it.

RedSparr0w:
Definitely worth it, as it doesn't change the image at all but reduces the file size by a decent amount.

For example:
\- | Before | After
:-- | :--: | :--:
Image | AllTreeSprout | AllTreeSprout (1)
Size | 434 B | 343 B (-21%)
Image | ApicotTreeBerry | ApicotTreeBerry (1)
Size | 662 B | 494 B (-25%)

src/scripts/farming/BerryDeal.ts
```diff
@@ -0,0 +1,166 @@
...
+        return ItemList[EvoItems.EvoItem[evoItem]];
+    }
+
+    private static randomUndergroundItem(): UndergroundItem {
+        return UndergroundItem.list[Math.floor(UndergroundItem.list.length * SeededRand.next())];
+    }
+
+    public static getDeals(region: GameConstants.Region) {
+        switch (region) {
+            case GameConstants.Region.johto:
```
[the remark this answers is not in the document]
HLXII: Removed this to instead set the available regions manually, and also split the berry deals into a recordd, so no more slicing

src/scripts/farming/BerryDeal.ts
```diff
@@ -0,0 +1,166 @@
...
+            this.berries.push({berryType: berry, amount: berryAmount[idx]});
+        }, this);
+        this.item = {itemType: item, amount: itemAmount};
+    }
+
+    private static randomBerry(berryList: BerryType[]): BerryType {
+        return berryList[Math.floor(berryList.length * SeededRand.next())];
+    }
+
+    private static randomAmount(base: number, variance: number): number {
```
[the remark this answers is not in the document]
HLXII: Found SeededRand has intBetween, which I can just use instead

RedSparr0w:
Awesome work, definitely makes the Farm a lot more interesting!

<hr>

What is the purpose of harvesting the berries?
It seems to always remove the berry, where as if you just leave it to "die", it will possibly auto replant and grow more berries.

<hr>

Maybe add a notification for when you unlock a new berry _(doesn't need a setting to disable/enable)_

<hr>

I think we could probably remove the `shovel all` button, as I feel like it will get miss clicked too often.
Maybe move the shovel button to the left near the berries, so it's more obvious that it changes the selection between the berries and the shovel.
_Which could also allow us to add an extra row of plots?_
image

<hr>

Not sure what's causing this, but when I open the berry master shop, there is quite a lot of lag when opening the shop, and while the shop is open.

<hr>

I purchased the `Squirt Bottle` but it doesn't seem to do anything, or be anywhere?

<hr>

Possibly change this to be a table:
image

<hr>

I think we also need to either take these berries away from the player and replace them with tokens or something else,
Or just unlock them if they already have some:
image

<hr>

Maybe align the text to the left:
image
image

<hr>

I'm not sure what has happened here, maybe the berry images are different sizes and not square?
image

HLXII:
Harvesting Berries returns the full harvest amount. If you let the Berry plant die, you'll get half of the Berries, and there's a chance for the Berry to replant itself. This allows for passive Berry gain if you don't want to micromanage the Farm, and also rewards active play with full harvests.

Added a notification

Removed the Shovel All button. I've moved the Berry Shovel option under the Berry/Mulch element.
image
I didn't want to move it into the Berry List since that would require additional work for handling change Berry pages.

I'm not sure about the Berry Modal lag either. I removed the footer as well as the Berry Trade tab and it seems to still happen. I assume it might be some Knockout problem but idk.

The SquirtBottle is an Oak Item that's unlocked after purchasing from the store. I might add a description hover tooltip for it.

Made the rest of the UI changes.

RedSparr0w:
> I'm not sure about the Berry Modal lag either. I removed the footer as well as the Berry Trade tab and it seems to still happen. I assume it might be some Knockout problem but idk.

It only seems to happen while in the Shop tab.
Did you use the same layout and everything as the other shops?

> I've updated Update.ts to unlock the already obtained berries, however I'd need to update the dev version for that code to run.

Looks like it's working now, just wasn't running the update function before as the `package.json` version wasn't `0.5.9`

RedSparr0w:
After playing through the Farm from a fresh save,
I think we probably should just reset farm progress.
Otherwise you are just missing/skipping a ton of the content for Farm.

Possibly:
- Refund Farm tokens spent on plots
- Convert all their berries to Farm tokens at a lowish rate _(10 - 100 berries per token)?_

src/scripts/farming/BerryType.ts
```diff
@@ -1,11 +1,85 @@
...
+    Wepear,
+    Pinap,
+
+    Figy,
+    Wiki,
+    Mago,
+    Aguav,
+    Iapapa,
+
+    Lum,
```
RedSparr0w, line 28: Is there any reason for all the extra spacings in this file?
```suggestion
    Pinap,
    Figy,
    Wiki,
    Mago,
    Aguav,
    Iapapa,
    Lum,
```

HLXII: I separated them into groups for determining properties. The spaces can be removed.

src/scripts/oakItems/OakItems.ts
```diff
@@ -18,14 +23,26 @@ class OakItems implements Feature {
...
+            new OakItem(OakItems.OakItem.Shiny_Charm, 'Shiny Charm', 'Encounter shinies more often',
+                true, [1.50, 1.60, 1.70, 1.80, 1.90, 2.00], 1, 70, 150),
+            new OakItem(OakItems.OakItem.Blaze_Cassette, 'Blaze Cassette', 'Hatch eggs faster',
+                false, [1.50, 1.60, 1.70, 1.80, 1.90, 2.00], 1, 80, 10),
+            new OakItem(OakItems.OakItem.Cell_Battery, 'Cell Battery', 'More passive mining energy regen',
+                false, [1.5, 1.6, 1.7, 1.8, 1.9, 2], 1, 90, 20),
+            new BoughtOakItem(OakItems.OakItem.Squirtbottle, 'Squirtbottle', 'Increases the chance of berry mutations', 'Johto Berry Master',
+                true, [1.15, 1.3, 1.45, 1.6, 1.75], 1, 5, undefined, undefined, AmountFactory.createArray([500, 1000, 2000, 3000, 5000], GameConstants.Currency.farmPoint)),
+            new BoughtOakItem(OakItems.OakItem.Sprinklotad, 'Sprinklotad', 'Increases the chance of berry replants', 'Hoenn Berry Master',
+                true, [1.5, 1.75, 2, 2.25, 2.5], 1, 7, undefined, undefined, AmountFactory.createArray([500, 1000, 2000, 3000, 5000], GameConstants.Currency.farmPoint)),
```
RedSparr0w, line 45: Are these 2 the wrong way around?

As with the Sprinklotad, if the multiplier is 2+ then it's a 100% replant rate with all the Kanto berries.
Also seems to be able upgrade way too fast IMO.
```suggestion
            new BoughtOakItem(OakItems.OakItem.Squirtbottle, 'Squirtbottle', 'Increases the chance of berry mutations', 'Johto Berry Master',
                true, [1.15, 1.3, 1.45, 1.6, 1.75], 1, 5, undefined, undefined, AmountFactory.createArray([500, 1000, 2000, 3000, 5000], GameConstants.Currency.farmPoint)),
            new BoughtOakItem(OakItems.OakItem.Sprinklotad, 'Sprinklotad', 'Increases the chance of berry replants', 'Hoenn Berry Master',
                true, [1.5, 1.75, 2, 2.25, 2.5], 1, 7, undefined, undefined, AmountFactory.createArray([500, 1000, 2000, 3000, 5000], GameConstants.Currency.farmPoint)),
```

HLXII: The Oak Items were implemented before I fully set all the properties for the berries. I'm always up for rebalancing (mostly because I didn't do that many test runs to check if they were balanced to being with). 

Experience required probably needs to go way up, especially since there are a lot more berries to work with now, so we can extend the Berry mechanic way into the next regions (aka slower growth).

We do want to balance the Sprinklotad to ensure that at the end game the player can have 100 replant so the farm can become fully passive. Whether that means 2x, or something less with Replant aura berries around is to be decided still.

RedSparr0w: IMO it shouldn't ever be a 100% replant chance from just the Oak Item, especially with how quickly the oak items level up.

RedSparr0w: We could probably also increase the upgrade cost, as that also seems quite low.

RedSparr0w: Maybe these amount would work okay, What do you think?
```suggestion
            new BoughtOakItem(OakItems.OakItem.Squirtbottle, 'Squirtbottle', 'Increases the chance of berry mutations', 'Johto Berry Master',
                true, [1.5, 1.75, 2, 2.25, 2.5], 1, 10, undefined, undefined, AmountFactory.createArray([1000, 2000, 5000, 10000, 20000], GameConstants.Currency.farmPoint)),
            new BoughtOakItem(OakItems.OakItem.Sprinklotad, 'Sprinklotad', 'Increases the chance of berry replants', 'Hoenn Berry Master',
                true, [1.15, 1.3, 1.45, 1.6, 1.75], 1, 2, undefined, undefined, AmountFactory.createArray([1000, 2000, 5000, 10000, 20000], GameConstants.Currency.farmPoint)),
```

src/scripts/items/EvolutionStone.ts
```diff
@@ -59,3 +59,35 @@ ItemList['Razor_fang']    = new EvolutionStone(GameConstants.StoneType.Razor_fan
...
+        'Magmarizer',
+        'Protector',
+        'Reaper_cloth',
+        'Razor_claw',
+        'Razor_fang',
+        'Prism_scale',
+        'Sachet',
+        'Whipped_dream',
+    }
+}
```
RedSparr0w, line 93: I don't think we should do this, as it just creates more places that need to be edited when we add a new evo item, and it's the same output as `GameConstants.StoneType` (without the `None`).

src/scripts/items/BattleItem.ts
```diff
@@ -20,4 +20,15 @@ ItemList['xClick']          = new BattleItem(GameConstants.BattleItemType.xClick
...
+namespace BattleItems {
+    export enum BattleItem {
+        'xAttack' = 0,
+        'xClick',
+        'Lucky_egg',
+        'Token_collector',
+        'Item_magnet',
+        'Lucky_incense',
+    }
+}
```
RedSparr0w, line 34: I don't think we should do this, as it just creates more places that need to be edited when we add a new battle item.
If we need a numeric enum, we should probably modify `GameConstants.BattleItemType` to fit our needs.

src/scripts/items/ShovelItem.ts
```diff
@@ -0,0 +1,18 @@
...
+    gain(amt: number) {
+        GameHelper.incrementObservable(App.game.farming.shovelAmt, amt);
+    }
+
+    use() {
+    }
+}
+
+// TODO: HLXII - Balance cost based on highest yield strategy
+ItemList['Berry_Shovel']   = new ShovelItem(500, 'Berry Shovel');
```
RedSparr0w, line 18: Would be good to get all the TODOs done before merge as well where possible,
Not sure what this should cost though.
```suggestion
ItemList['Berry_Shovel']   = new ShovelItem(500, 'Berry Shovel');
```

src/scripts/items/BattleItem.ts
HLXII: Fixed, though I couldn't use SeededRand.fromEnum, since the enum values for BattleItemType aren't numbers.

RedSparr0w (approved):
Changes look great :tada:

Thanks for the PR!

PR #974, opened 2020-10-21
feature(update): improve consistency of updates to saves

SpenserJ:
This fixes a bug that could occur if you import and old save, but refresh between the save being updated, and the save being persisted. This will also prevent the game from loading if an update fails, and will provide additional details about which update failed and the state before and after the failure:
image

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -237,60 +237,18 @@ class Update implements Saveable {
         return compareVersion.localeCompare(version, undefined, { numeric: true }) === 1;
     }
 
-    backup() {
+    check() {
         if (this.saveVersion === this.version || this.saveVersion === '0.0.0') {
             return;
         }
-
-        const playerData = this.getPlayerData();
-        const saveData = this.getSaveData();
-        const settingsData = this.getSettingsData();
-        const backupSaveData = { player: playerData, save: saveData };
-
-        try {
-            const button = document.createElement('a');
-            button.className = 'btn btn-block btn-danger';
-            button.innerText = 'Click to Backup Save!';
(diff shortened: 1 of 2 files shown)


RedSparr0w (changes requested):
Changes look good to me!

Could we also add a download backup save button to the error message so it will be easier for us to debug if we do encounter any problems.
image

Edit:
Another thing I just realized, you can't access the start menu if your save has an error, which means you are stuck in the loading screen.

src/scripts/Update.ts
```diff
@@ -53,311 +239,86 @@ class Update implements Saveable {
...
+                    callback(updateData);
+                    updateData.saveData.update.version = version;
+                    return updateData;
+                } catch (e) {
+                    console.error(`Caught error while applying update v${version}`, e, { beforeUpdate, updateData });
+                    Notifier.notify({
+                        title: `Failed to update to v${this.version}!`,
+                        message: 'Please check the console for errors, and report them on our Discord.',
+                        type: NotificationConstants.NotificationOption.danger,
+                        timeout: 6e4,
```
RedSparr0w, line 276: Maybe we should extend the timeout of this notification.
```suggestion
                        timeout: GameConstants.HOUR,
```

SpenserJ: I set it to a full day, just in case someone walks away from the computer mid-update. If they take longer, the notification will appear, but most people will refresh the page and get the error again

SpenserJ:
> Could we also add a download backup save button to the error message so it will be easier for us to debug if we do encounter any problems.
> Another thing I just realized, you can't access the start menu if your save has an error, which means you are stuck in the loading screen.

I've added a download button, changed the colors (because it was red on red), and added a reset button to the save because the start menu would have needed to be reworked so most of the items were disabled, and the z-index would need to be lifted for the menu and settings modals.
image

RedSparr0w (approved):

PR #975, opened 2020-10-21
Make Underground a Feature

fujnw:
Move `Underground` into a part of `Game`, so energy, upgrades and mine state can be saved in the same way `Feature` does. With this change, `Mine` and `mineInventory` remain static objects.

### DIFF
--- a/src/scripts/Game.ts
+++ b/src/scripts/Game.ts
@@ -25,6 +25,7 @@ class Game {
         public oakItems: OakItems,
         public party: Party,
         public shards: Shards,
+        public underground: Underground,
         public farming: Farming,
         public logbook: LogBook,
         public redeemableCodes: RedeemableCodes,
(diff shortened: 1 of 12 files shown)


src/scripts/Game.ts
```diff
@@ -152,8 +154,8 @@ class Game {
         if (Underground.counter >= GameConstants.UNDERGROUND_TICK) {
             Underground.energyTick(Math.max(0, Underground.energyTick() - 1));
             if (Underground.energyTick() == 0) {
-                Underground.gainEnergy();
-                Underground.energyTick(Underground.getEnergyRegenTime());
+                this.underground.gainEnergy();
+                Underground.energyTick(this.underground.getEnergyRegenTime());
             }
```
RedSparr0w, line 159: Should we move this to be in the `Underground` code?
Similar to how the other features handle the ticks with a `tick()` function.

RedSparr0w (approved):
Changes look good to me 👍 
Thanks for the PR!

_I was so confused how the data was being loaded, forgot that it was already a part of `save` not `player` 😅_

PR #976, opened 2020-10-21
Safari Zone changes

RedSparr0w:
Double the weight of any uncaught Pokémon so they have a higher chance of appearing.
Add a 50% chance to spawn a Pokémon in the overworld every 10 steps to make it easier to choose what you want to encounter.
Shiny overworld Pokémon will have the animated sparkle emoji.
SafariShiny

image
image

### DIFF
--- a/src/scripts/safari/Safari.ts
+++ b/src/scripts/safari/Safari.ts
@@ -327,7 +327,7 @@ class Safari {
         pokemon.element.style.left = `${offset.left}px`;
         document.getElementById('safariBody').appendChild(pokemon.element);
 
-        pokemon.steps = 40 + Math.floor(Math.random() * 21);
+        pokemon.steps = this.sizeX() + this.sizeY() + Math.floor(Math.random() * 21);
         this.pokemonGrid[y][x] = pokemon;
     }
 
--- a/src/scripts/safari/Safari.ts
+++ b/src/scripts/safari/Safari.ts
@@ -2,9 +2,11 @@
 
 class Safari {
     static grid: Array<Array<number>>;
+    static pokemonGrid: Array<Array<SafariPokemon>>;
     static player: Point = new Point(12, 20);
     static lastDirection = 'up';
     static nextDirection: string;
+    static steps = 0;
     static walking = false;
     static isMoving = false;
     static queue: Array<string> = [];
@@ -25,6 +27,7 @@ class Safari {
 
     public static load() {
         Safari.grid = [];
+        Safari.pokemonGrid = [];
         Safari.playerXY.x = 0;
         Safari.playerXY.y = 0;
         Safari.lastDirection = 'up';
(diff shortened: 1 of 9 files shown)


src/scripts/safari/Safari.ts
```diff
@@ -284,6 +291,46 @@ class Safari {
...
+        const offset = {
+            top: 32 * y - 8,
+            left: 32 * x - 4,
+        };
+        pokemon.element = document.createElement('div');
+        pokemon.element.classList.value = `pokemonSprite walkDown ${pokemon.shiny ? 'shiny' : ''}`;
+        pokemon.element.style.backgroundImage = `url('assets/images/dynamic-background/pokemon/${pokemon.id.toString().padStart(3, '0')}${pokemon.shiny ? 's' : ''}.png')`;
+        pokemon.element.style.top = `${offset.top}px`;
+        pokemon.element.style.left = `${offset.left}px`;
+        document.getElementById('safariBody').appendChild(pokemon.element);
```
Aegyo, line 328: This looks like something we should be doing with knockout

src/scripts/safari/Safari.ts
```diff
@@ -25,15 +27,18 @@ class Safari {
...
         Safari.playerXY.y = 0;
         Safari.lastDirection = 'up';
         Safari.inBattle(false);
         Safari.inProgress(true);
         Safari.balls(this.calculateStartPokeballs());
         for ( let i = 0; i < this.sizeY(); i++) {
-            const row = [...Array(this.sizeX())].map(Number.prototype.valueOf, 0);
+            const row = [...Array(this.sizeX())].fill(0);
             Safari.grid.push(row);
+            const pokemonRow: Array<null> = [...Array(this.sizeX())].fill(null);
```
SpenserJ, line 40: What is the purpose of spreading here?
```suggestion
            const pokemonRow: Array<null> = Array(this.sizeX()).fill(null);
```

src/scripts/safari/Safari.ts
SpenserJ: I believe all of the safari minigame (other than the modal+wrapping content and battle) is being rendered with straight JS right now, since a decent amount of code is for setting positions and animation frames. I'm torn on this because I actually prefer it as pure JS, and porting to knockout would be a bit of work, but it is also inconsistent with the rest of the game. I'm content leaving this outside of knockout for now though

SpenserJ (approved):
:+1: Seems like a great way to prevent frustrating safaris

src/scripts/safari/Safari.ts
RedSparr0w: I've moved it to knockout bindings 👍

src/scripts/safari/Safari.ts
RedSparr0w: Sorted 👍

PR #981, opened 2020-10-22
Boost dungeon tokens gained

RedSparr0w:
Boost amount of Dungeon Tokens gained slightly as per Discord poll:
image

Graph of old vs new
https://docs.google.com/spreadsheets/d/1Kz4Bj4EV9gMVwhYYiwFMYscRhqnpny_hLWm7d3WeXpE/edit?usp=sharing
image

Each bump is a new region _(this graph shows all 8 regions)_
Previously there would only be a bump every 3 regions.
Region | Average Increase
:-- | --:
Kanto | 14.96%
Johto | 62.05%
Hoenn | 1.15%
Sinnoh | 24.67%
Unova | 48.35%
Kalos | 12.52%
Alola | 28.14%
Galar | 43.76%


anapple96:
I feel like the boosts are quite low in higher regions, especially Hoenn

RedSparr0w:
Hoenn is a low increase because of the previous region rounding, It's not supposed to be a massive boost, just a small amount to bump it up.

PR #991, opened 2020-10-23
Make Zorua and Larvesta obtainable

Ultima1990:
Zoroark and Volcarona are both dungeon bosses.
Zorua and Larvesta are not available anywhere.

Solution: make them babies.


RedSparr0w (approved):
Thanks for the PR!
👍 Looks good to me, seems to be the way to obtain them in the main series:
image
image

PR #992, opened 2020-10-23
Actually deduct shards when purchasing a shard upgrade

Mephistic:
On the dev branch, I notied that shard purchases weren't actually deducting shards when buying shard upgrades. It looks like this started happening a few days ago with the ArrayOfObservable changes. We moved the shard count update in `gainShards` into an `if (amt > 0)` block, even though we remove shards by passing a negative amount  into `gainShards`.


RedSparr0w (approved):
Nice catch!

Thanks for the PR 👍

PR #993, opened 2020-10-23
Adding Deoxys Forms to Battle Frontier Shop

Mephistic:
This PR adds the Deoxys forms as purchases in the Battle Frontier shop after Deoxys has been caught. This seems like a pretty logical place to put them given that Deoxys is unlocked in the Battle Frontier and it gives players something to save their BP for besides more master balls.

Screen Shot 2020-10-23 at 4 27 24 PM

I grabbed the images with the same method described for the other pokemon item images (tinified 120x120 pngs of Pokemon Shuffle sprites).


src/scripts/items/PokemonItem.ts
```diff
@@ -43,3 +56,7 @@ ItemList['Burmy (plant)']        = new PokemonItem(GameConstants.PokemonItemType
 ItemList['Cherubi']              = new PokemonItem(GameConstants.PokemonItemType.Cherubi, 5000);
 ItemList['Spiritomb']            = new PokemonItem(GameConstants.PokemonItemType.Spiritomb, 5000);
 ItemList['Meloetta (pirouette)'] = new PokemonItem(GameConstants.PokemonItemType['Meloetta (pirouette)'], 50000);
+
+ItemList['Deoxys (attack)']      = new PokemonItem(GameConstants.PokemonItemType['Deoxys (attack)'], 500, GameConstants.Currency.battlePoint, new ObtainedPokemonRequirement(pokemonMap.Deoxys));
+ItemList['Deoxys (defense)']     = new PokemonItem(GameConstants.PokemonItemType['Deoxys (defense)'], 500, GameConstants.Currency.battlePoint, new ObtainedPokemonRequirement(pokemonMap.Deoxys));
+ItemList['Deoxys (speed)']       = new PokemonItem(GameConstants.PokemonItemType['Deoxys (speed)'], 500, GameConstants.Currency.battlePoint, new ObtainedPokemonRequirement(pokemonMap.Deoxys));
```
RedSparr0w, line 62: Do you think 1000 would be too expensive?
```suggestion
// Battle Frontier
ItemList['Deoxys (attack)']  = new PokemonItem(GameConstants.PokemonItemType['Deoxys (attack)'], 1000, GameConstants.Currency.battlePoint, new ObtainedPokemonRequirement(pokemonMap.Deoxys));
ItemList['Deoxys (defense)'] = new PokemonItem(GameConstants.PokemonItemType['Deoxys (defense)'], 1000, GameConstants.Currency.battlePoint, new ObtainedPokemonRequirement(pokemonMap.Deoxys));
ItemList['Deoxys (speed)']   = new PokemonItem(GameConstants.PokemonItemType['Deoxys (speed)'], 1000, GameConstants.Currency.battlePoint, new ObtainedPokemonRequirement(pokemonMap.Deoxys));
```

RedSparr0w (approved):
Looks good to me 👍 

Do you think we should have `Deoxys (attack)` as a higher stage reward in the Frontier instead of a shop item, as you need more attack to reach the higher stages 🤷

src/scripts/items/PokemonItem.ts
Mephistic: Works for me. The 500 BP cost was semi-arbitrary - I was just going off the initial master ball cost as a benchmark for "expensive, but not crazy expensive".

Mephistic:
That's a fair question - moving them to milestones was the other option here. If we want to do that, it'd probably be better to keep things consistent and either make them all purchases or all milestones.

FWIW, my reasons for going for shop items over milestones were all fairly weak, so I'm open to switching things up:
* The other Deoxys forms are actually all weaker than base Deoxys with the current attack formula, so I didn't think they felt particularly milestone-y
* The milestones are more work to balance than shop items. Players in the discord seem to have a pretty wide range of attack values by the end of Hoenn, while all players will be able to eventually grind their way up to shop purchases.

That said, the forms aren't mandatory and there's still one legendary that seems to need a real home: Jirachi.

How's this for a counterproposal:
* Add the Deoxys Forms as new battle frontier milestones at levels 125, 150, and 175.
  * Without factoring in shards, it takes roughly 100K attack to reach stage 100, 200K attack to reach stage 125, 300K attack to reach stage 150, and 400K attack to reach stage 175. Those are just an upper bound - shards could drive that down substantially.
* Add Jirachi as a shop item for 1,000 BP

That lets us reward raw power with new milestones (the forms, which aren't mandatory and won't block the way to Sinnoh) and sheer grit with a necessary mythical (grinding up BP for Jirachi, which should still take less time and be less variable than the current ~1/12,000 chance of finding it as a roamer).

src/scripts/items/PokemonItem.ts
```diff
@@ -43,3 +56,7 @@ ItemList['Burmy (plant)']        = new PokemonItem(GameConstants.PokemonItemType
 ItemList['Cherubi']              = new PokemonItem(GameConstants.PokemonItemType.Cherubi, 5000);
 ItemList['Spiritomb']            = new PokemonItem(GameConstants.PokemonItemType.Spiritomb, 5000);
 ItemList['Meloetta (pirouette)'] = new PokemonItem(GameConstants.PokemonItemType['Meloetta (pirouette)'], 50000);
+// Battle Frontier
+ItemList['Deoxys (attack)']  = new PokemonItem(GameConstants.PokemonItemType['Deoxys (attack)'], 1000, GameConstants.Currency.battlePoint, new ObtainedPokemonRequirement(pokemonMap.Deoxys));
+ItemList['Deoxys (defense)'] = new PokemonItem(GameConstants.PokemonItemType['Deoxys (defense)'], 1000, GameConstants.Currency.battlePoint, new ObtainedPokemonRequirement(pokemonMap.Deoxys));
+ItemList['Deoxys (speed)']   = new PokemonItem(GameConstants.PokemonItemType['Deoxys (speed)'], 1000, GameConstants.Currency.battlePoint, new ObtainedPokemonRequirement(pokemonMap.Deoxys));
```
anapple96, line 62: I personally don't agree with these prices. They're extremely low. Imo, they should be around 15 - 25k BP

PR #996, opened 2020-10-24
Add Pokémon categories

### LINKED ISSUE #982: [SUGGESTION] add a way to custom sort pokemon in the hatchery
basically what the title says, add a favorites feature or maybe like a folder to put a select group of pokemon into so when your hatching you dont have to scroll through all the pokemon of a region to find it in the middle

**Current Behavior**
currently you can sort by attack and a few other methods and narrow down pokemon by region type and if they are shiny

**Improved Behavior**
id like an option added maybe to where you select region where instead of any of the regions its a category where it has your selected pokemon all in it

**Reason for change**
this would make it easier when trying to level up specific pokemon that are players favorites despite them being weaker and further down the list generally, this would also help when the auto incubate feature is added by making it easier to group which pokemon you want incubating by keeping them all next to each other when there is no other way to have them grouped close to each other based on the current filter options.

**Detailed Description**
just a folder or folders letting you group specific pokemon together making it easier to find select groups when using the day care.

RedSparr0w:
closes #982 

Allow users to set a custom category for each Pokémon,
By default there is a `None` and `Favorite` category.
image

You can add, remove and reset categories from the `Edit` menu option:
image

Clicking the color will bring up a color picker (different display depending on browser):
image

The category names are in a `contenteditable` `<td>` so they can be edited directly:
image

Clicking `+ New` will just create a basic category:
image

You can sort by category:
image

You can filter to display certain categories:
image


src/modules/koExtenders.d.ts
```diff
@@ -2,4 +2,8 @@ export declare module 'knockout' {
     export interface ExtendersOptions {
         numeric: number;
     }
+
+    export interface BindingHandlers {
+        contentEditable;
+    }
```
RedSparr0w, line 8: Note that I have no clue what I am doing here, so this may be wrong 😅

PR #1001, opened 2020-10-24
Make Sinnoh stones usable on Unova mons

Ultima1990:
I searched for "StoneType.none" in PokemonList.ts
I've changed the stonetype for Lampent to Dusk stone, and Minccino to Shiny stone.
I've changed Bonsly and Mine Jr. to levelup evolutions. Because why not. Not that it changes anything in practice, oh well. They are both move based evolutions, both based on Mimic. Bonsly learns it at 17, Mime Jr. at 18.

I also changed some Kalos Pokémon, best to avoid potential future problems with that:
I've changed the stonetype for Floette to Shiny stone, and Doublade to Dusk stone.
There are still several StoneType.none situations in there. Happiny and Mantyke aren't very important. 

The rest is Alola stuff:
Charjabug and Crabrawler are Alola location based evolutions. I haven't touched those. I have changed Steenee, which is also a move based evolution. Stomp, at 28. Poidole is also move based. From Dragon Pulse. Which it learns at level 1. Honestly, I have no idea for this one. Level 1 is a bit early for a UB. Finally there is Meltan. Which. Uhm. Yeah. 400 Meltan Candy in PoGo. IDK.

### DIFF
--- a/src/scripts/pokemons/PokemonList.ts
+++ b/src/scripts/pokemons/PokemonList.ts
@@ -8434,7 +8434,7 @@ const pokemonList: PokemonListData[] =
             'levelType': LevelType.mediumfast,
             'exp': 58,
             'catchRate': 255,
-            'evolutions': [new StoneEvolution('Bonsly', 'Sudowoodo', GameConstants.StoneType.None)],
+            'evolutions': [new LevelEvolution('Bonsly', 'Sudowoodo', 17)],
             'baby': true,
             'base': {
                 'hitpoints': 50,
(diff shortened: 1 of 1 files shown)


Mephistic:
Ooh, Poipole is an interesting one - the most game-accurate way is probably to make Heart Scale an evolution item (this would also technically be more accurate for Mamoswine).

Meltan could be a new evolution type based on Times Captured?

RedSparr0w (approved):
Changes look good to me 👍 
Thanks for the PR!

It may be worth opening an issue for the other evolutions so they aren't forgotten about.

PR #1002, opened 2020-10-24
fix(breeding): Don't reset party Pokémon level if not breeding

### LINKED ISSUE #951: [BUG] Hatching a shop egg resets your existing pokemon level 1
**Info**
 - Version 0.5.7

**Description:**
When an egg (when its a mystery egg or a bought egg from a store)  hatches and its already a pokemon I own, the new one and its stats, ie attack power, replaces the older pokemon.
Example: my dragonite who was level 100 got replaced with a level 1

**Expected Behavior:**
I thought it would be like when I capture that pokemon for a second time on any of the routes

aufde:
closes #951


aufde:
Fix for https://github.com/pokeclicker/pokeclicker/issues/951

RedSparr0w:
This will also prevent shop eggs from increasing Pokémon attack when hatched though too.

RedSparr0w (approved):
Changes look good to me 👍 

Thanks for the PR!

RedSparr0w (commented):
Sorry after a second look, we might also need to wrap the evolution stuff in this too?

PR #1004, opened 2020-10-25
Adding Hoenn Mythicals as Battle Frontier Rewards

Mephistic:
Based on feedback in the last PR, this change:
* Adds the Deoxys Forms as rewards for new Battle Frontier milestones
* ~Makes Jirachi purchasable in the Battle Frontier shop~

Screen Shot 2020-10-26 at 11 52 25 PM

~Open to tweaking costs/stages here - The important concern is just making sure Jirachi is affordable by the end of Hoenn. The Deoxys forms feel a bit on the easy side, but I thought it'd be better to err on the side of making them available in Hoenn (even if they aren't mandatory for moving on).~ Based on Discord feedback, I've decoupled the Jirachi change from this PR. Still interested in doing something more interesting with Jirachi, but it should probably be pushed off until we stop using the Hoenn roamers as a pseudo-DPS check for early Sinnoh.


Mephistic:
Some changes here based on Discord feedback:
* Removed the Jirachi change - that was really more of a side suggestion anyway, and it sounds like we're using Hoenn roamers as a pseudo-DPS check and shouldn't mess too much with those until we've decoupled those concepts.
* Increased the Deoxys form milestones by a good bit. Bumped them from 125/150/175 to 151/251/386 because they're not blocking the way to the next region and it looks like I was underestimating how far people were getting into the Battle Frontier. Back-of-the-napkin says you need around ~450K raw attack before max shards and X-attacks up to beat 386, probably less with good type matchups.

RedSparr0w (approved):
👍 Changes look good to me, Thanks for the PR!

PR #1005, opened 2020-10-25
feat(pokeball-selector): Add quantity to the Pokéball selector modal

RedSparr0w:
Display current quantity in the Pokéball selector modal:
image


anapple96 (approved):
Haha, *pickle*

PR #1006, opened 2020-10-25
Bump version to 0.6.0 for semantic versioning

RedSparr0w:
Start using [semantic versioning](https://semver.org/)

## Summary
Given a version number `MAJOR.MINOR.PATCH`, increment the:

`MAJOR` version when you make incompatible API changes,
`MINOR` version when you add functionality in a backwards compatible manner, and
`PATCH` version when you make backwards compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the `MAJOR.MINOR.PATCH` format.

_We will generally only bump the `MINOR` version with each live website release._
_And the `PATCH` version if there are any small semi-urgent changes needed on the live website._


SpenserJ (approved):
Looks good to me! :+1: I pushed a fix for guarding against routeKills being converted already, just so that dev saves will also update

src/scripts/Update.ts
```diff
@@ -146,24 +146,33 @@ class Update implements Saveable {
...
-                return acc;
-            }, {});
-            saveData.statistics.routeKills = result;
+            // Re-map route kill statistics
+            if (
+                typeof saveData.statistics.routeKills === 'object' &&
+                Array.isArray(saveData.statistics.routeKills) === false
+            ) {
+                console.log('Skipping 0.6.0 statistics routeKills update because it is already an object');
+            } else {
```
RedSparr0w, line 156: I'm not sure about adding this part, seems unnecessary as dev saves shouldn't be main saves, and they can easily be fixed by just manually changing the version number from `0.5.8` -> `0.6.0`.

It's similar to when we add other data that will need to be modified, if the save is already on the latest version, the update functions won't be run, and the save is unlikely to work with the new version.

PR #1007, opened 2020-10-25
fix(statistics): Fix bug with length property

### LINKED ISSUE #947: [BUG] Statistics complaining in console with a fresh save
Lots of this on a fresh save with the current develop branch (48d1cbd425493b344b19dbfa430d9c6755c5d586)
```
console.trace() [Statistics] [berriesHarvested] Invalid property requested: length StatisticStore.ts:74:37
    get webpack:///./src/modules/DataStore/StatisticStore.ts?:74
    toJSON webpack:///./src/modules/DataStore/StatisticStore.ts?:145
    toJSON webpack:///./src/modules/DataStore/StatisticStore.ts?:144
    load http://localhost:3000/scripts/script.min.js?v=0.5.7:19194
    load http://localhost:3000/scripts/script.min.js?v=0.5.7:19190
    initialize http://localhost:3000/scripts/script.min.js?v=0.5.7:19208
    start http://localhost:3000/scripts/script.min.js?v=0.5.7:21
```

SpenserJ:
This is a small fix that should prevent the warning about `length` being an invalid property on new saves (because `0` is a falsey value). This will return whatever value is set, as long as it isn't undefined, so both `0` and an observable will be properly returned

closes #947

### DIFF
--- a/src/modules/DataStore/StatisticStore/index.ts
+++ b/src/modules/DataStore/StatisticStore/index.ts
@@ -133,7 +133,7 @@ export default class Statistics implements Saveable {
             // We use a proxy to generate new array observables on the fly.
             this[array] = new Proxy([], {
                 get: (statistics, prop: string) => {
-                    if (statistics[prop]) {
+                    if (typeof statistics[prop] !== 'undefined') {
                         return statistics[prop];
                     }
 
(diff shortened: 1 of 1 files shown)


RedSparr0w (approved):
Changes look good to me, thanks for the PR 👍

PR #1008, opened 2020-10-26
change(encounters): Use Pokémon Yellow version encounters

Ultima1990:
Changed many encounters in Kanto routes and dungeons to make them more accurate to Yellow.
You'll notice encounters for routes 8 and 7 do not match what you will find on Bulbapedia. That's because those routes are in the wrong place. They need to be swapped. With this change to Routes.ts the Pokémon are in the right place, the route numbers are not.
I kept Tangela on route 21 and Chansey in Cerulean Cave as a mercy. They aren't there in Yellow. If the new Safari change works out this might be something to come back to. (In that case the Venonat line should also be removed from Safari.)
There are wild Pikachu and Raichu where there shouldn't be, even after these changes. Yellow has 0 wild Pikachu. Can't have that of course, gotta keep the yellow mouse available. Could move it to an egg...
I've kept version exclusives in mind. Everything should still be obtainable though it's very possible I missed something so please look it over.

### DIFF
--- a/src/scripts/dungeons/Dungeon.ts
+++ b/src/scripts/dungeons/Dungeon.ts
@@ -75,7 +75,7 @@ dungeonList['Digletts Cave'] = new Dungeon('Digletts Cave',
 );
 
 dungeonList['Mt. Moon'] = new Dungeon('Mt. Moon',
-    ['Sandshrew', 'Clefairy', 'Zubat', 'Paras', 'Geodude', 'Pidgeotto'],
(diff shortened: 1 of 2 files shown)


RedSparr0w (approved):
Looks good to me, did some manual checking and ran this to check all the Kanto mons were still obtainable:
```js
((region) => pList.filter(p => p.nativeRegion == region).filter(p => !Object.keys(p.locations).length))
(GameConstants.Region.kanto)
```
But still might have missed some, 🤞 hopefully not!

PR #1013, opened 2020-10-27
Add Liberty Garden

Ultima1990:
Also removed Strange House, Desert Resort, Virbank Complex, Undella Bay and Wellspring Cave dungeons.


RedSparr0w (changes requested):
Could you also remove these from the GameConstants,
Will be under the `UnovaDugneons` and the different area types _(Forest, Cave etc)_.

Ultima1990:
> Could you also remove these from the GameConstants,
> Will be under the `UnovaDugneons` and the different area types _(Forest, Cave etc)_.

Right, yup, forgot about that. Done.

RedSparr0w (approved):
Changes look good to me!

Thanks for the PR 👍

PR #1015, opened 2020-10-27
Add Achievements and Quest Completes to log book

jaahay:
Helpful for if you idle for extended periods of time. Also, fosters a great sense of satisfaction.

### DIFF
--- a/docs/scripts/script.min.js
+++ b/docs/scripts/script.min.js
@@ -21629,6 +21629,9 @@ class Achievement {
                 timeout: 1e4,
                 sound: NotificationConstants.NotificationSound.achievement,
             });
+            App.game.logbook.newLog(
+                LogBookTypes.ACHIEVEMENT,
+                `Earned "${this.name}".`);
             player.achievementsCompleted[this.name] = true;
             this.unlocked = true;
         }
@@ -27181,12 +27184,18 @@ class Quest {
                     message: `You have completed your quest and claimed ${this.pointsReward} quest points!`,
                     type: NotificationConstants.NotificationOption.success,
                 });
+                App.game.logbook.newLog(
+                    LogBookTypes.QUEST_COMPLETE,
+                    `Completed "${this.description}" for ${this.pointsReward} quest points.`);
             }
             else {
                 Notifier.notify({
                     message: 'You have completed a quest!',
                     type: NotificationConstants.NotificationOption.success,
                 });
+                App.game.logbook.newLog(
+                    LogBookTypes.QUEST_COMPLETE,
+                    `Completed "${this.description}".`);
             }
             GameHelper.incrementObservable(App.game.statistics.questsCompleted);
             return true;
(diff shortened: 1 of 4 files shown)


src/scripts/logBook/LogBookTypes.ts
```diff
@@ -24,4 +24,12 @@ const LogBookTypes: Record<string, LogBookType> = {
         display: 'primary',
         label: 'FOUND',
     },
+    ACHIEVEMENT: {
+        display: 'success',
+        label: '\'CHIEVE',
```
RedSparr0w, line 29: I'm not sure about shortening this to `'CHIEVE`

jaahay: `ACHIEVE` or `ACHIEVEMENT`
I wanted to keep the text width same as the other labels

docs/scripts/script.min.js
```diff
@@ -29480,29 +29489,29 @@ SpecialEvents.events = [];
 // TODO: Fetch events from a server each 1/2/3/6/12/24 hours?
 // Create our events here for now
 // Yearly
-SpecialEvents.newEvent('Flying Pikachu', 'Encounter Flying Pikachu for a limited time roaming Kanto.',
```
jaahay, line 29483: I think Atom on Windows is proactively trying to change some file endings. Also, I think git hates me

jaahay:
Feel free to tweak any text or other UI stuff. Much appreciated

docs/scripts/script.min.js
RedSparr0w: The `docs` file shouldn't be update, we only update this when we push to master for the live website.

jaahay: Understood! Thanks for (another) assist :)

src/scripts/logBook/LogBookTypes.ts
```diff
@@ -24,4 +24,12 @@ const LogBookTypes: Record<string, LogBookType> = {
...
         label: 'FOUND',
     },
+    ACHIEVEMENT: {
+        display: 'success',
+        label: 'ACHIEVE',
+    },
+    QUEST_COMPLETE: {
+        display: 'success',
+        label: 'QUEST',
+    },
```
RedSparr0w, line 34: Maybe we could use something like `warning` and `info` here instead to make them stand out from the rest of the logs:
```suggestion
    ACHIEVEMENT: {
        display: 'warning',
        label: 'ACHIEVE',
    },
    QUEST_COMPLETE: {
        display: 'info',
        label: 'QUEST',
    },
```

Before | After
:-- | :--
image | image

jaahay: Looks sleeker with the `warning` and `info` colours

RedSparr0w (approved):
Looks good 👍,

Thanks for the PR!

PR #1028, opened 2020-10-31
Fix/standardize infobox

HLXII:
Previously, we had multiple locations show helpful information using this button:
image
image

However, there are a couple of locations that don't use this standard. I've updated them so it's cleaner for the user to know that "Question mark button in the top right means info".

Hatchery:
Before:
image
After:
image

Damage Calculator:
Before:
image
image

Quest Line:
Before:
image
After:
image


RedSparr0w:
We need the `x` on modals for mobiles, otherwise it becomes a huge chore trying to close the modals by clicking the background.

Maybe we could move the `?` to the left hand side instead?

HLXII:
The Shard Modal doesn't have the X either? There's a Close button in the footer of all the modals as well

RedSparr0w:
As long as the updated modals have the footer and close button, then it should be fine _(I can't tell from the files changed tab)_.

I don't think we need for it on the Quest Lines though, as it's not really help information, just some minor information/flavor text.

RedSparr0w (approved):
Looks good 👍 

Thanks for the PR!

Issue #950, opened 2020-10-16
[SUGGESTION] Loadouts/Hotkeys


adapap:
As a midgame player, I find that a lot of the time I spent when playing actively is switching between Pokeball types and Oak Items to complete quests. I think for most, if not all players, a system to save and load "loadouts" will be very useful. If not loadouts, hotkeys that allow for quick micro-management of Oak Items.

**Current Behavior**

Pokeballs and Oak Items require mouse interaction (also the Hatchery and Berry Farm, but those are different). There is no quick way to change the settings.

**Improved Behavior**

One of two solutions:
- Hotkeys: 1-8 for different Oak Item toggles, and perhaps Shift+(1-4) to scroll through the Pokeball selector for each respective condition.
- Loadouts: Let the user use X currency to buy up to Y loadout slots. These loadouts save your Pokeball, Oak Item, Hatchery Filter/Sorting, etc. configuration.

**Reason for change**

It would improve accessibility and efficiency during active play. I haven't looked much into it, but I would assume this is a common thing to write scripts for.

**Detailed Description**

See above.

**Screenshots:**

anapple96:
Close, as it's already implemented?

RedSparr0w:
Closed via #1261

PR #1029, opened 2020-10-31
Hotkeys

adapap:
This PR aims to address one component of #950, hotkeys.
The change adds support for `O` and `1-8` to toggle Oak Items,
as well as `P` + `1-4` together to scroll through Pokeballs.

The idea is to improve micro-management during active play without needing
to write scripts or use the mouse to change these items.

Code review appreciated, as well as suggestions for potentially doing loadouts!


RedSparr0w:
👍 Works well for the Oak items,
Maybe we could make it that `P + x` opens the Modal for that particular setting instead of cycling through the Pokeball types, as once more Pokeball types are added it will definitely become impractical.

adapap:
Alright, now it works as described - P + x opens the correct modal, and then any of 1-9 or Numpad 1-9 will set the pokeball. I wasn't sure if 0 should be used to choose no pokeball, but it's a simple change if we wanted that. Also, could be my lack of experience with Knockout, but loading the Pokeball selectors felt very contrived since the current state is difficult to discern at first.

RedSparr0w (approved):
Changes look good, thanks for the PR 👍

PR #1033, opened 2020-11-02
fix maxAmount of masterballs in league and BF shop

LuqDragon:
masterballs2
TO
masterballs


RedSparr0w (approved):
Changes look good, Thanks for the PR!

PR #1035, opened 2020-11-02
Switch kanto route 7 and 8 positions

Ultima1990:
Let's see if I did this right.


RedSparr0w (approved):
Changes look good to me 👍 

But before I merge, are the route encounters still correct for each?

Ultima1990:
> Changes look good to me 👍
> 
> But before I merge, are the route encounters still correct for each?

Yes encounters are correct.

PR #1036, opened 2020-11-02
Fix Sinnoh route difficulty

Ultima1990:
Routes 211-215 basically unlock in reverse order.


RedSparr0w (approved):

PR #1037, opened 2020-11-02
Add hint npcs that tell about egg queue

Ultima1990:
I'm no writer. Please tell me what you think.

Cherrygrove City:
Mr. Pokémon
Welcome to Johto! This is where the first ever Pokémon egg was found long ago.
Astounding breakthroughs have been made since then. We can now store Pokémon eggs for longer and queue them up for breeding.
This new technology only allows up to four stored eggs, for now.

Littleroot Town:
Professor Birch's Aide
We have received word from Mr. Pokémon in Johto! He has made another breakthrough.
You can now store an additional four eggs in the queue! His research has really gained speed.
He wants you to know that he will have an additional eight slots ready by the time you reach Sinnoh.

The implication is that, before the invention of this technology, eggs would need to be sent to the hatchery immediately or they would... rot... or something. Now we can keep them "fresh" for a while.
The bit about Sinnoh is to prevent having to repeat this NPC in every region.

### DIFF
--- a/src/scripts/towns/Town.ts
+++ b/src/scripts/towns/Town.ts
@@ -335,6 +335,13 @@ const BlackthornCityShop = new Shop([
 ]);
 
 // Johto NPCs
+
+const CherrygroveMrPokemon = new NPC('Mr Pokémon', [
+    'Welcome to Johto! This is where the first ever Pokémon egg was found long ago.',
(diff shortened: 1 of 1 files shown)


RedSparr0w (approved):
Nice!

PR #1041, opened 2020-11-05
toggle shiny form to all pokedex or in statistic

LuqDragon:
I like to implement QoL suggestions, so i was think to get suggestions in discord and implement it

Suggested by (and others):
qol

this is not intuitive but is pretty, so i dont know what put in here
uhd


src/components/pokedex.html
```diff
@@ -84,6 +84,13 @@ <h4 style="margin-bottom: 0">Pokédex</h4>
                                       value="false" onclick="PokedexHelper.updateList()"> Rare Hold Item
                            </label>
                        </div>
+                       <!--Show all shiny form-->
+                       <div class="col-sm-2">
+                        <label class="form-check-label">
+                            <input class="form-check-input" type="checkbox" id="pokedex-filter-held-item"
+                                   value="true" checked onclick="PokedexHelper.toggleAllShiny(!PokedexHelper.toggleAllShiny())"> Show all shiny form
+                        </label>
+                    </div>
```
RedSparr0w, line 93: I'm not sure about this settings placement, it feels out of place here.

LuqDragon: i placed aligned with all others, have a better sugestion? Maybe a icon in front of "Showing X pokemon"?

RedSparr0w: I'm honestly not sure where else would make sense,
But being a part of the filtering options seems strange +(although tbf, those also need to be updated)_
Maybe just update the text to something like:
`Use non-shiny Pokémon images`
`Use default Pokémon images`
`Hide shiny Pokémon images`

RedSparr0w (approved):
Looks good, Thanks for the PR!

PR #1042, opened 2020-11-06
fixed jumpy battle screen

LuqDragon:
This bug was reported by Rishab on #bugs on discord, and happen on mobile or tiny screens

Example of the bug:
bug2


RedSparr0w (approved):

PR #1043, opened 2020-11-07
Make route 22 available earlier

Ultima1990:
It is how it be in RBY and FRLG. Mankey should help with clickless brock, might even be possible with itemless charmander with this! Not sure tbh, didn't do the math.


RedSparr0w (approved):

PR #1044, opened 2020-11-07
Update Unova route difficulty.

Ultima1990:
This should cover everything except 19 and 20. I'm really not sure how to do those correctly.
My gut tells me to go with 0.1 and 0.2, but I don't know if it will accept values below 1. Going with 1.1 and 1.2 might work, but route 1 is valued as 23.1, so wouldn't that make 19 and 20 too difficult? Maybe that's not how it works, IDK.


RedSparr0w (approved):

PR #1050, opened 2020-11-09
Saving breeding filters to settings

LuqDragon:
I applied to Region, Shiny Status, Types and Display Value. The categories dont fit on general method, so i dont put it and i dont know if need to include the search
filter


RedSparr0w (approved):

PR #1039, opened 2020-11-04
[WIP-Draft] Party Slots

apple096:
Created this PR to keep track of the progress

- [ ] Party Slots
   * [ ] UI
     * [ ] Container
       * [ ] Party Images
       * [ ] Ability Tooltips
       * [ ] Cooldown timer (see Battle Items Container)
     * [ ] Pokemon List Modal to assign to Party Slot. May want to create base template for Pokemon lists since they're used in a couple of locations now (Pokedex, Hatchery, now Party Slots).
   * [ ] Update PartyPokemon to store location (e.g. In Party Slots).
- [ ] Abilities
   * [x] Base Class - Probably needs just name and description.
     * [ ] Passives - Will probably need to store all current passives and add references to them in correct locations. (egg, berry growth, etc).
     * [ ] Custom - Some abilities might have additional modals to open for options. Not sure of this one
   * [ ] Base Abilities
     * [ ] General passive multipliers (egg, harvest, money, etc)
     * [ ] General damage actives (damage, type, cooldown)
   * [ ] Custom Abilities
     * [ ] TBD
   * [ ] Pokemon Assignment - Assigning Pokemon abilities. Will need to update PokemonList.
 - [ ] Ability Capsules
   * [ ] Implement Item
   * [x] Add to Shop
   * [ ] Implement backend usage (probably will be used in new Pokemon List modal for Party Slots)

PR #1051, opened 2020-11-10
Party List Refactor

HLXII:
This refactor is mostly setting up for using the Party List for Party Slots (#1039) and the PokeLab (#1048).

It refactors the current Breeding Modal elements into two templates: the PartyListFilterTemplate and the PartyListTemplate. This is so the sorting/Pokemon display elements can be shared between multiple modals (e.g. the Party Slot assignment modal and the Research assignment modal).

This also refactors the breeding property of PartyPokemon into a location property, to expand the different locations the pokemon can be in (e.g. the Party Slots or Researching).

### DIFF
--- a/src/components/breeding.html
+++ b/src/components/breeding.html
@@ -34,117 +34,14 @@ aria-labelledby="breedingModalLabel" aria-hidden="true">
 
             <div class="tab-content p-3">
                 <div class="tab-pane active" id="breeding-pokemon">
-                    <div class="text-left">
-                        <div class="form-row collapse" id="breeding-filter">
-                            <div class="form-group col-md-6 col-6">
-                                <label>Search</label>
-                                <input autocomplete="off" class="form-control" oninput="BreedingController.filter.search(new RegExp(`(${this.value})`, 'i'))" placeholder="search here"/>
-                            </div>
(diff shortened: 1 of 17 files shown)


anapple96 (approved):
Party Slots is bottlenecked behind this. Gotta get this merged fast

KrushGames:
So i'm assuming the pokelab is gonna be used for kyurem in some way, right? if so, it could be used for necrozma and calyrex as well hopefully

RedSparr0w:
I'm not sure we need these to be templates, as they may end up looking quite different and templates tend to be a lot more laggy than just inline code.

HLXII:
How did you intend for selecting pokemon to the Party Slots or the Research Workers to work?

RedSparr0w:
Not entirely sure, but it doesn't need to be the exact same as the breeding modal, otherwise it may as well just use the same modal.

HLXII:
The issue was that the Breeding modal contained a lot of extra stuff like the hatchery banners / tabs for unnamed eggs and fossils. I moved all the shared elements into the templates.

Is there a way to change the HTML at compile time to just inline the templates if they're that laggy? Otherwise there'll be a lot of code duplication if we don't use templates.

apple096:
This is long overdue

PR #1058, opened 2020-11-13
Hotfix to issue with pokemon evolution stones

LuqDragon:
image


RedSparr0w (approved):
🤦 OOF, Thanks for the PR!

PR #1061, opened 2020-11-13
Fix calculation of native pokemon region

LuqDragon:
Some pokemons (especially Deoxys) have the forms passing the region calculation, so `Math.floor` makes it work now

image


RedSparr0w (approved):

LuqDragon:
that was fast lol

PR #1067, opened 2020-11-14
Nerf eggCycle for some Pokemons

anapple96:
Heatran, Manaphy and Phione are legendaries. They have too little steps
Karp-line is extremely broken, because of the new Protein and the extremely low 5 cycle


DaveYognaught:
Karp-Line is broken, but it's not 200 > 3200 steps broken. Dear lord. W-Was that intentional?

I think 400 at max. Anything over that is overkill since by nature of Protein, it's just targetting lowest step cycle. This just puts it on level with the next powerful evolution line. Also, we keep all evolution lines on the same step cycle, don't overnerf Gyarados.

I do agree legendaries should have large step cycles though. I'd put those 3 legendaries on 3200 though in my opinion.

RedSparr0w:
I would rather not have random individual egg step changes applied to Pokémon that can breed in the games.

The legends being changed is fine, as their egg step data isn't really used in the games, so it is mostly just placeholder data.

If anything for the other Pokémon, it would be better to make the evolved forms take `1.x` the amount of steps as the previous form, and babies `0.x` amount of steps of their base form.

PR #1077, opened 2020-11-16
Fix console error jumping from dungeon to town

LuqDragon:
just a little thing i saw when playing on anapple unova map
console


RedSparr0w:
I haven't seen this problem on the main game, there's likely a problem somewhere with the Unova stuff which needs to be fixed.

We also can't use `?.` in our javascript (the data-bind stuff isn't parsed by typescript), as the support just isn't quite fully there yet.

LuqDragon:
i tested on my desktop version and have the same issue, you need to go to a town that dont have a dungeon access

RedSparr0w:
Ah yeah, that is strange, it seems to be the left overs of the old dungeon/town, even though the town should have changed already 🤔
This line _should_ already be handling it, not sure why it's not being picked up from there
image

LuqDragon:
Maybe it update faster than the knockout collapse/remove

RedSparr0w (approved):

PR #1079, opened 2020-11-17
Berry fixes round 5

### LINKED ISSUE #1078: [BUG] TypeError in EnigmaMutation unlock requirement check
**Info**
 - OS: Windows 10
 - Browser: Chrome/Firefox
 - Version: v0.6.3/v0.6.4

**Description:**
After unlocking the Wepear berry, the console gets spammed with TypeErrors caused by the EnigmaMutation trying to check whether we unlocked berry #67, which doesn't exist.

**Steps to Reproduce:**
1. Unlock the Wepear berry
2. Check the console

**Screenshots:**
image

HLXII:
- Fix Hondew mutation to match actual Berry flavor and be harder.
 - Update Enigma mutation algorithm to use 2nd and 3rd gen berries. So nobody can get Cheri and Starf as two requirements (which would suck lol). The berry timers should be closer with 2nd and 3rd gens. This also closes #1078 
 - Fix typo from "Ringo" to "Rindo"
 - Change some wording in the Farm Help menu to be more clear
 - Mutation balance changes
 - Nerf Auras quite a bit
 - Boost wandering pokemon shiny chance.

### DIFF
--- a/src/components/farmModal.html
+++ b/src/components/farmModal.html
@@ -233,6 +233,9 @@
 
                     <h4><u>Trades</u></h4>
                     <p>Berries can be traded with Berry Masters in later regions for other useful items</p>
+
+                    <h4><u>Wandering Pokémon</u></h4>
+                    <p>Ripe Berry plants may attract wild Pokémon that will join your party. Some Pokémon can only be found using certain Berry species.</p>
                 </div>
             </div>
 
--- a/src/components/farmModal.html
+++ b/src/components/farmModal.html
@@ -195,13 +195,13 @@
                     <br/>Once fully grown, the Berry plant can be clicked to manually harvest the Berries and gain some Farm Points.</p>
 
                     <p>Otherwise after some time being fully grown, the Berry plant will wither away,
-                    dropping only a fraction of the Berries but may also be re-planted automatically.
+                    dropping half of the expected Berries but may also be re-planted automatically.
                     <br/><i>(Farm Points are only gained by manually harvesting Berries)</i><p>
 
                     <p>Additional plots can be unlocked by spending Berries.</p>
 
                     <h4><u>Mutations</u></h4>
-                    <p>Berries have the chance to mutate into other Berries based on certain conditions.
+                    <p>Berries have the chance to cause mutations and create new Berry species based on certain conditions.
                     <br/><i>(the Berry Master in Kanto will give you daily hints to what these conditions may be)</i></p>
 
                     <h4><u>Mulches</u></h4>
--- a/src/components/farmModal.html
+++ b/src/components/farmModal.html
@@ -201,7 +201,7 @@
                     <p>Additional plots can be unlocked by spending Berries.</p>
 
                     <h4><u>Mutations</u></h4>
-                    <p>Berries have the chance to cause mutations and create new Berry species based on certain conditions.
+                    <p>Berries have the chance to cause mutations and create new Berry species in empty plots based on certain conditions.
                     <br/><i>(the Berry Master in Kanto will give you daily hints to what these conditions may be)</i></p>
 
                     <h4><u>Mulches</u></h4>
--- a/src/scripts/farming/mutation/Mutation.ts
+++ b/src/scripts/farming/mutation/Mutation.ts
@@ -11,7 +11,7 @@ abstract class Mutation implements Saveable {
 
     defaults: Record<string, any>;
 
-    mutationChance: number;
+    _mutationChance: number;
     mutatedBerry: BerryType;
     _hint?: string;
     showHint: boolean;
@@ -20,7 +20,7 @@ abstract class Mutation implements Saveable {
     _hintSeen: KnockoutObservable<boolean>;
 
     constructor(mutationChance: number, mutatedBerry: BerryType, options?: MutationOptions) {
-        this.mutationChance = mutationChance;
+        this._mutationChance = mutationChance;
         this.mutatedBerry = mutatedBerry;
         this._hint = options?.hint;
         this._unlockReq = options?.unlockReq;
@@ -74,6 +74,14 @@ abstract class Mutation implements Saveable {
     }
 
     /**
+     * Handles getting the mutation chance
+     * @param idx The plot index
+     */
+    mutationChance(idx: number): number {
+        return this._mutationChance;
+    }
+
+    /**
      * Update tag for mutations. Returns true if this mutation will occur
      */
     mutate(): boolean {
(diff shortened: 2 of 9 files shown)


src/scripts/farming/mutation/Mutation.ts
```diff
@@ -73,6 +73,14 @@ abstract class Mutation implements Saveable {
         return '';
     }
 
+    /**
+     * Handles getting the mutation chance
+     * @param idx The plot index
+     */
+    mutationChance(idx: number): number {
+        return this._mutationChance;
+    }
```
RedSparr0w, line 82: This doesn't seem to be doing anything different than the old variable?
`idk` is never used either.

HLXII: Adding it so it can be overriden in the GrowNearFlavorMutation class.

src/scripts/towns/Town.ts
```diff
@@ -911,9 +911,6 @@ const CanalaveCityShop = new Shop ([
 const PalParkShop = new Shop([
     ItemList['Razor_claw'],
     ItemList['Razor_fang'],
-    ItemList['Combee'],
-    ItemList['Burmy (plant)'],
-    ItemList['Cherubi'],
```
RedSparr0w, line 916: I don't think we should completely remove these from the shops yet.

src/scripts/farming/mutation/mutationTypes/GrowNearFlavorMutation.ts
```diff
@@ -48,4 +48,30 @@ class GrowNearFlavorMutation extends GrowNearMutation {
...
+    /**
+     * Handles getting the mutation chance.
+     * Will decrease the mutation chance if the mutatedBerry already exists around this one.
+     * @param idx The plot index
+     */
+    mutationChance(idx: number): number {
+        const plots = Plot.findNearPlots(idx);
+        if (!plots.length) {
+            return 0;
+        }
```
RedSparr0w, line 60: In the unlikely event this does happen, should it not just return the default mutation chance?
```suggestion
        if (!plots.length) {
            return super.mutationChance(idx);
        }
```

src/scripts/farming/mutation/mutationTypes/GrowNearFlavorMutation.ts
```diff
@@ -48,4 +48,30 @@ class GrowNearFlavorMutation extends GrowNearMutation {
...
+            if (!plot.isUnlocked) {
+                return;
+            }
+            if (plot.isEmpty()) {
+                return;
+            }
+            if (plot.berry === this.mutatedBerry) {
+                sameBerries += 1;
+            }
+        });
```
RedSparr0w: This can be simplified a bit:
```suggestion
        const sameBerries = Plot.findNearPlots(idx).reduce((acc, plotIndex) => {
            return App.game.farming.plotList[plotIndex].berry === this.mutatedBerry ? acc + 1 : acc;
        }, 1);
```
or
```suggestion
        const sameBerries = Plot.findNearPlots(idx).filter(plotIndex => {
            return App.game.farming.plotList[plotIndex].berry === this.mutatedBerry;
        }).length;
```

src/components/farmModal.html
```diff
@@ -195,13 +195,13 @@ <h4><u>Farm</u></h4>
...
                     <p>Otherwise after some time being fully grown, the Berry plant will wither away,
-                    dropping only a fraction of the Berries but may also be re-planted automatically.
+                    dropping half of the expected Berries but may also be re-planted automatically.
                     <br/><i>(Farm Points are only gained by manually harvesting Berries)</i><p>
 
                     <p>Additional plots can be unlocked by spending Berries.</p>
 
                     <h4><u>Mutations</u></h4>
-                    <p>Berries have the chance to mutate into other Berries based on certain conditions.
+                    <p>Berries have the chance to cause mutations and create new Berry species based on certain conditions.
```
RedSparr0w, line 204: This seems to also be a common source of confusion:
```suggestion
                    <p>Berries have the chance to cause mutations and create new Berry species in empty plots based on certain conditions.
```

HLXII: Only issue is that some mutations don't require empty plots. Not sure if we want that distinction in the help tab though.

RedSparr0w: Maybe in the hint text for those few we can specify that it will actually change the berry?

src/scripts/farming/mutation/mutationTypes/GrowNearFlavorMutation.ts
```diff
@@ -48,4 +48,30 @@ class GrowNearFlavorMutation extends GrowNearMutation {
...
```
RedSparr0w: This can be simplified a bit with a reduce or filter:
```suggestion
        const sameBerries = plots.reduce((acc, plotIndex) => {
            return App.game.farming.plotList[plotIndex].berry === this.mutatedBerry ? acc + 1 : acc;
        }, 1);
```
or
```suggestion
        const sameBerries = plots.filter(plotIndex => {
            return App.game.farming.plotList[plotIndex].berry === this.mutatedBerry;
        }).length;
```

src/components/farmModal.html
HLXII: I'm okay with changing it to "in empty plots". Hopefully that doesn't confuse the player later on with evolve mutations though.

src/scripts/farming/mutation/mutationTypes/GrowNearFlavorMutation.ts
RedSparr0w: 😓 Updated this comment because the old one was outdated, but you've already updated it lol, too quick

src/scripts/farming/mutation/mutationTypes/GrowNearFlavorMutation.ts
```diff
@@ -54,23 +54,9 @@ class GrowNearFlavorMutation extends GrowNearMutation {
...
-                return;
-            }
-            if (plot.berry === this.mutatedBerry) {
-                sameBerries += 1;
-            }
-        });
+        const sameBerries = Plot.findNearPlots(idx).filter(plotIndex => {
+            return App.game.farming.plotList[plotIndex].berry === this.mutatedBerry;
+        }).length;
         return super.mutationChance(idx) * Math.pow(10, -sameBerries);
```
RedSparr0w, line 60: Is it expected that this will basically destroy the chances of mutation for the same berry if there is 1 or more?

HLXII: Yea it'll kill the invasiveness. I added 10 just randomly. We could scale it down to 5 or something if you still want some reproduction occurring.

RedSparr0w: I feel like anywhere between `2` and `5` should cut down the odds enough that it doesn't become so invasive,

HLXII: I'll put it at 4 for now, hopefully that'll curb it still.

RedSparr0w (approved):
Changes look good, thanks for the PR!

PR #1080, opened 2020-11-17
Protein updates

RedSparr0w:
Nerf protein (5 per region, down from 10)
Is now used from within the item menu
Removed from the hatchery
Fixed category list showing in hatchery

### DIFF
--- a/src/components/breeding.html
+++ b/src/components/breeding.html
@@ -130,11 +130,6 @@ aria-labelledby="breedingModalLabel" aria-hidden="true">
                                         <a class="dropdown-item bg-dark py-1" href="#categoryModal" data-toggle="modal"><i>Edit</i></a>
                                     </div>
                                 </div>
-                                <div class="sidebar">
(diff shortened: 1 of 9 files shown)


KrushGames:
honestly this is probably fair. But 1 question: does this take away the effects of protein we have already used?

RedSparr0w:
> honestly this is probably fair. But 1 question: does this take away the effects of protein we have already used?

Yes, but as with most updates it refunds any overused Proteins.

RedSparr0w:
> honestly this is probably fair. But 1 question: does this take away the effects of protein we have already used?

Oh, nevermind, misread your comment.
Nope, it won't reduce the current attack gains you already have, but will reduce the total proteins used on Pokemon, so in the future they won't be gaining that same amount as before.

KrushGames:
actually you may have misread it, but it answered what i wanted to ask perfectly. I phrased the question wrong too

PR #1082, opened 2020-11-17
Berry Fixes Round 6

HLXII:
I just want to pad my total commits in the contributors page :)

 - Change Chilan Mutation so Chople doesn't spontaneously change. However the new mutation is also equally annoying :)
 - Added time display setting for Farm. Can now choose between time between stages, or time from ripe/death.


src/scripts/farming/mutation/mutationTypes/EvolveNearBerryMinMutation.ts
```diff
@@ -0,0 +1,55 @@
...
+            const plot = App.game.farming.plotList[idx];
+            if (!plot.isUnlocked) {
+                return;
+            }
+            if (plot.isEmpty()) {
+                return;
+            }
+            if (plot.stage() !== PlotStage.Berry) {
+                return;
+            }
```
RedSparr0w, line 32: Do we need all the extra checks in here?
If it's not unlocked or is empty, then this should also be true:
```suggestion
            if (plot.stage() !== PlotStage.Berry) {
                return;
            }
```

src/scripts/farming/Plot.ts
```diff
@@ -55,6 +56,20 @@ class Plot implements Saveable {
...
+        this.formattedTimeLeft = ko.pureComputed(() => {
+            if (this.berry === BerryType.None) {
+                return '';
+            }
+            let timeLeft = 0;
+            if (this.age < this.berryData.growthTime[3]) {
+                timeLeft = Math.ceil(this.berryData.growthTime[3] - this.age);
+            } else {
+                timeLeft = Math.ceil(this.berryData.growthTime[4] - this.age);
+            }
```
RedSparr0w, line 68: Are we able to do anything about the magic numbers here (`3`, `4`)?
```suggestion
            if (this.age < this.berryData.growthTime[3]) {
                timeLeft = Math.ceil(this.berryData.growthTime[3] - this.age);
            } else {
                timeLeft = Math.ceil(this.berryData.growthTime[4] - this.age);
            }
```

src/scripts/farming/mutation/mutationTypes/EvolveNearBerryMinMutation.ts
HLXII: I have a lot of sanity checks all over the place. Feel free to remove them.

src/scripts/farming/Plot.ts
HLXII: Technically we could use the PlotStage enum, but those are named Bloom/Berry for 3/4, which doesn't really make sense in this context (4 being time to death).

RedSparr0w: Only way I can think of to use PlotStage _(without it becoming confusing)_, is if we added 0 to the `growthTime` in the class:
```js
        this.growthTime = [0, ...this.growthTime];
```

HLXII: I don't want to think right now, so you can refactor it if you want. I've been using magic numbers for this in other locations as well.

Maybe add a getter for ripeTime/deathTime since I only reference 3 and 4 directly in those locations?

RedSparr0w: Yeah, no worries, we can get it done another time, just mentioning it now so it doesn't become too prevalent.

RedSparr0w (approved):

PR #1083, opened 2020-11-18
Add new pokeball trades to the Johto berry master

RedSparr0w:
Add new Pokeball trades to the Johto berry master.
Random trade each day.

image

The method of obtaining them can be changed later on once we have Apricorns or however we implement them.


HLXII:
I think the prices are a bit too high right now, especially since we buy Pokeballs in 1000s in the shop currently, but we can lower it after I guess.

Now that we're actually adding the new Pokeballs in, do we need to update the Pokeball selection logic? Or will it just be Special Ball > Ultraball > etc.

RedSparr0w:
Do you have some suggestions for prices?
I forgot to actually change the amounts 😅

The selection code has already been updated in a previous PR, basically any special ball will default to Pokeball once it runs out, otherwise the selection logic remains the same as before.

HLXII:
Cutting in half might be good.

Do we want any more complex logic for the player to choose? Or I guess the current new pokeballs will probably be used for grinding, so they'd want to default to pokeballs after anyways.

RedSparr0w:
> Do we want any more complex logic for the player to choose? Or I guess the current new pokeballs will probably be used for grinding, so they'd want to default to pokeballs after anyways.

I'd rather not make it too complex, I feel like defaulting to standard Pokeball is a good middle ground for it.
Although If anyone feels like making it more complex, then not a problem lol.

PR #1087, opened 2020-11-18
Hotfix to start at false instead of true

LuqDragon:


src/scripts/pokedex/PokedexHelper.ts
```diff
@@ -1,7 +1,7 @@
 import TypeColor = GameConstants.TypeColor;
 
 class PokedexHelper {
-    public static toggleStatisticShiny = ko.observable(true);
+    public static toggleStatisticShiny = ko.observable(false);
```
RedSparr0w, line 4: Wouldn't this make it show the non-shiny image on the individual Pokémon modal?

src/components/pokedex.html
```diff
@@ -88,7 +88,7 @@ <h4 style="margin-bottom: 0">Pokédex</h4>
                        <div class="col-sm-2">
                         <label class="form-check-label">
                             <input class="form-check-input" type="checkbox" id="pokedex-filter-held-item"
-                                   value="false" checked onclick="PokedexHelper.toggleAllShiny(!PokedexHelper.toggleAllShiny())"> Hide shiny image
+                                   value="true" onclick="PokedexHelper.toggleAllShiny(!PokedexHelper.toggleAllShiny())"> Hide shiny image
```
RedSparr0w, line 91: I've also just pushed this, as I forgot to remove the `checked` attribute when I flipped the setting name.

src/scripts/pokedex/PokedexHelper.ts
LuqDragon: Oh god, my head is exploding this days, sorry! I will push the right tested things

src/components/pokedex.html
LuqDragon: I removed the checked and keep `false`, this `value` isnt used and for me represent if checkbox is checked or not

RedSparr0w (approved):

PR #1090, opened 2020-11-18
Fix to convertion of seconds to time

LuqDragon:
This method was with `Math.ceil` instead of `Math.floor` so always it round up the number, so 1,3 turns 2 instead of 1.


RedSparr0w (approved):
Oh no, Thanks for the PR!

PR #1091, opened 2020-11-18
Fix protein in older versions

LuqDragon:
Older versions, especially 0.5.8, dont have `Protein` in player data so just a fix to avoid fix all the saves that comes in `#bugs`

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -276,9 +276,6 @@ class Update implements Saveable {
         },
 
         '0.6.5': ({ playerData, saveData }) => {
-            // check if playeData have Protein
-            playerData._itemList.Protein ? null : playerData._itemList['Protein'] = ko.observable(0);
-
             // nerf amount of proteins used per Pokemon
             const maxProteins = (playerData.highestRegion + 1) * 5;
             let proteinsToRefund = 0;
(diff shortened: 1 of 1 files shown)


RedSparr0w:
`Protein` is a part of older saves, as the item has existed the entire time, for example it still exist in this `v0.4.4` save:
[[v0.4.4] Poke Clicker Backup Save (1).txt](https://github.com/pokeclicker/pokeclicker/files/5562016/v0.4.4.Poke.Clicker.Backup.Save.1.txt)

And just a side not, we can't use observables in the update functions as we are working with the raw JSON.

LuqDragon:
Sorry for the observable in update but what happened so? The two saves you mentioned is giving this error:
image

RedSparr0w:
Ah, it's because of these lines:

https://github.com/pokeclicker/pokeclicker/blob/416bb10f8473e2d6b96bf4302bb3698ad040ca70/src/scripts/Update.ts#L284-L289

`proteinsToRefund` becomes NaN.

We should probably be extending the item observables to be `numeric` as well so it would avoid this possibly happening in the future too, as it would just become `0` if it was a non number.

LuqDragon:
ok, javascript bug me sometimes, i come from C# very typed, everything is warned, javascript/typescript is like, do what you want and you will see the consequences after

LuqDragon:
Updated the branch if want to use

RedSparr0w (approved):
Changes look good, thanks for the PR!

PR #1100, opened 2020-11-19
Fix near berry strict mutation

HLXII:
Added shallowEqual


RedSparr0w:
This will not work if the keys are in a different order though:
```js
a = {
  'test2': 2,
  'test1': 1,
}

b = {
  'test1': 1,
  'test2': 2,
}

console.log(JSON.stringify(a)) // {"test2":2,"test1":1}
console.log(JSON.stringify(b)) // {"test1":1,"test2":2}
JSON.stringify(a) === JSON.stringify(b) // false
```

HLXII:
That doesn't seem to be the case for numbers?

image

RedSparr0w:
Numbers seem to vary from browser to browser, I would rather use a deep/shallow equal than stringify, just so we can be certain it's comparing them correctly.

RedSparr0w:
And then we can also use the same function in the future if needed anywhere else.

HLXII:
Added

RedSparr0w (approved):
👍 Changes look good to me, Thanks for the PR!

RedSparr0w:
Whoops, I forgot to merge this 😅

PR #1104, opened 2020-11-20
Add two more Oak slots.

silasary:
With the addition of the two new Berry related oak items (to a total of 10), a max of 3 active feels too low.
This adds a fourth at 300 (mid-Hoenn) and a fifth at 450 (mid-Sinnoh)

### DIFF
--- a/src/scripts/App.ts
+++ b/src/scripts/App.ts
@@ -25,7 +25,7 @@ class App {
                 new Wallet(),
                 new KeyItems(),
                 new BadgeCase(),
-                new OakItems([20, 50, 100]),
+                new OakItems([20, 50, 100, 300, 450]),
                 new PokemonCategories(),
                 new Party(),
                 new Shards(),


RedSparr0w:
Sorry, I don't think we should just add 2 new slots because we have 2 new items.
Allowing too many to be active at a time removes much reason to ever change between them once you have maxed them for the achievement.

silasary:
I chose two because the current limit of three was implemented when there were only six options, and I felt that the original intention was that you could have half active at once.  

I do think we need at least one more (Even if it's granted at 100% Sinnoh), as three feels very cramped, and some of the more niche oak items will never be used under the current system.

My personal loadout is *always* Shiny Charm and Magic Ball, then either Cell Battery or Sprinklotad, depending on whether I'm idle.  I cannot think of a reason (other than quests telling me to) that I'd ever switch away from them, as the rest are simply not worth removing these three)

PR #1105, opened 2020-11-21
Weather

HLXII:
- Weather now changes every 4 hours.
 - Weather can be seen in the top right corner of the Town Map
image
 - Weather has type attack multipliers associated with each weather condition
 - Castform forms will now appear based on weather
   * Currently normal Castform only appears in Clear weather. This can be changed since Clear weather has 30% chance of showing up right now (distribution also up for change).
 - Cherrim (sunshine) evolution changed to depend on Sunny weather (at level 1 or greater)
 - Goodra evolution now depends on Rain or Fog weather (as well as level 50)

Some things to think about:
 - How much effect will weather have on attack? (I have it around .9~1.2 multiplier at the moment).
 - What other Pokemon can be modified to be dependent on weather? I haven't played the canon games in years, so I don't have that much knowledge anymore. IIRC there's a deer pokemon for the seasons in some generation?
 - This mechanic is quite narrow right now. Would be nice to expand it a bit further using Party Slots, PokeLab, or some other additional content.


src/scripts/party/Party.ts
```diff
@@ -117,7 +117,7 @@ class Party implements Feature {
         return Math.round(attack);
     }
 
-    public calculateOnePokemonAttack(pokemon: PartyPokemon, type1: PokemonType = PokemonType.None, type2: PokemonType = PokemonType.None, region: GameConstants.Region = player.region, ignoreRegionMultiplier = false, includeBreeding = false, useBaseAttack = false): number {
+    public calculateOnePokemonAttack(pokemon: PartyPokemon, type1: PokemonType = PokemonType.None, type2: PokemonType = PokemonType.None, region: GameConstants.Region = player.region, ignoreRegionMultiplier = false, includeBreeding = false, useBaseAttack = false, includeWeather = false): number {
```
RedSparr0w, line 120: I may be incorrect, but I think by default this should be assuming a typical route attack situation:
```suggestion
    public calculateOnePokemonAttack(pokemon: PartyPokemon, type1: PokemonType = PokemonType.None, type2: PokemonType = PokemonType.None, region: GameConstants.Region = player.region, ignoreRegionMultiplier = false, includeBreeding = false, useBaseAttack = false, includeWeather = true): number {
```

src/scripts/party/Party.ts
```diff
@@ -104,10 +104,10 @@ class Party implements Feature {
      * @param type2 types of the enemy we're calculating damage against.
      * @returns {number} damage to be done.
      */
-    public calculatePokemonAttack(type1: PokemonType = PokemonType.None, type2: PokemonType = PokemonType.None, ignoreRegionMultiplier = false, region: GameConstants.Region = player.region, includeBreeding = false, useBaseAttack = false): number {
+    public calculatePokemonAttack(type1: PokemonType = PokemonType.None, type2: PokemonType = PokemonType.None, ignoreRegionMultiplier = false, region: GameConstants.Region = player.region, includeBreeding = false, useBaseAttack = false, includeWeather = false): number {
```
RedSparr0w, line 107: I may be incorrect, but I think by default this should be assuming a typical route attack situation:
```suggestion
    public calculatePokemonAttack(type1: PokemonType = PokemonType.None, type2: PokemonType = PokemonType.None, ignoreRegionMultiplier = false, region: GameConstants.Region = player.region, includeBreeding = false, useBaseAttack = false, includeWeather = true): number {
```

src/scripts/party/evolutions/WeatherEvolution.ts
```diff
@@ -0,0 +1,41 @@
+class WeatherEvolution extends Evolution {
```
RedSparr0w, line 1: I think we should keep the naming similar to other Evolution restrictions, and also include a level for this as it could just be set to 1 by default where it's not needed.
```suggestion
class WeatherRestrictedLevelEvolution extends Evolution {
```
Only other weather evolution I could find
image

src/scripts/weather/WeatherType.ts
```diff
@@ -0,0 +1,14 @@
...
+    Rain,
+    Thunderstorm,
+    Snow,
+    Hail,
+    Blizzard,
+    Sunlight,
+    Sandstorm,
+    Fog,
+    Shadow,
+    Winds,
```
RedSparr0w, line 13: I think we can safely remove `Shadow` as it's only in Pokémon XD,
And maybe change the following:
`Cloudy` → `Overcast`?
`Sunlight` → `Sunny`?
`Winds` → `Windy`?

src/scripts/weather/Weather.ts
```diff
@@ -0,0 +1,91 @@
...
+        [WeatherType.Thunderstorm]:     0.5,
+        [WeatherType.Snow]:             0.54,
+        [WeatherType.Hail]:             0.57,
+        [WeatherType.Blizzard]:         0.6,
+        [WeatherType.Sunlight]:         0.7,
+        [WeatherType.Sandstorm]:        0.8,
+        [WeatherType.Fog]:              0.9,
+        [WeatherType.Shadow]:           1.0,
+        [WeatherType.Winds]:            1.0,
+    };
```
RedSparr0w, line 61: We could likely just tie this in with the above `weatherConditions` instead of being separate?

HLXII: Did we want separate distributions between regions? or just a global distribution?

RedSparr0w: We could also just use weights, rather than having to manually set the percentages ourselves, that way if we add more weather in the future _(unlikely)_ it will just automatically tie in.

RedSparr0w: 🤔 Not too sure tbh, I'll leave that up to you to decide, I don't think it really matters too much, especially in the early stages.

RedSparr0w: Although it would be good to have different weather in each region.

You could just do something like:
```js
for (i = 0; i <= player.region; i++) rand = SeededRadn.next();
```

RedSparr0w:
Also not a big deal, but if possible could we find some `.svg` images to use instead of the `.png`?
[svgrepo](https://www.svgrepo.com/) has a ton available.

HLXII:
I got the images from [SwSh icons](https://bulbapedia.bulbagarden.net/wiki/Weather#Overview). Feels a bit more authentic than some random weather icons.

RedSparr0w:
> I got the images from [SwSh icons](https://bulbapedia.bulbagarden.net/wiki/Weather#Overview). Feels a bit more authentic than some random weather icons.

Yeah that makes sense 👍

src/scripts/pokemons/PokemonList.ts
```diff
@@ -8087,7 +8087,7 @@ const pokemonList = createPokemonArray(
         'levelType': LevelType.mediumfast,
         'exp': 158,
         'catchRate': 75,
-        'evolutions': [new WeatherEvolution('Cherrim', 'Cherrim (sunshine)', WeatherType.Sunny)],
+        'evolutions': [new WeatherRestrictedLevelEvolution('Cherrim (overcast)', 'Cherrim (sunshine)', 1, [WeatherType.Sunny])],
```
RedSparr0w, line 8090: Would it also be worth adding `Cherrim (sunshine)` as a possible weather restricted level 25 evolution for `Cherubi`?
And `Cherrim (overcast)` as a possible evolution for `Cherrim (sunshine)`?

RedSparr0w: Oh wait, nvm, I just noticed that `Cherrim (overcast)` is not weather dependent, it's just a level 25 evolution. 😅

src/scripts/quests/QuestHelper.ts
```diff
@@ -117,7 +117,7 @@ class QuestHelper {
         const routes = Routes.getRoutesByRegion(region).map(r => r.number);
         const first = Math.min(...routes);
         const last = Math.max(...routes);
-        const attack = Math.max(1, App.game.party.calculatePokemonAttack(PokemonType.None, PokemonType.None, false, region, true));
+        const attack = Math.max(1, App.game.party.calculatePokemonAttack(PokemonType.None, PokemonType.None, false, region, true, undefined, false));
```
RedSparr0w, line 120: I feel like we should probably just explicitly set this as false.
```suggestion
        const attack = Math.max(1, App.game.party.calculatePokemonAttack(PokemonType.None, PokemonType.None, false, region, true, false, false));
```

src/scripts/underground/SeededRand.ts
```diff
@@ -10,7 +10,11 @@ class SeededRand {
     }
 
     public static seedWithDate(d: Date) {
-        this.state = Number((d.getFullYear() - 1900) * d.getDate() + 1000 * d.getMonth() + 100000 * d.getDate());
+        this.state = this.getDateSeed(d);
+    }
+
+    public static getDateSeed(d: Date): number {
+        return  Number((d.getFullYear() - 1900) * d.getDate() + 1000 * d.getMonth() + 100000 * d.getDate());
```
RedSparr0w, line 17: Could we maybe change this to be `getSeedFromDate`
```suggestion
        this.state = this.getDateSeed(d);
    }

    public static getSeedFromDate(d: Date): number {
        return Number((d.getFullYear() - 1900) * d.getDate() + 1000 * d.getMonth() + 100000 * d.getDate());
```

Or just add a new method something like `seedWithDateHour`:
```suggestion
        this.state = Number((d.getFullYear() - 1900) * d.getDate() + 1000 * d.getMonth() + 100000 * d.getDate());
    }

    // hours specifies how many hours this seed should remain the same
    public static seedWithDateHour(d: Date, hours = 1): number {
        this.seedWithDate(d);
        const newHours = new Date(d - d % (GameConstants.HOUR * hours)).getHours();
        this.state += 1000000 * newHours;
```

RedSparr0w (approved):

PR #1107, opened 2020-11-21
Berry shop rebalance

HLXII:
- Decrease item cost multiplier by berry exp.
 - Lower shovel cost from 400 to 300 Farm Points

### DIFF
--- a/src/scripts/Player.ts
+++ b/src/scripts/Player.ts
@@ -160,10 +160,10 @@ class Player {
         this._itemList[itemName](this._itemList[itemName]() - amount);
     }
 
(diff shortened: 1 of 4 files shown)


src/scripts/Player.ts
```diff
@@ -160,10 +160,10 @@ class Player {
         this._itemList[itemName](this._itemList[itemName]() - amount);
     }
 
-    public lowerItemMultipliers(multiplierDecreaser: MultiplierDecreaser) {
+    public lowerItemMultipliers(multiplierDecreaser: MultiplierDecreaser, n = 1) {
```
RedSparr0w, line 163: Could we make `n` more descriptive of what this variable represents?

RedSparr0w (approved):

PR #1108, opened 2020-11-22
Fix Yaches

HLXII:
We were filtering out any nearby berries that weren't at the Berry stage. The EvolveMutation also checks for evolutions for a Berry at Taller stage or higher. This means having a whole field of Passho Berries at Taller stage would allow for the mutation to occur, as no "neighbors" were found.

I've added a neighborStageReq property (set to Seed for Yache), so that all nearby berries are captured.


RedSparr0w (approved):
👍 thanks for the PR!

PR #1110, opened 2020-11-23
Berry changes 7

HLXII:
- Fixing auras on reloading page and shoveling
 - Change Wacan mutation so Pinaps don't mess them up.


RedSparr0w (approved):

PR #1119, opened 2020-12-02
Update PokemonItem.ts

Ultima1990:
This is... uhm... a bug fix. Yeah. Let's go with that. This fixes a bug.

Nothing sus. Don't look into it.


KrushGames:
You're joking right? This isn't a serious thing?

RedSparr0w (approved):
👍

PR #1121, opened 2020-12-03
Add new fossils to the underground

RedSparr0w:
Added the Unova fossils to the Underground,
They have different style images, but seems to be the only ones really available.
https://bulbapedia.bulbagarden.net/wiki/Fossil

I've adjusted the IDs of the items as well so everything can be grouped a little easier.


silasary (approved):

RedSparr0w:
Updated the images thanks to @HLXII

PR #1122, opened 2020-12-05
[WIP] Balance round 5 : Unova

anapple96:
This PR will (try) to balance PR and the entire game, as in the process of doing this, I noticed Johto and Hoenn was too easy, as they were balanced pre-dynamic reduction and Protein. Also bumped Sinnoh up a bit, also because of Protein


src/modules/GameConstants.ts
```diff
@@ -835,7 +835,7 @@ export const KalosDungeons = [
     'Glittering Cave',
     'Reflection Cave',
     // 'Tower of Mastery',
-    'Azure Bay', // Should really be a route
+    // 'Azure Bay', Should really be a route
```
RedSparr0w, line 838: We can just completely remove this, don't think it will ever be needed to bring it back, as it is now a route.

Will also need to remove it from the `dungeonList`
```suggestion
```

src/scripts/towns/Town.ts
```diff
@@ -1845,11 +1845,11 @@ TownList['Reflection Cave'] = new DungeonTown(
...
-TownList['Azure bay'] = new DungeonTown(
-    'Azure bay',
-    GameConstants.Region.kalos,
-    [new RouteKillRequirement(10, GameConstants.Region.kalos, 12)]
-);
+// TownList['Azure bay'] = new DungeonTown(
+//     'Azure bay',
+//     GameConstants.Region.kalos,
+//     [new RouteKillRequirement(10, GameConstants.Region.kalos, 12)]
+// );
```
RedSparr0w, line 1852: We can just completely remove this, don't think it will ever be needed to bring it back, as it is now a route.
```suggestion
```

RedSparr0w (approved):

PR #1124, opened 2020-12-05
Roamer delay

Ultima1990:
Delays the first appearance of Tornadus and Thundurus, locking them behind the Legend (7th) badge of the region. Also adds a hint to Humilau City.

The National Gymquirer is a real in-universe show you can catch on TV in the gen 5 games btw. Thought it'd be a nice reference. Though the in-universe show is about gym leaders, not gym challengers. Creative liberty.


RedSparr0w (approved):
👍 Looks good to me, thanks for the PR!

PR #1125, opened 2020-12-06
Burmy forms

Ultima1990:
Makes Burmy forms available.
Just level up Burmy (grass) in any cave for sand cloak, and in any building for trash cloak.

For the record: I did not count ruins as caves OR buildings. So no forms in Alph or Sky Pillar.
Also for the record: old does not = ruin, there's a difference. So yes, you can get trash in Old Chateau and Pokémon Mansion.
Isn't Mt Pyre a cave? Nope, it is not. It is a mountain, obviously, but it has zero interior cave areas. It does have a well maintained interior graveyard area just like Pokémon Tower, so I counted it as a building.

Does this mean it has to be manually changed any time any new cave or building dungeons are added? Yeah. I guess. I imagine we can stop at some point and assume people will have obtained these forms by the time Kalos roles around. Or not. IDK.

Will the bot create walls of text if you search for Burmy (sand)? Yeah, probably lol.


RedSparr0w:
I'm not sure this is the best way to go about it,
I think we need to add a new evolution method which is based on the dungeon/route environment type.
As otherwise this way we would need to add every new dungeon which uses these types 🤔

RedSparr0w (changes requested):
Could you update this PR to make use of the new evolution methods implemented in #1138.

Can be used like this:
```js
new EnvironmentRestrictedLevelEvolution('Cave', 'Burmy (plant)', 'Burmy (sand)', 1)
```

RedSparr0w (approved):

PR #1126, opened 2020-12-07
Added empty queue notification

unclebanks:
* Added notification for when queue is empty.


RedSparr0w:
There doesn't seem to be any sound files in this PR?

unclebanks:
Sorry about that Red, I did these changes on the same fork that I'm doing map changes on so I deleted/re-forked and forgot the sounds file. :/

src/scripts/breeding/Breeding.ts
```diff
@@ -266,6 +266,15 @@ class Breeding implements Feature {
             if (this.queueList().length) {
                 const nextEgg = this.createEgg(this.queueList.shift());
                 this.gainEgg(nextEgg);
+                if ((this.queueSlots() - this.queueList().length) === this.queueSlots()) {
```
RedSparr0w, line 269: Not sure as I haven't fully tested, but aren't we able to simplify this by just checking if the queue is now empty?
```suggestion
                if (!this.queueList().length) {
```

unclebanks: If you're saying so I am sure we can. I'll commit it now.

RedSparr0w (approved):

PR #1128, opened 2020-12-09
Remove parasite Berries from Enigma pool

HLXII:
Occa, Kebia, and Colbur berries won't show up in Engima mutations anymore.

This will probably affect player's current Enigma mutations, but 🤷


src/scripts/farming/mutation/mutationTypes/EnigmaMutation.ts
```diff
@@ -58,7 +58,22 @@ class EnigmaMutation extends GrowMutation {
...
+            if (berry == BerryType.Occa) {
+                return false;
+            }
+            if (berry == BerryType.Kebia) {
+                return false;
+            }
+            if (berry == BerryType.Colbur) {
+                return false;
+            }
+            return true;
```
RedSparr0w, line 75: Could we simplify this to something like this
```suggestion
            return ![BerryType.Occa, BerryType.Kebia, BerryType.Colbur].includes(berry);
```

src/scripts/farming/mutation/mutationTypes/EnigmaMutation.ts
```diff
@@ -58,7 +58,22 @@ class EnigmaMutation extends GrowMutation {
      */
     static getReqs(): BerryType[] {
         SeededRand.seed(+App.game.discord.ID());
-        const berryTypes = Farming.getGeneration(2).concat(Farming.getGeneration(3));
+        // Getting possible Berries
+        // Only Gen 3 and 4 Berries
+        let berryTypes = Farming.getGeneration(2).concat(Farming.getGeneration(3));
+        // Remove parasite Berries
```
RedSparr0w, line 64: Can we make this a little more descriptive of why these certain berries are being removed.

HLXII: Updated.
Though theoretically with my Babiri changes this isn't that necessary, but it would be arbitrarily harder for some players, so it's probably better to remove these from the pool.

RedSparr0w (approved):

PR #1129, opened 2020-12-09
Buff wander rate

HLXII:
By my calculations Starf with maxed out farm should have a 10% chance every 24 hours.


src/modules/GameConstants.ts
```diff
@@ -68,7 +68,7 @@ export const AMAZE_MULCH_GROWTH_MULTIPLIER = 1.25;
 export const AMAZE_MULCH_PRODUCE_MULTIPLIER = 1.5;
 export const AMAZE_MULCH_MUTATE_MULTIPLIER = 1.25;
 
-export const WANDER_RATE = 0.00001;
+export const WANDER_RATE = 0.0005;
```
RedSparr0w, line 71: This seems like quite a large increase, 50x larger than before?

HLXII: I balanced it with Starf in mind.

Originally with .00001 chance, there was a .2% chance of getting Starf every 24 hours. Buffed it up to 10%.

RedSparr0w: Could we not just boost the Starf mutation rate instead of the wondering Pokémon chance?

HLXII: The starf mutation always occurs on wandering shiny pokemon. There isn't a mutation rate for that. To increase the chances of obtaining a Starf plant, we can either buff the wander rate, or buff the shiny chances on the farm.

RedSparr0w (approved):

PR #1131, opened 2020-12-09
Updated breedingModal.html

unclebanks:
* Added dex number to pokemon in hatchery selector


RedSparr0w:
I'm not sure about this one, as some names are already quite large, and this just expands the problem.

I think this would be better as an option for the Hatchery "Display Value":
image

unclebanks:
I agree but multiple people specifically requested this so would it be ok to set those to be able to select multiples? Or lock it to just allow dex number with whatever other standard was picked?

unclebanks:
I cant find the file to edit that. :/ Can you give me the name of it?

RedSparr0w:
> I cant find the file to edit that. :/ Can you give me the name of it?

The main area is:
`src\scripts\settings\SortOptions.ts`

And you will also need to add it to:
`src\scripts\settings\Settings.ts`

RedSparr0w:
Wrong file, updated above 😅

PR #1132, opened 2020-12-09
Change some envirnments

Ultima1990:
Removed Shoal Cave from water area. It was getting two environment types, which might be why it defaulted to default grass area. It is now ice area only.
Moved Whirl Islands from ice area to gemcave area.
Moved Mistralton Cave from cave area to gemcave area.
Moved Cave of Being from cave area to gemcave area.
Moved Seafloor Cavern from water area to cave area.
Added Sealed Chamber to gemcave area.


RedSparr0w (approved):

PR #1136, opened 2020-12-10
Babiri buff

HLXII:
Babiri plants will not mutate anymore (thus stopping parasite mutations)
Babiri Mutation Aura buffed, so Fully grown will stop all mutations around it.
Because Babiri can't mutate, Enigma/Starf mutation will evolve from Haban instead.

Requires buff because:
1. Aura currently isn't beneficial at all, nor does it have an effect on the player's decisions.
2. Buff will make Petaya a bit easier.


RedSparr0w (approved):

PR #1137, opened 2020-12-11
Add Flabble forms and change some Kalos route mons

Ultima1990:
Adds forms for Flabébé and evolutions. Including AZ's eternal form, though it has no in game source. Might find a use for it later.
Removes some mons from special encounter types such as rock smash and rustling grass, while adding mons from things that are just grass with a different sprite (think flower patches instead of grass. Functionally just grass with slightly different encounter pools, nothing special about it aside from the sprite, no other mechanics or anything). Also changes some mons on various routes to make it more accurate to the games. Removes several super rod catches, added a couple old rods here and there. Also reordered several mons on several routes to be in dex no order.

No guarantee that every mon will be obtainable with this. Also will need to remove some later for eggs.


src/scripts/pokemons/PokemonList.ts
```diff
@@ -13133,13 +13205,102 @@ const pokemonList = createPokemonArray(
...
+            'attack': 45,
+            'specialAttack': 75,
+            'defense': 47,
+            'specialDefense': 98,
+            'speed': 52,
+        },
+    },
+    {
+        'id': 670.05,
+        'name': 'Flabble',
```
RedSparr0w, line 13298: Could this be?!.. the legendary flabble!

RedSparr0w (approved):

PR #1138, opened 2020-12-11
Feat/environment evo

Aegyo:
Adds a couple of new evolution methods so that #1125 Burmy evolutions can be a bit more general.

This pr doesn't actually use the new evolution methods anywhere, so you will have to create an evolution in the console for testing, eg
```js
test = new EnvironmentGymLevelEvolution('Ice', 'Charizard', 'Icezard', 1)
// and running this in various scenarios
test.isSatisfied()
```

Also groups the environments onto a single data structure, with some typings in a few places to restrict usage and enforce updating the EnvironmentCssClasses if a new Environment is added.

Moves the constructed evolution methods into a single file to make seeing what we have a little easier, and also means we don't have to update the references in PokemonList every time we make a new method. Renames some of those evolution files since they weren't an accurate description anymore after moving the methods out of them.

LocationRestriction is looking more and more unnecessary the longer I look at it, but I'll leave it for now.


RedSparr0w (approved):
👍 Awesome work, Thanks for the PR!

PR #1139, opened 2020-12-11
Resetting Aura Fix

HLXII:
Comment in Farming.ts explains most of it.

For some reason, the stage of the plant wasn't being updated in the same tick, so `resetAura()` would see the plant that triggered the update in its previous stage. This affects aura plants because if there's only one plant, and the aura Berry goes from Sprout to Taller, the `resetAura()` call will think the Berry is still in Sprout stage, and won't apply the aura correctly.

To fix this, we queue up an aura reset instead, so that the auras will be reset in the next tick, where the plot stages are updated correctly. This leads to a minor visual delay, however overall the player will still have the same amount of ticks with the aura applied.

There might be a cleaner solution if there's a way of updating the Stage KnockoutComputed variable, if that's actually the issue here.

### DIFF
--- a/src/scripts/farming/Farming.ts
+++ b/src/scripts/farming/Farming.ts
@@ -13,6 +13,13 @@ class Farming implements Feature {
     mutationCounter = 0;
     wanderCounter = 0;
 
+    // You may be wondering why this is necessary.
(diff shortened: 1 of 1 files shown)


RedSparr0w (approved):
👍

PR #1141, opened 2020-12-13
Add notification when berries are about to wither

Awec4:
Add notification when berries are about to wither
Quality of life improvement which has been mentioned a few times.


RedSparr0w (approved):

PR #1142, opened 2020-12-13
Save achievement filters

Awec4:
Save achievement filters so that they are preserved after a browser refresh


src/scripts/Save.ts
```diff
@@ -7,6 +7,7 @@ class Save {
         localStorage.setItem('player', json);
         localStorage.setItem('settings', Settings.save());
         localStorage.setItem('save', JSON.stringify(this.getSaveObject()));
+        localStorage.setItem('achievements', AchievementHandler.save());
```
RedSparr0w, line 10: I don't think we should add this as another localStorage object, would be better to keep it a part of the `settings` with the other filters.

Awec4: That's going to be a lot of extra work to refactor the Achievements filter to be stored in Settings, and I'm not sure I see what difference it will make?

RedSparr0w: It's a lot cleaner to have all the filters stored in the same way, here's the PR that was used for the hatchery filters:
https://github.com/pokeclicker/pokeclicker/pull/1050/files

Eventually these might also be stored in the save file or as a separate settings file that could be exported/imported, so easier keeping everything together where possible.

Awec4: Resolved

RedSparr0w (approved):

PR #1143, opened 2020-12-14
Added new vitamins

Awec4:
Adds HP Up, Iron, Calcium, Zinc, and Carbos as well as adding a use for Rare Candies.
Using the vitamins will increase the bonus damage from breeding proportional to the Pokemon's relative stat in the game e.g. Mewtwo will get way more bonus damage from Zincs than most due to it having one of the highest Sp. Atk stats in the game.

Probably needs some fine tuning, since with the current formula defensive Pokemon are way worse than offensive, and it might be quite imbalanced. And while the modals work, it's a little annoying to have to switch between modals for each vitamin type.
Vitamins are currently only available at Celadon City. This could be extended to all cities with the department store (Goldenrod, Lilycove, etc.) or alternatively moved to Lilycove to give the shop there a purpose and also region-lock the vitamins until Hoenn & Later

Also suggested as part of this is an "egg steps bonus" which slightly skews the attack bonus depending on the number of egg cycles the Pokemon has (reduced bonus for fast hatching eggs, increased bonus for slow hatching eggs).


RedSparr0w:
Are you able to provide some sort of examples on how this will change a Pokémons attack compared to previous values?

Awec4:
> Are you able to provide some sort of examples on how this will change a Pokémons attack compared to previous values?
> 
> As from a quick read of the code, it seems like it might be a little too OP.

I haven't put too much thought into the formula yet. I figured it would be better to start with seeing if anyone else liked the general idea of this

RedSparr0w:
I had a quick play around with this, seems to work pretty decent as is.

Will definitely need to double check how each one affects some Pokémon though as some didn't have any increase after using almost 10 of each type.

Will also need to make sure we are only calculating the `damage` value once when it should be updated, as currently it causes lag every time you defeat a Pokémon due to recalculating all the Pokémon's damage/base attack each time.

Awec4:
I've been playing using this for a while and I think the formula probably needs a big change. Rare candies are now pretty overpowered - they basically take protein's spot and make low egg cycles mons incredibly efficient - and the others just aren't very good.
I was thinking instead for Rare Candies, if I implemented my idea of having extra exp over lvl 100 very slowly contribute to a temporary (until next breed) bonus multiplier - i was thinking log(expOverLvl100) ~ 1% bonus damage multiplier, so even a 10% bonus requires a huge amount of exp - then Rare Candies could increase that bonus per level.

RedSparr0w:
Maybe the Rare Candies could increase the Maximum overlevelled level they can reach instead of increasing the effectiveness, which would limit how long someone can just AFK and have everything overlelled.

Possibly the other vitamins need to increase the base stats by 2 per each?
Or double the amount that can be used?

PR #1144, opened 2020-12-16
Make the shop 'Max' button able to work beyond 1 million items

Aegyo:
Uses a binary search to find the max amount of items
Seems to take around 53 iterations in most cases

Tried the make the binary search part of the code general so that it can be used in other parts of the game.

When you try testing if it is really giving the max, make sure you are not on a route collecting money  :sweat_smile:

### DIFF
--- a/src/modules/GameHelper.ts
+++ b/src/modules/GameHelper.ts
@@ -110,6 +110,16 @@ export default class GameHelper {
         return keys1.every((key) => object1[key] === object2[key]);
     }
 
+    // Find the largest integer between min and max that does not return true when passed to testTooHigh, using a binary search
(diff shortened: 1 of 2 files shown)


unclebanks:
What will happen if you're on a route collecting money?

PixLSteam:
> What will happen if you're on a route collecting money?

Well, you might be able to afford more than before 😅 The 53 iterations make sense to me, because `Math.log(Number.MAX_SAFE_INTEGER) / Math.log(2)` equals 53.

Aegyo:
The "max" number won't update with your new money total, and you will then be able to add 1 to the amount the button suggested and find you can buy more... making you think the code is broken (as I did when testing it myself)

RedSparr0w (approved):

PR #1145, opened 2020-12-17
Berry Update: Life and Death

HLXII:
Kasib wasn't annoying enough. Petaya was extremely difficult to obtain with no payoff.

I've removed the Wandering Pokemon Attract Debuff from Kasib, and replaced it with faster growth rate for Berry plants in the Berry stage. This way Kasib Berries will be very annoying as players will have less time to manually harvest their plants.

The Petaya Berry will now keep Berry plants alive indefinitely. It does not however affect itself or any other Petaya plant, so once the Petaya plant withers, the rest of the plants will start aging again. This brings some interesting end-game strategies, as you can keep Auras around for much longer. This also helps mutations (as plants won't die anymore), but it won't help out that much since Petaya is already an end game plant to obtain (maybe for people who didn't want to try for Lum yet).

I've also added empty Mutations so that the Kasib/Starf mutations have hints that appear.

### DIFF
--- a/src/scripts/farming/Farming.ts
+++ b/src/scripts/farming/Farming.ts
@@ -305,8 +305,8 @@ class Farming implements Feature {
             [0, 10, 20, 0, 0], BerryColor.Purple,
             [
                 'Considered to have a special power from the olden days, this Berry is sometimes dried and used as a good-luck charm.',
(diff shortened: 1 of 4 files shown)


RedSparr0w (approved):

PR #1147, opened 2020-12-19
Fix nested modals locking the screen if clicked too quickly

Aegyo:
I tried spam clicking on my laptop with this for a few minutes, but couldn't get it to lock up like I can on the live site. Might need testing with an autoclicker to be sure.


RedSparr0w (approved):
👍 Looks good to me, couldn't get it to lock up either, Thanks for the PR!

RedSparr0w (approved):
👍 Looks good to me, couldn't get it to lock up either, Thanks for the PR!

PR #1148, opened 2020-12-19
[Egg Cycles] Buff baby Pokemon, Nerf evolutions

RedSparr0w:
Gave baby Pokémon `0.8x` their adult form values.
Gave evolutions `1.5x` their pre-evolved form values.

This makes evolutions not as OP as before, also allows legendary Pokémon to catch up a little bit in terms of efficiency.

Examples below:
image
image
image


Aegyo (approved):
Looks good, though we may want to lower the egg cycles on Cosmog before releasing Alola... 10.8k steps on Solgaleo with this :fearful:

RedSparr0w:
Nice catch!

Made the max egg cycles be whatever the highest default value is (120 at the moment).

PR #1150, opened 2020-12-20
Convert weather to modules

Aegyo:
The first of hopefully many PRs to move small sections of the code over to modules, to bring us closer to the promised land Spenser has sold us on.

Weather conditions are the same on this branch, and nothing seems to be throwing errors in the console, in my limited testing.

Small addition to the rewriting of emitted type declarations in gulpfile.babel.js to fix an issue I was having with modules that export multiple things.


RedSparr0w (approved):
1 step closer to the promised lands 👍 
Awesome work, Thanks for the PR!

PR #1151, opened 2020-12-20
Convert Notifcations, Settings, DynamicBackground to modules

Aegyo:
Everything needed to move Notifications into modules, except for `player`, `pokemonMap` and `App` which I have convinced typescript to not complain about with
```
// @ts-ignore
// eslint-disable-next-line no-undef
```
above any line where they are used.
This means those things lose some type checking, but is only a temporary measure until we have those things available to modules. It isn't ideal, but I haven't found a better solution yet. I wouldn't want to have too much of this in the code.

Settings.ts has been split into a file containing just the class, and another file (`settings/index.ts`) which adds all of the settings and re-exports the class. This is to avoid a circular dependency with Sound.


RedSparr0w (approved):

RedSparr0w:
We could also set them as globals?
```
    "globals": {
        "player": "readonly",
        "App": "readonly",
		"pokemonMap": "readonly",
    }
```
or at the top of each file where they are used:
```
/* global player, App, pokemonMap */
```

Aegyo:
That stops eslint complaining, but typescript still complains with
```
Cannot find name 'player'.ts(2304)
```
Though we can do this in a similar vein to stop it complaining
```
declare const player: any
```
That could be in some other module file and exported for whoever wants it. I'm not sure if that is any better though. We could give more detailed type info, but it might have to be maintained manually, and I don't know if it would cause issues elsewhere if this isn't the "real" Player type.

RedSparr0w:
Ah okay, that makes sense then 👍

PR #1155, opened 2020-12-21
Move Changelog to Module

unclebanks:


RedSparr0w (approved):
👍 Looks good to me, thanks for the PR!

RedSparr0w:
Just checked a new game, it seems to break the intro.js stuff.

unclebanks:
Crap, I figured that one might've. Is there a way I can fix it or should I just revert it for now?

unclebanks:
removed the changes in regards to the information.ts. hopefully everything is fine now.

src/modules/changelog/ChangelogType.ts
```diff
@@ -48,7 +43,7 @@ class ChangelogUpdate extends Changelog {
  * MINOR - Will increment for each feature refactor or large changes to a feature
  * PATCH - Increment for small changes, bugfixes, UI changes.
  */
-const changelogItems = [
+export const changelogItems = [
```
RedSparr0w, line 46: Could we maybe move these to their own file?

Just seems a little odd to be importing them from the `ChangelogType` file.

unclebanks: When i did that i was getting the complaints of circular dependencies. I can try to find a different way to do so. maybe i did it wrong in the first place?

unclebanks: I tried changing it to have its own file or putting it in changelog but i just got complaints of not finding Change in Changelogtype and stuff. I can't figure out a better way to do this. I can close the PR if you'd prefer as I can't really change (or fix?) it.

PR #1157, opened 2020-12-22
Fix crash in Settings when there is no save data

Aegyo:


RedSparr0w (approved):

PR #1158, opened 2020-12-22
Remove Babiri from Enigma Pool

HLXII:
Forgot to update this as Babiri will stop all mutations near it.


RedSparr0w (approved):
👍 Thanks for the PR!

PR #1163, opened 2020-12-23
P2 Buff

anapple96:
Realized P2 was extremely weak post-push, added a bunch of tough Pokemons and added Arceus (steel) as a reward (I feel like steel is appropriate here, as it's a lab)


anapple96:
Now that I think about it, Arceus-Steel may not be a great idea

Ultima1990:
I'm not a big fan of putting Arceus in there either. Also bloody hell that's a big boss pool.

anapple96:
Should be OK for review now

RedSparr0w (approved):
Looks good, Thanks for the PR

PR #1168, opened 2020-12-25
Fix parasite mutations

HLXII:
We were using EvolveNearBerryMutation with the originalBerry parameter undefined so all Berry plants nearby had the chance to mutate (and thus it becomes a parasite mutation). However, we didn't check if the nearby plants were the same Berry species, so the parasite mutation could occur on the same Berry type (e.g. Kebia Berries could mutate into Kebia Berries). This caused an infinite loop of mutations, as Evolve Mutations resets the Berry age.

To fix this, I created the ParasiteMutation subclass for EvolveNearBerryMutation, that overrides getMutationPlots to exclude the parasite mutation Berry itself from the mutation pool. This should stop the infinite mutations.

### DIFF
--- a/src/scripts/farming/Farming.ts
+++ b/src/scripts/farming/Farming.ts
@@ -630,8 +630,8 @@ class Farming implements Feature {
                 BerryType.Tamato,
                 BerryType.Spelon,
             ]));
-        // Occa Overgrow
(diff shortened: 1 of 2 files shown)


RedSparr0w (approved):

PR #1170, opened 2020-12-26
chore(refactor): Multipliers

Aegyo:
Adds a framework for registering and querying a multiplier for some value.

This is mainly to disentangle a few of the features for easier migration to modules, but it also has the benefit of making new multiplier sources a little cleaner (eg, you don't have to update Party :: calculatePokemonAttack when you add some new way of multiplying pokemon attack)

When getting the multiplier for something, you can pass a flag `useBonus`, defaulting to false, which I added so that we can gain oak item exp when calculating the bonus, without having to reference oakitems wherever we are when asking for the multiplier bonus. It only really makes sense for oak items so far though, so I'm not entirely convinced this is a good way to be doing things.

This also includes a fix for the double shiny charm application to egg hatches, and also applies the same "you only get full egg shiny odds by having the bonus on all the time" idea to all sources (farm aura has been exempt so far).

There are probably other things that we can plug into this Multiplier system, but I have just added a few of the obvious ones here.

The addition to .gitignore is to exclude crash reports from the ts compiler. Accidentally committed them one too many times...


RedSparr0w (approved):
Thanks for the PR, looks good to me,
I haven't had a chance to fully check everything out but don't have much time over the next few of weeks.
Feel free to self-merge it though assuming you've tested everything's working as expected 👍

RedSparr0w (approved):
Never mind, this was a lot easier to review than I initially thought!

Thanks for the PR! 🎉

PR #1173, opened 2020-12-26
Fix quests changing when going to a new region.

MrGrote:
When going to a new region, if you reload the game it will regenerate the quest list using the highest region.
This pr adds the region for which a quest list was generate to the save file and uses it to recreate the quest list as it should be.
This means you'll no longer see route kill quests with negative number of kills when refreshing after advancing a region.


unclebanks:
something has to be wrong with travis and this pr. it should not take this long. not for anything.

silasary:
I'm assuming Travis has run out of build minutes for this month

RedSparr0w (approved):
Looks good to me, thanks for the PR 🎉

PR #1177, opened 2020-12-27
Switch CI to Github Actions

silasary:
Travis [recently ](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing)dropped support for free plans, and now has a limit to how many minutes per month it'll run for.

This PR adds a Github Actions equivalent we can use instead.


RedSparr0w:
@Aegyo @SpenserJ Do either of you have any opinion on this?
Haven't really used GitHub actions, so not sure what I need to be looking for 😅

silasary:
This PR just adds the CI runs.  It doesn't remove travis (So all builds will be duplicated across both platforms for the time being).

You can see what it looks like on my branch here: https://github.com/silasary/pokeclicker/runs/1613588947

RedSparr0w (approved):

PR #1181, opened 2021-01-01
Fix Poison Barb gaining exp

Aegyo:


RedSparr0w (approved):

PR #1184, opened 2021-01-02
Fix Map Tooltip not appearing

### LINKED ISSUE #995: [BUG] Moving to a new region does not show initially show names for towns and dungeons
**Info**
 - OS: macOS Catalina
 - Browser: Chrome
 - Version: 86

**Description:**
When taking the Dock between Kanto and Johto, hovering over a town or dungeon does not show the name in the upper left corner of the Town Map until you move to a new location. At that point it continues to work as expected.

**Steps to Reproduce:**
1. Start in Johto or Kanto
2. Take the Dock to the other region
3. Hover over any town or dungeon
4. No name will appear in the upper left corner
5. Navigate to any route, town or dungeon
6. Names will start appearing in the upper left corner again when you hover over a town or dungeon

**Expected Behavior:**
I'd expect to see town and dungeon names immediately after moving to a new region.

**Screenshots:**
Video of the behavior in an incognito window:

https://gofile.io/d/StOCZN

HLXII:
Closes #995 

The Map Tooltip is setup in `applyRouteBindings::GameController.ts` to display when hovering over towns on the map. This method was called on initialization, as well as in `moveToRoute` and `moveToTown`.

However, when we switch regions, `applyRouteBindings` was being called before the new region map was loaded in the DOM. This means the bindings were applied on the old map, and then the new map was loaded, thus the tooltip doesn't appear initially. When moving to a route/town in the region though the method was called, which is why the issue fixes itself then.

There were two options of fixing this:
1. Call `applyRouteBindings` after the new region map has loaded
2. Don't remove the region maps on switching regions

I didn't implement option 1 as I didn't know how to call a method after Knockout updates the DOM.

For option 2, this was done by changing the Knockout binding for the region maps from `if` to `visible`. the `if` binding removed the DOM elements entirely from the document, which is why the bindings were lost on moving to a new region. If we use `visible` instead, all the maps are loaded on initialization, and all bindings are applied correctly. Thus switching to a new region will only hide the previous and display the current region (with the bindings already added).

This solution fixes the bug (and also slightly increases performance on navigating to new routes/towns/regions), however it means that all the maps are loaded in the DOM and are never removed. This might be slightly less performant.

### DIFF
--- a/src/components/AlolaSVG.html
+++ b/src/components/AlolaSVG.html
@@ -1,4 +1,4 @@
-<g data-bind='if: player.region == GameConstants.Region.alola'>
+<g data-bind='visible: player.region == GameConstants.Region.alola'>
     <image width="100%" xlink:href="assets/images/background.png"></image>
 
     <!-- ------ -->
--- a/src/components/HoennSVG.html
+++ b/src/components/HoennSVG.html
@@ -1,4 +1,4 @@
-<g data-bind='if: player.region == GameConstants.Region.hoenn'>
+<g data-bind='visible: player.region == GameConstants.Region.hoenn'>
     <image width="100%" xlink:href="assets/images/hoenn.png"></image>
 
     <!-- ------ -->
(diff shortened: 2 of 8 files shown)


RedSparr0w (approved):

PR #1188, opened 2021-01-03
Berry Update 9

HLXII:
#### Fixed Aura Updates
It seems like we need to wait for more than one tick before calling `resetAuras`. I've modified the queue logic to do this after 2 ticks, which seems to be working now.

#### Nerf Kasib
The original modification to Kasib in #1145 was a bit too far, considering it made Petaya even harder to get. I didn't take this into account, so I'm nerfing the effects.

Kasib will now have a `Death` aura instead, which acts like the other normal auras in that it'll only affect nearby Berry plants. In addition, the `Death` aura will not be multiplicative, and will only take the max value available (e.g. a Berry plant near a blooming Kasib and a ripe Kasib will only have the effects of the ripe one). This makes the max `Death` aura only cause a growth rate increase of `2x`

#### Add External Auras Tooltip
There's currently no way of knowing what the full effects of external aura Berries are, unless you manually multiply all the auras in the farm. I've added a minor info icon in the modal footer that only appears when there are external auras present.

image

#### Fix Maranga Image
Maranga's Taller plant image was way too fast. I must have forgotten to slow it down when creating the image. I've updated the image to match the other plants now.

#### Add Unova Wandering Pokemon
Added Sewaddle and Karrablast to base wandering Pokemon. Added Unova Starters to Occa/Passho/Rindo wanders.


RedSparr0w (approved):
Looks good, thanks for the PR

PR #1189, opened 2021-01-03
Fix Underground Quest Old Amber reward

HLXII:
We updated the Underground Item IDs in 0.7.1, but didn't update the Quest to return the new item IDs.

We should really think about creating an enum or something similar to make sure this doesn't happen again...

For now, I replicated how we found `Mind Plates` for the Deoxys quest line to find the `Old Amber` item.


RedSparr0w (approved):

PR #1191, opened 2021-01-04
Quest Generation Refactor

HLXII:
The current implementation for the Quest system uses only the current time, as well as the player's highest available region to generate the seed. This has a couple of issues:

1. When moving onto a new region, reloading the game would change the quest generation seed (as the highest available region has changed). This caused some bugs with the available quests having negative values (see #1173 for more details). This was fixed in the aforementioned PR, however the base cause is due to how we generated the quests.
2. There have been complaints for Berry quests being generated where the player hasn't unlocked that Berry yet. This is due to the quest generation not looking at the unlock status of the berries. If we did look at the player's current progress there, the quests would be changed once the player unlocks new berries (as the quest generation would be dependent on the unlock status, meaning any change in the unlock status would change the resulting quest generation).

To fix this, we should instead store the current quest information into the save, so that we aren't dependent on the seed being the same every time we reload the game. This will make sure the current available quests don't suddenly change on the player.

In regards to the Berry quest situation (and less common the route kill quests as well), modifying the quest generation to check current player progress will be done in a later development, pending this change.

### Changes

I've refactored `QuestHelper.ts` to instead use a factory-esque generation (see[ this stackoverflow page](https://stackoverflow.com/questions/34655616/create-an-instance-of-a-class-in-es6-with-a-dynamic-name) for more details).  The `createQuest` method will now be used to both generate the quests on refresh, as well as reinitialize them on reloading the game.

The quest generation details were moved into the individual quest type class files, as it pertains to them more than the quest helper itself.

Due to the refactoring of the `Quest` class, the `QuestLine` framework also had to be updated. This was mostly just refactoring the quest initialization. I didn't touch the `QuestLine` framework other than that due to the fact that we don't need to store additional quest data for quest lines, as they're always regenerated the same way on reloading the game.

For the individual subclass for the quests, I stored the additional initialization parameters in the quest instance on construction. I also moved the `description` variable into a getter, as every description is either a custom description (see quest lines), or dependent on the initialization parameters.

To ensure we restore the quests correctly on reload, we must store all the initialization parameters in the JSON, as well as the type of quest. This was done by creating a new `data` and `name` property in the JSON. The `data` property will store all the necessary initialization parameters for the quest, while the `name` property stores the quest class name. These properties will be used in `createQuest` to regenerate the quests exactly how they were originally stored.

### Testing

Test saving quests in various states (complete, in progress, etc) and reloading.
Regression test #1173.
I modified how the descriptions were handled, so test that the descriptions aren't broken in any way.
Also test the quest lines weren't broken.

### Additional Note

Due to this refactoring, I added an update method to regenerate the quests. This will clear any quest progress the player currently has on updating. I couldn't find an easy way of storing the correct quest data with the new refactor, so hopefully this is okay.

To make sure no one complains (or spams the bugs channel in discord), maybe we should add another popup like in the previous update? In addition to that, maybe also add a redeemable code (and start adding redeemable codes to the update popup), so there's more incentive to actually read the popup this time.


RedSparr0w (approved):
Looks good, Thanks for the PR!

PR #1192, opened 2021-01-04
"Fill Queue" button

PixLSteam:
A "Fill Queue" button requested by many.
Screenshot: 
image
The button itself is currently wider than the "List" button, but this can easily be changed.
It uses `BreedingController.visible` to apply the selected filters. If I understand it correctly, the sorting automatically gets applied to `App.game.party.caughtPokemon` so it should also work as expected.
Feedback is welcome!


unclebanks:
Is there a reason you have put it on the left?

PixLSteam:
No particular reason, no. I'm open for suggestions, this was just the easiest place to put it for now.

unclebanks:
I can't speak for everyone but I personally would like it next to the other.

IanWalder:
I agree with Banks, but if you don't want the buttons to be together (they do look kinda similar) you could set the button to be as wide as the hatchery modal, between the visible eggs and the queue display (just a suggestion).
Side note, do we want any restrictions on this functionality? Should players be able to do it upon reaching Johto (i.e. getting their first queue slots)? I feel we may need to discuss this.

unclebanks:
Could always set the button color to something else too. Blue/white or something

Gavin-fp:
I agree with @IanWalder, a restriction will be good for this feature, and this is a really nice feature. I think either reaching Johto or beating the Johto Elite Four will be good options.

PixLSteam:
Something similar to Ian's suggestion:
image
Banks' suggestion:
image
Any favorites or other suggestions?
A restriction (i.e. Johto+) should be trivial to implement.

mog-kupo:
I would probably lean towards Ian's look personally.
Suggest not showing the button till Johto as there is no queue until that point. I would just do a check for max queue > 0 for showing the button.

unclebanks:
Tbh I like Ian's suggestion better too. Gives a nice defined border between the two windows

PixLSteam:
In that case I'll probably implement it and wait for a dev to take a look at this. I also managed to get rid of the background behind the button, feedback appreciated:
image

unclebanks:
Looks good to me. Send it in. Good job buddy.

Ultima1990:
Here I am to do what I do best.

I don't like this. The queue was never supposed to be full automation. It was for convenience. To allow you to line up a bunch of pokémon for breeding if you go afk or something. I get that people are just constantly using it now, not just for those afk situations, but it's still not supposed to be automation. This button brings it one step closer to automation. With the current queue size it's not so bad. This is a clicker after all. Get clickin'. 32 isn't that much.

I can hear you say. It's 32 NOW, imagine with a couple more regions! We are going to need this button! I think 32 is a good size for the queue. It could double up one more time to 64, but that should really be the final max size.
Additionally, IMO, no more free slots. It's about time we change it up for this feature, do what other idlers and clickers do: stop giving convenience features away for free, sell em! Like, for example, sell the last 32 slots in PWT if that's ever implemented.

I want to meet you halfway tho. So I propose this: batch fill. Queue up a couple Pokémon with one click. Not all tho. Like, 4, max.

Qwertypop4:
Trying to fill a queue can, and often does, cause lag. If this would stop that, even if it was only be allowing you to queue for per click instead of one

PR #1195, opened 2021-01-06
Tooltip additions

mog-kupo:
Noticed there were some tooltips missing on big numbers that could likely use them. Also, hatchery eggs/fossils cause.... knowledge is power?


RedSparr0w (approved):
Looks good, Thanks for the PR.

_I've removed the breeding display tooltip for now, as the typed eggs should remain hidden until hatched IMO_

mog-kupo:
I can add logic for that, I couldn't think of a good reason to leave typed eggs out... If people are gonna refresh to get the right egg, the pattern gives enough away.

PR #1197, opened 2021-01-08
Underground Images Cleanup

HLXII:
#### Remapping Image names
Currently we store all the Underground images by the ID number. This is a bit fragile, since we've already moved around the IDs a couple of times already. Who's to say it won't happen again. The names of the items are a lot more constant, so by renaming the images by the actual names, we won't need to touch them again for a while. It'll also make it easier to find the correct images, rather than having to remember which ID matches to which item.

#### Dream World Images
The Underground items had the same images regardless of whether they appeared in the Underground, or in other locations. This doesn't match the original games where the Underground items have different sprites compared to when they're in your inventory. I've added the Dream World images for all the items and updated the display locations so that the Underground images appear in the Underground, while the Dream World images appear everywhere else.

#### Resizing Underground objects
The Underground images were not fully pixelated, and seem to be stretched for some reason. I've changed all the images to be the perfect pixel versions. I've also replaced some of the Underground items with their actual Underground image counterparts (see all evolution stones). I've also updated the sizes of the items to match the original games. Not sure why they were different to begin with...

#### Shard Images
Now that we're using the Dream World images for the Plates, the current Shard images are basically pixel duplicates of those. I've replaced the shard images with the Dream World shard images. I've hue-shifted the images to match the correct colors.

#### Some final notes
Going through the Underground codebase made my skin crawl a bit. The way we store the Underground items is a bit awkward, leading to a lot of calls to `Underground.getMineItemById`, which loops through the entire list which is inefficient.

We're differentiating different types of underground items by the `itemType` parameter, and the `value` parameter also takes different meanings based on the `itemType`. It might be better to create subclasses to handle this differentiation instead.

The `imagePath` method I added is a bit ugly due to the fact that we store images all over the place currently. There's not much I can do about that though.


src/scripts/underground/UndergroundItem.ts
```diff
@@ -43,7 +43,20 @@ class UndergroundItem {
...
+            const evostone: EvolutionStone = (ItemList[this.valueType] as EvolutionStone);
+            return evostone.imagePath;
+        } else if (this.valueType == 'Mine Egg') {
+            return `assets/images/breeding/${this.name}.png`;
+        } else {
+            return `assets/images/items/${this.name}.png`;
+        }
+    }
+
+    get underGroundImage() {
```
RedSparr0w, line 58: Could we change this to be one word:
```suggestion
    get undergroundImage() {
```

RedSparr0w (changes requested):
Looks good, just a minor change

RedSparr0w (approved):

PR #1198, opened 2021-01-08
Fix Event description display

HLXII:
We were using the `text` bind for displaying the event description. This caused the line break to not be rendered. I changed it to the `html` binding.


RedSparr0w (approved):
👍 Thanks for the PR

PR #1206, opened 2021-01-12
Dungeon boss breakout

mog-kupo:
Quick edit for showing dungeon bosses with some separation.
Can likely be done differently, just wanted to give at least a small amount of separation.


RedSparr0w:
closed via #1434

PR #1209, opened 2021-01-13
Add Solaceon Ruins

Ultima1990:
Added Solaceon Ruins with 4 unown types (solacen - discor = laen), 2 of them are brand spanking new. Also moved Hippopotas to these ruins. Reasoning for that is Hippopotas is currently on 214 because that's where the Maniac Tunnel is in the main games. The tunnel does connect to some secret rooms in the Solaceon Ruins, so while the Maniac Tunnel is still not in the game (probably won't happen either), I think having Hippopotas in a dungeon makes sense, and having them in the Solaceon Ruins also makes sense.

Now who wants to do some map making magic? I need a cave just to the right of Solaceon Town.


RedSparr0w:
Before this is merged, `Update.ts` will also need to be modified as the index of any dungeons after this one will be different, so it will affect players statistics.

src/scripts/dungeons/Dungeon.ts
```diff
@@ -1022,6 +1022,18 @@ dungeonList['Mt. Coronet South'] = new Dungeon('Mt. Coronet South',
...
 
+dungeonList['Solaceon Ruins'] = new Dungeon('Solaceon Ruins',
+    ['Zubat', 'Geodude', 'Natu', 'Bronzor', 'Hippopotas'],
+    [GameConstants.BattleItemType.xAttack, GameConstants.BattleItemType.xClick, GameConstants.BattleItemType.Lucky_incense],
+    960000,
+    [
+        new DungeonBossPokemon('Unown (E)', 4100000, 30),
+        new DungeonBossPokemon('Unown (L)', 4100000, 30),
+        new DungeonBossPokemon('Unown (N)', 4100000, 30),
+        new DungeonBossPokemon('Unown (O)', 4100000, 30),
```
RedSparr0w, line 1033: Maybe we could replace the `O` with `U`?

DISCOR
ALPH
~~SOLAC~~E~~O~~N ~~R~~U~~I~~N~~S~~
```suggestion
        new DungeonBossPokemon('Unown (U)', 4100000, 30),
```

Ultima1990: Could do. But to be consistent, we should then also add it to Alph as well, along with N and a new F.

D I S C O R
~~R~~ U ~~I~~ N ~~S~~   ~~O~~ F   A L P H 
~~S O~~ L A ~~C~~ E ~~O~~ N   ~~R~~ U ~~I~~ N ~~S~~

RedSparr0w: Sounds good 👍

Ultima1990: Don't know why the O was there tbh. Twas supposed to be A E L N. That's my B.

PR #1210, opened 2021-01-13
Challenge settings

PixLSteam:
Adds a new tab to the settings modal, containing these settings:
 - Disable click attacks
 - Disable shard bonuses
 - Disable Masterballs
 - Disable Oak Items
 - Disable Battle Items

This will probably require some changes, at the very least some visual ones.
Most of the features were tested, but I'd appreciate any further testers.
Disabling Oak Items should deactivate any active ones, while disabling Masterballs should hide them in the modal and select Ultraballs instead.
Disabling Battle Items currently does NOT end any active effects.
Feedback is welcome!


RedSparr0w:
IMO the challenge modes shouldn't be togglable, they should either be enabled or ignored at the beginning of the game.

PixLSteam:
> IMO the challenge modes shouldn't be togglable, they should either be enabled or ignored at the beginning of the game.

These are just meant to allow people to enforce these challenges while there's no official challenge mode. I agree, an actual challenge mode would make this obsolete.

RedSparr0w:
Ah, I see what you mean, I would rather not implement something as a temporary measure., although it shouldn't be too far off how you have already set this up to move to something more official.

PR #1212, opened 2021-01-14
[WIP] Region Gate Overhaul

### LINKED ISSUE #1086: [WIP - FEATURE OVERHAUL] New Region Gate Balancing
**NOTE: This is _extremely_ tentative, and will require quite a lot of balancing/idea generation to fully implement. This issue is mainly for a location for throwing ideas into.**

# Abstract
The current incentives for region progress limit the design space for other ideas for capturing Pokemon. Since we made it so that all 151 pokemon are required to advance, there's a hard incentive to catch every single one of them. This hard incentive is a bit unfun to the player, as they don't really have any other option if they want to progress through the game. It also dilutes the incentives for catching pokemon. 
Catching any pokemon, regardless of species, will have the same priority since all of them are needed for progression anyways. Since the priority to the player is the same, having some pokemon being more difficult to catch than others is more annoying, since the gain/work ratio difference is so different (looking at legendaries/safari pokemon).

If we instead allow the player to advance without requiring every pokemon, that removes the hard incentive, which then opens up the design space to add soft incentives to catch harder pokemon.

# Design Considerations

## New Region Gates
With the removal of requiring catching every Pokemon in the region removed, this will alter how much attack the player will have to accumulate to advance to the next region, as they won't have to wait to catch the all the roaming Pokemon.

Some ideas for region gates:
 - Buffing E4/Champion to be an attack-gate before advancing to the next region. This is currently the least interesting one, as it will be an obvious attack-gate for the player to overcome.
 - Modifying E4/Champion to become a gauntlet (#1038). A little more interesting, however still an obvious attack-gate.
 - Adding End Region Trainers (#1026 also covers some of the ideas above)

Most of these are basically attack-gates, but that will at least allow for easy balancing, as we'll know what the minimum total attack the player will be at when they enter a new region, rather than estimating it based off of how long it takes to catch roaming Pokemon.

## Pokemon Locations
We currently add previous region's Pokemon into the route Pokemon pool in other regions, due to the fact that we assumed that the player as already caught every Pokemon (due to the current region requirement). This won't be the case anymore, which will cause some issues. Pokemon that are hard to catch in one region cannot be easy to catch in newer regions. This defeats the purpose of the difficulty in the earlier regions, which will be a disincentive for player to even attempt catching the Pokemon there.

If this idea is implemented, the Pokemon locations will need to be updated to ensure difficult Pokemon remain difficult. (e.g. Scyther still only exists in the Kanto Safari). We could also make the newer region encounters dependent on already catching the Pokemon as well, if we don't want to remove the Pokemon's additional locations entirely (e.g. Scyther exists in the Kanto Safari, and also in a Johto Route. However Scyther will only appear in Johto if it was already caught).

## Soft Incentives
With the removal of the requirement to catch all Pokemon in the region, some Pokemon will have no incentive to catch anymore (looking at Safari/Dungeon Bosses,Shop Pokemon). There is still a minor soft incentive due to adding to the total attack, however a player can just rely on breeding in those instances anyways.

While this is a negative in terms of current design, it can become a positive, as we can now buff Pokemon that are difficult to catch. Previously, since we needed to catch every Pokemon, there wasn't really a work/reward ratio for every Pokemon, as catching a Rattata had the same amount of incentive as the Scyther (both are needed to advance to the next region). 

In the new implementation, both aren't necessarily needed. Rattata can be caught easily for attack, however people can choose to catch Scyther or not. We can now buff Scyther in some way to give additional incentive to the player to attempt to catch it, rather than forcing them to catch it (which isn't fun).

This, however, means we need to implement some additional incentives to catch Pokemon. Here are a couple of ideas:
 - Add global buff for catching every Pokemon in a region. This will partially bring back the original incentive to catch every Pokemon. However now instead of being a hard incentive to advance through the game, it'll be a soft incentive to boost overall attack.
 - Modify Attack/Steps to reflect obtaining difficulty. Not too sure about this one, however more difficult Pokemon could be more efficient for breeding, which gives incentive to catch them.
 - Stronger Abilities (#814) would give incentive to catch those Pokemon to use their abilities.

## Catch Requirement Design
The original implementation had a limit to how difficult a Pokemon could be to catch. Specifically, the hardest Pokemon would be the Roamers of each region, as those took the most time. This limited the design space, as we couldn't make any Pokemon too hard to catch, otherwise they would be the bottleneck for that region.

The new implementation expands the design space for requirements, as if they're too difficult, they can be skipped until later when the player has more attack/experience/more willpower-to-catch-that-last-pokemon.

However, with additional difficulty, comes additional benefits. The reward will have to match the work, which is one of the harder parts of this implementation.

Anyways, this opens the design space for even harder challenges. One issue with the current design is that Legendaries don't feel, well, Legendary. They're usually bosses at the end of dungeons (and some dungeon bosses aren't even legendary). Legendaries should be separated from the riff-raff, and be much harder to catch. Here are a couple of ideas:
 - Victory Road Gauntlet: Have a mini BF-like stage in the Victory Road where the player must beat 100 floors to reach Moltres. The catch rate will be 100%, as it's more about beating the stages, rather than the catch chance.
 - Story Quests: Currently the only one implemented is for Deoxys. Additional ones could be added to unlock the Legendary in the Dungeon. This would give both additional QP resources, as well as additional tasks for the player to do besides for beating routes/gyms/E4 and getting more attack.
 - Weird Requirements: I had the idea that the player would have to get the Farm in a certain state before a Legendary was unlocked. E.g. Shaymin unlocked after getting a full field of blooming Lansat Berries or something. Maybe something like that for the Underground. Could be connected to additional mechanics (Wailord/Relicanth in party for regis or something).

HLXII:
Closes #1086 

## Remove Pokedex Requirement for Region Progression

I simply updated `MapHelper.ableToTravel` to check for Champion Badges, rather than checking Pokedex completion. This will need to be updated every time a new region is added, as there's no current static mapping between region and champion badges (we could theoretically use the region string like `BadgeEnums['Elite_'+<regionName>+'Champion']`, but that seems fragile to me (especially if we have other region names like Sevii Islands or something idk).

## Remove Prof. Oak NPC

He won't do anything anymore since region progress will be fairly straight forward to the player now, since it goes straight from Champion to new region.

## Re-Balance Region Progression

Because the original roadblock for most players were Roaming Pokemon, most players are stuck idling and gaining additional attack via breeding for quite some time. This leads to players steamrolling through new regions after catching the last roamer. Since roamers won't be a roadblock to progression anymore, it's up to the other roadblocks to slow the players' progression.

### Gyms/E4

 - [ ] - Update Gym/E4 Power Level

The Gyms/E4 were fairly underpowered, as most players finish new regions usually in less than a day. Since Gyms (and especially the E4/Champion) are now the main roadblocks for the player, they should be the benchmark for the players progress. This is especially true for the Champion, as they are the last battle the player needs to beat before advancing to the next region. This means we can directly control the minimum attack needed to advance to the next region.

### Dungeons

 - [ ] - Update Dungeon Power Level
 
 Dungeons will also be part of the roadblocks for the player, and they'll have to be buffed since they exist in between Gym runs.
 
 ### Routes

 - [ ] - Update Route Power Level
 
Routes probably need a buff in general anyways. They aren't hard roadblocks for the player, since their only requirement is defeating 10 Pokemon and there's no time limit. However keeping the power level slightly inline with the new Gym/E4/Dungeon power levels would probably be good.

## Update Exclusive Pokemon Locations

 - [ ] - Update Pokemon Locations

Due to removing the Pokedex requirement, some Pokemon which are more difficult to catch in their native region can be easily caught in newer regions. The main types can be seen below:

#### Safari Pokemon
These are kind of a wash, since I doubt we want to keep the Safari Zone grind in the game for much longer. Its probably better to leave these in the Safari and allow them to be caught later on. The Safari Zone should probably be reworked from the grind it currently is, into a new mechanic with other incentives than the exclusive Pokemon it offers.
* Exeggcute
* Rhyhorn
* Scyther (though they're in the farm as well)
* Pinsir
* Kangaskhan
* Tauros
#### Dungeon Bosses
Too many to count. I'm not sure if these are going to be an issue. Dungeon bosses are mostly RNG machines anyways, so the player can just not gain easy attack from catching them, or wait until they're more easily caught in the next region. If it's that important to have them, it's probably easier to scrub their existence from newer regions (though that opens a whole new can of worms to make sure that the Dungeon Boss exclusives aren't used in any new encounter pools).
#### Egg-Exclusive Pokemon
Note: None of the starters are in this group, as they're only obtainable via eggs (or from the farm). Some other Pokemon also remain Egg exclusive just due to them not existing in any other encounter pool. The [wiki page](https://pokeclicker.miraheze.org/wiki/Egg) was very helpful in this. 

For the remaining ones, there's not much we can do as they exist in newer Routes/Dungeons.
* Lapras
* Hitmonchan (obtained from leveling Tyrogue)
* Hitmonlee (obtained from leveling Tyrogue)
* Dratini
* Houndour
* Slugma
* Sunkern
* Roselia
* Feebas
* Electrike
* Bagon
* Carnivine
* Pachirisu
* Riolu
* Gible
* Throh
* Sawk
#### QP Pokemon
Every single one besides for Togepi and Zorua. Not sure what to do here either. Some of the purchases can be modified into their own story quest (looking at Spiritomb), but for the others not sure what should be done about them.

#### Scrubbing

We remove the exclusive Pokemon from appearing in any encounter pools in newer regions. This is a bit of a dangerous approach, as we'd need to ensure we never use those pokemon in any encounter pools in the future. It'll also cut down on the variety in the encounter pools.

#### Locking

We could remove the exclusive Pokemon from later encounter pools until the player catches the pokemon from their original location. This lets us keep the current encounter pools, however there are a couple of drawbacks:
1. This might be confusing to the player, as there won't be any indication for why a pokemon isn't appearing on a route (unless they know that they need to catch it from a previous region).
2. Some players may game the system by not catching a previous region's pokemon, which will affect the encounter pool in a later region (increasing the overall chances of finding a pokemon). This is especially problematic if there are still Dungeon Boss exclusives (and the other dungeon boss is locked since the player didn't catch it yet in a previous region).

#### Ignoring/Reshuffling

Even more radical. We ignore the issue semi-entirely. 

Safari will be (hopefully) reworked one day, so it's okay if the current exclusive Pokemon aren't exclusive anymore (and the new mechanic will hopefully have other incentives). 

Dungeon Bosses are just RNG machines anyways, so it's probably okay if there's an easier way to catch them in a farther region. The player will just have to either lose out on easy attack from capturing it and wait for an easier capture, or spend time in the RNG machine for the free attack to progress further.

Egg Pokemon were originally another RNG region gate and they're heavily hated in the community. It might just be easier to scrap most of the egg-exclusives and disperse them into other encounter pools. Some of the Egg-exclusives were due to not having a good location to add them in the game (see Starters/Lapras). It might be easier to have a story quest (at least for non-starters) to obtain these Pokemon. The story quests would also be optional for most of them, as the player can just wait until the reaching the newer region to obtain them.

QP Pokemon also fall into the same category as Egg Pokemon, where some can be relegated to story quests, while the others can be added back into encounter pools.


silasary:
Safari:  If we make routes 19 and 21 require clearing the safari at least once (to collect the Surf HM), hopefully that will give the player enough incentive to get some of the pokemon there.  Other than that, shrug.

I agree with you about ignoring Dungeon Bosses. Eliminating the RNG grind there is the main reason I like this change.

Egg Pokemon:  As long as they're still in the egg pool, I'm not fussed.  People will hatch them while grinding for starters.  

QP Pokemon:  These are the complicated ones, as they are intended to be a QP gate.  I'd want to do something special here.  Probably questlines, as we really don't have enough of them right now.

PR #1214, opened 2021-01-15
Ash Greninja

Ultima1990:
This adds Ash Greninja.
This also adds the beginnings of how to obtain the Greninja, which is done by adding six Ash Ketchum gyms. One in every region up to and including Kalos, with teams inspired by the anime. To be defeated in order. With the Greninja rewarded by defeating the final Ash in Kiloude City.

Currently I've added each Ash gym to the bottom of that region's list of gyms. Maybe it'd be better to have them all together in one place below the Kalos gyms tho. Let me know.

What still needs to happen:
The Ash's need to be balanced. They are post-KalosChamp fights, and should also increase in difficulty somewhat as you go along. Order of difficulty should be obvious, but just in case: Kanto --> Johto --> Hoenn --> Sinnoh --> Unova --> Kalos, with Kanto being easiest and Kalos being hardest.
The Ash's need sprites.
A questline needs to be added, pointing players in the direction of each Ash and actually rewarding the Greninja at the end.
I would like to find a way to have each Ash gym hidden from the player until the quest is started.
Kiloude needs to be added to the Kalos map.

Finally I'm currently having a bit of trouble with the BadgeEnums. I'll probably figure it out at some point... maybe...


src/scripts/temporaryBattle/TemporaryBattleRunner.ts
```diff
@@ -19,6 +19,9 @@ class TemporaryBattleRunner {
         this.timeLeft(GameConstants.TEMP_BATTLE_TIME * this.timeBonus());
         this.timeLeftPercentage(100);
 
+        player.route(0);
+        Battle.route = 0;
+        Battle.catching(!(battle.optionalArgs.isTrainerBattle ?? true));
```
RedSparr0w, line 24: I'm not quite sure what this part is supposed to be doing?
Could we add a comment explaining please

RedSparr0w (approved):
PR looks good to me 👍 

Is there any way we could use the same image for all the battles instead of duplicating the same one?

src/modules/requirements/QuestLineStartedRequirement.ts
```diff
@@ -0,0 +1,23 @@
...
+export default class QuestLineStartedRequirement extends Requirement {
+    questLineName: string;
+
+    constructor(questLineName: string, option = AchievementOption.equal) {
+        super(1, option);
+        this.questLineName = questLineName;
+    }
+
+    public getProgress(): number {
+        return (App.game.quests.getQuestLine(this.questLineName).state() === QuestLineState.started
```
RedSparr0w, line 15: Are we able to bind `App.game.quests.getQuestLine(this.questLineName)` to something in the constructor so we don't have to keep on calling it here?

Jaaslet: I could not get it to work. If i remember correctly, the quest lines are not created there. But might be able to add some caching. Might want to do it in a new PR, as i can do the same for other requirements

PR #1215, opened 2021-01-15
Preliminary fix for Unova dungeon difficultyRoute values

PixLSteam:
I haven't finished Unova yet, so I can't judge the difficulty of most of these routes and dungeons. I simply used the closest routes for now. I haven't tested all the dungeons, but Relic Castle now gives 22 egg steps per defeated enemy.


RedSparr0w (approved):
Looks good, thanks for the PR

PR #1218, opened 2021-01-17
Minor refactor

HLXII:
For some reason we were creating observables for Towns and Dungeons. I'm not sure why considering everything works without it anyways. I've removed the observables and replaced them with original values. This shouldn't have any affect on the game as Town/Dungeon data shouldn't be modified anyways (thus, having the data stored in observables doesn't really make sense).

We were also storing both the `Town`, as well as a `currentTown` property on the `player` object. I've refactored the logic to remove the `currentTown` property, as it seems to be unnecessary.

There was also something weird in `calculateTownCssClass` with key items. I didn't see any key items defined in calling this method, so I removed that as well.


RedSparr0w (approved):

Issue #987, opened 2020-10-23
[SUGGESTION] "Trainer Dungeons"


TrueWolves:
Right now, the Kanto wall can be pretty harsh, needing potentially more than 100k Quest Tokens to progress, and delaying the game long after you've already finished the Elite Four.

**Suggestion:**
Trainer Dungeons! Think like mini-versions of the Battle Frontier where you have multiple progressively more difficult Trainer Battles in a row. In places where you'd see a lot of trainer battles and virtually no wild pokemon in the game. Nugget Bridge can be the first "easy" one at just 6 trainers, while the SS Anne, and most of the Team Rocket dungeons (and other Team dungeons in regions outside of Kanto) can be Trainer Dungeons in the future.

**Reason for change**
By default, Trainer Dungeons would just extend the play time of each region a bit further, giving more time to complete quests and a bit of new variety to the game mechanics... but Trainer Dungeons could also grant one-time rewards. If each Trainer Dungeon is like a non-endless Battle Frontier, a certain number of trainers could be all you need to "Clear" the dungeon and progress further in the game... but there could be additional awards for some dungeons, like one-time pokemon (Lapris or a Water Egg for example) and/or items, like the Battle Frontier. This would help cut down on the needed amount of Quest Tokens in Kanto without influencing Quest Token prices later in the game. By both giving alternate goals during/past the Elite Four, and some 1-time sources of Evolutionary Stones/Eggs to cut down on the total Quest Tokens needed for just Kanto.

TrueWolves:
In particular, a single Fire Egg could be on Nugget Bridge to represent the free Charmander in Pokemon Yellow, and a single Fighting Egg could be in a Fighting-themed Trainer Dungeon in Saffron City.

jk13pclick:
Closed as not planned. Temporary Battles is "kind of" what this is going for, trainer-only dungeons also exist, but overall this is not a feature we are looking to actively pursue.

Issue #965, opened 2020-10-19
Trainers


Ultima1990:
My suggestion is to have trainers that can appear instead of wild pokémon in routes, and perhaps later on dungeons.

Trainers pokémon would not be able to be captured, and as such they would not be able to give DT either, but they would give increased money, exp and steps. We could also have quests that target trainers specifically.

In the main series not every route has trainers, so this system should be able to keep this in mind. In theory it would work like this:

If the route/dungeon has no trainers: 100% wild encounters.
If the route/dungeon has no wild pokémon: 100% trainers
If the route/dungeon has both trainers and wild pokémon: 10% trainers, 90% wild pokémon.
When a dungeon is generated, and the dungeon has trainers, and the dungeon has rolled trainers in tiles, for every trainer in the dungeon the dungeon's timer gains +X(20ish?) seconds. This, of course, assumes that dungeon encounters are decided when the dungeon is generated, not when the tile is entered. TBH I have no clue how this works.

This would allow us to get the villain teams involved properly, with several dungeons containing only trainers. Mt. Chimney could become a proper trainer dungeon. Mostly it would be a great help for making Kalos work, where the cover legendaries are inside a Team Flare dungeon with no wild encounters (even then that's a whole thing, TBH I don't know how to make it function with Lysandre as a boss but also having Xerneas and Yveltal there... we'll see what becomes of this).

Additionally, since challenges have become quite a popular topic of discussion (also involved in a potential future suggestion I'll post on the subject), this would allow for a true Team Rocket Challenge, where you can only catch Pokémon from gyms and trainers, but not wild encounters.

If groundwork is put in place to make this function, I'll gladly go in myself and manually fill in all the trainer data and all of their pokémon.

unclebanks:
I like the idea of adding trainers but if you make them rng then it'll make it that much harder to find roamers... Which are already kind of rough sometimes. I think adding main villains to the game wouldn't be too hard. That could probably be achieved as is by editing the maps and adding a gym type battle that sets a flag instead of a gym badge to unlock other villains.

jk13pclick:
Closing this as stale + outdated. The idea of Trainers on routes is approved, but we would like a new issue for that concept exclusively.

PR #1219, opened 2021-01-17
[FEATURE] Trainer Dungeons

HLXII:
Closing #987 and #965 

This PR will implement the framework for adding Trainers into Dungeons.

A couple of design considerations that need to be resolved before this PR can be fully submitted:

## Visuals

image

Here is an image of the current implementation.

~~1. Where should the Trainer's name be added in the view? For Gym battles, the Gym Leader name is placed where the Dungeon name is in the image.~~ Added
~~2. Does the Trainer image need to be moved anywhere? I set it to 65% from the left, but it can be adjusted if necessary~~
~~3. The Caught Pokeball icon appears next to the Trainer Pokemon's name. This can be easily removed, but not sure if necessary (Gym Battles don't have this).~~ This is fixed
~~4. How should trainers be handled in the initial Dungeon view? Currently we display the available pokemon. Should Trainers be added to the lineup? Should we display their team somewhere?~~ Trainers not added to encounter pool view

### Pokemon Health/Level

~~The current `DungeonBossPokemon` implementation has hard-coded health and levels, while the normal Dungeon Pokemon use a separate system for creating constructing the `BattlePokemon` instance. How should this be handled for both normal Dungeon Trainers, as well as Dungeon Trainer Bosses?~~

While the `DungeonTrainer` class uses `GymPokemon` to store the Pokemon data, for minion Pokemon the `maxHealth` and `level` properties aren't used, and the minion Trainers will have Pokemon generated using the original `Dungeon` values. The `maxHealth` and `level` properties *are* used when generating a Boss Trainer Pokemon.

### Rewards

~~What should the rewards be for defeating Trainers?~~

I've included a `reward` property on the `DungeonTrainer` class which will be used to gain money after defeating them, much like Gym leaders.


Ultima1990:
Displaying the trainer name instead of the dungeon name is probably fine.

I'm gonna need some explanation of what's happening here, as I still barely speak code.

Can dungeons now exist with only trainers and no wild pokémon?

What are the odds of a trainer showing up? IMO, if a dungeon has wild pokémon and trainers, the player should still encounter _mostly_ wild pokémon. To get that good old feel of encountering like 5 annoying bats before getting to the next trainer in mt moon. If everything has equal chance to show up, and a dungeon has a large trainer list, I can see a situation where you'd encounter mostly trainers and barely any wilds. Which would lead to frustration with catching and with the dungeon timer. No matter how many possible trainers can appear in a given dungeon, I think the odds of encountering one instead of a wild pokémon should be in the 20%-30%ish ballpark.
Different for bosses though. Equal chance for every possible outcome there. If Viridian ends up with a boss wild Pikachu and a boss Bug Catcher Sammy with his level 9 Weedle, that should be a 50/50.

As for rewards, I'm thinking just more money, exp and shards. 150% or 200%, not sure.

As for balancing, I'm not sure really. Just keep them the same difficulty as equivalent wild/boss pokémon I guess? Trainers are already made more difficult by the fact that they often have more than 1 pokémon. In fact, maybe a reward for trainers should be to get some seconds back on the dungeon timer. To compensate for the fact that they'll take longer to defeat. Who knows, maybe the true benefit of trainers in dungeons with wilds is that they'll return so much time that it becomes an actual strategy to seek out as many wilds as possible, hope that they are trainers, so that with the extra time you can hopefully finally take down and potentially catch that boss you've been struggling on. Something to think about.

If you think anything else could do with some explaining, please do, because honestly I have no idea what's happening here.

HLXII:
> Can dungeons now exist with only trainers and no wild pokémon?

If you want yes.

> What are the odds of a trainer showing up? IMO, if a dungeon has wild pokémon and trainers, the player should still encounter _mostly_ wild pokémon. To get that good old feel of encountering like 5 annoying bats before getting to the next trainer in mt moon. If everything has equal chance to show up, and a dungeon has a large trainer list, I can see a situation where you'd encounter mostly trainers and barely any wilds. Which would lead to frustration with catching and with the dungeon timer. No matter how many possible trainers can appear in a given dungeon, I think the odds of encountering one instead of a wild pokémon should be in the 20%-30%ish ballpark.

This can be controlled by the `weight` property.

```JS
    [
        'Caterpie',
        {pokemon: 'Bulbasaur', options: { weight: 2 }},
        new DungeonTrainer('Joey',
            [
                new GymPokemon('Rattata', 200, 5),
                new GymPokemon('Raticate', 210, 7),
            ], 1, { weight: 100 }),
    ],
```

Here's an example `enemyList` with the new Dungeon framework. Here, we have a `PokemonNameType` Caterpie, which means the `weight` defaults to `1`. Next, we have a `DetailedPokemon` interface with a Bulbasaur with `weight` set to `2`. Finally, we have a `DungeonTrainer`, whose `weight` is set to `100`.

When determining the random enemy to generate, we pool all the weights together (`1+2+100 = 103`). Then we generate a random number between 0 and that number to determine which enemy to create. In this instance, the Trainer has `100/103` chance to appear (since the weights are so heavily scaled).

If you wanted Trainers to show up 50/50, make sure the sum of their weights is half of the entire sum. Note that the `weight` value can be fractional as well, you don't need to use integers.

> Different for bosses though. Equal chance for every possible outcome there. If Viridian ends up with a boss wild Pikachu and a boss Bug Catcher Sammy with his level 9 Weedle, that should be a 50/50.

Bosses have the same `weight` property as well, so the same calculations can apply.

> As for balancing, I'm not sure really. Just keep them the same difficulty as equivalent wild/boss pokémon I guess? Trainers are already made more difficult by the fact that they often have more than 1 pokémon. In fact, maybe a reward for trainers should be to get some seconds back on the dungeon timer. To compensate for the fact that they'll take longer to defeat. Who knows, maybe the true benefit of trainers in dungeons with wilds is that they'll return so much time that it becomes an actual strategy to seek out as many wilds as possible, hope that they are trainers, so that with the extra time you can hopefully finally take down and potentially catch that boss you've been struggling on. Something to think about.

Currently we generate minion pokemon using a formula:

```JS
    public static generateDungeonPokemon(name: PokemonNameType, chestsOpened: number, baseHealth: number, level: number): BattlePokemon {
        const basePokemon = PokemonHelper.getPokemonByName(name);
        const id = basePokemon.id;
        const maxHealth: number = Math.floor(baseHealth * (1 + (chestsOpened / 5)));
        const catchRate: number = this.catchRateHelper(basePokemon.catchRate);
```
While for `DungeonBossPokemon`, we use hard coded values for the health/level:

```JS
new DungeonBossPokemon('Pikachu', 510, 7),
```

How should `DungeonTrainers` in both the minion and boss positions be handled? I'm leaning towards hard coding the health/level values like `DungeonBossPokemon` for both, since using the minion calculation means that every enemy Trainer pokemon will have the same health, which doesn't really make sense.


Final question that wasn't answered. Do we add the Trainers/their party to the Dungeon encounter list at the dungeon start screen? It would look weird for 100% Trainer only dungeons to have nothing there, but also it'll look weird if random trainer images appear in the encounter pool (like they're Pokemon themselves) without knowing their party.

Ultima1990:
Having all the wild pokémon with the same health already doesn't make sense. So long as that doesn't change, why not do the same nonsense with trainers.

IMO don't put trainers on the encounter list at the start screen. I can see that becoming very cluttered with the amount of trainers in the game.

RedSparr0w:
> Having all the wild pokémon with the same health already doesn't make sense. So long as that doesn't change, why not do the same nonsense with trainers.

Maybe we could make the health scale up/down a little bit _(maybe +/- 5%)_ based on each Pokémons base HP stat?

HLXII:
LOL

IIRC Apple mentioned this a *while* back when considering route pokemon, to have their health vary ~10% based on HP stat. I think that was rejected though.

If we do add that, it should probably apply across all locations.

RedSparr0w:
Yeah, something for another PR though anyway 😅

Ultima1990:
Changed my mind. Trainer _bosses_ should be on the encounter list. Bosses only tho.

RedSparr0w (approved):
Looks good, Thanks for the PR! 🎉

PR #1223, opened 2021-01-19
Item images cleanup

### LINKED ISSUE #1200: [SUGGESTION] Clean up Item Images
I realize that this is mostly my fault, but maybe someone wants to do some work for the game (aka I want to offload some busy work so I don't have to do it).

The item images right now are all sorts of sizes right now because they were ripped directly from Bulbapedia. The original item images were standardized to 120x120 pixel images, which makes rendering them in the game much easier. I didn't do this with the new items, which makes some items look wonky.

Here's a list of item images to update:
* Berries
* Mulches
* Possibly new Farm Oak Items, didn't check
* New Underground images from (#1197)

Also make sure to run the updated images through tinypng.com to minimize their size.

I'll probably get to this later when I have time, but like I said if someone wants to do this beforehand go ahead.

HLXII:
### Image cleanup

Closes #1200 

Updated Berry (thanks @Piotisgood), Farm, and Underground item image sizes to match standard 120x120 or 80x80. (both currently exist, and scaling up the 80x80 will make their resolution worse. It's probably better to just have the same relative spacing within every image).

### Item image directory cleanup

Moved related images to their own subdirectory within `images`. Added `imageDirectory` as an optional parameter for `Item` to consolidate the `image` getter mostly into `Item.ts`. Some items override the base `image` getter (see `buyKeyItem.ts` and `buyOakItem.ts`). This allows us to remove the additional images in the `items` directory and use the original ones.

This is still technically an issue with `EggItem.ts`, however it seems like there are two different Egg images, one with additional spacing and one without (to be used either in the Hatchery, or in the Shop). 

### Code Cleanup

Update similar methods to use `image`, rather than `imagePath`. Both were used in the codebase, so I moved everything to `image`. I've also updated all references to item images in the HTML scripts to use the `image` getter instead.

Removed `abstract` modifier from `Item` class. The only abstract method was `use`, which some items didn't require anyways. Doing this cleaned up the empty `use` methods in most subclasses.

Update `name` property in `Item` to be a `string`, rather than a `KnockoutObservable<string>`. This is related to #1218, where the item names never change, so there's no reason for it to be an observable.


RedSparr0w:
Ah, I think I may have created some conflicts here from the Galar PR.

RedSparr0w (approved):

PR #1224, opened 2021-01-19
Berry Update 10

HLXII:
### Log Book Entry

Added a `WANDER` Log Book entry for wandering Pokemon

### Lum Update

Lum Berries will now have a `Boost` Aura, scaling `1.01/1.02/1.03` in effect. The `Boost` Aura will multiply the Aura effect of surrounding Berries. This Aura acts like the `Death` Aura, in that it is the max magnitude received (multiple Lum Berries will have no additional effect).


RedSparr0w (approved):

PR #1226, opened 2021-01-19
Remove Flabbles

Ultima1990:
Removed Blue and Yellow Flabébé from routes. For reasons. Potentially garden related reasons.


RedSparr0w (approved):
RIP Flabble.

PR #1231, opened 2021-01-21
Remove regional forms from Battle Frontier if region has yet to be reached.

RedSparr0w:
Remove regional forms (Alolan, Galarian) from Battle Frontier if region has yet to be reached.


src/scripts/pokemons/PokemonList.ts
```diff
@@ -20817,6 +20817,17 @@ const pokemonMap: any = new Proxy(pokemonList, {
...
                 };
+            case 'randomRegion':
+                return (_max = GameConstants.Region.kanto, _min = GameConstants.Region.kanto) => {
+                    // minimum 0 (Kanto)
+                    const min = Math.max(GameConstants.Region.kanto, Math.min(_min, _max));
+                    const max = Math.max(GameConstants.Region.kanto, _min, _max);
+                    const filteredPokemon: PokemonListData[] = pokemon.filter(p => p.id > 0 && (p as PokemonListData).nativeRegion >= min && (p as PokemonListData).nativeRegion <= max);
+                    const random = Math.floor(Math.random() * filteredPokemon.length);
+                    const poke: PokemonListData = filteredPokemon[random];
+                    pokemon.find(p => p.id == 0);
```
Aegyo, line 20828: Did you want to do anything with the pokemon you are looking for here?

RedSparr0w: 🤦 Was supposed to return that if no Pokémon found 😅

RedSparr0w: Should be fixed now 👍

PR #1235, opened 2021-01-23
Fixed the Grammer for Let's Go Event Pokemon

kushpatel0703:
Fixed the grammar for Let's Go event Pokemon. I thought it look odd that the capitalization and apostrophe was missing.


RedSparr0w (approved):

Would you also mind updating the other places we have text saying `Lets go` → `Let's go` in this PR too?
Think there's a couple of buttons in modals and the title of the event.

kushpatel0703:
Here you go boss. I'll let you take final look @RedSparr0w I think I got all of them though

RedSparr0w:
Looks good, Thanks 🌮

PR #1237, opened 2021-01-23
Update Quest Generation Limitations

HLXII:
### Defeat Pokemon Quest
We now prioritize accessible routes
### Gain Shards Quest
We now allow for all shard types, and scale the reward based on shard type
### Defeat Gym Quest
We now prioritize accessible Gyms
### Defeat Dungeon Quest
We now prioritize accessible Dungeons
### Harvest Berries Quest
We now select from unlocked Berries (we also always include all Gen 1 Berries)
I've also updated the reward calculation to be more balanced.

## Testing

For testing, you can:
1. Remove all unnecessary quests from `QuestHelper.quests`
2. Set `uniqueQuestTypes` to false in `Quests.generateQuestList`

Then just reroll the quests in-game.

### DIFF
--- a/src/scripts/quests/questTypes/GainShardsQuest.ts
+++ b/src/scripts/quests/questTypes/GainShardsQuest.ts
@@ -33,7 +33,7 @@ class GainShardsQuest extends Quest implements QuestInterface {
     }
 
     public static generateData(): any[] {
-        const type = SeededRand.fromArray(GameHelper.createArray(1, GameHelper.enumLength(PokemonType) - 1, 1));
+        const type = SeededRand.fromArray(GameHelper.createArray(0, GameHelper.enumLength(PokemonType) - 2, 1));
         const amount = SeededRand.intBetween(200, 600);
         const reward = this.calcReward(type, amount);
         return [amount, reward, type];
--- a/src/scripts/quests/questTypes/GainShardsQuest.ts
+++ b/src/scripts/quests/questTypes/GainShardsQuest.ts
@@ -10,25 +10,37 @@ class GainShardsQuest extends Quest implements QuestInterface {
         this.focus = App.game.statistics.shardsGained[this.type];
     }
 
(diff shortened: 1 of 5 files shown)


src/scripts/quests/questTypes/GainShardsQuest.ts
```diff
@@ -10,25 +10,37 @@ class GainShardsQuest extends Quest implements QuestInterface {
...
-            PokemonType.Normal,
-            PokemonType.Poison,
-            PokemonType.Water,
-            PokemonType.Grass,
-            PokemonType.Flying,
-            PokemonType.Fire,
-            PokemonType.Fighting,
-        ];
-        const type = SeededRand.fromArray(possibleTypes);
+        const type = SeededRand.fromArray(GameHelper.createArray(1, GameHelper.enumLength(PokemonType) - 1, 1));
```
RedSparr0w, line 36: This looks like it will be able to generate the type `18` which is `undefined`?
```js
GameHelper.createArray(1, GameHelper.enumLength(PokemonType) - 1, 1)
(18) [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18]
```

HLXII: Fixed to
```JS
const type = SeededRand.fromArray(GameHelper.createArray(0, GameHelper.enumLength(PokemonType) - 2, 1));
```

If you find a cleaner solution feel free to add it. I don't know of a clean way of not including the `None` enum value.

RedSparr0w: Just realized we already have a method in `SeededRand` actually:
```
SeededRand.fromEnum(PokemonType);
```
Which filters out any strings and numbers below 0.

RedSparr0w (approved):

PR #1239, opened 2021-01-23
Fixing Underground Messages during Tutorial

### LINKED ISSUE #1199: [BUG] Underground notification appears before underground is unlocked
**Info**
 - OS: Windows 10
 - Browser Chrome
 - Version v0.7.2 (Most recent as of this submission)

**Description:**

Before the underground is unlocked the notification "Your mining energy has reached maximum capacity!" can appear on screen. This can confuse the player since they wouldn't know what the notification means.

**Steps to Reproduce:**

1. Start a brand new game
2. Either wait a few minutes or execute `App.game.underground.gainEnergy()` a few times until mining energy is 50.
3. Notice the notification: "Your mining energy has reached maximum capacity!"

**Expected Behavior:**

No notifications for the underground should appear until the underground is unlocked

**Screenshots:**

image

**Additional Information:**

The simplest solution is to have the underground initialize with max energy rather than zero, so that this notification cannot be triggered until after the player first mines. Other solutions such as checking if the underground is unlocked before notifications are made seems convoluted.

kushpatel0703:
Addressing #1199 as it's super simple fix

closes #1199

### DIFF
--- a/src/scripts/underground/Underground.ts
+++ b/src/scripts/underground/Underground.ts
@@ -6,7 +6,7 @@ class Underground implements Feature {
 
     upgradeList: Array<Upgrade>;
     defaults: Record<string, any>;
-    private _energy: KnockoutObservable<number> = ko.observable(50);
+    private _energy: KnockoutObservable<number> = ko.observable(Underground.BASE_ENERGY_MAX);
 
     public static itemSelected;
     public static energyTick: KnockoutObservable<number> = ko.observable(60);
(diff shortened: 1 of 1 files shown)


src/scripts/underground/Underground.ts
```diff
@@ -6,7 +6,7 @@ class Underground implements Feature {
 
     upgradeList: Array<Upgrade>;
     defaults: Record<string, any>;
-    private _energy: KnockoutObservable<number> = ko.observable(0);
+    private _energy: KnockoutObservable<number> = ko.observable(50);
```
RedSparr0w, line 9: Could you use `Underground.BASE_ENERGY_MAX` here instead of magic numbers please.

RedSparr0w (approved):

PR #1242, opened 2021-01-23
Fix shop egg items

HLXII:
Minor fix to get shop egg items to display correctly. Forgot to add their subdirectory.


RedSparr0w (approved):
👍 Thanks for the PR!

PR #1246, opened 2021-01-24
Trainers for Kanto Dungeons

Ultima1990:
This adds trainers to existing dungeons only. Adding the trainer only dungeons (Game Corner, Silph Co., SS Anne, etc) will be a whole different thing. I used FRLG for the trainers, as trainers didn't have proper names in RBY.

A couple things have come to light while doing this:

Trainer sprites are not working in battle, but they are working in the dungeon preview.
viridiantrainer

Full name seems to be needed when looking for the sprite. This would result in a massive list of duplicate sprites.
trainersprites
As such, all of the trainers I added are only named for their trainer class. No Joeys, only Youngsters. Maybe we should look into doing something with trainer class. Bug Catcher, Hiker, Lass, etc, to fetch the correct sprite. Then I can add their proper names afterwards.

Dungeon encounters are acting all strange-like currently. Sometimes when I enter a red dungeon tile there will be no sprite and no HP bar, though the battle does take place. I can also escape these spriteless battles by simply moving to a different tile during the battle. Seems to only happen with wild Pokémon I have not caught yet. Doesn't happen with trainers.

ClearDungeonRequirement is not working as a requirement to lock a boss. Maybe because that requirement doesn't work in dungeons at all, maybe because the requirement is the dungeon the boss resides in (Mt. Moon requirement inside Mt. Moon).

Joey seems to be hard coded. Could be the cause of some of these issues.


Ultima1990:
New push containing the fix from #1247. Trainers have names now. Yay.

Ultima1990:
Almost everything is working now! Yay! Many thanks to HLXII! All that remains is this bit:

> ClearDungeonRequirement is not working as a requirement to lock a boss. Maybe because that requirement doesn't work in dungeons at all, maybe because the requirement is the dungeon the boss resides in (Mt. Moon requirement inside Mt. Moon).

Ultima1990:
All problems fixed, all images ready, all dungeons filled with trainers where appropriate. Rewards are not where they should be, but that can be part of a larger change to dungeon trainer rewards. I'm gonna say this is done.

KrushGames:
im gonna use your formatting for some of the other regions, so thanks for being the first one to do this (besides HLXII of course).

src/scripts/dungeons/Dungeon.ts
```diff
@@ -100,6 +120,21 @@ class Dungeon {
...
+        return this.enemyList.map((enemy) => {
+            if (typeof enemy === 'string') {
+                return 1;
+            } else if (enemy.hasOwnProperty('pokemon')) {
+                return (<DetailedPokemon>enemy).options.weight ?? 1;
+            } else {
+                return (<DungeonTrainer>enemy).options?.weight ?? 1;
+            }
+        });
+    }
```
RedSparr0w, line 136: @HLXII 
Now that the standard dungeon enemies can have requirements, should we be checking those here too?

src/scripts/dungeons/Dungeon.ts
```diff
@@ -115,34 +150,17 @@ class Dungeon {
...
-            if (typeof enemy === 'string') {
-                return 1;
-            } else if (enemy.hasOwnProperty('pokemon')) {
-                return (<DetailedPokemon>enemy).options.weight ?? 1;
-            } else {
-                return (<DungeonTrainer>enemy).options?.weight ?? 1;
-            }
-        });
+    get availablePokemon(): PokemonNameType[] {
+        return this.pokemonList.concat(this.bossPokemonList);
```
RedSparr0w, line 157: Should this be called something like `allPokemon`?
as this is a list of all possible Pokémon, not necessarily available (locked by requirements)

src/scripts/dungeons/Dungeon.ts
HLXII: Updated to handle those (though there aren't any use cases yet, so it isn't tested).

You'd probably also need to update `getPokemonDungeons` to handle the requirements like `getPokemonBossDungeons` for the bot as well.

src/scripts/dungeons/Dungeon.ts
HLXII: Updated

RedSparr0w:
If anyone thinks any other ribbons would work better for the challenges, just download them from here and let me know:
https://msikma.github.io/pokesprite/overview/misc.html

(preferably gen 8 ribbons)

RedSparr0w (approved):

PR #1247, opened 2021-01-24
Update Trainer class for more details

HLXII:
Updated the `Trainer` and `DungeonTrainer` classes to include more details.

[trainerClass] - String for the class name
[subTrainerClass] - additional details for the class
[name] - Unique name of trainer

This way you could have `("Bug Catcher", undefined, "Charlie")`, which would be `Bug Catcher Charlie`.
You could also have `('Team Rocket Grunt', undefined, undefined)` for `Team Rocket Grunt`.
And also `('Team Rocket Grunt', '(female)', undefined)`, which would still display the name as `Team Rocket Grunt`, but the additional subClass will display the female picture instead (with the image name being `Team Rocket Grunt (female)`.


RedSparr0w:
closed via #1246

PR #1251, opened 2021-01-25
Don't allow leaving dungeon until moved off first tile

RedSparr0w:
Prevent people accidently leaving the dungeons..


Aegyo (approved):
Looks good :+1:

PR #1253, opened 2021-01-26
Add map overlay template

RedSparr0w:
Allows you to add temporary dungeons or other stuff that shouldn't always be on the map.
Also added example with Kanto SS-Anne that appears once you reach Johto.


src/components/KantoSVG.html
```diff
@@ -544,4 +544,14 @@
         , databind: "click:function(){MapHelper.openShipModal()}, css:{ unlockedTown: player.highestRegion() >= 1}"
         })
      %>
+
+     <%- include('templates/mapOverlay',
+         { name: "SS-Anne"
+         , x: 50, y: 35
+         , width: 8, height: 3
+         , databind: "click:function(){MapHelper.openShipModal()}, css:{ unlockedTown: player.highestRegion() >= 1}"
+         , visible:  "player.highestRegion() >= 0"
```
Aegyo, line 553: You say that this only appears once you reach Johto, did you mean it only *unlocks* or was this meant to be a 1 here? I can see it as soon as I unlock the map on a new save.

Aegyo (commented):
The ejs looks good, though if you are going to push a change for my other comment, since you went to the trouble of adding parameter documentation some mention of the default height and width + letting people know the `visible` thing is going into a knockout data-bind would be nice. I'm sure every future mapOverlay will just be copy pasted from the existing one though, so no big deal.

src/components/KantoSVG.html
RedSparr0w: 😅 Yeah, supposed to be a `1` forgot to change it after testing
```suggestion
         , visible:  "player.highestRegion() >= 1"
```

RedSparr0w:
> some mention of the default height and width + letting people know the visible thing is going into a knockout data-bind would be nice.

Ah yeah, I probably should make width/height required as, all the images used aren't likely to be the same size.

Aegyo (approved):

PR #1259, opened 2021-01-29
Pokémon League number key fix

PixLSteam:
This should allow using the number keys to start Elite Four battles.


RedSparr0w (approved):

PR #1260, opened 2021-01-29
Fix checking all regions for current environment

Aegyo:
Just checks the current region environment info, so Kanto 23 doesn't pick up Galar 23, or any other routes that may be affected.


RedSparr0w (approved):
Looks good, Thanks for the PR 👍

PR #1270, opened 2021-02-02
Fix leaving dungeon without moving

PixLSteam:
Pressing the down arrow would allow the player to leave the dungeon, even though they haven't moved. This should fix that issue.


RedSparr0w (approved):

PR #1271, opened 2021-02-02
Fix oak item challenge

PixLSteam:
While playing with the "No Oak Item" challenge, you could still toggle them using the number keys. This should fix that issue.
Additionally, this should also fix the issue where holding down the `O` key would prevent any further inputs until the page is refreshed.

### DIFF
--- a/src/scripts/GameController.ts
+++ b/src/scripts/GameController.ts
@@ -127,7 +127,7 @@ class GameController {
     static addKeyListeners() {
         // Oak Items
         const $oakItemsModal = $('#oakItemsModal');
-        $oakItemsModal.data('allowToggle', true);
+        $oakItemsModal.data('allow-toggle', true);
         $oakItemsModal.on('hidden.bs.modal shown.bs.modal', _ => $oakItemsModal.data('allow-toggle', true));
         const oakItems = App.game.oakItems;
         // Pokeball Selector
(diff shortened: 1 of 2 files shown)


PixLSteam:
It would probably be a good idea to include an update mechanism to make sure people abusing this bug would have their oak items deactivated.

src/scripts/GameController.ts
```diff
@@ -127,6 +127,8 @@ class GameController {
     static addKeyListeners() {
         // Oak Items
         const $oakItemsModal = $('#oakItemsModal');
+        $oakItemsModal.data('allowToggle', true);
```
RedSparr0w, line 130: Just for consistency with the other areas, can we change this to be:
```suggestion
        $oakItemsModal.data('allow-toggle', true);
```

PixLSteam: The way this works is by setting that value to `false` when toggling the modal, and then waiting for the event `hidden.bs.modal` or `shown.bs.modal` until we allow toggling again. I'm pretty sure we need to set the initial value to `true`, otherwise we need to open the modal manually at least once before we can use the hotkey.

RedSparr0w: It might be better to instead flip the flag, so its `disable-toggle`?
That way you don't need anything there initially?

PixLSteam: Done.

RedSparr0w (approved):

PR #1275, opened 2021-02-07
Initialise settingsData to empty object if not found when updating

Aegyo:
This is to fix the game crashing in a certain situation when loading an old save file.

1. If the localStorage is cleared, or you haven't played pokeclicker on that browser/device, we don't have a settings object.
2. The settings object is not included with the save file.

When updating, we are were checking if we found a settings object, but if we are in the situation described by (1), and we import an old save file, we would skip the updates for that save file. This usually means the game crashes, because we haven't applied necessary updates.

<sub>Sorry for the whitespace changes, my editor is doing it automatically and I am too lazy to figure out how to make it stop</sub>


RedSparr0w (approved):

PR #1276, opened 2021-02-08
Logbook settings

unclebanks:
*Created tab for logbook settings
*Added relevant options

### DIFF
--- a/src/modules/logbook/LogBookSettings.ts
+++ b/src/modules/logbook/LogBookSettings.ts
@@ -1,0 +1,20 @@
+import BooleanSetting from '../settings/BooleanSetting';
+
+/**
+ * Contains all LogBook settings for easy access.
+ */
+
+const LogBookSettings = {
+    LogBookOptions: {
+        logbook_new: new BooleanSetting('logbook_new', 'Newly encountered Pokémon', true),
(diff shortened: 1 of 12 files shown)


src/modules/logbook/LogBookSettings.ts
```diff
@@ -0,0 +1,20 @@
+import BooleanSetting from '../settings/BooleanSetting';
+
+/**
+ * Contains all LogBook settings for easy access.
+ */
+
+const LogBookSettings = {
+    LogBookOptions: {
+        logbook_new: new BooleanSetting('logbook_new', 'Newly encountered Pokémon', true),
```
Aegyo, line 9: I think it would be a good idea to have these generated based on all of the log book types available, so that when a new type is added we automatically get a setting to go with it.

In order to make these automatically created settings work with minimal effort, I think you should move the setting checks inside of the `logbook.newLog` function, using the log book type to calculate what the setting is for this type. This would also guarantee we haven't missed checking any newLog calls, or allow us to create any in the future that don't check a setting first.

You will probably lose all of these nice descriptions by doing this though, maybe log book types should be required to have a human readable description in their definition, and you can just reference it here?

I haven't tested doing it this way, so I am unsure if some part of this will cause module headaches.

unclebanks:
closing pr because its easier to start over for different filter function :/

PR #1277, opened 2021-02-09
[CODES] Only allow obtaining a shiny from within your maximum region

RedSparr0w:
Changed the free shiny code to only being able to obtain a shiny from within your maximum region, you can still obtain a unreleased Pokémon such as the alternate Arceus forms, or Unowns, but don't think it's too much of an issue.


silasary:
That code can currently give you a pokemon that is already shiny. Do we want to prevent that?

PR #1278, opened 2021-02-10
Use custom prompt and confirm

RedSparr0w:
Use custom prompt/confirm, needed due to electron not supporting `prompt`.


Aegyo:
May want to add `'no-alert': 'error'` into the main eslint config <https://eslint.org/docs/rules/no-alert>, so we don't accidentally add more later.

RedSparr0w:
Good point!

PR #1290, opened 2021-02-19
Fix quests refreshing without confirming

PixLSteam:
This should fix the issue that clicking the refresh button would generate new quests immediately without waiting for user input.


Aegyo (approved):
:+1: Looks good, and I can't find any other places with the same issue.

PR #1297, opened 2021-02-26
A better solution to global references in modules

Aegyo:
## AKA No more ugly `@ts-ignore` on all of the App references!

Instead, we have
 - An entry in the eslintrc `globals`, to stop eslint complaining
 - An entry in the `declare global` statement of `modules/globals.ts`, to stop typescript complaining
 - With type information for those globals, that we can update as more becomes available in modules

The advantage of this method is that we can have some type checking on these global references, and we don't have to add comments above every single one. We don't even need to import the globals to the module we want to use them in, they are just magically there (don't ask me how, I am as suprised as you).

There is a downside - we will have to maintain the temporary type structures when any changes occur to the real types/classes. This still seems preferable to the current situation though, where we have no type checking at all for these globals.

When the real type/class for a global is available in modules, we can reference it directly in `modules/globals.ts` to gain the benefits of the full type checking wherever the global is referenced, without needed to make changes to the files it is referenced in.

I haven't provided any type info for `player`, it looked like it may take a few minutes to figure out all of the properties available, and I wanted to just get this up for review first. The `any` type I have given it for now is no worse than the current situation of ignoring ts errors.


RedSparr0w (approved):
Awesome work, thanks for the PR! 🎉

RedSparr0w:
> We don't even need to import the globals to the module we want to use them in, they are just magically there (don't ask me how, I am as suprised as you).

I guess just because they are globals in `window`? _maybe_
_That's also assuming they actually exist in window before they are used in the modules too_

> I haven't provided any type info for player, it looked like it may take a few minutes to figure out all of the properties available, and I wanted to just get this up for review first. The any type I have given it for now is no worse than the current situation of ignoring ts errors.

Works good enough for now, Probably better to eventually move the `player` data to their own respective modules so we can possibly remove `player` entirely?

Aegyo:
I was expecting that I would need to import `globals.ts` *somewhere* at least, so that the module gets evaluated. If the module isn't being used anywhere, why does it get run at all? Maybe typescript sees the `declare globals` thing and thinks it should do that first?

Yeah, moving everything in player to their own modules is a good idea. Player used to be way more bloated and has already gone through some refactoring into separate features. It doesn't seem like there is a lot left to the player class on second look, even less that I can't think of a better place for.

PR #1299, opened 2021-02-28
chore(modules): Move Wallet, Amount, AmountFactory

Aegyo:
An example for banks of moving something to modules that accesses `App`. Nothing hugely special, there were no compiler complaints I had to deal with for those lines.

Wallet is the only thing that was using `GameController.animateCurrency`, so I have pulled it out into a new file for modules. The rest of GameController needs a few other things before it can be moved.

Also in this, I'm trying out a potential way to avoid bloating the temporaryWindowInjection file, by having an `index.ts` file in the root of some feature folder (in this case, `wallet`) which imports the classes we want to expose globally, and exporting an object with those classes. Then in temporaryWindowInjection, we can import the folder name (`import WalletClasses from './wallet';`) and add everything from that `index.ts` with one line (`...WalletClasses,`). If we want to use this `index.ts` feature for some other purpose (like in Settings), we can probably just have another file which does the same thing, and specify it exactly (eg, `./wallet/inject.ts`)
Not hugely necessary for Wallet, but some folders are going to have a whole bunch of files we may need to expose, like the Achievements/Requirements.


Aegyo:
On second thought, an explicit `wallet/inject.ts` may be better, so we dont accidentally import from somewhere that is only intended to be temporary.

RedSparr0w (approved):
Looks good, Thanks for the PR! 🎉

PR #1301, opened 2021-02-28
refactor(PokemonNameType): Define PokemonNameType separately to PokemonList

Aegyo:
This gives us type checking on pokemon names in evolutions, which is nice, but the main reason I want it is so we don't have to move a million things to modules all at once because of the PokemonList -> Evolution -> <various classes all over the game> dependencies (and those various classes want PokemonNameType, so we can't just move those before doing PokemonList)

The downside of this is having to maintain a separate definition of pokemon names, which I do not expect to be a huge chore. The pokemon list won't be changing too much, just for new events (only a few pokemon at a time) and new generations (doesn't happen very often).

Typing pokemonList -> name as PokemonNameType also prevents us adding pokemon without remembering to also update PokemonNameType

The other alternative to breaking this knot in the dependency chain is declaring all of the pokemon evolutions outside of PokemonList, but I can't convince myself that is better. I like having the evolutions defined with the rest of the pokemon data.

This has also highlighted an issue with Galarian Meowth's evolution, a nice bonus to validate doing this :smile:


src/modules/pokemons/PokemonNameType.ts
```diff
@@ -0,0 +1,1155 @@
+export type PokemonNameType
```
RedSparr0w, line 1: Should we also leave some commented code here for people to use to update the list?

something like:
```suggestion
/*
copy(`export type PokemonNameType
    = ${pokemonList.map(p => `'p.name'`).join('\n    | ')};`);
*/
export type PokemonNameType
```

Although I'm not sure if there is a way to force run ts and ignore any type errors 🤔

Aegyo: That is the exact code I used :sweat_smile: (it leaves a bunch of unescaped `'` in names like Farfetch'd though)

Having that would be useful, we can get around the type errors when updating the list by making a temporary copy of PokemonList without any type restrictions, and run the code on that list instead.

RedSparr0w: Ahh, forgot about those ones, we should just be able to use `${p.name.replace(/'/g, "\\'")}` to escape those.

Good point, that should do the trick.

RedSparr0w (approved):

PR #1302, opened 2021-03-01
fix(event-roamers): Add Type information to pokemonMap

Aegyo:
Depends on #1301

This adds a mostly truthful type for pokemonMap, I had to make a new type wrapper for proxy as the existing type assumes what we return is the same kind of thing we put in, but we are adding extra functionality.

Doing this highlighted a few places in the code where the `any` type we used to have was hiding issues, the biggest (and motivator for this PR) was special event roamers - at some point we updated RoamingPokemon to just take a name, but didn't update special events.

A weird one I am not sure about is EggItem calling Breeding::gainPokemonEgg with a random PokemonListData, instead of the PartyPokemon it was expecting. I don't think that bit of code is ever run, but I have updated gainPokemonEgg to handle the case if it ever is.

(This is likely how somebody got a MissingNo recently)


RedSparr0w (approved):

PR #1336, opened 2021-04-06
Make Flabble Yellow and Blue available

RedSparr0w:
Make Flabble Yellow and Blue available from the Farm via Wandering.


HLXII:
The original plan was for Blue and Yellow to be wanders for Blue and Yellow berries respectively, but having them as Kalos wanderers is probably fine.

RedSparr0w:
👍 That makes more sense, i'll update it to do that.

PR #1344, opened 2021-04-13
[BUG FIX] Fixed Kalos roaming birds

PixLSteam:
This should fix the requirement for the Kanto birds in Kalos. I don't have a Kalos save for testing but this should be a trivial change.


RedSparr0w (approved):
👍 Looks good thanks for the PR!

PR #1346, opened 2021-04-15
[BUG FIX] Added home: true to gen 5+ starting towns

### LINKED ISSUE #1345: [BUG] Unova+ maps have no town flagged as the home town
**Info**
Sinnoh and below all have a town with `home: true` like this:
```js
    <!--Twinleaf Town-->
    <%- include('templates/mapTown',
        { name: "Twinleaf Town"
        , x: 7, y: 52.5
        , height: 4.5, width: 5
        , home: true
        })
    %>
 ```
 
 Unova+ do not, which means nothing flashes when you are able to move on to the next region (even though we have starting towns specified in GameConstants.

**Expected Behavior:**
All region maps should have a town flagged as the home town. There might be something we can do using GameConstants.StartingTowns, but it will be simpler to just add the `home: true` to the relevant towns in the map svg/ejs

**Additional Information:**
Home towns should match up with what GameConstants.StartingTowns claims.

PixLSteam:
Unova: Aspertia
Kalos: Vaniville
Alola: Iki
Galar: Postwick
Apparently my editor removed some whitespace, but that shouldn't be a problem.

Fixes #1345 

I tested this and Aspertia immediately started flashing once I got the final Unova Pokémon.

### DIFF
--- a/src/components/AlolaSVG.html
+++ b/src/components/AlolaSVG.html
@@ -223,6 +223,7 @@
         { name: "Iki Town"
         , x: 36, y: 0
         , width: 4, height: 4
+        , home: true
         })
     %>
 
--- a/src/components/GalarSVG.html
+++ b/src/components/GalarSVG.html
@@ -59,7 +59,7 @@
         , noText: true
         })
     %>
-    
+
     <!--West Lake Axwell-->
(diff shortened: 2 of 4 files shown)


RedSparr0w (approved):

PR #1353, opened 2021-04-19
Add pathing to tiled

RedSparr0w:
Make creating the paths a little bit quicker

Select the terrain brush
image

Select terrain sets (down bottom right)
Select the path type you want
image

Will auto fill in the edges/corners while drawing paths

Note: holding CTRL will fill 3x3 paths
Note2: You need to hold CTRL to fill using the "Land Mass" mode, which will randomly choose between all the "grass only" tiles.


RedSparr0w:
Examples:
paths
paths
paths

Although it can get a bit messed up if you push different paths over existing paths:
paths2

PR #1357, opened 2021-04-20
Hide unobtainable achievements

### LINKED ISSUE #1265: [SUGGESTION] Remove unobtainable achievements based on challenge mode settings
**Current Behavior**

Can't complete oak item achievements with `No Oak Items` or click attack achievements with `No Click`
Not sure which others may be affected by challenge modes 🤔 

**Improved Behavior**

Don't include these achievements in total percentage for challenge mode, Hide the affected achievements.

**Reason for change**

Impossible to complete some achievements with challenges activated

PixLSteam:
This PR should hide unobtainable achievements and recalculate the bonus, making sure you can reach the maximum achievement bonus even with challenges enabled. This fixes #1265.
The only affected achievements right now are the oak item level achievements and the click count achievements. If I missed any, feel free to point them out.


RedSparr0w:
Seems to work well, the only problem I've noticed is when starting a new game, the achievement bonuses percentages aren't calculated correctly.

For example a new all challenge run:
image
compared with the same save after reloading:
image

RedSparr0w (approved):

PR #1359, opened 2021-04-22
Fix some requirements

Ultima1990:
Fix Lumiose opening early.
Fix Camphrier opening early.
Prevent skipping of route 10.


RedSparr0w (approved):
Thanks for the fixes!

PR #1371, opened 2021-04-29
fix(hatchery): hatched counter not updating

RedSparr0w:
Hatched counter _should_ update correctly now in the hatchery.
Also changed `getDisplayValue` to be public as it is used out of the class (for getting the value in the html).


Aegyo (changes requested):
This doesn't seem to work. The problem is likely because on that first run of timesHatched check, the statistics observable for that id does not exist yet, so there is nothing to subscribe to and then we don't get an update when it is eventually created.

Suggested solution: If one of these dynamic statistics observables does not exist when checking it, force the creation. I can't think of a situation when we wouldn't want it to be created - if we are looking at it, we probably want updates later.

To test it, I am doing this:
1. Create new save
1. Use the code to complete kanto
1. `App.blah.gainExp(1e12)` to get everything to level 100
1. Open the hatchery list, switch to `Times Hatched` if not already selected
1. Put Bulbasaur in egg
1. `App.blah.progressEggs(9999)` to complete egg
1. `App.blah.gainExp(1e12)` again to get Bulba back to 100
1. Check Hatchery to see if times hatched updated

RedSparr0w:
> Suggested solution: If one of these dynamic statistics observables does not exist when checking it, force the creation. I can't think of a situation when we wouldn't want it to be created - if we are looking at it, we probably want updates later.

I _think_ we are already doing this in the Statistics class:
https://github.com/pokeclicker/pokeclicker/blob/b7b87484e480647c9cb66a6040de40aad7cc6fab/src/modules/DataStore/StatisticStore/index.ts#L151-L169

As if I enter something like `App.game.statistics.pokemonHatched[1234]()` it returns 0, which shouldn't have already been set by the game.

RedSparr0w:
Found the problem, we were only doing that for Array observables, not the Object observables..

https://github.com/pokeclicker/pokeclicker/blob/b7b87484e480647c9cb66a6040de40aad7cc6fab/src/modules/DataStore/StatisticStore/index.ts#L216-L222

PR #1390, opened 2021-05-06
[No test - May not work] Static Encounters

anapple96:
0 test, may not work. Will test later


anapple96:
Closing, as static encounters will probably be made in a different way

PR #1402, opened 2021-05-10
Small Daily Deals QOL Change

catie-cat-3183:
Added a column so you can see how much of an item you have before you trade for it :)


RedSparr0w (approved):

PR #1404, opened 2021-05-14
Kanto Trainer fix + Kalos Trainer Name Fix

jk13pclick:
One last tiny Kanto Trainer fix


RedSparr0w (approved):

PR #1409, opened 2021-05-20
First half of the Alola map

Ultima1990:
Including all the SVG stuff.

Extra bonus: removed "Alola" from Akala Outskirts and added ' to Kala'e Bay.


RedSparr0w (approved):

PR #1413, opened 2021-05-23
Balancing EXP and money

anapple96:
(Most of these ideas are from HLXII, I'm just doing the code work)
This will (hopefully) make the EXP and money scaling less severe, by solving:
 - EXP/Money scaling from achievement. They're now capped at 100%. In return, the achievements also give a 50% boost to dungeon tokens and egg steps
 - Lucky Incense removed from shops. Previously, this was a very overpowered, as you're almost guaranteed to make more money than the cost of one Lucky Incense for every Lucky Incense use, thus creating a very ROI rate. Now, the Incense is only obtainable in dungeons
 - Slightly nerfed the cost and effectiveness of Lucky Egg


anapple96:
This is going to stay here for a bit, as this will probably make sense in the future, but not now

anapple96:
Screw this, found the problem with coins, will push a separate PR

PR #1414, opened 2021-05-23
Add sub-regions

RedSparr0w:
Just added Alola's island for the example as they will be 2 maps.
(not sure if I've labelled them correctly either)

When moving between main regions, you will always go to the main island first. (sub-region resets to 0)

Still needs a few things done:
- [x] Move to a modal?
- [x] Move the player to a certain route/city in the new sub-region when moving
- [ ] Probably something else I have missed too 🤷


RedSparr0w:
Moved to a Modal:
image

Button shown just above map (if multiple subregions are unlocked):
image

RedSparr0w:
Included players selected trainer in the modal now too:
image
Also added this to the usual dock.

PR #1415, opened 2021-05-24
Rebalance all gyms and E4 money

anapple96:
Someone completely messed up the values for these, some of these are 8x higher than what they should be. The worst case is Kalos' first gym, with up to 100x coin inflation


RedSparr0w (approved):

Issue #1405, opened 2021-05-14
[SUGGESTION] Mass queue Pokémon for breeding  


CIH137:
Buttons added to Day care to quickly fill hatchery queue add 5 10 25 max after chosen pokemon

**Current Behavior**
When you hit the last region you have to click over 100 times to fill the queue 

**Improved Behavior**
One button to add all to queue starting with the 1st on the list based on your preferred sorting method

**Reason for change**
With a farm full of chople berries, a blaze cassette, and farming high routes, 60 egg steps per Pokémon is possible leading to a need to fill the queue more often.

**Detailed Description**

Next to filter and sort, a third option for breeding could be added. The drop down would show added clicked Pokémon and [5] [10] [25] [Max] after it in descending order

amative1:
Per Hatchery Helpers ( #2053 ), no longer necessary.

Issue #1179, opened 2020-12-31
[SUGGESTION] "Fill Hatchery" button in Hatchery


IanWalder:
**Current Behavior**

Players have to manually click a maximum of 36 times to fully fill up their egg slots and queue as of v0.7.0.

**Improved Behavior**

A button that auto-fills the egg slots and queue with the Pokémon currently displayed in the Hatchery modal.

**Reason for change**

With the current rate of queue slot scaling (as of the current version), players have access to 32 queue slots in the endgame. Assuming the number doubles every region, players have to click an exponentially increasing number of times to get the most out of the breeding system. Considering how often the hatchery is used, this would be a useful QoL feature, at least for the later regions.

It's also important to note that egg-hatching rates vary wildly from player to player. For example, one who has access to the Chople berry (boosts egg steps) and a Level 5 Blaze Cassette would currently need to click about 2.25x more frequently per unit time than a player who hasn't progressed that far.

**Detailed Description**

Have a button on the Hatchery modal that, when pressed, automatically fills the egg slots and queue (whichever is applicable) with the Pokémon currently displayed in the Hatchery modal (with any filters and sorting applied).

To prevent a loss of emphasis on clicking in the early regions, there could be any number of restrictions barring access to this button, from a flat Quest Point fee to locking it behind later regions, or even a dedicated questline (iffy, not sure if it's supported by canon games).

**Screenshots:**

Ultima1990:
Just gonna point out, don't count on getting double free slots every region going forward. That's gonna get out of hand real quick. I don't even know what would happen if players managed to slot their entire roster.

unclebanks:
There's like almost a thousand mon. I don't see it becoming an issue with Galar being the latest region. Also, the more you have in queue the less attack you have.

PixLSteam:
I already prepared some code for this, but I'm not quite sure where to put it. Does it deserve its own file? Anyway, here's my current implementation (feedback is welcome): 
```ts
function fillHatcheryQueue() {
	let index: number = 0;
	while (App.game.breeding.hasFreeQueueSlot() && index < App.game.party.caughtPokemon.length) {
		let pokemon: PartyPokemon = App.game.party.caughtPokemon[index];
		if (BreedingController.visible(pokemon)()) {
			App.game.breeding.addPokemonToHatchery(pokemon);
		}
		index++;
	}
}
```

PR #1435, opened 2021-06-13
Fill Queue Button

Mephistic:
## Problem

In later regions, it takes dozens of clicks to fill your hatchery queue for attack grinding.

## Solution

This PR adds a "Fill Queue" button to the hatchery that will attempt to fill the hatchery queue with all breedable pokemon that match the current filters, prioritized by sort order.

This option only becomes available once the player has at least 16 queue slots because I felt that was the point where the needed clicks started getting annoying.

Screen Shot 2021-06-13 at 1 03 28 PM

## Related Issues

https://github.com/pokeclicker/pokeclicker/issues/1179
https://github.com/pokeclicker/pokeclicker/issues/1405


IdlerDude:
I am happy to see that this issue is still open. I finally broke down and got a github id just to add this suggestion, but it's already here :) I play this game basically as it was designed - no mouse macros, no auto-clickers, etc... The amount of "required" clicking in the rest of the game is warranted, focused, and the player has discretion when and if they want to do it. This is not the case for filling up the hatchery queue. It is a mindless task of placing the desired eggs into the queue. There is nothing special from the first to last click EVERY time a person wishes to fill the queue. This is why a "Fill Queue" button greatly increases the playability of the game. In my mind, it is a one time replacement of 16 or 32 continuous clicks. It is not a request to do it on a continual basis. 

If the devs wish, they can optionally tie this to some event or quest line that requires players to fork out money or effort to show they know how to play and this rewards them with some benefit.

One factor to consider is that there is a fairly strong rule to not talk about automation tools in discord. The lack of some kind of "Fill Queue" button will push even die-hards like myself to look for one... it's only a matter of time before I get fed up of continually doing 32 continuous clicks to fill up the queue. Should the queue increase in size,   without such a feature... that would just simply be crazy.

A minor point: As I stated in 1724 - this issue has nothing to with that one. My humble opinion of course.

JapuDCret:
bump

Ultima1990:
> bump

Don't do that.

amative1:
Per Hatchery Helpers ( #2053 ), no longer necessary.

PR #1437, opened 2021-06-20
[FIX] Underground size upgrade breaking the current layer

PixLSteam:
I replaced several calls to `Underground.getSizeY()` with calls to a new function `Mine.getHeight()`, which simply reads the height of the current `rewardGrid`. I did some testing myself and it seemed to work fine, but feel free to try it out yourself - maybe I missed something.


RedSparr0w (approved):
Nice work, Thanks for the PR! 🎉

Issue #1368, opened 2021-04-28
[SUGGESTION] Mystery egg shop in Unova


pipamoo:
**Current Behavior**
There is currently no shop that sells Mystery Eggs in Unova.

**Improved Behavior**
I would like to see a shop in Unova selling Mystery eggs

**Reason for change**
I have not been to Kalos yet, but from the wiki, I have noticed that every region except Unova has a shop that sells mystery eggs.

**Detailed Description**
Adding Mystery eggs to a shop in Unova makes it streamlined with the other regions. I have also noticed a pattern in what can be bought at shops. Pokeballs tend to be the first thing purchasable in the region, being the only thing the starter town's shop sells. The general pattern continues with Mystery Eggs and Greatballs. The Mystery Egg tends to be sold at a shop with a gym and Aspertia, the Unova starter town is also a gym. Following with this pattern, Virbank should sell Mystery Eggs as well, but it is already selling Greatballs. I think it would be a good idea to break tradition and sell Mystery Eggs in Floccesy, being where in canon, former Champion Alder lives.

RedSparr0w:
closed via #1440

PR #1440, opened 2021-06-28
[ADD] Mystery egg shop in Unova #1368

switchlove:
Created the FloccesyTownShop const with the Mystery egg item and added it to Floccesy Town.


RedSparr0w (approved):
👍 Thanks for the PR!

PR #1444, opened 2021-07-02
[FIX] 12 Oak Items achievement unobtainable for No Click Attack challenge

PixLSteam:
This change should properly mark the achievement as unobtainable if the No Click Attack challenge is activated. Please note that I could only test it with new saves, but it should also work with older ones.


RedSparr0w:
Maybe instead of removing the achievement, we could make a variable to track how many oak items are being taken off and reduce it by that many (in this case only 1).

something like this:
_pseduo code_
```js
		let maxOakItems = 12;
		maxOakItems -= (+challenges.list.disableClickAttack.active());
        AchievementHandler.addAchievement('Prepared for anything!', `Level ${maxOakItems} Oak Items to the maximum level`, new MaxLevelOakItemRequirement(maxOakItems), 0.18, GameConstants.Region.none, () => !challenges.list.disableOakItems.active());

```

PixLSteam:
> Maybe instead of removing the achievement, we could make a variable to track how many oak items are being taken off and reduce it by that many (in this case only 1).
> 
> something like this:
> _pseduo code_
> 
> ```js
> 		let maxOakItems = 12;
> 		maxOakItems -= (+challenges.list.disableClickAttack.active());
>         AchievementHandler.addAchievement('Prepared for anything!', `Level ${maxOakItems} Oak Items to the maximum level`, new MaxLevelOakItemRequirement(maxOakItems), 0.18, GameConstants.Region.none, () => !challenges.list.disableOakItems.active());
> ```

Unfortunately, the save has not been loaded yet when achievements are being initialized, so we don't have access to that information. Is there anything else we can do?

RedSparr0w:
Ah, maybe instead we could do something like:
```js
        AchievementHandler.addAchievement('Prepared for anything!', 'Level 11 Oak Items to the maximum level', new MaxLevelOakItemRequirement(11), 0.18, GameConstants.Region.none, () => !challenges.list.disableOakItems.active() && challenges.list.disableClickAttack.active());
        AchievementHandler.addAchievement('Prepared for anything!', 'Level 12 Oak Items to the maximum level', new MaxLevelOakItemRequirement(12), 0.18, GameConstants.Region.none, () => !challenges.list.disableOakItems.active() && !challenges.list.disableClickAttack.active());
```

PixLSteam:
I added the alternative achievement, but it might break the achievement tracker, since it tracks the achievement returned by `findByName`. Should we include a check for `achievement.achievable()`? Similarly, having hidden achievements that are in fact achievable might provide the achievement bonus, maybe `achievement.isCompleted` should also check `achievable()`.

RedSparr0w:
We could also add a function that runs once the save data is loaded which removes any non achievable achievements from the list entirely, that way it shouldn't ever have the same achievement twice.
Also frees up a tiny bit of memory.

RedSparr0w (approved):
👍 Thanks for the PR, Looks good!

PR #1451, opened 2021-07-11
region and type correction for 1 each

ValorBeastFlame:
??? to alola, and galar meowth to steel from nromal


RedSparr0w (approved):

PR #1461, opened 2021-07-24
[FIX] Obtaining key item by badge requirement

PixLSteam:
Reloading the page while on certain "Gym Badge obtained" popups would prevent the player from gaining the corresponding key items. This change should add the item silently while delaying the item popup.
I was only able to test the Safari ticket, not the Shard case.
I'm not sure if an update function for broken saves is needed, but feel free to request it.


Askellown:
can confirm after loading an old challenge save this also applies to Gio and the shard case

RedSparr0w (approved):

RedSparr0w:
> I'm not sure if an update function for broken saves is needed, but feel free to request it.

Yeah, it's probably worth adding one, just in case any people not in the Discord encountered this problem too, and may not have ever noticed 😅

PixLSteam:
I implemented the update function.

RedSparr0w:
Looks great thanks 👍

PR #1467, opened 2021-07-29
Adding player offline earnings

Symi001:
(money only atm)


src/scripts/Game.ts
```diff
@@ -87,9 +87,36 @@ class Game {
...
+        this.computeOfflineEarnings();
+
         this.gameState = GameConstants.GameState.fighting;
     }
 
+    computeOfflineEarnings() {
+        const now = Date.now();
+        const timeDiffInSeconds = (now - player._previousStartLastSeen) / 1000;
+        if (timeDiffInSeconds > 1) {
+            const timeDiffOverride = timeDiffInSeconds > 86400 ? 86400 : timeDiffInSeconds;//The maximum value is 24h
```
RedSparr0w, line 99:
```suggestion
            // Only allow up to 24 hours worth of bonuses
            const timeDiffOverride = Math.min(86400, timeDiffInSeconds);
```

src/scripts/Game.ts
```diff
@@ -87,9 +87,36 @@ class Game {
...
+            const maxHealth: number = PokemonFactory.routeHealth(route, region);
+            let hitsToKill = 0;
+            for (const pokemon of availablePokemonMap) {
+                const type1: PokemonType = pokemon.type[0];
+                const type2: PokemonType = pokemon.type.length > 1 ? pokemon.type[1] : PokemonType.None;
+                const attackAgainstPokemon = App.game.party.calculatePokemonAttack(type1, type2);
+                const currentHitsToKill: number = maxHealth / attackAgainstPokemon;
+                hitsToKill += currentHitsToKill;
+            }
+            const numberOfPokemonDefeated = (timeDiffOverride / 0.9) / hitsToKill;//We attack every 0.9 secs, so we have to divide by 0.9
```
RedSparr0w, line 112: We attack every 1 second, it's just limited to 900ms in the code as a max speed.
```suggestion
            const numberOfPokemonDefeated = timeDiffOverride / hitsToKill;
```

src/scripts/Player.ts
```diff
@@ -29,6 +29,7 @@ class Player {
         const saved: boolean = (savedPlayer != null);
         savedPlayer = savedPlayer || {};
         this._lastSeen = savedPlayer._lastSeen || 0;
+        this._previousStartLastSeen = savedPlayer._lastSeen || 0;
```
RedSparr0w, line 32: I think instead of adding this, we should just not bother updating the `_lastSeen` variable to the current time in the constructor below.

As it's already updated on save anyway, and not used anywhere else.

src/scripts/Game.ts
```diff
@@ -87,9 +87,36 @@ class Game {
...
+            const currentMoney: number = PokemonFactory.routeMoney(player.route(), player.region);
```
RedSparr0w, line 113: May be worth updating `PokemonFactory.routeMoney` to include an variable that doesn't include a random modifier to the amount you receive.

Also maybe rename the variable from `currentMoney` → `routeMoney`

src/scripts/Game.ts
```diff
@@ -87,9 +87,36 @@ class Game {
...
+            const baseMoneyToEarn = numberOfPokemonDefeated * currentMoney;
+            const moneyToEarn = baseMoneyToEarn * 0.5;//Debuff for offline money
+            App.game.wallet.gainMoney(moneyToEarn);
```
RedSparr0w, line 116: Will need to add a variable to not include bonuses in the amount received, as if people had battle items running then those bonuses would still be counted without using up any of the timer.

src/scripts/Game.ts
```diff
@@ -87,9 +87,36 @@ class Game {
...
+
         this.gameState = GameConstants.GameState.fighting;
     }
 
+    computeOfflineEarnings() {
+        const now = Date.now();
+        const timeDiffInSeconds = (now - player._previousStartLastSeen) / 1000;
+        if (timeDiffInSeconds > 1) {
+            const timeDiffOverride = timeDiffInSeconds > 86400 ? 86400 : timeDiffInSeconds;//The maximum value is 24h
+            const route: number = player.route();
```
RedSparr0w, line 100: I think we should add some logic here to be idling the current regions first route if they weren't on a route when they quit the game.

src/scripts/Game.ts
```diff
@@ -87,9 +87,36 @@ class Game {
...
         RoamingPokemonList.generateIncreasedChanceRoutes(now);
 
+        this.computeOfflineEarnings();
+
         this.gameState = GameConstants.GameState.fighting;
     }
 
+    computeOfflineEarnings() {
+        const now = Date.now();
+        const timeDiffInSeconds = (now - player._previousStartLastSeen) / 1000;
```
RedSparr0w, line 97: Just to have a nice round number
```suggestion
        const timeDiffInSeconds = Math.floor((now - player._previousStartLastSeen) / 1000);
```

RedSparr0w (changes requested):

src/scripts/Game.ts
Symi001: Is this a safety if the player's ctor changes ? It's handled by the player ctor, meaning it should never be invalid. I can make the change though ^^ Would you rather I turned the code in the player ctor into a free function or duplicate it for the time being ?

RedSparr0w: I'm not sure what you mean by `ctor` sorry 😅.
The `player.route()` value changes to `0` if the player last saved in a town.
If this is the case, we should assume they idled the first route of the region, rather than no route.

Symi001: Sorry ^^" ctor = constructor.
In the player  `constructor(savedPlayer?) ` we switch the 0 value to a correct one (if needed), and my code should be executed after (as I'm using the _lastSeen that is updated some lines before). That's why I wanted to know if you wanted me to add the security you were talking about nonetheless ;) 
Sorry my comment lacked clarity :/

RedSparr0w: I don't think we need to change the actual value that's being saved, just calculate the start route here if they aren't on a route.
Otherwise it will affect other aspects of the game.

Symi001: I'm not talking about changing the saved value but I'm talking about the code starting line 45 of Player.ts 
If I understand correctly, as soon as the player is constructed, its route and region are switched to a "correct" one. Meaning, when the part described in the comment will be executed, the route should always be correct. I've made the change though, I'll update the review soon ;)

RedSparr0w (approved):

RedSparr0w:

PR #1473, opened 2021-07-30
[HOTFIX] Kalos gyms have much higher HP than intended

anapple96:
Self-explanatory title I guess


RedSparr0w (approved):

PR #1476, opened 2021-07-30
Last Vivillon not showing up

NiCeDiCe90:
This SHOULD fix the Problem with the Last Vivi not showing up in the BF as intended.


RedSparr0w (commented):
This won't change anything, as the `ViviBallAdd` isn't used inside of the `catchNormal` quest, it's only used after it.

PR #1482, opened 2021-07-31
Fixes #1433

### LINKED ISSUE #1433: [BUG] Incorrect Kalos Route Order
**Info**
 
Version: 0.8.2

**Description:**

Kalos Route 22 is unlocked after clearing Kalos Route 3 and should be the 4th ordered route in Kalos. However, it is not and is instead the 23rd ordered Route in Kalos

**Expected Behavior:**

Kalos Route Order should go 1->2->3->22->4->5...etc.

**Additional Information:**

Additional Information: Fixing this issue could make some Kalos Dungeons slightly unbalanced with respect to the Closest Route difficulty.

(Reopening issue with correct tag)

jk13pclick:
Hopefully I did this right this time! Fixes the Kalos Route Order.

### DIFF
--- a/src/scripts/wildBattle/Routes.ts
+++ b/src/scripts/wildBattle/Routes.ts
@@ -1341,6 +1341,15 @@ Routes.add(new RegionRoute(
     [new ClearDungeonRequirement(1, GameConstants.getDungeonIndex('Santalune Forest'))]
 ));
 Routes.add(new RegionRoute(
+    'Kalos Route 22', GameConstants.Region.kalos, 22,
+    new RoutePokemon({
+        land: ['Psyduck', 'Farfetch\'d', 'Azumarill', 'Dunsparce', 'Azurill', 'Bidoof', 'Bibarel', 'Riolu', 'Bunnelby', 'Diggersby', 'Litleo'],
+        water: ['Magikarp', 'Goldeen', 'Azumarill', 'Carvanha'],
+    }),
+    [new RouteKillRequirement(10, GameConstants.Region.kalos, 3)],
+    3.1
+));
+Routes.add(new RegionRoute(
     'Kalos Route 4', GameConstants.Region.kalos, 4,
     new RoutePokemon({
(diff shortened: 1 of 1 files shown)


Ultima1990:
I'm not exactly sure what you are trying to do here.

First off, route 22 can be accessed this early in pokémon X and Y. It's exactly the same situation as route 22 in Kanto.
Second, just moving the route down in the file and removing the 3.1 isn't going to change its unlock requirements, which you still have set as clearing route 3. The current result of this pr will be the route still unlocks after clearing route 3, but its difficulty is now higher than route 21.

Basically I don't think this route needs fixing, and this fix doesn't fix anything.

Ultima1990:
> I'm not exactly sure what you are trying to do here.
> 
> First off, route 22 can be accessed this early in pokémon X and Y. It's exactly the same situation as route 22 in Kanto.
> Second, just moving the route down in the file and removing the 3.1 isn't going to change its unlock requirements, which you still have set as clearing route 3. The current result of this pr will be the route still unlocks after clearing route 3, but its difficulty is now higher than route 21.
> 
> Basically I don't think this route needs fixing, and this fix doesn't fix anything.

Wait, no, I got that backwards. The route moved up and the 3.1 was _added_. Don't mind me.

RedSparr0w:
> > I'm not exactly sure what you are trying to do here.
> > First off, route 22 can be accessed this early in pokémon X and Y. It's exactly the same situation as route 22 in Kanto.
> > Second, just moving the route down in the file and removing the 3.1 isn't going to change its unlock requirements, which you still have set as clearing route 3. The current result of this pr will be the route still unlocks after clearing route 3, but its difficulty is now higher than route 21.
> > Basically I don't think this route needs fixing, and this fix doesn't fix anything.
> 
> Wait, no, I got that backwards. The route moved up and the 3.1 was _added_. Don't mind me.

Glad you spotted it, looked exactly the same to me 😅, that's the problem when it's been moved you can't really see what exactly changed.

RedSparr0w (approved):
Thanks for the PR, looks good.

PR #1483, opened 2021-07-31
Implements #1392 - Adding an NPC that lets you know how many Pokemon are left in a regio…

### LINKED ISSUE #1392: Add an NPC that lets you know how many Pokemon are left in a region
Add an NPC to a town that lets you know how many more Pokemon there are to discover are catch in your current region.

**Current Behavior**
This is ambiguous with no in game way to check without having already encountered all Pokemon in the region. 

**Improved Behavior**
You go to a town, presumably the one where the regions professor is based and there is an NPC who lets you know how many Pokemon are left in the region. Either to catch or to discover, the exact implementation of that can be decided by the devs. 

**Reason for change**
I recently ran into an issue in Unova where I hadn't seen a Pokemon but had no Idea that the Pokedex only showed encountered Pokemon. I also noticed a user in the discord had the exact same issue for the Hoenn Pokedex.
Ultimately this is to help players wrap up the last few things in a region.

**Detailed Description**
```
let caughtPokemonCount = new Set(App.game.party.caughtPokemon.filter(p => p.id > 0 && PokemonHelper.calcNativeRegion(p.name) <= player.highestRegion()).map(p => Math.floor(p.id))).size;
let remainingPokemon = GameConstants.TotalPokemonsPerRegion[player.highestRegion()] - caughtPokemonCount;
```
An NPC in a town should be able to at least give you a clue as to how many more Pokemon remain in a region, either to be discovered or caught. This probably should only be unlocked at a late stage in the region, possibly if one of the following conditions are met:
 - trainer has become the region master
 - all route Pokemon have been caught
 - all roaming Pokemon have been caught

benanderson1:
closes #1392 

Added a new NPC to each region's starting town.

Once you are the region's champion, he will tell you how many pokemon are left to catch in that region.

Feel free to post suggestions (different names, different text)

Thanks


RedSparr0w:
It may be worth just updating `ProfOakNPC` as it already serves a similar function.

benanderson1:
> 
> 
> It may be worth just updating `ProfOakNPC` as it already serves a similar function.

Can do this, but this would make the player go back to Kanto to ask Prof. Oak how they are doing in future regions.

If that is acceptable (or if we just add Oak to every region), let me know and I'll update

RedSparr0w:
We could add Oak to Every region, or change it to `ProfNPC` and give each regions their respective professor names?

benanderson1:
Sorry there are no tracked changes on ProfOakNPC -- I used Ctrl+R to rename but I noticed renaming the class doesn't automatically rename the file, I did that manually, so rip.

Images are not working. I left the reference to ProfOak's image in comments but I think we would need images for the other Professors to add it as a parameter. For now the field shows undefined in game. I'm really not good with HTML so please feel free to tweak it so it's prettier.

PR #1486, opened 2021-07-31
[FIX] Show item count for non-UG items

PixLSteam:
Tested with today's water stone trade:
image
Feel free to tell me if you'd rather want to modify `player.getUndergroundItemAmount()` or add a new function.

### DIFF
--- a/src/components/underground.html
+++ b/src/components/underground.html
@@ -174,7 +174,7 @@
                                     </td>
                                     <td class='vertical-middle' data-bind='text: amount1 + " × " + item1.name'></td>
                                     <td class='vertical-middle'>→</td>
-                                    <td class='vertical-middle text-center' data-bind='text: player._itemList[Underground.getMineItemById(item2.id).valueType]?.() ?? player.getUndergroundItemAmount(item2.id)'></td>
+                                    <td class='vertical-middle text-center' data-bind='text: player.getUndergroundItemAmount(item2.id)'></td>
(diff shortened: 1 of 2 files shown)


src/components/underground.html
```diff
@@ -174,7 +174,7 @@ <h4><u>Daily deals</u></h4>
                                     </td>
                                     <td class='vertical-middle' data-bind='text: amount1 + " × " + item1.name'></td>
                                     <td class='vertical-middle'>→</td>
-                                    <td class='vertical-middle text-center' data-bind='text: player.getUndergroundItemAmount(item2.id)'></td>
+                                    <td class='vertical-middle text-center' data-bind='text: player._itemList[Underground.getMineItemById(item2.id).valueType]?.() ?? player.getUndergroundItemAmount(item2.id)'></td>
```
RedSparr0w, line 177: I think we may need to update the `getUndergroundItemAmount` function to check stones amount instead, as we have had problems with peoples browsers not quite yet supporting `??` and `?.` in the past.

We are able to use those in typescript because it converts them to something like:
`foo = bar != undefined ? bar : other;`
But when it's in the HTML/knockout, it doesn't get converted.

PixLSteam: I had a feeling this could be an issue. I modified the `getUndergroundItemAmount` function, I'll push it in a few minutes.

RedSparr0w (approved):
Looks great, thanks for the PR! 🎉

PR #1496, opened 2021-08-03
Alola Maps draft

NiCeDiCe90:
I will add/edit Alola Maps and upload them here for feedback.
Thanks in advance. :)

Step1: Melemele


RedSparr0w:
Only feedback so far, maybe name the maps something like `alola-melemele.png` so it's easier to know which map is which.

RedSparr0w:
There seems to be 1 row of ocean missing up the top?
Which also seems to make the overlays appear slightly off too 🤔 
image

Not sure if it's just me, but these 5 houses feel out of place without any interactions on them:
image

Something also feels a little off with these 3 dungeons, maybe just the top one, not really sure what though 🤷 
image

Would it be worth extending these routes/dungeons to the beaches just so they have some coverage?:
image

RedSparr0w (approved):

PR #1500, opened 2021-08-04
Fix of LevelType

### LINKED ISSUE #1499: [BUG] Some pokemon have incorrect levelType
**Info**
Several pokemon in PokemonList.ts have a levelType of `4` instead of the expected `LevelType.something`. All cases appear like they should be `LevelType.fast`, but `4` is actually the "slow" level type.

**Steps to Reproduce:**

1. Go to PokemonList.ts
2. Ctrl+f `'leveltype': 4`

**Expected Behavior:**
`levelType` key should be `LevelType.something`, probably `LevelType.fast` in all cases, but double check with bulbapedia.

jk13pclick:
Changed all 4s to LevelType.fast as per Bulbapedia, fixes #1499


jk13pclick:
Closing because duplicate of #1501

jk13pclick:
Reopening :)

NiCeDiCe90:
This is comedy that can't be scripted. 👍

RedSparr0w (approved):
Thanks for the PR!, Looks good.

Issue #1003, opened 2020-10-24
Remember to change some evolution methods for Alola.


Ultima1990:
Crabrawler is a location based evolution in Mount Lanakila.
Charjabug is a location based evolution in Vast Poni Canyon.
Poipole is a move based evolution, based on Dragon Pulse, which it learns at level 1. This is way too early for an Ultra Beast, and I don't know if the game even supports level 1 evolutions at this moment. Will need to think of something for this.
Meltan only evolves in Pokémon GO by feeding it 400 Meltan Candy. Will need to think of some way of making this work in Clicker.

RedSparr0w:
@Ultima1990 Is this still an issue?

Ultima1990:
Yes. Crabrawler and Charjabug are easily done just like any other location evo, just not done yet. Nothing has been decided for Poipole and Meltan. Also the Lycanroc Midday evolution has not been fully implemented yet.

RedSparr0w:
Should Meltan be Alolan or Galarian?

Ultima1990:
Meltan is Alolan as far as I'm concerned. Kind of a weird situation, a gen 7 mythical that wasn't obtainable at all in gen 7, but what can you do. Could always move it to gen 8 if that's more convenient. Either way it doesn't have an obvious obtaining or evolution method we can use for clicker.

amative1:
I'm also seeing Steenee --> Tsareena as having no evo path, though that's less of an issue as Tsareena can be separately obtained. Meltan/Melmetal are sort of a Gen 7+, since they were part of PoGo, so I can see holding them back to give time for discussion about how to implement them (there's a thread on the Discord discussing it).

Ultima1990:
Closing this. Meltan is the only one that's not solved yet and I'll leave that one to the greater Meltan obtaining project.

PR #1504, opened 2021-08-05
Alola evolutions update

amative1:
Per #1003, updating Charjabug -> Vikavolt (as location-sensitive), Crabrawler -> Crabominable (as location-sensitive), Steenee -> Tsareena (as level-sensitive), and Poipole -> Naganadel (as level-sensitive).

Still pending is Meltan -> Melmetal, though this is not Alola-launch critical.


src/scripts/pokemons/PokemonList.ts
```diff
@@ -15865,7 +15865,10 @@ const pokemonList = createPokemonArray(
         'levelType': LevelType.mediumfast,
         'exp': 140,
         'catchRate': 120,
-        'evolutions': [new StoneEvolution('Charjabug', 'Vikavolt', GameConstants.StoneType.Thunder_stone)],
+        'evolutions': [
+            new StoneEvolution('Charjabug', 'Vikavolt', GameConstants.StoneType.Thunder_stone),
+            new DungeonRestrictedLevelEvolution('Vast Poni Canyon','Charjabug','Vikavolt', 20),
```
RedSparr0w, line 15870: So `Charjabug` can be evolved via either method?

amative1: That's the idea. The stone is a Gen 8 method that was already there, I just added the Gen 7 "exposure to a special magnetic field" rule.

Ultima1990: I don't agree with having stone evolution as an option for Charjabug.

Askellown: While dex entry is restricted by region, I agree it should be like Leafeon and restricted to area, but when it stops being restricted I believe the stone should be fine.

amative1: Looking at the reqs on Bulbapedia, I'll be re-doing that portion later today. Closing the pull request for now, I'll resubmit when it's done.

PR #1505, opened 2021-08-05
Add an s to some achievements

Penguindude2000:
Added an s to the plural quest achievements


RedSparr0w (changes requested):
You will also need to add something to the Update.ts script for the new achievements.

RedSparr0w (approved):
nvm, description update doesn't need anything 😅 

👍 Looks good, thanks for the PR! 🎉

PR #1509, opened 2021-08-05
Alola evolutions update

amative1:
Third time's the charm!

Per #1003, updating Charjabug -> Vikavolt (as location-sensitive, removing stone restriction), Crabrawler -> Crabominable (as location-sensitive), Steenee -> Tsareena (as level-sensitive), and Poipole -> Naganadel (as level-sensitive).

Still pending is Meltan -> Melmetal, though this is not Alola-launch critical.


RedSparr0w (approved):

PR #1513, opened 2021-08-06
Decrease Shop Price Multiplier

### LINKED ISSUE #1466: [SUGGESTION] Possible Bug, But Pokemart Item Price Reset Issue
So, I've noticed that Gyms & Elite 4 don't provide any progress toward Poke Mart item price resets. If it's not a bug, well I'd like to make a suggestion that it be changed to do so.

jk13pclick:
Fixes the lack of a Multiplier Decrease in Gyms, Dungeons, and Battle Frontier. Also decreased the timer on the Battle Frontier notification box because 5 minutes is way too long. Fixes #1466 

Note: The Battle Frontier multiplier decrease might be a little too quick for repeat battles as it is with every Pokémon defeated

### DIFF
--- a/src/scripts/battleFrontier/BattleFrontierBattle.ts
+++ b/src/scripts/battleFrontier/BattleFrontierBattle.ts
@@ -56,6 +56,7 @@ class BattleFrontierBattle extends Battle {
         } else {
             this.enemyPokemon(null);
         }
+        player.lowerItemMultipliers(MultiplierDecreaser.Battle);
     }
 
     public static generateNewEnemy() {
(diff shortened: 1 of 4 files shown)


src/scripts/battleFrontier/BattleFrontierBattle.ts
```diff
@@ -56,6 +56,7 @@ class BattleFrontierBattle extends Battle {
         } else {
             this.enemyPokemon(null);
         }
+        player.lowerItemMultipliers(MultiplierDecreaser.Battle);
```
RedSparr0w, line 59: Battle frontier shouldn't decrease normal shop prices, just the BF shop prices.

jk13pclick: Good point, I reverted the battle frontier change as the shop already has the price reduction in the town.ts file.

RedSparr0w (approved):

Issue #1366, opened 2021-04-28
[SUGGESTION] Provide NPC/telegraphs for obscure mechanics


IanWalder:
**Current Behavior**
Assuming gameplay does _not_ require actively using Discord or the Wiki, certain obscure game mechanics are not hinted at in the game itself. This is particularly prominent in Kalos.

**Improved Behavior**
Most of the below mechanics can be telegraphed or at least hinted to the player through NPCs.

**Reason for change**
Other new mechanics (e.g. how to unlock Celebi and Ho-oh in Johto, how to obtain Leafeon and Magnezone in Sinnoh) are at the very least hinted at via NPCs, and these should be no different, for consistency's sake if nothing else.

**Detailed Description**
To my knowledge, the below gameplay aspects aren't adequately represented in-game.

- [X] The fact that Sligoo and Cherrim's evolutions are locked behind weather
- [X] Needing to beat Volcanion's dungeon multiple times to unlock it
- [X] Certain Flabebe forms being exclusive to the Farm (I'm aware there's a reference to wandering Pokemon in the Help section, but nothing even hints as to which Berries attract which Pokemon. Perhaps add a Kalos Berry Master for this very purpose? Note that all other wandering Pokemon can be obtained outside the farm)

Please feel free to correct me in the comments if these are accounted for in the game. This list may be updated as time goes on.

Namadu7:
I hugely support this, even for non-evolution-related mechanics. There are a lot of mechanics that you wouldn't know except by looking at the Wiki or hearing about them in the Discord - things like the Item Magnet increasing held item drop rates, some of the more specific Berry requirements, maybe even the fact that drop/Shiny rates are higher in dungeons. 

Would also make it more exciting to reach new towns - never know what new thing you'll learn next!

PR #1515, opened 2021-08-08
Lots of NPCs

jk13pclick:
Addresses #1366. Added a whole bunch of new NPCs. More can be added, just thought this was a good start. Let me know if the tone/verbiage of the NPCs needs refining!

### DIFF
--- a/src/scripts/towns/Town.ts
+++ b/src/scripts/towns/Town.ts
@@ -691,7 +691,7 @@ const LittlerootAide = new NPC('Professor Birch\'s Aide', [
 ]);
 
 const OldaleTrackingScientist = new NPC('Tracking Scientist', [
-    'Tracking Scientist - Hey trainer, look at these footprints! Huh, I’ve never seen footprints like these before…They look like they came from two different Pokémon, and I saw two blurs, one red and one blue, quickly fly away just as I exited the Pokémon Center.',
+    'Hey trainer, look at these footprints! Huh, I’ve never seen footprints like these before…They look like they came from two different Pokémon, and I saw two blurs, one red and one blue, quickly fly away just as I exited the Pokémon Center.',
     'They were flying really fast, I bet Pokémon that fast will only challenge trainers who have proven they are as strong as Champion Wallace...',
 ]);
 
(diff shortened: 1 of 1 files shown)


src/scripts/towns/Town.ts
```diff
@@ -639,22 +683,40 @@ const HoennBerryMaster = new Shop([
...
     'You can now store an additional four eggs in the queue! His research has really gained speed.',
     'He wants you to know that he will have an additional eight slots ready by the time you reach Sinnoh.',
 ]);
 
-const SootopolisWallace = new NPC('Gym Leader Wallace', [
-    'The creators of the lands and ocean slumber within the Cave of Origin.',
-    'However, they will only awaken when in the presence of a truly great trainer.',
-    'You will have to overcome the Pokémon League before you have any chance to encounter them.',
+const OldaleTrackingScientist = new NPC('Tracking Scientist', [
+    'Tracking Scientist - Hey trainer, look at these footprints! Huh, I’ve never seen footprints like these before…They look like they came from two different Pokémon, and I saw two blurs, one red and one blue, quickly fly away just as I exited the Pokémon Center.',
```
RedSparr0w, line 694: Doesn't need the name as already shows.
```suggestion
    'Hey trainer, look at these footprints! Huh, I’ve never seen footprints like these before…They look like they came from two different Pokémon, and I saw two blurs, one red and one blue, quickly fly away just as I exited the Pokémon Center.',
```

RedSparr0w (approved):

src/scripts/towns/Town.ts
```diff
@@ -995,15 +1060,24 @@ const SinnohBerryMaster = new Shop([
 ]);
 
 //Sinnoh NPCs
+
+const TwinleafContestChampion = new NPC('Contest Champion', [
+    'Welcome to Sinnoh! There are many legends and myths here. For example, it is said that trainers that conquer the Lake nearby, Lake Verity, will meet a mythical Pokémon known for Emotion roaming around the region. It sure would have been awesome to partner with that Pokémon in one of my routines!',
```
Ultima1990, line 1065: Do we want to keep the distinction between legendary and mythical? Mythical being an event-only legendary Pokémon. By that definition Mesprit is not a mythical.

PR #1518, opened 2021-08-08
Bunch of shops in Alola

Ultima1990:


RedSparr0w (approved):
Looks good to me, thanks for the PR.

Did you want to also update the cost for `Type: Null` and `Poipole` in this PR or separately?

Ultima1990:
Those shop prices are probably fine as they are. Unless someone else has opinions on them.

PR #1519, opened 2021-08-11
Pledge Grove Fix

jk13pclick:
Pledge Grove was super underpowered relative to when it is unlocked. I made it roughly equivalent in difficulty to the Moor of Icirrus which is a pre-req for unlocking the dungeon. Ultima liked this change better so I reverted it back to this one.

~With help from Dev-chat I made a better change. Now Pledge Grove is unlocked with Floccesy Town and is an "optional" dungeon. Two Sawsbuck forms (and the Deerlings) can be obtained early, which is a good reward for early Unova. Keldeo Resolute is a dungeon boss still locked behind Keldeo standard form in the Moor of Icirrus.~


RedSparr0w (approved):

PR #1528, opened 2021-08-17
buffing "CapturePokemonTypesQuest"

NiCeDiCe90:
Lowered the amount of required Pokemon to catch by 50% because some types take forever to catch if you roll a 400+.
Buffed the base rewards by 20% so it's still worth it.


RedSparr0w (approved):

PR #1531, opened 2021-08-21
Updated Berries with Kalos/Alola/Galar starters to match other regions

amative1:
Added starters from later regions to Occa, Passho, and Rindo to match availability with the starters from earlier regions.


RedSparr0w (approved):

PR #1548, opened 2021-09-01
Decrease dungeon size for every 10, 100, 1000 etc defeats to a minimum of 5x5

RedSparr0w:
Decrease the dungeon size for every 100 clears (to a minimum of 3x3).
Show all chest when (floored) dungeon size / 3 chest have been opened
Show entire dungeon when (ceiling) dungeon size / 2 chest have been opened


scolpitts536:
Tbh I'm not a fan of 3x3 dungeons, one kanto becomes a cake walk easily. I feel like 5x5 minimal should clear any dungeon with ease but not be so easy babies can do it.

anapple96:
Same, 3x3 too small, 5x5 min is probably best

RedSparr0w:
There's currently a vote in the Discord for sizes/method of size reduction.

RedSparr0w:
As per the current poll in Discord have updated to be minimum size of `5 x 5`, and decrease in size once cleared 10, 100, 1000 etc times.

GarrikFel:
It makes me sad that the log_10 option won the poll.  I wonder how many of those people are actually doing dungeons.  When we get to Galar I was hoping to not spend most of my dungeon time in a 10x10 dungeon...

Claygame:
reducing size on 10 & 100 clears is ok but next reduction needing 1000 clear is not useful. This feature is so that people can reach 1000 clear in later region dungeons easily where they are so big that reduction of size just 2 times wont really make it any easier/faster to get the 1000 clears.

GarrikFel:
It's still technically useful if you're shiny hunting.  There's a good chance that you won't have all the shinies after 1,000 clears, and in gen4+ your egg hatching shinies are still probably lagging behind so that it makes sense to do those dungeons more than 1,000 times.  I agree that the log_10 option isn't super useful, but it is still kind of useful.

PR #1550, opened 2021-09-01
Added some missing trainer sprites and Alola backgrounds

IanWalder:
There are two backgrounds for the dungeon "Nebby" (listed in Dungeon.ts as of 1st Sept 2021) as the corresponding location in the mainline games is ambiguous. The image titled "Altar of the Sunne and Moone" refers to the town / gym.


RedSparr0w (approved):
👍 Looks great, Thanks for the PR! 🎉

PR #1552, opened 2021-09-02
Add setting for highlighting locations with uncaught Pokémon/Shinies

RedSparr0w:
Also made settings menu a little more compact.
image
image


IanWalder:
Shouldn't there be an option for all caught and all shiny?

RedSparr0w:
> Shouldn't there be an option for all caught and all shiny?

"All Caught" would just be "Uncaught Shiny"

"All Shiny" is just transparent.

PR #1557, opened 2021-09-03
Added NPCs for Sliggoo weather evo and Farm-exclusive Flabebes

IanWalder:
Also fixed some spacing in the other NPCs' dialogue

### DIFF
--- a/src/scripts/towns/Town.ts
+++ b/src/scripts/towns/Town.ts
@@ -439,7 +439,7 @@ const CianwoodPhotographyAide = new NPC('Photography Aide', [
 ]);
 
 const MahoganySouvenirShopAttendant = new NPC('Souvenir Shop Attendant', [
-    'We’ve got stuff here nobody else has got! But keep any Item Magnets you have away from the merchandise…especially the RageCandyBars. Keep ‘em outside where they belong, I’ve heard magnets can attract Pokémon with held items more often, and even more so in Dungeons!',
+    'We’ve got stuff here nobody else has got! But keep any Item Magnets you have away from the merchandise… especially the RageCandyBars. Keep ‘em outside where they belong, I’ve heard magnets can attract Pokémon with held items more often, and even more so in Dungeons!',
 ]);
 
 const JohtoRoamerNPC = new RoamerNPC('Pokéfan Trevor', [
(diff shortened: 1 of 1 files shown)


RedSparr0w (approved):

src/scripts/towns/Town.ts
```diff
@@ -1901,7 +1901,18 @@ const KalosRoamerNPC = new RoamerNPC('Hex Maniac Melanie', [
 ], GameConstants.Region.kalos);
 
 const KiloudeConfusedHiker = new NPC('Confused Hiker', [
-    'Whoa! What the- Where am I? How did I get here? Last thing I can remember I was in Reflection Cave when this little Pokémon with hoops threw something at me…Oh you’ve beaten the Pokémon League? Maybe you can find it roaming around the region so you can tame that little prankster. Now how am I gonna get home…',
+    'Whoa! What the- Where am I? How did I get here? Last thing I can remember I was in Reflection Cave when this little Pokémon with hoops threw something at me… Oh you’ve beaten the Pokémon League? Maybe you can find it roaming around the region so you can tame that little prankster. Now how am I gonna get home…',
+]);
+
+const DendemilleGoomyEnthusiast = new NPC('Goomy Enthusiast', [
```
jk13pclick, line 1907: There is already an NPC in Laverre that I intended to hint at Sliggoo's evolution, one of the two NPCs should be removed

IanWalder: Oh, so that’s what that NPC was referring to. I thought it was just a comment on Fairy types, I didn’t make the gloomy -> Goomy connection.
Are you saying delete FurisodeGirl and put GoomyEnthusiast in Laverre instead?

While we’re at it we should discuss evo hint vagueness. To my knowledge, every other hint mentions either the base Pokemon or its evo or both by name except this and Cherrim (overcast) -> (sunny). Should the Pal Park NPC be changed to mention Cherrim by name?

jk13pclick: For the Goomy one that is what I am saying. The gloomy -> Goomy hint is probably a little too vague (even though I am proud of it!). Just replace the NPC because Goomy is found on the route before Laverre, it makes sense logically to put it there.

As for the Cherrim NPC, in my opinion I think it's kinda boring and super "hand-holdy" to tell the player exactly what evolves and where, but that's really up to the devs. Initial Sinnoh NPCs were written that way, so I suppose it's fine to change the Cherrim one to something a little more direct like the others (ie: Mt. Coronet, Sinnoh Eeveelutions)

PR #1568, opened 2021-09-06
Basic auto rebattle gym system

RedSparr0w:
Unlocks after clearing a gym 100 or more times.
Gym leader won't award money when defeated.
Cost is money, amount per battle is double what the Gym Leader would normally award you for winning,
_(not including achievement/battle item bonuses)_.
image
image

Added a `Stop Auto` button which will stop continuously fighting the gym after this battle ends.
image

If you lose the Gym battle, it won't continue auto fighting it, although if you've defeated it 100 times, chances are unlikely of that happening.


GarrikFel:
Quick note about the odds of losing a gym battle after 100 wins.  It's more common than you think.  In Kalos+, it's possible to add so many pokemon to your egg queue that you're no longer able to win gym battles anymore.  It's a good call to have an auto-stop for a loss

Claygame:
> Quick note about the odds of losing a gym battle after 100 wins. It's more common than you think. In Kalos+, it's possible to add so many pokemon to your egg queue that you're no longer able to win gym battles anymore. It's a good call to have an auto-stop for a loss

well thats upto the player to know if they can defeat gym or not when they do auto battle, we cant be spoonfeeding every little thing.

GarrikFel:
> well thats upto the player to know if they can defeat gym or not when they do auto battle, we cant be spoonfeeding every little thing.

So... you want to remove a feature that was already coded in, because humans can make mistakes and we should waste their money?  I'm confused by your statement.

Would it be possible to add a setting to the game's settings so that you would stop autoing when you reach 1,000 clears?  If I still needed a feature like this, I would love to auto while I sleep, and just have the gym done overnight, without doing several thousand extra clears (since they cost money)

Claygame:
well i can only say what i feel, i cant actually put anything in game as i m nit a contributor, hehe
but yeah i just said what i felt was good, how it is done is upto the coders.

RedSparr0w:
> Would it be possible to add a setting to the game's settings so that you would stop autoing when you reach 1,000 clears? If I still needed a feature like this, I would love to auto while I sleep, and just have the gym done overnight, without doing several thousand extra clears (since they cost money)

I think it should still require some management by the user to make sure they don't waste resources themselves.

PR #1571, opened 2021-09-07
[WIP-EXP] Dungeon rework

apple096:
Yes, it's happening, it may not finish, but it's happening


apple096:
If you read the previous message in the Party PR, you'll know I'm taking a break from this as a whole. The same applies here, this is effectively over, unless Jessica decides it's time to make a follow up to that issue more than a year ago, which is unlikely. So, again, it means you can't use this as a reason to delay the dungeons fixes and don't expect this to come at all.
Bye.

Issue #1398, opened 2021-05-10
[SUGGESTION] Achievements I would like to see added.


NiCeDiCe90:
**Current Behavior**
Achievements fpr several Tasks are missing.

**Reason for change**
Achievements can motivate to finish certain tasks. For example: I never even thought about mutating every berry except for the ones I want to use.

**Detailed Description**

**Farm unlocks**
- [x] My new dirty hobby (Farm unlocked)
- [x] allotment gardener (9 Plots unlocked)
- [x] Horticulture (25 Plots unlocked)

**Berry mutating**
- [x] Found them along the way (have all Gen1 Berries)
- [x] Something weird happend to my berry (Have all Gen2 Berries)
- [x] Big Berry Business (Have alle Gen 3 Berries)
- [x] Mutation Madness (Have all gen 4 berries)
- [x] Shiny Miracle berries (have all gen 5 berries)

**finished Quests**
- [x] can you do this for me? (1 quest done)
- [x] just one more favor (10 quests)
- [x] YES MAN (100 quests)
- [x] I just love green coins (1000 quests)

**climbing Battle frontier**
- [x] Fighting Novice (100 Battle Frontier Stages)
- [x] Competent fighter (151 Battle Frontier Stages)
- [x] let them come (251 Battle Frontier Stages)
- [x] a real Master (386 Battle Frontier Stages)
- [x] unstoppable fighting machine (500 Battle Frontier Stages)
- [x] Living Legend (1000 Battle Frontier Stages)

**using Proteines**
- [x] let's try this out (use 1 proteine)
- [x] training progress (use 10 proteine)
- [x] Strong like the mountain  (use 100 proteine)
- [x] An actual mountain of muscles (use 1000 proteine)

**Pokedex related**
I dont have names for these but having unique catches an unique shinys should be moved from general to reginal and added for every region. Should something like:

- [x] 10; 25; 50; 75; 100 unique catched in Johto
- [x] 10; 25; 50; 75; 100 unique shiny in Johto
- [x] 10; 25; 60; 100; 135 unique catched in Hoenn
- [x] 10; 25; 60; 100; 135 unique shiny in Hoenn
- [x] 10; 25; 50; 80; 107 unique catched in Sinnoh
- [x] 10; 25; 50; 80; 107 unique shiny in Sinnoh
- [x] 10; 30; 65; 110; 156 unique catched in Unova
- [x] 10; 30; 65; 110; 156 unique shiny in Unova
[...]

PR #1574, opened 2021-09-08
Updating Region Achievements

### LINKED ISSUE #1322: [SUGGESTION] An achievement  for completing a pokedex
Currently theres an achievement for completing the kanto pokedex but none of the other regions have an achievement for completing the pokedex 

i think i would be better if all the regions had an achievement for completing the pokedex

this would an improvement as currently its weird that only kanto has the achievement and i feel like games can always be in need of new achievements

### LINKED ISSUE #1386: "Caught Unique Pokemons By Region" only having Kanto region even after unlocking other regions [BUG]
I've moved past the Kanto region (Now into Sinnoh) yet still don't have extra achievements unlocked for catching unique Pokemon by each region making it hard to tell when I have enough to move on.

urCe3ZF 1

amative1:
Removing some of the "filler" general achievements for catching unique 'mons, and adding 10%, 50%, and 100% catch achievements for each region.

Weights are 1/2/3 to mirror the route/gym/dungeon breakout, but that can be adjusted if need be.

closes #1386 
closes #1322 
related #1398


RedSparr0w (approved):
👍 Looks great, thanks for the PR! 🎉

PR #1576, opened 2021-09-08
Badge case level limit text

### LINKED ISSUE #433: add badge documentation
in this version, nothing tell you about lvl limit, it may be said when winning badges, and passing cursor over it in badge case

IanWalder:
Draws attention to start menu after Brock is beaten and tooltip with explanatory text added to badge case
Closes #433


RedSparr0w (approved):
Thanks for the PR, Looks good! 🎉

PR #1580, opened 2021-09-09
Protein achievements added

### LINKED ISSUE #1398: [SUGGESTION] Achievements I would like to see added.

IanWalder:
Added new statistic for tracking proteins obtained.
Something weird happened after I added a retroactive check for proteins obtained to Update.ts though:

- Set the version to 0.8.9 temporarily so the new totalProteinsObtained stat gets assigned a value
- Changed it back to ${VERSION} when everything worked (as far as I could see)
- Imported a 0.8.6 file and found that the stat had the correct value

Everything seems correct so far, but could someone confirm? Thanks

Closes #1398

### DIFF
--- a/src/scripts/achievements/AchievementHandler.ts
+++ b/src/scripts/achievements/AchievementHandler.ts
@@ -242,6 +242,11 @@ class AchievementHandler {
         AchievementHandler.addAchievement('Don\'t stop trying', 'Complete 2500 total stages in Battle Frontier', new BattleFrontierTotalStageRequirement(2500), 0.25);
         AchievementHandler.addAchievement('King of the hill', 'Complete 5000 total stages in Battle Frontier', new BattleFrontierTotalStageRequirement(5000), 0.4);
 
+        AchievementHandler.addAchievement('Let\'s try this out', 'Obtain your first Protein', new ProteinObtainRequirement(1), 0.20);
+        AchievementHandler.addAchievement('Training progress', 'Obtain 10 Proteins', new ProteinObtainRequirement(10), 0.30);
+        AchievementHandler.addAchievement('Strong like the mountain', 'Obtain 100 Proteins', new ProteinObtainRequirement(100), 0.40);
+        AchievementHandler.addAchievement('A literal mountain of muscle', 'Obtain 1000 Proteins', new ProteinObtainRequirement(1000), 0.50);
+
         /*
          * REGIONAL
          */
--- a/src/scripts/achievements/AchievementHandler.ts
+++ b/src/scripts/achievements/AchievementHandler.ts
@@ -242,10 +242,15 @@ class AchievementHandler {
         AchievementHandler.addAchievement('Don\'t stop trying', 'Complete 2500 total stages in Battle Frontier', new BattleFrontierTotalStageRequirement(2500), 0.25);
         AchievementHandler.addAchievement('King of the hill', 'Complete 5000 total stages in Battle Frontier', new BattleFrontierTotalStageRequirement(5000), 0.4);
 
-        AchievementHandler.addAchievement('Let\'s try this out', 'Obtain your first Protein', new ProteinObtainRequirement(1), 0.20);
-        AchievementHandler.addAchievement('Training progress', 'Obtain 10 Proteins', new ProteinObtainRequirement(10), 0.30);
-        AchievementHandler.addAchievement('Strong like the mountain', 'Obtain 100 Proteins', new ProteinObtainRequirement(100), 0.40);
-        AchievementHandler.addAchievement('A literal mountain of muscle', 'Obtain 1000 Proteins', new ProteinObtainRequirement(1000), 0.50);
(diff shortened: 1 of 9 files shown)


src/scripts/achievements/AchievementHandler.ts
```diff
@@ -242,6 +242,11 @@ class AchievementHandler {
         AchievementHandler.addAchievement('Don\'t stop trying', 'Complete 2500 total stages in Battle Frontier', new BattleFrontierTotalStageRequirement(2500), 0.25);
         AchievementHandler.addAchievement('King of the hill', 'Complete 5000 total stages in Battle Frontier', new BattleFrontierTotalStageRequirement(5000), 0.4);
 
+        AchievementHandler.addAchievement('Let\'s try this out', 'Obtain your first Protein', new ProteinObtainRequirement(1), 0.20);
+        AchievementHandler.addAchievement('Training progress', 'Obtain 10 Proteins', new ProteinObtainRequirement(10), 0.30);
+        AchievementHandler.addAchievement('Strong like the mountain', 'Obtain 100 Proteins', new ProteinObtainRequirement(100), 0.40);
+        AchievementHandler.addAchievement('A literal mountain of muscle', 'Obtain 1000 Proteins', new ProteinObtainRequirement(1000), 0.50);
```
RedSparr0w, line 248: I feel like the achievement bonus amounts should be reduced a bit.

We should probably add some higher amounts too, maybe similar to BF total stages achievements, but up to 10k?

amative1: Maybe bonuses at every 1 and 5 (so 1,5,10,50,100,500,1k,5k,10k), and have the bonus start lower (like 0.05 or 0.01) and finish at 0.5?

src/scripts/achievements/ProteinObtainRequirement.ts
```diff
@@ -0,0 +1,15 @@
...
+    constructor(value: number, option: GameConstants.AchievementOption = GameConstants.AchievementOption.more) {
+        super(value, option, GameConstants.AchievementType['Protein']);
+    }
+
+    public getProgress() {
+        return Math.min(App.game.statistics.totalProteinsObtained(), this.requiredValue);
+    }
+
+    public hint(): string {
+        return `${this.requiredValue} proteins need to be obtained in total. Used proteins stay counted.`;
```
RedSparr0w, line 13: I think we should keep the hint similar to the Diamonds/Money etc hints

IanWalder: If the other currency chieves count total cumulative amounts instead of what the player has at the moment, then keeping the hints the same makes sense. No problemo, I’ll get on those tasks soon.

RedSparr0w (approved):
Look good, thanks for the PR! 🎉

PR #1581, opened 2021-09-09
Update ClearGymRequirement.ts

amative1:
Fixed Hint pointer per comment made in Discord


RedSparr0w (approved):
How'd this go un noticed for so long 😅 
Guess we've just been using badge requirements instead of this for routes etc..

Thanks for the PR! 🎉

Issue #1017, opened 2020-10-28
[SUGGESTION] Add highlight to quest locations


jaahay:
**Current Behavior**
1st gen / scrubs like me do not have region maps memorized. I have to dig around for `clear gym` and `clear dungeon` locations.

This could be undesirable, as, "learning" the maps could be an integral part of the player's experience. Hence, a suggestion!

Since routes are integer sequential, not named, they aren't so troublesome.

**Improved Behavior**
A `go to quest location` hyperlink that takes me directly to where I need to be.

**Reason for change**
QoL - saves a few clicks and a few awkward moments. App feels more responsive when it provides helpful links.

**Detailed Description**
Potentially, add links to applicable quests found in the top-level `Quests` modal. Other navigation options are definitely possible. I feel the least astonishing change would be to fully qualify locations i.e., `${region} > ${location}`.

Nit: could also be applied to `Achievement` records / `Achievement Tracker` modal.

**Screenshots:**
![pokeclicker quests](https://user-images.githubusercontent.com/10636658/97505563-52063900-1936-11eb-8eb6-6a64396f627d.PNG)
Imagine I also had a `Gym` quest, and that all three of these quests had some form of `go to` or other.

Mephistic:
This has been requested a few times before, but the devs have pushed back in favor of promoting more map interaction.

I wonder if there's a middle ground where any routes/dungeons/gyms with an active quest would:
* Flag the region in the Dock modal
<img width="336" alt="Screen Shot 2020-10-29 at 1 23 19 AM" src="https://user-images.githubusercontent.com/2694761/97529007-66255700-1985-11eb-8cb1-ddb1874c2192.png">

* Highlight the location in the map
![questTest2](https://user-images.githubusercontent.com/2694761/97529958-79392680-1987-11eb-89af-42f93d15eebf.gif)

RedSparr0w:
I'm definitely opposed to a straight link to go directly to the area,
But I think an option for highlighting the area on the map and indicating the region in the dock modal would be fine for quest.

erdnaxeli:
Maybe a quick (& dirty) solution could be to just add the region name in the quest name, like "default elite karen 17 times (Johto)".

jaahay:
This suggestion was implemented elsewhere. Thank you!

RedSparr0w:
Not sure what you mean by this? 😅

jaahay:
> Not sure what you mean by this? 😅

[...]

PR #1585, opened 2021-09-10
Quest QuickTravel

AevitasDragonkin:
Created a quickTravel option for certain quests, with a quickTravelAvailable function to control when it should show (can lock it behind region completion)

Updated MapHelper to update the region when using moveToTown

### DIFF
--- a/src/components/questDisplay.html
+++ b/src/components/questDisplay.html
@@ -31,9 +31,14 @@
                     </td>
                     <td width="30%" class="p-0">
                         <!-- ko ifnot: $data.isCompleted() -->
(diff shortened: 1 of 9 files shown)


RedSparr0w:
This also needs some sort of check for players that have quest out of the current region before reaching the dock

IanWalder:
Related #1017.
Wouldn't this reduce map interaction, which is not what we want (see the above related issue)?

AevitasDragonkin:
Alright, didn't see that link. In case it's still up for debate (since we're almost a year later now), I'd make the following suggestion for requirements:

- A required quest level of 15 or 20
- Current region Dock access
- Maybe a 10% reward reduction when using it?

If it's not up for discussion then the pull request can be closed

PR #1588, opened 2021-09-11
Add Alola Roamers

Ultima1990:
Just putting in all the mythicals except Meltan as roamers for now. Maybe someone will come up with something better for one or more of them, but at least they are now obtainable.


RedSparr0w (approved):

PR #1589, opened 2021-09-11
Update Champion Blue's team to yellow version

Ultima1990:
Sandslash, Alakazam and Exeggutor as the common mons.

Cloyster, Ninetales, Magneton and the Eeveelutions against the various starters.


Qwertypop4:
In that case shouldn't the starters be changed to be only Pikachu?

Ultima1990:
One step at a time.

Qwertypop4:
I'm being unclear. My point is, I feel this is likely a negative change, as it removes one of the most iconic fights in the series in favour of a mostly forgotten one.

GarrikFel:
I'm also not a fan.  I agree that his original fight is more iconic.  While switching to his Yellow team doesn't matter at all, it's just seems like an odd choice to make.  Why are you choosing his Yellow team instead of his Let's Go team (besides the obvious charizard problem)?  Since this game let's you pick from Red's choice of starters, it feels natural that Blue should base his team on the Red&Blue logic, and not the Yellow logic.  To go on a tangent, it's always weird to me to see Blue without his pidgeot... there's a reason why in Master's his first pokemon partner is pidgeot.  I'm one of those oddballs who prefer's to see Blue with his pidgeot, to help differentiate him from Gary.  

It doesn't matter at all, this won't stop anyone from playing the game, and this isn't going to get anyone new to play the game.  I'm just not sure why you want this.

Qwertypop4:
I personally feel like, even though we are in most cases using the enhanced versions (Crystal, Emerald, etc). That that isn't really needed here.
(Yellow isn't really an enhanced version anyway imo, it's just a special edition made to capitalise on the anime. Hell, it would make more sense to say Red and Blue are enhanced versions similar to USUM, given that they are based on Japanese Blue, which is the enhanced version of Japanese Red and Green)

Ultima1990:
I don't know why this change is controversial. Everything else in our Kanto region is based on yellow, except starters and some wild encounters that would otherwise be unobtainable due to version exclusivity. This is just one step closer to consistency.

Qwertypop4:
The question is... Why is it actually needed?
Away from from consistency, is this an improvement to the game? Yellow Blue, as previously stated, uses a far less iconic team, even removing his signature Pokémon (Pidgeot and his starter) It just.... seems like a pointless step, slightly in the negative direction. This wouldn't do any real harm to the game, but neither would a small amount of inconsistency.

Ultima1990:
I still don't see how it is negative. Eevee is his starter, and I don't see how Pidgeot is iconic.

jk13pclick:
The real improvement with this change comes once the variable Eeveelution mechanism (based on rival battle outcomes) is implemented for Blue (Which Ultima has mentioned wanting to do IIRC). Then the 3 elemental starters can be removed and it will feel like in game results have a personalized outcomes for each player. IMO this is a worthwhile change because it will eventually make the game less formulaic and standardized -> player variability is something good.

GarrikFel:
I don't see this change as a negative.  I just don't like it.  That's entirely on me.  I don't like Yellow.  I too feel like it's a cheap cash grab, and I try not to think about it.  My surfing pikachu is nice, but it's just the creators selling the same game a 4th time.  I don't care that the pokemon on routes use their Yellow locations.  I bet most people didn't even notice... but they'll notice this.  The starter, and the rival are two of the most memorable elements of the games, and there's bound to be strong feelings about those two.

Removing the three starters is going to have a bigger pushback... but that seems interesting.  I assume we'll be able to pick from a lot more pokemon for our starter.  Are you planning on having Blue pick a fighting starter if we pick a normal starter?  Ground for gastly?  Ice for dratini?

I have some questions about making the game more personalized.  Since the game requires completion to move on to the next region, there's not a lot of choices to make.  You have to get everything anyway, does it really matter who you pick as a starter?  You'll catch them all by the end, and you use every pokemon on every fight... This seems like a change that would help with teams, if that ever happens.  

Explaining Blue's Pidgeot is going to be tricky.  Gen 1 didn't have a lot of written story.  Gen 1 had a story, but it literally couldn't fit on the cartridge.  So the game does this thing where they include a lot of world building elements without explicitly stating them.  
For example:  Team Rocket takes over Silph Co. because they want the masterball.  They want the masterball to catch Mewtwo, which they had a hand in creating.  They want Cubone's dead mother because it would be the only ghost type in gen 1 not weak to psychic.  They hang out on nugget bridge (just outside of Cerulean Cave) looking for strong trainers... I wonder why?  Did you notice that team rocket is known for using poison type pokemon, and the leader of team rocket is known for using ground type... a type strong against poison?  Talk about an easy way to keep your criminal organization in line!  The games never really go into details about any of these things, but it has all of these little details that help imply the story they were trying to tell.  Gen 1 is very "Show, don't tell" to a fault.  Even so... I'll still take it compared to some of the later game story lines... (sometimes limitations make the final product better).
Blue's Pidgeot is one of those "Show, don't tell" elements of the story, and says a lot about his character.  Pidgeot is... not the best pokemon to use, and yet it's on Blue's **champion team**.  It implies that (unlike Silver) he cares about his pokemon, even if they're not the strongest.  "Truly skilled trainers try to win with their favorites", is one of the core themes of the games, expressed in this one small decision.  Pidgey is probably the first pokemon that Blue caught, and he brought it all the way to the end.  Why?  Who knows?  Maybe that Pidgeot is the reason he's been 1 step ahead of you the entire journey (constantly flying to new cities instead of walking).  Maybe Blue is just sentimental, especially since who knows what the deal is with his Raticate.  This is why Blue's pidgeot is iconic, because it's one of the things that makes him more than just a bully-rival character.  It gives this character with a dozen lines of dialogue a little bit of depth.  You may not have noticed it, but your brain did.  
It's the reason why Master's gave Blue pidgeot first.  It's why it's on his GSC team.  It's why it's on his FR/LG team.  It's why it's on his HG/SS team.  It's why it's on his Battle Tree team.  While he doesn't always have it on his team, if Blue was a real person, it would definitely be one of his favorites.

Once again, this isn't a bad change, it seems like a neutral change.  No one is going to start, or stop, playing this game because of this change.  Everyone is going to get over it.  I don't care for it, but I'll get over it.

Ultima1990:
Cancelled

PR #1592, opened 2021-09-12
Fix for log book types missing

### LINKED ISSUE #1591: [BUG] Unable to load save file if log book contains an entry without a specified type
**Info**
 - OS: Windows 10
 - Browser: Chrome
 - Version: v0.8.9

**Description:**
Save files with log book entries that do not contain type information cannot be loaded in version 0.8.9.

**Steps to Reproduce:**
1. Get a save file containing one of those entries
2. Try to load it

**Expected Behavior:**
There should be fallback behavior for those entries.

**Screenshots:**
The exception:
image

Entries causing this error:
image

**Additional information:**
I'm not sure what's causing these entries to not contain the `type` property - is this worth investigating?
Edit: could this be related to the ACHIEVEMENT and QUEST_COMPLETE types getting renamed?

RedSparr0w:
Not sure how/why this error is caused, but this should mostly remedy the problem.

closes #1591

### DIFF
--- a/src/modules/logbook/LogBookLog.ts
+++ b/src/modules/logbook/LogBookLog.ts
@@ -1,13 +1,12 @@
 import { LogBookType } from './LogBookTypes';
 
 export default class LogBookLog {
-    public type: LogBookType;
-    public description: string;
-    public date: number;
-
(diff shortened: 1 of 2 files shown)


Aegyo (approved):
LGTM

PR #1593, opened 2021-09-13
Limit to +32 queue slots per new region

RedSparr0w:
Give a maximum of 32 extra queue slots when travelling to a new region instead of the current doubling each region.

### DIFF
--- a/src/scripts/breeding/Breeding.ts
+++ b/src/scripts/breeding/Breeding.ts
@@ -389,7 +389,7 @@ class Breeding implements Feature {
     }
 
     public queueSlotsGainedFromRegion(region: GameConstants.Region): number {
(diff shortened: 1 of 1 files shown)


Aegyo:
Aren't we currently giving 64 slots for kalos? Do 32 of those need to be removed for existing saves?

GarrikFel:
Kalos has a total of 64 slots.  Unova had 32, and Kalos added an additional 32.  This update just makes it so every upgrade is only as good as the Kalos hatchery queue upgrade.

RedSparr0w:
I believe we only have total 64 slots in Kalos.

The function is kind of misleading in the fact that it's calculating based off your current region to how many extra you will have in the next region.

Aegyo (approved):

PR #1595, opened 2021-09-13
Pacing out Pokémon Unlocks

Ultima1990:
Very WIP. Very Experimental.

It started with locking Headbutt Pokémon behind Ilex forest. And then... yeah. Fishing rods are split, surfing mons are split, and given that the rods are named after the region you can see where this is going.

Many things need doing. Every other region. Also I want to do the Dive mons in Hoenn. Maybe honey tree mons in Sinnoh. Etc.

Or nothing, this is mostly for messing around. Posting the PR to let people see what I'm messing with, see if they like it.

It's not just out of the blue tho. I think it's good for pacing tbh.


Ultima1990:
Today I learned something interesting.

If there is no land pokémon category on a route, every other category is allowed regardless of requirements.
This caused Lanturn to show up on route 41 before beating Champion Lance. Same with Kingler on 40. But Qwilfish on 32 and Remoraid on 42 did not show up, since those routes do have land mons.

Solution: on water only routes, which are not accessible before surf anyway because of route requirements, surf mons are now land mons.

Qwertypop4:
Perhaps to stop HMs and Fishing Rods from flooding the key item bag, perhaps put them in some sort of TM case and Fishing Bag respectively?

Ultima1990:
New idea. Instead of flooding the key item inventory, why not have Oak pop up and go:

> Congratulations!
> 
> Your [Fishing Rod] has been upgraded! It can now be used to find more Pokémon!

or

> Congratulations!
> 
> Your [Fishing Rod / HM03 Surf] has been upgraded! It can now be used in this region!

Then I can just turn all of the mons that would be locked behind these key items into special encounters (see Castform), locking them behind the thing that would give the key item. Same effect, less key items.

So now to just make that... I'm lost. Gonna need some help with this one.

Ultima1990:
The new idea in effect, Kanto Johto and Hoenn done. I decided to not do the special encounters thing. Instead I just updated RouteHelper to the requirements that the key items used to have.

Now to just get Oak to show up to announce the newly available encounters, cuz in most cases the key item announcement is gone. I have no idea how to pull that off.

PR #1596, opened 2021-09-14
Sinnoh encounters, route difficulty, other things

Ultima1990:
Eliminated a couple of encounters from routes because they don't exist there in Platinum version. Ignored Diamond/Pearl exclusive Pokémon that would become unobtainable.

Changed the order of routes 211-215 in the file to match the unlock order.

Moved 218 in the file and changed its difficulty to match the moment it unlocks. It shouldn't be harder than 216 and 217.

Removed Noctowl boss from Lake Valor, since it doesn't exist there in Platinum version. Replaced with Bibarel. Yay, Bibarel gets to be a boss somewhere, I'm so proud.

Moved a bunch of Pokémon from water to land, as they are encountered in tall grass.

Changed the order of a bunch of mons for no real reason other than I prefer them ordered by dex number.

Cleaned up a bunch of requirements that didn't need to be there.

### DIFF
--- a/src/scripts/dungeons/Dungeon.ts
+++ b/src/scripts/dungeons/Dungeon.ts
@@ -1474,7 +1474,7 @@ dungeonList['Wayward Cave'] = new Dungeon('Wayward Cave',
     56500, 230, 100);
 
 dungeonList['Mt. Coronet South'] = new Dungeon('Mt. Coronet South',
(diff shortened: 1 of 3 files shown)


RedSparr0w (approved):
Looks good, thanks for the PR

PR #1599, opened 2021-09-15
Updating Gym Display Names for Johto

amative1:
Fixed display/identification issue for Johto League to prevent "Bruno2", "Koga2", and "Lance2" from being reported as bugs.


anapple96:
Trust me, it's not that simple

AevitasDragonkin:
What would be the preferred way to solve this? Would we add another property to the constructor (displayName) and use that one only to display text and still use the leaderName for everything else.

Or could we alter the leaderName constructor property and handle it in the constructor?
Something like Bruno2:Bruno would be split on : with the first param used as leaderName and the second as displayName, and if there is no : present, then displayName = leaderName?

amative1:
Closing for now, need a more thorough fix (such as adding a "display name" field or something to all gyms).

PR #1600, opened 2021-09-16
Auto-restart CSS

AevitasDragonkin:
Added CSS to the auto-restart gym button so aligned better on the E4


RedSparr0w (approved):

PR #1602, opened 2021-09-17
Add basic web worker for game ticks

RedSparr0w:
Make use of web workers, should hopefully not slow the game down while running in the background and should also provide a more reliable tick rate.

Option in settings to disable it.


Aegyo (approved):
Doesn't seem to be any problems with this, I wonder if there are other things worth offloading to a worker?

RedSparr0w:
Only other things I can think of at the moment is:

Dynamic background _(nothing time sensitive, all just doing it's own thing)_
Notifications _(for if we add desktop notifications, although not sure if it would be necessary)_

PR #1604, opened 2021-09-18
fix (lag): Use a copy of caughtPokemon for sorting

Aegyo:
This change makes things run smoother for me while sorting by attack, could do with testing from other people though.

The difference isn't coming through in gifs very well, but here is a comparison with some auto defeat hacks
```js
setInterval(() => {if (Battle.enemyPokemon()?.isAlive()) Battle.defeatPokemon()}, 10)
```
[Before](https://imgur.com/FUz4rNy)

[After](https://imgur.com/ILvS3jS)

### DIFF
--- a/src/scripts/party/Party.ts
+++ b/src/scripts/party/Party.ts
@@ -75,9 +75,6 @@ class Party implements Feature {
 
         App.game.logbook.newLog(LogBookTypes.CAUGHT, `You have captured ${GameHelper.anOrA(pokemon.name)} ${pokemon.name}!`);
         this._caughtPokemon.push(pokemon);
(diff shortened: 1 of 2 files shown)


RedSparr0w (approved):
As far as my testing went, everything worked a lot better.

Thanks for the PR! 🎉

Aegyo:
Do you want anything more on this before merging? I'm going to leave the exp buffer thing out for now, I'm not convinced it does much more than the existing deferUpdates or rateLimit on sortedList.

I'm still going to look for other things causing lag, but they can be separate PRs if I find anything worth doing.

PR #1605, opened 2021-09-19
fix (lag): Prevent knockout updating damage calculator when the modal is hidden

Aegyo:
When putting pokemon into eggs/queue, there seem to be two main causes of lag:
1. DamageCalculator updating things
2. Achievements for total damage updating their progress

This deals with the first case, and will be fairly easy to use for other modals if they need it.


RedSparr0w (approved):
Working great, thanks for the PR!

PR #1606, opened 2021-09-19
fix (lag): Make Achievements use a single attack observable for common parameters

Aegyo:
Part 2 of #1605

A few achievements are creating identical observables that all want to run calculatePokemonAttack with the same parameters. This makes them use a single common observable instead.

According to firefox performance analysis there are still some lag spikes when putting pokemon in eggs, but I can't decipher where the remaining lag is coming from with the information it gives me.

With these two PRs, I'm not noticing any breeding modal lag while auto-defeating every 200ms and sorting by attack.

Leaving them as separate PRs because they each have a noticeable lag improvement on their own and I don't want to prevent one because of issues in the other.


RedSparr0w (approved):
Nice catch, thanks for the PR!

PR #1607, opened 2021-09-20
Alola routes/dungeons

amative1:
Added Bounsweet, Cosmog, Sandygast

Still need to determine UBs (possible quest?), confirm Alolan Raichu & Exeggutor evos


RedSparr0w (approved):

We can always change how Cosmog is obtained before releasing Alola (or later) if needed.

PR #1609, opened 2021-09-21
Shop amount reset setting

AevitasDragonkin:
Adds a setting to not reset the shop amount after a purchase.
Default behaviour stays the same as is (reset on purchase)


RedSparr0w (approved):

PR #1610, opened 2021-09-28
Prevent shoveling fully grown crops

Lear85:
Fully grown crops are instead harvested as if the shovel were not selected.


RedSparr0w (approved):

Issue #1556, opened 2021-09-02
[SUGGESTION] Dungeon Loot Improvement


jk13pclick:
**Current Behavior**
Dungeon Loot is currently limited to X-Items. 

**Improved Behavior**
Similar to the proposed incremental Dungeon shrinkage #1548 with X number of clears, "Rare Loot" could be implemented in a similar way with an increasing chance of finding rare items as loot with more clears of a dungeon.

**Reason for change**
Dungeons are kind of boring right now. Apart from the occasional Pokémon that has to be captured, Quests, and achievement hunting for the bonus, there is little reason for players to go into Dungeons of their own interest once they have been cleared. While the system may be improved with an eventual Dungeon Rework #1249, this is still a ways off.

**Detailed Description**
X-Items are good, especially in the early game. However, they are not particularly unique reward for Dungeons especially later in the game when they are easily purchasable. By making Rare Items, in particular QP required items, available in Dungeons the game economy of DT, Money, and QP will become more fluid. Additionally, I envision "Berry Stashes" as loot that could yield rarer berries for those less inclined to mutate in the Farm. While there have been many proposals to diminish the existing QP grind early in the game, this could be an easy solution that doesn't necessarily have to make purchasing QP items irrelevant if the base odds are low enough. Finally, as mentioned in the Rework issue, this is an implementation that can be transferred to a new Dungeon system when it arises.

PR #1612, opened 2021-09-30
Loot System Improvements

jk13pclick:
My attempt at #1556. The gist is that more item types can now be used as Loot in the Underground, making Dungeons a more viable strategy for getting rare Items. In general, the goal is to have the intended mechanism (ie: Farm, Underground, etc.) to always be more efficient ways to collect these items, but to allow players other methods to encourage Dungeon grinding for the achievements and for players that really dislike certain mechanisms.

~What is not currently done is adding revised loot for every Dungeon in the game. I thought it prudent to get approval before moving forward + I thought it could be fun to add in some game lore for item locations (ie: Moon Stone in Mt Moon,  Dragon Scale in Mt. Mortar, etc) and could use some input from others + the Discord.~ **This is done now**

As I know far too little about coding and Typescript, this may look super bad and any revisions/improvements would be very welcome. While I worked hard to figure out how to do this, if you don't like it at all then feel free to disregard this PR. I feel that with an aggressive weight curve like the one implemented it won't disrupt the balance of the game too much - hopefully this change could make it into #1571 at the least. Also, if I messed up please let me know, as I mentioned I do not claim to be good at this at all! Let me know what you all think.


src/scripts/dungeons/Dungeon.ts
```diff
@@ -136,7 +136,7 @@ class Dungeon {
             if (typeof loot === 'string') {
                 return 1;
             } else {
-                return ((<DetailedLoot>loot).weight / (App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]() + 1) + 1) ?? 1;
+                return (10 ^ (<DetailedLoot>loot).weight / (App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]() + 1) + 1) ?? 1;
```
PixLSteam, line 139: This might not do what you want it to do. (Hint: ^ is not exponentiation)

jk13pclick: True, I totally forgot about Math.pow!

jk13pclick:
One more tweak to the system, Item Magnet is now nerfed and tied to Dungeon Clears and Weight, would be super broken to have a 50% chance get 2x rare items.

jk13pclick:
One final thought I had while making the loot for all the regions is that this system could be used for some of the currently absent alt forms. Orbs, Drives, etc. could be made into extremely rare loot only obtainable at the end of questlines. Mega Stones could also be obtained using this system (Megas themselves should have their own feature of course).

RedSparr0w:
👍 Nice work, Everything looks good as far as I can tell!
I haven't checked all the items that have been made available, but will trust your judgement on them 😅 

Made a couple of changes:
```
Force all loot to use the new formatting. ({loot: 'name', weight: 4})
Limit minimum weight to original weight specified.
Adjusted rate for Item Magnet based on original weight specified.
Updated description for Item Magnet.
```
Let me know if the changes are all right.

RedSparr0w:
If you want to do them, that's cool, but otherwise I can just merge as is and they can be done in a separate PR?

RedSparr0w (approved):
Forgot to approve it 👍

PR #1614, opened 2021-09-30
Professor Oak Item Exp Share not unlocking

Dasorine:
Have 112 pokedex completed but have not been given the Exp Share item despite having the further unlocks available
Untitled
.


Aegyo:
It looks like you have an adblocker which is removing the image. You can disable your adblocker for pokeclicker.com, there are no ads on the page.

Also, you should open an [issue](https://github.com/pokeclicker/pokeclicker/issues/new/choose) for this kind of thing, pull requests are for submitting your code changes

Dasorine:
my bad, was just going by what the "report a bug" button in game lead to,
might want to look into that.

Issue #1572, opened 2021-09-07
[FEATURE] Ultra Beast quest


Ultima1990:
Most Ultra beasts (or UBs for short) are currently not obtainable. The only ones that are currently take care of are Guzzlord as a dungeon boss and Poipole (and by extension Naganadel).

Our options are:
- Put them in their USUM location. Which for most of them is Ultra Space. Which would probably mean a whole new mechanic for that minigame.
- Just put them in their SM locations after beating the champ. Except for Stakataka and Blacephalon, which are USUM additions and conveniently are not located in Ultra Space. So that's nice of them. They'd get to go in their USUM location.
- Do a SM style quest line.

The quest line is my personal favorite option. Problem is I don't know how to do any of this, so I could use a hand.
Here's what I'm currently envisioning:

The first quest in the line starts upon beating the champion. Nihilego unlocks in Wela Volcano Park and Diglett's Tunnel. Catch X number of it.
Second quest, Buzzwole unlocks in Melemele Meadow. Catch X number of it.
Third quest, Pheromosa unlocks in Verdant Cavern. Catch X number of it.
Fourth quest. Xurkitree unlocks in Memorial Hill and Lush Jungle. Catch X number of it.
Fifth quest. Kartana unlocks in route 17 and Malie Garden. Catch X number of it.
Sixth quest. Celesteela unlocks in Haina Desert (route 23) and Malie Garden. Catch X number of it.
Seventh quest. Blacephalon unlocks in Poni Grove (route 27). Catch X number of it.
Eighth quest. Stakataka unlocks in Poni Grove (route 27). Catch X number of it.
Ninth quest. Catch Guzzlord in Resolution Cave.

All UBs that get added to dungeons would be added as random encounters, not bosses.
All unlocks are permanent. The UBs stay available after the quest is finished. In case people want to farm them for shinies or whatever.

If we can do quests with multiple objectives, combine quests 2 and 3, 5 and 6 and 7 and 8.

Resolution Cave would be locked until the Guzzlord part of the quest is active.

For extra fun we could add the route 8 motel as a location and start the quest from there. In that case for extra extra fun we could add a new first quest in the line to defeat Anabel in a battle.

We could extend the quest by adding filler "defeat X many Y types* and "catch X many Y type" quests in between. You know, the usual. Personally, I'd say no, don't do that. We have tons of that already, it's boring, meh. If we still do really want to extend the quest, consider this: 
[...]

PR #1019, opened 2020-10-28
Alola

Ultima1990:
It begins.

Oricorio forms.
Lycanroc forms.
Wishiwashi forms.
Very early route requirements. None of the Pokémon on the routes are correct. OMG ended up with a whoppin' 27 routes!

PR #1615, opened 2021-09-30
UB Questline

jk13pclick:
An Ultra Beast Questline as outlined in #1572 . Task List as follows:

- [x] Make Beast Balls a thing
- [x] Lock Resolution Cave behind UB Questline
- [x] Make the Route 8 Motel battle with Anabel (Temporary OTB)
- [x] Combine quests 2 and 3, 5 and 6, 7 and 8
- [x] Weight the Dungeon stuff correctly
- [x] Custom Beast Ball svg for the Pokeball selector created, Beast Ball png uploaded
- [x] Need Route 8 Motel Background Image
- [x] Update the tmx
- [x] Final Steps - Make sure the Champion trigger works, ``Update.ts`` line, balance QP <-> Beast Ball exchange rate
(Post crew test)
- [x] Add story NPCs (waiting for #2083)
- [x] Change the quests to not include breeding (waiting for #2380)
- [x] Change to MultipleQuestQuest (waiting for #2618)


jk13pclick:
This is ready, all that is left to do is balance and maybe an Update.ts line if this isn't included on release.

src/scripts/gym/Gym.ts
```diff
@@ -1449,6 +1449,21 @@ gymList['Elite Kahili'] = new Gym(
...
+    [
+        new GymPokemon('Alakazam', 2500000, 61),
+        new GymPokemon('Weavile', 2500000, 61),
+        new GymPokemon('Mismagius', 2500000, 61),
+        new GymPokemon('Salamence', 2500000, 61),
+        new GymPokemon('Snorlax', 2500000, 61),
+    ],
+    BadgeEnums.Agent_Anabel,
+    64000,
+    'Oh you\'re good! You might just have a shot, here is a Beast Ball. Go hunt down those strange Ultra Beasts! If you collect any shards from them you can trade them for more Beast Balls here.',
```
Ultima1990, line 1464: Should probably change this mention of "shards". Might confuse players. Because of that other game mechanic called "shards".

jk13pclick: Changed text to say Beast Points

Ultima1990:
We do have the beginnings of a questline based requirement check. I completely forgot about it. It was added as part of the initial Alola PR. #1019. It's intended to be used as part of the potential eventual expansion of Mina's trial. Maybe you can use this (or a modification of this) as the Resolution Cave requirement, instead of naming those UBs specifically? https://github.com/pokeclicker/pokeclicker/pull/1019/commits/3c6c3f5a795d1189a5ae80e15620d4106069ab6d

RedSparr0w:
Just a heads up, this is starting to get quite a few conflicts 😅

Jaaslet:
The 'always use beastballs on beasts' feature works a little weird.
- It's never explained that it will always use beastballs on beasts, no matter what is selected. And it's never explained that you can never use other pokeballs on beasts. Maybe add a NPC explaining the balls a bit better?
- It's not possible to not use beasts balls. I have "already caught" set to "no catch", but it still uses my beastballs
- It says that it has reduced catch rate for other pokemons. But if it can never be used on other pokemons, it makes no sense.
- Non-beast pokemons currently always has a 0% catchrate

Jaaslet:
- The notification, when you get a new quest in the questline, does not stay for long. It's a lot of text, and not much time to read it. The Notify-method has a "time"-parameter, where you can make it stay longer
- All the quest descriptions ends with "..". In the "createUltraBeastQuest"-method, you add a dot to the end of the hint, but all the hints already has a dot

src/scripts/dungeons/Dungeon.ts
```diff
@@ -308,14 +308,17 @@ class Dungeon {
...
                     pokemonName = (<DetailedPokemon>enemy).pokemon;
+                    hideEncounter = (<DetailedPokemon>enemy).options?.hide ? ((<DetailedPokemon>enemy).options?.requirement ? !(<DetailedPokemon>enemy).options?.requirement.isCompleted() : (<DetailedPokemon>enemy).options?.hide) : false;
                 } else {
                     pokemonName = <PokemonNameType>enemy;
                 }
                 const encounter = {
                     image: `assets/images/${(App.game.party.alreadyCaughtPokemonByName(pokemonName, true) ? 'shiny' : '')}pokemon/${pokemonMap[pokemonName].id}.png`,
                     shiny:  App.game.party.alreadyCaughtPokemonByName(pokemonName, true),
+                    hide: hideEncounter,
                     hidden: !App.game.party.alreadyCaughtPokemonByName(pokemonName),
```
Jaaslet, line 322: Using both hide and hidden is confusing. One of them need to change name

src/scripts/pokeballs/Pokeballs.ts
```diff
@@ -98,6 +98,12 @@ class Pokeballs implements Feature {
                 return Math.min(15,Math.pow(amountCaught,2) / 5000);
             }, 1250, 'Increased catch rate with more catches', new RouteKillRequirement(10, GameConstants.Region.johto, 34)),
 
+            new Pokeball(GameConstants.Pokeball.Beastball, () => {
+                if (typeof GameConstants.UltraBeastType[Battle.enemyPokemon().name] === 'number') {
+                    return 10;
+                }
```
Jaaslet, line 104: Don't think we need this code anymore, as ultra balls can only be used on Ultra beasts. It's confusing to have code that don't do anything

src/scripts/dungeons/Dungeon.ts
jk13pclick: @RedSparr0w Your system, any thoughts on this?

src/scripts/towns/TownList.ts
```diff
@@ -3054,14 +3064,13 @@ TownList['Melemele Woods'] = new Town(
...
-//TODO: Change requirement when UB questline is merged.
 TownList['Roadside Motel'] = new Town(
     'Roadside Motel',
     GameConstants.Region.alola,
-    [],
+    [TemporaryBattleList.Anabel, RoadsideMotelShop],
     {
-        // requirements: [new GymBadgeRequirement(BadgeEnums.Elite_AlolaChampion)],
-        requirements: [new NullRequirement()],
+        requirements: [new GymBadgeRequirement(BadgeEnums.Elite_AlolaChampion)],
```
Jaaslet, line 3072: This should be changed to QuestlineStartedRequirement too

src/scripts/dungeons/Dungeon.ts
RedSparr0w: We should probably change `hidden` to `caught` or `uncaught` as that is more inline with what the option actually does.
_(would also need to be changed on the boss encounter const too)_

RedSparr0w:
Could we also get the beast ball SVG to look more similar to our other current balls?
_(front on view, no perspective)_
image

Jaaslet (approved):
I think we are finally there. Time to release the beasts!
Thanks for the PR

PR #1620, opened 2021-10-14
Update Slowpoke drops

amative1:
Add King's Rock to Slowpoke drops, to match original game behavior


amative1:
Also removed from Galarian version (which has no drops)

RedSparr0w (approved):

PR #1626, opened 2021-10-27
Update underground.html

### LINKED ISSUE #1625: [SUGGESTION] Underground's 'Total Net Worth' should be formatted with commas
Basically the "Total net worth" value displayed in the underground looks something like 123456789 and should be 123,456,789.

**Current Behavior**

image

**Improved Behavior**

image

**Reason for change**

It would be easier to see what my total net worth is.

**Detailed Description**

The number just needs to be formatted to match the style of other numbers with commas (like how money uses commas when it's displayed)

**Screenshots:**

Used above appropriately

amative1:
Fixes #1625


RedSparr0w (approved):

PR #1633, opened 2021-11-04
Give wild route pokemon variable health based on base stat.

Ultima1990:
Because seeing the same unchanging number on the bar regardless of what Pokémon you are fighting is just a little odd. It's for flavor mostly.

What it does is grab some of the route health, change it by a percentage based on the base hitpoints stat, then puts it back.
0.001 * PokemonFactory.routeHealth(route, region) * basePokemon.hitpoints + 0.9 * PokemonFactory.routeHealth(route, region)

This has actually made all routes easier because turns out most wild Pokémon have less than 100 hp. So that must be compensated for before this goes out. No idea how to tackle that tbh.


Symi001:
I don't know if we want to change offline earnings following this. We're using the route health for it atm, so as long as it represents the mean value it shouldn't be an issue. I'm only afraid of how health could be altered by the last part of your comment about rebalancing ;)

KrushGames:
well obviously this is mainly to have some diversity, not a real balance change or anything, so probably some rebalancing based on the average of the pokemons' hp stats on every route, but thats not too hard, its just numbers and the time it takes to calculate them.

Ultima1990:
I want to involve the average hp of the route. The formula would look like this:

(PokemonFactory.routeHealth(route, region) - (PokemonFactory.routeHealth(route, region) / 10)) + ((PokemonFactory.routeHealth(route, region) / 10 ) / ([average route hp] * basePokemon.hitpoints))

Maybe too many brackets, whatever.

The result should be this:
VariableRouteHP

Now to just get the game to grab this average route hp for me...

Ultima1990:
This now functions. Mostly. With a little (big) help from our lovely #development-chat, the following updated table of HP values for 208 is now true ingame!
208HPrecalc10again
HP values vary a bit based on what Pokémon you are fighting, but are still fairly close to the routeHealth value.

Math.round((PokemonFactory.routeHealth(route, region) - (PokemonFactory.routeHealth(route, region) / 10)) + (PokemonFactory.routeHealth(route, region) / 10 / routeAvgHp(region, route) * basePokemon.hitpoints))

The formula could probably be cleaned up quite a bit, but it does its job and that makes me happy.

The only problem now is that our friendly build checking bot has some prefer-const complaints about poke and total, which I have no idea how to fix.

src/scripts/pokemons/PokemonFactory.ts
```diff
@@ -22,9 +22,14 @@ class PokemonFactory {
...
         const id = basePokemon.id;
+        const routeAvgHp = (region, route) => {
+            const poke = [...new Set(Object.values(Routes.getRoute(region, route).pokemon).flat().map(p => p.pokemon ?? p).flat())];
+            const total = poke.map(p => pokemonMap[p].base.hitpoints).reduce((s, a) => s + a, 0);
+            return total / poke.length;
+        };
 
         // TODO this monster formula needs to be improved. Preferably with graphs :D
-        const maxHealth: number = PokemonFactory.routeHealth(route, region);
+        const maxHealth: number = Math.round((PokemonFactory.routeHealth(route, region) - (PokemonFactory.routeHealth(route, region) / 10)) + (PokemonFactory.routeHealth(route, region) / 10 / routeAvgHp(region, route) * basePokemon.hitpoints));
```
RedSparr0w, line 32:
```suggestion
        // Health has a +/- 10% variable based on base health stat compared to the average of the route
        const maxHealth: number = Math.round((PokemonFactory.routeHealth(route, region) - (PokemonFactory.routeHealth(route, region) / 10)) + (PokemonFactory.routeHealth(route, region) / 10 / routeAvgHp(region, route) * basePokemon.hitpoints));
```

Ultima1990: See, I knew it could use some cleaning up. I don't math gud. I computer math even less gud.

RedSparr0w: Yeah didn't change anything 😅 
Is the comment correct though?

Ultima1990: Yes looks to be an accurate description.

RedSparr0w (approved):

PR #1634, opened 2021-11-06
Git those filfy Dungeons outa ma taaawn!

Ultima1990:
The most important thing here is giving Team Flare Secret HQ a presence on the map. People seem to get confused after the Anistar City gym, seeing this dungeon light up might convince people to go check it out.
Did Slowpoke Well and Castelia Sewers too because why not. All of these dungeons do also still have the option to access them from inside the town. Can't do Team Plasma Assault right now, would need a sprite of the plasma frigate in flight for that one, so I'd say this is done.


RedSparr0w (approved):
Great work, Thanks for the PR!

PR #1637, opened 2021-11-10
Kasib Occa fix

amative1:
Updated Plot.ts to fix Kasib mutations (should be any Berry @ Occa or higher, but wasn't set to allow Occa)


amative1:
Updated hint notification timing for this as well as for Starf's hint to show when Occa obtained (rather than Passho)

RedSparr0w (approved):

PR #1640, opened 2021-11-13
Hide fake gyms in the badge case

PixLSteam:
Changes the badges displayed in the badge case to be based on `GameConstants.RegionGyms`, while still filtering out `Elite` badges.
This would hide potential one-time gym battles or dummy badges.
Also fixes a capitalization inconsistency I found during tests.


RedSparr0w (approved):

PR #1642, opened 2021-11-14
Alola Badges

Ultima1990:
Gives proper badges to the current gym-style trials. 5 of them. Bit ugly for the badge case. Also starts experimenting with options to have the trial dungeons give badges. That would land us on 12 badges, which would fit neatly in the badge case I'd say. 2 rows of 6 might be nicer, but that's not a thing I'm gonna tackle.


Ultima1990:
This PR now includes the SVG for the second half of our current Alola maps. Doesn't look great, especially route 12 and Poni Gauntlet. Turns out curved routes are an absolute pain in the groin area.

Ultima1990:
Looks like all badges are now properly rewarded.
The testing that can now be done with the Alola2 map has brought some more issues to light, but that will be a different PR. Don't want to make this one messy with a bunch of unrelated fixes. It's already sort of bloated with the SVG.

I'd say this is done.

RedSparr0w (commented):
> Turns out curved routes are an absolute pain in the groin area.

Sure are...

Everything looks good, but noticed a few of the "badges" don't show in the badge case?
image

Is there anything different about these badges compares to the others?
Is it just the dungeon badges missing?

Ultima1990:
DarkiniumZ, FairiumZ and GroundiumZ are regular gyms. Can't help but notice that none of the dungeon badges are in that image at all...

PixLSteam:
I'd assume that it's because of my recent PR that hides fake badges (well, badges for gyms that are not listed in RegionGyms). Adding the dummy gyms to that list *should* show them.

Ultima1990:
Worth a shot.

Question tho. Are we gonna get quests based on these dummy gyms?

PixLSteam:
Probably, unless we add impossible requirements to them. I'm also not sure if those are used for the bot quiz.

Ultima1990:
K. So. I have added temporary ObtainedPokemonRequirement. I think it is the only requirement that will work, can't just have a dummy requirement. Can't have the thing require a gym or route that does not exist, and if the gym or route *does* exist it means we get the same problem again.
So ObtainedPokemonRequirement it is. It is currently set to Meltan, which is currently impossible to get but of course that will not always be the case. I was kind of hoping to use ??? since it is never meant to be obtained but the code got all confused by that.

Next problem is achievements.

Before this stuff is sorted out this is NOT ready for merge.

Ultima1990:
Massive thanks to PixLSteam and RedSparr0w for all the help!

I think this is ready now.

RedSparr0w (approved):
Testing all seemed to go well, looks good!

Thanks for the PR!

PR #1645, opened 2021-11-17
Fix hatchery sorting, Separate hatchery sort from party sorting

### LINKED ISSUE #1647: Pokemon not sorting correctly in Hatchery sorting area
Since the most recent update, while the list of the pokemon on the main page, default is left, sorts with no problem, the list that displays when you open the Hatchery no longer works. It will show by the filters but it won't sort.

image

No matter what I click the order doesn't change and as you can see in the image everything is completely out of order. Thanks so much, love the game!!

RedSparr0w:
Fixed hatchery sorting.
Hatchery can now be sorted independently of party.

@Aegyo
~~Any chance you could take a look at not sorting while the modal is hidden?
I tried using the `modalUtils.observableState['breedingModal'] === 'show'` but it just caused a ton of lag when trying to re-open the modal.~~

Fixes #1647

### DIFF
--- a/src/components/breedingModal.html
+++ b/src/components/breedingModal.html
@@ -91,7 +91,7 @@ aria-labelledby="breedingModalLabel" aria-hidden="true">
                             </div>
                         </div>
                         <div class="form-row collapse" id="breeding-sort">
-                            <div class="form-group col" data-bind="with: Settings.getSetting('partySort')">
+                            <div class="form-group col" data-bind="with: Settings.getSetting('hatcherySort')">
                                 <label>Sort</label>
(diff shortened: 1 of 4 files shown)


RedSparr0w:
Also added a fix for Protein sorting (independent of party)

Aegyo (approved):
Looks good 👍

Is it worth doing the `if (modal is open) { sort }` thing for the protein list too? I know you have that around the actual modal content, but I assume the list will continue re-sorting in the background if it is being sorted by attack. Probably doesn't cause much lag by itself anyway.

PR #1648, opened 2021-11-18
Add image selection for Trainer and Background on Profile

### LINKED ISSUE #1646: [SUGGESTION] Make Trainer Profile Sprite Settings a Visual List
**Current Behavior**
Currently in settings, you can pick your trainer sprite from a list of numbers, which goes all the way from Trainer 1 to Trainer 118 

**Improved Behavior**
You don't need to remove the "Trainer X", but it would be nice to see what each trainer number sprite looks like.

**Reason for change**
Frankly, for those that care about customizing their profiles, it's extremely tedious to have to pick every trainer one after the other to see if you like it, all the way to 118.

**Detailed Description**
It doesn't exactly need much detail, and I don't know how hard it would be, but all you would need to do is add the picture of the sprite you would get from the trainer number to the side. That way all you need to do is scroll down the list until you see a sprite you like, without having to check out each one before you decide.

**Screenshots:**
It won't show the dropdown list when I take a screenshot, but you know how it's just the list of "Trainer X, Trainer X+1...
Screenshot 2021-11-17 at 11-03-54 Pokéclicker
"
It might be nice to do the same for the backgrounds, but since there's way fewer of them compared to trainers, it's not as important. Of course, this whole suggestion is by no means urgent, but it doesn't seem too hard to fix. Although, I know nothing of coding, so maybe it would be.

RedSparr0w:
Allow players to select a trainer or background image by clicking an image rather than a dropdown of numbers.

Closes #1646 

image
image
image

### DIFF
--- a/src/components/profileBackgroundModal.html
+++ b/src/components/profileBackgroundModal.html
@@ -1,0 +1,24 @@
+<div class="modal noselect fade" id="profileBackgroundModal" tabindex="-1" role="dialog" aria-labelledby="profileBackgroundModalLabel">
+    <div class="modal-dialog modal-dialog-scrollable modal-dialog-centered modal-sm" role="document">
+        <div class="modal-content">
+            <div class="modal-header" style='justify-content: space-around;'>
+                <h5 class="modal-title">Select a Background</h5>
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close">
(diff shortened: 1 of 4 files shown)


Aegyo (approved):

blacknazgul13:
Thanks for implementing my suggestion! I'm surprised that no one's actually suggested that until me... XD

Issue #1153, opened 2020-12-20
[MODULES] - The easy ones 


Aegyo:
We want to move all of the code from `src/scripts/` into typescript modules, for the various improvements that will give us in typechecking and circular dependency avoidance. More information on #871

Anything moved into a module will become unavailable to anything that does not import it. To keep other things happy without also converting them, we have `src/modules/temporaryWindowInjection.ts` to make available anything used in `src/scripts/` and `src/components/` that we have moved into a module.

The general process I have done so far is:
1. Move files into `src/modules/`, creating a folder if necessary
2. Add `export ` in front of anything that should be exported. If there is only one thing, use `export default `
3. Add `import` statements for anything the file has lost knowledge about. An IDE like visual studio code makes identifying those things very easy
4. The eslint config in `src/modules/` is different to `src/scripts/`, so there may be some code style changes you need to fix. Sometimes these stricter eslint rules force you into something which looks worse, in these cases you can disable the rule for that line or try altering the config to something more in line with the existing code style we have
5. Update temporaryWindowInjection with anything used outside of `src/modules/`
6. Update any  `/// <reference` in `src/scripts/` pointing to a file you have moved to the equivalent `src/declarations/` file
6. Run `npm start`, to confirm everything compiles, and fix anything it complains about
7. Check the browser console for any errors, and test the functionality of whatever you have converted is unchanged

*Still adding to this list, but submitting now in case I get distracted*

- [x] **Changelog**
Entirely self contained, should only need changeLogItems exported for window injection (used in changelog.html)
- [x] **Amount and AmountFactory**
Super simple classes, both need to be injected into window. Wallet references App multiple times, so we'll leave this one in scripts for now.
- [x] **SpecialEvent and SpecialEventsNotifiedStatus**
These two shouldn't have any problems converting, but SpecialEvents wants RoamingPokemon which will pull in a whole chain of things, so leave that one in scripts.
- [x] **utilities/Information.ts**
- [x] **party/Category.ts**
Doesn't look like it has any dependencies.
Can update the type of `TmpGameType -> categories` in `module/globals.ts` after moving this.
- [x] **scripts/oakitems/*.ts**
Can be moved when #1299 is merged.
Can update the types of `TmpGameType -> oakItems, oakItemLoadouts` after moving this.
- [x] **Most Requirements**
I haven't looked through all of them, but most don't seem like they will be any problem.
[...]

PR #1650, opened 2021-11-19
Category to modules

RobinLaevaert:
#1153


RedSparr0w (approved):
Thanks for the PR! 👍 

Sorry it's taken so long to merge this, but thanks to @Aegyo for fixing the gulp script that was causing the tests to fail 🎉

PR #1664, opened 2021-12-04
Johto dungeon trainers

Ultima1990:
It begins...


Ultima1990:
Team Rockets Hideout in Mahogany is here! Immediately undoing my work in #1634 by adding a new dungeon that exists only in a town. There's just no space around Mahogany.

Anyway, requirements are updated to include the hideout. Everything that was once locked behind Mineral+Glacial is now locked behind the hideout, which itself is of course locked behind Mineral+Glacial. Did the update.ts thing to move the dungeon indexes.

This is where I hit a bit of a roadblock. I have to think about how to handle the Goldenrod situation. That place is effectively 3 dungeons. Radio tower, underground tunnel/warehouse, and then radio tower again. So yea, gotta figure that out. Currently I'm leaning towards just smashing that all together into one dungeon. Let me know any other ideas.

Askellown:
it makes sense to me to just make it Radio Tower, since the only battle worth talking about in the UG path (Rival) not really being relevant in Pokeclicker

Ultima1990:
Dramatic finish!

RedSparr0w (approved):

PR #1666, opened 2021-12-08
Challenge modes - Regional Dex/Attack

RedSparr0w:
Added 2 new "challenge" modes.
Both of these challenges are how the game is already currently played.

`Regional Attack Debuff`: Pokémon receive an attack debuff based on if it's their native region and the highest region you have reached.
`Require Complete Pokedex`: Requires completing the current regions pokédex before moving on to the next region.

Added a popup for enabling/disabling the challenges upon save loading:
image

Challenges can now be disabled before you have selected a starter.
_(may change it in the future that they can always be disabled but not enabled)_

@Ultima1990 If you could please check the requirements I set for the starter towns, would be much appreciated.

I haven't tested the `Require Complete Pokedex` challenge yet, as I don't have any saves at that point 😅


RedSparr0w:
> I haven't tested the Require Complete Pokedex challenge yet, as I don't have any saves at that point 😅

Have now tested and fixed it, should be all good.

Ultima1990:
The starter town requirements look all right to me.

PR #1667, opened 2021-12-10
The Castform Shuffle

Ultima1990:
Base Castform removed from route 119, now a boss in the Weather Institute. Requires at least one Weather Institute clear so that players will always have to defeat Aqua Admin Shelly at least once.

The other Castform forms remain on 119 but now require owning base Castform in addition to the weather requirement.

Removed fog from the possible weathers for Castform (snowy). Don't know why that was there.

Additionally a small bug fix, added Weather Institute to the requirements for route 120.


RedSparr0w:
> Removed fog from the possible weathers for Castform (snowy). Don't know why that was there.

I think fog was there as that is how it can also appear in Pokémon GO IIRC

amative1:
That is correct, it's a POGO addition due to the possibility of lack of snow in tropical areas of the world (since it's tied to IRL weather)

Ultima1990:
Fair enough. The fog's back.

RedSparr0w (approved):
Looks good 👍 Thanks for the PR!

PR #1672, opened 2021-12-12
Change several dungeon pokémon in prep for dex removal

Ultima1990:
We have a bunch of dungeons in locations that have no wild encounters in the main series. These have been filled by the creative (or sometimes lazy) minds of our developers. The result is that sometimes there's some whacky stuff.

This PR tackles a bunch of encounters that would allow players to skip a bunch of grinds for stones, eggs and other QP related purchases. We already get people saying QP is worthless, we don't want to make it worthlesser (that's a word now).

Yes, situations like Gabite in Victory Road already let you skip some of the dragon egg grind, yes several routes in Sinnoh have Roselia so it's only eggsclusive in Hoenn. But that's just how these things are in Platinum. This PR _only_ tackles dungeons that have had their entire roster of Pokémon (sans that one legendary that you go to that dungeon for) thought up by us because there are just no random wild pokémon in those locations in the source. So here we go.

Umbreon and Espeon in Newmoon and Fullmoon island have been replaced. Go buy some Soothe Bells!

Vileplume, Exeggutor, Bellossom and Sunflora in Flower Paradise have been replaced. Go get those Leaf and Sun stones!
The bosses Venusaur and Meganium have been replaced. Better grind those Grass Eggs!
Roselia has NOT been replaced because it exists in the wild in Sinnoh.

I have not tackled Hall of Origin, it is a bit of a special case.

Sunflora in Liberty Garden has been replaced. Sunstone.
The bosses Ninetales and Alakazam have been replaced. Fire and Trade Stone.
Sunkern gets to exist in a dungeon now despite being Eggsclusive before, since it also exists as a wild encounter in Unova.

Scizor and Porygon2 in P2 Laboratories have been replaced. Metal Coat and Upgrade.
The bosses Porygon, Porygon-Z and Escavelier have been replaced. Go buy that Porygon, that Dubious Disc and that Trade Stone!
Metang has NOT been replaced, since it is a canon wild encounter in Giant Chasm.


RedSparr0w (approved):
Nice work, everything looks good to me, thanks for the PR! 👍

PR #1673, opened 2021-12-12
Remove battle items from stores

RedSparr0w:
Unsure of this PR, please provide any feedback.

Currently battle items can be easily obtained in great quantities fairly early in the game,
This PR attempts to address this by removing them from stores, so the only way to obtain them is via running dungeons _(I think some route mons also drop them as rare drops?)_
Could also be added as rewards for defeating dungeon trainers, which allows you to get the items, without increasing the dungeon mons health.

It removes another use for money which could be a problem, and also makes the department stores kind of useless.

### DIFF
--- a/src/scripts/towns/Town.ts
+++ b/src/scripts/towns/Town.ts
@@ -58,12 +58,6 @@ const pokeMartShop = new Shop([
     ItemList['Pokeball'],
     ItemList['Greatball'],
     ItemList['Ultraball'],
-    ItemList['xAttack'],
-    ItemList['xClick'],
-    ItemList['Lucky_egg'],
-    ItemList['Item_magnet'],
-    ItemList['Token_collector'],
-    ItemList['Lucky_incense'],
     ItemList['SmallRestore'],
     ItemList['MediumRestore'],
     ItemList['LargeRestore'],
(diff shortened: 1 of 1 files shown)


scolpitts536:
One. Expect a ton of complains of people having to run dungeons to get battle items
Two. Might as well just get rid of battle items in the case.
Three. That is a stupid thing to change.

amative1:
Suggestion: either limit the number that can be bought early-game, or remove them from early-game (rather than just removing them outright).

apple096:
I feel like the better way is to make battle item price scaling harsher so it's harder to bulk buy. This is too much

Claygame:
make it so only 1 battle item can be bought in a region, like xattack is in kanto, xclick is in johto, etc along with the price change that apple suggested, but you can still find them all in all chests of all dungeons of all regions, so that people in later regions can bulk buy all but people in early regions can buy only some

NiCeDiCe90:
i don't know about this. It feels like removing a part of the game just to make it go slower.
I can understand that the items feel OP for the early game but still not a fan of this solution.

Another option could be to level them up over time like oak items. Starting with 10 seconds time and/or 10% buff and making them stronger over time.
Could make it cost some gen4/5 berries or give them as an reward for 1k dungeons clears. To give the player an reward for doing these things.

RedSparr0w:
Thanks for the feedback people 👍

PR #1677, opened 2021-12-14
More evolution locations

Ultima1990:
This covers the following location based evolutions:

Moss Rock: Leafeon
Ice Rock: Glaceon
Special magnetic field: Magnezone, Probopass (and in Alola Vikavolt)

Since soon people can choose to move on from Sinnoh before evolving these Pokémon, I thought it'd be nice to add these locations to the later regions.
These are dungeon evolutions, so I have taken some creative liberties. Similar to the current Sinnoh Ice Rock situation, which we moved from route 217 to Lake Acuity.

I have moved the Kalos Moss Rock from route 20 to Pokémon Village.
I have moved the Kalos Special magnetic field from route 13 to Kalos Power Plant.

ORAS added these locations to Hoenn, but I have not added them because these are firmly Sinnoh or later evolutions. For the same reason the Vikavolt evolution has not been added to the Sinnoh, Unova and Kalos Special magnetic fields.


RedSparr0w (approved):

PR #1678, opened 2021-12-15
Remove Mahogany Fighting Egg

Ultima1990:


RedSparr0w:
Is there a reason for the removal?

Qwertypop4:
There's no reason to have the same egg twice in the same reason

Ultima1990:
It's a duplicate. The same egg is also in the CianwoodCityShop, which makes more sense since Cianwood has the fighting gym.

RedSparr0w (approved):
Sweet, thanks for the PR 👍

PR #1681, opened 2021-12-15
Change several Johto requirements

Ultima1990:
I got the order of events wrong when adding the Team Rocket dungeons. Hideout before Pryce. You can't even enter Pryce's gym before doing the Hideout. Some dude blocks the way. So the order is now route 43, Hideout, Pryce(+ Jasmine), Radio Tower. 43 first because I am counting Lake of Rage as part of 43 and the red Gyarados thing must be done before Hideout.
Radio Tower now requires Mineral+Glacier.

I have also changed other things for the sake of why not.
You see, in the gen 2 games you can walk all the way up to Lake of Rage before beating Morty. You can't actually go into the lake, because that requires surf, but you can walk up to the shoreline if you want. This is possible because you can avoid the water on route 42 by going through Mt Mortar. You can enter Mt Mortar before encountering any water (or grass). SO:

Mt Mortar now requires 37, same as Ecruteak.
Mahogany now requires 42 OR Mt Mortar.
43 is special, because like I said before I count Lake of Rage as part of 43, which requires surf, so
Route 43 now requires (Mt Mortar AND Fog badge) OR route 42. 42 requires Fog as usual so all good there.

As mentioned above the Mahogany Gym now requires Hideout and Hideout requires 43, so in the end you can now get to Mahogany early and... do nothing. I guess you can now get Upgrade early. Yay! Doesn't actually change much, but the option is there if you want. I didn't touch Mt Mortar's difficulty at all and it's tuned to be rather difficult for this stage of the game, so if you want to do this you'd better get your grind on.

The quest now starts after Fog badge. I'd love for it to start after route 43, but I don't see a way to start a quest after a route clear and can't be arsed creating one. Fog implies you can clear 43 because surf, so Fog is the quest starter.

I think this is the first instance of a MultiRequirement inside a OneFromManyRequirement. The reverse has happened a couple of times.

### DIFF
--- a/src/scripts/quests/QuestLineHelper.ts
+++ b/src/scripts/quests/QuestLineHelper.ts
@@ -237,7 +237,7 @@ class QuestLineHelper {
     }
 
     public static createRocketjohtoQuestLine() {
-        const RocketjohtoQuestLine = new QuestLine('Radio Tower Takeover', 'Team Rocket is up to no good again!');
+        const RocketjohtoQuestLine = new QuestLine('Team Rocket', 'Team Rocket is up to no good again!');
 
         const clearTeamRocketHideout = new CustomQuest(1, 0, 'Clear the Team Rockets Hideout dungeon in Mahogany Town', () => App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex('Team Rockets Hideout')](), 0);
         RocketjohtoQuestLine.addQuest(clearTeamRocketHideout);
--- a/src/scripts/wildBattle/Routes.ts
+++ b/src/scripts/wildBattle/Routes.ts
@@ -396,7 +396,6 @@ Routes.add(new RegionRoute(
         water: ['Tentacool', 'Tentacruel', 'Krabby', 'Magikarp', 'Staryu', 'Corsola', 'Kingler'],
     }),
     [
-
         new RouteKillRequirement(10, GameConstants.Region.johto, 39),
         new GymBadgeRequirement(BadgeEnums.Fog),
     ]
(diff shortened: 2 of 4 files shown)


src/scripts/wildBattle/Routes.ts
```diff
@@ -424,7 +423,15 @@ Routes.add(new RegionRoute(
...
         headbutt: ['Venonat', 'Exeggcute', 'Hoothoot', 'Pineco'],
     }),
-    [new RouteKillRequirement(10, GameConstants.Region.johto, 42)]
+    [
+        new OneFromManyRequirement([
+            new MultiRequirement([
+                new ClearDungeonRequirement(1, GameConstants.getDungeonIndex('Mt Mortar')),
+                new GymBadgeRequirement(BadgeEnums.Fog),
+            ]),
+            new RouteKillRequirement(10, GameConstants.Region.johto, 42)
```
amative1, line 432: build fail, should this be `new RouteKillRequirement(10, GameConstants.Region.johto, 42),` instead (comma at the end)

Ultima1990:
Commas are the bane of my existence lol.

Also my develop branch is out of date oooops.

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -237,7 +237,7 @@ class QuestLineHelper {
     }
 
     public static createRocketjohtoQuestLine() {
-        const RocketjohtoQuestLine = new QuestLine('Radio Tower Takeover', 'Team Rocket is up to no good again!');
+        const RocketjohtoQuestLine = new QuestLine('Team Rocket', 'Team Rocket is up to no good again!');
```
RedSparr0w, line 240: Aw I liked the old name, although it kinda sounds like you're taking over the radio tower 😅

Ultima1990: Name change because the questline doesn't just involve the tower anymore.

RedSparr0w (approved):
Thanks for the PR! 👍

PR #1683, opened 2021-12-16
Some balance tweaks for the new loot system

NiCeDiCe90:
- removed "Passho" from Kantos diglett cave and added the "Wiki" berry instead
- increased chance to find Dome and Helix fossils in Mt. Moon (weight: 0 -> 1)
- removed Omanyte and Kabuto as boss encounter from Mt. Moon


scolpitts536:
I'm okay with everything except removing the bosses, they are there in-game.

NiCeDiCe90:
I don't understand your point.
In the original games you only get the fossils in Mt. Moon. That will be the case in Pokeclicker too with this change.

scolpitts536:
Well you would be making them underground or mt.moon/cerulean cave chests and ngl i already know how bad the complaints for not getting underground rng is with fossils.

NiCeDiCe90:
People always complain but:
In Kanto these fossils are the only fossils in the underground pool so it's very likely to get them early.
With the dex requirement removed you don't need them to move on. There are a also lot of dungeons where you get these fossils later in the game.
So there are plenty of ways to get them.

scolpitts536:
Hmm fair point.

RedSparr0w:
Any changes should still be balanced around the dex requirement being a thing, as that's still the default way to play.

We could make it that when you defeat the dungeon for the first time you are awarded 1 of the fossils though as you would normally get 1 in game.

RedSparr0w (approved):

PR #1684, opened 2021-12-17
Fixup several sprites and names in Alola

Ultima1990:
Moved a couple of sprites from Trainers to gymLeaders.
Fixed every instance of "Hao".
Created double battle sprites for Aether Foundation Employees.
Combined some Aether Foundation Employees into their double battles.
Fixed sprites for Faba, Guzma, Lusamine and the Actor. Meredith really gets around.
Removed some Pokémon from Faba's team. He appears there in a double battle with a grunt but I wanted to have him solo in his first appearance so I removed the Pokémon that I assume belong to the Employee based on Faba's teams in postgame.
Renamed the Nebby dungeon and gave it a new background for shits and giggles.


RedSparr0w:
Everything seems to be all good, although not sure about "The Bag"

Ultima1990:
I wasn't sold on "The Bag" either. But the same was true for "Nebby". So here's a complete overhaul of where and when to obtain Cosmog, Cosmoem, Lunala and Solgaleo.

The player does obtain Lunala or Solgaleo at the altar in USUM, but so far I have not found a way to make that work. So I switched to the Lake of the Sunne/Lake of the Moone where the player can obtain Cosmog in the postgame. The Bag/Nebby is now the Lake of the Sunne and Moone, requires Elite_AlolaChampion, and has Cosmog as a catchable boss. Cosmog will evolve into Cosmoem and then the box legendaries. I also devolved all of the stone evos in that dungeon.

Since I do want Lunala and Solgaleo to be catchable someplace for future DNA shenanigans, they also exist in this dungeon, but are locked behind themselves. Which means you have to catch and fully evolve Cosmog before Lunala and Solgaleo can be encountered.

Qwertypop4:
Isn't it kinda weird that Solgaleo/Lunala have significantly less health than Cosmog?

Ultima1990:
Isn't it weird that absolutely nothing in Alola is balanced yet?

Qwertypop4:
Oh right, I just saw that they had been decreased and presumed it was a weird intentional decision lol

Ultima1990:
A quick behind the scenes look into the order of operations:

First I renamed the dungeon, removed one of the bosses and turned the other into Cosmog, and changed the requirement. Then later on I decided that I do want Lunala and Solgaleo as initially unavailable bosses, so I copied Ho-oh down because that conveniently also copies the requirement I needed. Which is why the health difference happened.

PR #1685, opened 2021-12-17
Magnet/Collector swap

### LINKED ISSUE #1575: [SUGGESTION] Swap Magnet and Token Collector In Pokemart
In the explorer's pokemart (the one you can access anywhere by hitting the button) everything is in order. The balls are ordered worst to best, the potions are ordered lowest to highest, and all of the Battle Items are in the same order they are shown in their menu with the exception of these two being swapped. It has led to me making mistakes when trying to buy a specific amount of each Battle Item and overall it would just make more sense for them to be arranged this way.

jk13pclick:
Fixes #1575. Swaps magnet and token collector in shop menu to reflect correct battle item modal order. This also fixes item order with respect to price.


RedSparr0w (approved):
Thanks for the PR! 👍

PR #1690, opened 2021-12-19
Change the order of Evolution Stones

Ultima1990:
Not a big deal. It's just been sorta bugging me.

Trade stone should really be higher up. It is used in Kanto, but somehow is listed after Sun Stone.
And how about that Prism Scale?! All the way down below the gen 4 stones! How silly.

And then I just re-ordered some other stones within their generation based on some vague combination of when they become available, when they become useful and the dex number of the Pokémon they are used on.

While the change really only needed to be done in Gameconstants.ts in order to make this happen, I decided to do it in Gameconstants.d.ts and EvolutionStone.ts too for the sake of consistency.


RedSparr0w (approved):

RedSparr0w:
> While the change really only needed to be done in Gameconstants.ts in order to make this happen, I decided to do it in Gameconstants.d.ts and EvolutionStone.ts too for the sake of consistency.

👍 Thank you 👍 

Also worth noting, that `Gameconstants.d.ts` does need to be updated when updating `Gameconstants.ts` as this is the one only declarations file that doesn't get automatically updated.

PR #1693, opened 2021-12-19
Change Distortion World dungeon appearance

Ultima1990:
Here's what happens. You first enter Sinnoh and it looks like this:
Sinnoh1
You'll notice an empty spot where the Distortion World dungeon used to be.

Then you defeat Snowpoint Gym and suddenly the map looks like this:
Sinnoh2
Yep. Just on top of the Hall of Origin. Because I can't think of any situation where that dungeon would be accessed before becoming champion of Sinnoh.

So it blocks a dungeon. But that's fine, because after one Distortion World clear:
Sinnoh3
Crisis averted. Back in its old location.

Cuz it's just cooler this way.


Ultima1990:
I thought the 214 extension looked kinda weird without the cave there.

RedSparr0w (approved):
PR looks good, thanks! 👍 

> I thought the 214 extension looked kinda weird without the cave there.

It does look a little odd, but also looks odd with it missing once it is there (which will be a majority of the time)

Issue #1668, opened 2021-12-10
[SUGGESTION] Implement new berry aura to improve Roamer encounter odds


amative1:
**Current Behavior**
Currently there are numerous berries with aura effects, and a number of Gen 4 and 5 berries with no effect.

**Improved Behavior**
I propose to add an additional effect to Jaboca, to slightly increase Roamer encounters.

**Reason for change**
Jaboca currently has no effect, even though its description states "The sound of these Berries attracts wild Pokémon", implying that there should be some kind of effect.

**Detailed Description**
The increase could be as little as 1.02x or 1.03x, but would provide additional incentive to working towards a full Berrydex.

NiCeDiCe90:
We had a discord thread about possible berry effects (Roamer chances was one of suggestions) and it got more or less denied because "the farm is flawed from the start" and needs rework. 

So it doesn't make sense to add new berry effects until this rework happens I guess...

amative1:
Closed per #1694

PR #1694, opened 2021-12-19
Roaming Pokémon Berry

jk13pclick:
This one might be controversial, so please provide feedback.

Addresses #1668. Roaming Pokémon Aura added, along with bonus to appropriate multiplier files. I did a bit of testing, and it does seem that the changes make an impact on Roamer odds (although more eyes/testing would be ideal). Starf bonuses (0.005, 0.01, 0.015) used instead of Lum, for math reasons below.

Min Roamer Chance (no Jaboca, min route, no increased chance) - 1/8192
Max current Chance (no Jaboca, max route, increased chance) - 1/1365

**Jaboca Modifications**

Max 1.015x Jaboca Roamer Chance (21 Jaboca w/ Lum Boost, max route, increased chance) ~ 1/394 - This seems too high for the absolute max, basically triple what the current max is. Lum boost is OP.

Max 1.03x Jaboca Roamer Chance (21 Jaboca w/ Lum Boost, max route, increased chance) ~ 1/536 - This seems good (maybe even still too high) for the absolute max, more than double what the current max is.

**_Summary_**

Jaboca is a very good berry for a niche reason, it is excellent for Roaming events and region completion. To address NiceDice's concerns with the Farm, I generally agree; the Farm is flawed and needs a rework. I just don't think that means we can't improve upon the flawed concept in the meantime. No devs have picked up the Farm and a rework would likely come after other reworks/rebalancing in progress. As mentioned in the beginning I expect some resistance against this change, Let me know what you all think.

### DIFF
--- a/src/modules/multiplier/MultiplierType.ts
+++ b/src/modules/multiplier/MultiplierType.ts
@@ -6,6 +6,7 @@ enum MultiplierType {
     dungeonToken,
     shiny,
     eggStep,
+    roaming,
 }
 
 export default MultiplierType;
--- a/src/scripts/farming/Farming.ts
+++ b/src/scripts/farming/Farming.ts
@@ -51,9 +51,11 @@ class Farming implements Feature {
         this.externalAuras[AuraType.Attract] = ko.observable<number>(1);
         this.externalAuras[AuraType.Egg] = ko.observable<number>(1);
         this.externalAuras[AuraType.Shiny] = ko.observable<number>(1);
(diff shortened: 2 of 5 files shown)


NiCeDiCe90:
A lot of people hate roamers for the low odds. I would be fine with 1/536.
With this it would be very doable to farm shiny roamers. A6CSRQ players will love and hate you for it.

scolpitts536:
Lmao

RedSparr0w (approved):

We can adjust the boost later on if it seems too high/low.

PR #1696, opened 2021-12-20
Loot Hotfix #1

jk13pclick:
Magnet nerf, will need a buff. I want to experiment a bit to get a good formula, but this will nerf it for now


src/scripts/dungeons/DungeonRunner.ts
```diff
@@ -98,7 +98,7 @@ class DungeonRunner {
             const magnetChance = 0.5 / (4 / (loot.weight + 1));
             if (Rand.chance(magnetChance)) {
                 // Gain more items in higher regions
-                amount *= (GameConstants.getDungeonRegion(DungeonRunner.dungeon.name) + 2);
+                amount += (Math.round(Math.max(loot.weight,2) / 8 * (GameConstants.getDungeonRegion(DungeonRunner.dungeon.name) + 1)));
```
RedSparr0w, line 101: Should we make this:
```suggestion
                amount += Math.max(1, Math.round(Math.max(loot.weight,2) / 8 * (GameConstants.getDungeonRegion(DungeonRunner.dungeon.name) + 1)));
```
So it will always add at least 1 extra?

jk13pclick: Yeah that makes sense, good catch

PR #1704, opened 2021-12-20
Fixing the OP Loot and NERFing it

scolpitts536:
fixing the loot


Ultima1990:
Those are some big diffs lol. Should probably fix the path on these. They are in the wrong place. Should be pokeclicker/src/scripts/dungeons/Dungeon.ts, not just pokeclicker/Dungeon.ts.

jk13pclick:
Oh lol, I was wondering what was going on with that lol

scolpitts536:
> Those are some big diffs lol. Should probably fix the path on these. They are in the wrong place. Should be pokeclicker/src/scripts/dungeons/Dungeon.ts, not just pokeclicker/Dungeon.ts.

yeah i just fixed that. gonna work on johto now

PR #1705, opened 2021-12-20
Miguel Must be Stopped (Nerf the Mt. Moon Fossil Reward)

jk13pclick:
Nerfs the Reward function to a 1/10 chance, could be updated to allow a custom value


RedSparr0w:
I messed up, It was meant to be a 1 time reward, not a constant thing, will fix it in a different PR.

Didn't realise the dungeon rewards were an always thing, thought they were the same as Gyms 😅

PR #1707, opened 2021-12-20
Loot fixup observable

jk13pclick:
Pokéballs obtained in dungeons should probably add to the observable, thanks Vic for the report.


RedSparr0w:
Should we be doing this for berries too?

Not sure if any of the other things are tracked at all 🤔

jk13pclick:
Berries should be tracked as well for quest reasons, might even make berry quests more doable

RedSparr0w:
Ah didn't realise it's already incrementing the berries stats 👍

RedSparr0w (approved):

PR #1719, opened 2021-12-21
Update index.html

amative1:
Hover text in Battle Frontier currently displays Hoenn-damage rather than region-free-damage. This should fix it.


RedSparr0w (approved):

PR #1722, opened 2021-12-22
reduce BP cost of Furfrou (Kabuki)

NiCeDiCe90:
Turns out to be a massive grind and I dont want to keep people hostage in the BF forever.
So BP cost: 150k -> 75k


RedSparr0w (approved):
👍 Thanks for the PR!

PR #1725, opened 2021-12-23
Add item name types

RedSparr0w:
This adds "Item Name" types

Should stop any items in dungeons having incorrect names.


RedSparr0w:
We should probably move to using BagItem as the item at some point, as then we could make use of the functions there:
`BagHandler.gainItem()`
`BagHandler.displayName()`
`BagHandler.image()`

jk13pclick (approved):
All looks good, if I understand this correctly the massive chunk of if else statements in DungeonRunner can now be condensed using BagItem. Maybe some observables will have to be checked for, but this should make it way simpler (and efficient)

RedSparr0w:
Pretty much, the `BagHandler` should deal with the observables/statistics _(although tbf I haven't ever checked if it was actually added for everything)_

But would remove the `DungeonRunner` item adding stuff, so it would all be more consolidated

Issue #1700, opened 2021-12-20
[SUGGESTION] New Challenge - "Harder Dungeons"


NiCeDiCe90:
Some of the newest additions to the game makes the game (dungeons) too easy for people who love the insane challenges like A6CSRQ (or now A8C i guess).
Maybe we can make it harder again. 

So here is my Idea:
Add a 9th challenge: "Harder Dungeons"

-> removes all dungeon shrinking
-> removes "Flash" effect on 200+ clears

-> removes all items with a weight of 2 or lower

Not sure about the last one though...
The last point is a bit weird for "All Challenges" since they would get mostly battle items again that they can't use.

The Problem for SRQ Challengers are the high numbers of the weight 1 and 0 Items you get after 10k++ clears. So removing these and let them only get "useless" Items might be the way they want to play.

scolpitts536:
Instead of removing dungeon clears make it kalos, alola, and galar can go down to 9x9 or we won't finish the dungeons when alola and galar comes out, I would keep flash, and make it remove item weight with 1 or lower that way we can get small restores and its only underground items and small restores and still excludes eggs, fossils, and evo items.

jk13pclick:
You don't need to remove the items, just adjust the (currently) existing max clears restriction on the weight formula. Ie: Max it at 10 and low weight items will still be hard to get even at 1000 clears. Removal of the Dungeon perks is good, maybe even bump the health modifier from opening chests.

NiCeDiCe90:
closing this in favor of the new suggestion for challenge reworks: #2332

PR #1726, opened 2021-12-23
[WIP] Hard Dungeon Challenge

jk13pclick:
#1700. Experimental, ~~untested~~, and would like lots of feedback. Feel free to reject. I considered messing with health, but decided just a revert to Dungeons pre-buffs is probably good enough.


jk13pclick:
Tested, new Challenge works well. For some reason it doesn't stay activated, but I may be missing something obvious. Going to mark it as ready for review, but still needs a little work + an Update.ts activation.

src/scripts/dungeons/DungeonRunner.ts
```diff
@@ -36,11 +36,21 @@ class DungeonRunner {
...
+            // Decrease dungeon size by 1 for every 10, 100, 1000 etc completes
+            dungeonSize -= Math.max(0, App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]().toString().length - 1);
+            DungeonRunner.map = new DungeonMap(Math.max(GameConstants.MIN_DUNGEON_SIZE, dungeonSize), flash);
+        } else {
+            const flash = false;
+            DungeonRunner.map = new DungeonMap(Math.max(GameConstants.MIN_DUNGEON_SIZE, dungeonSize), flash);
+        }
+
+
+
```
RedSparr0w, line 54: Might be worth simplifying this a little bit and removing some duplication
```suggestion
		let flash = false;

        if (!App.game.challenges.list.hardDungeons.active()) {
        	// Reveal dungeon tiles nearby player
            const flash = App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]() >= 200;
            // Decrease dungeon size by 1 for every 10, 100, 1000 etc completes
            dungeonSize -= Math.max(0, App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]().toString().length - 1);
        }

	    DungeonRunner.map = new DungeonMap(Math.max(GameConstants.MIN_DUNGEON_SIZE, dungeonSize), flash);
```

src/scripts/dungeons/Dungeon.ts
```diff
@@ -137,10 +137,14 @@ class Dungeon {
             if (loot.requirement && !loot.requirement.isCompleted()) {
                 return 0;
             }
-            // Minimum of 1 times cleared for division
+            // Minimum of 1 times cleared for division, maximum of 500 times cleared
             const timesCleared = Math.min(500, Math.max(1, App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(this.name)]()));
             // Calculate total weight based on times cleared, minimum weight being original number specified
-            return Math.max(loot.weight, Math.pow(10, loot.weight) / timesCleared) + 1 || 1;
+            if (App.game.challenges.list.hardDungeons.active()) {
+                return Math.max(loot.weight, Math.pow(10, Math.pow(loot.weight, loot.weight)) / timesCleared) + 1 || 1;
```
RedSparr0w, line 144: I think this might be a little too intense? 😅 

image

src/scripts/dungeons/Dungeon.ts
```diff
@@ -137,10 +137,14 @@ class Dungeon {
...
+            } else {
+                return Math.max(loot.weight, Math.pow(10, loot.weight) / timesCleared) + 1 || 1;
+            }
```
RedSparr0w, line 147: Maybe something like this might work better?
```suggestion
			let power = App.game.challenges.list.hardDungeons.active() ? 20 : 10
			return Math.max(loot.weight, Math.pow(power, loot.weight) / timesCleared) + 1 || 1;
```
10 power:
image

15 power:
image

20 power:
image

NiCeDiCe90: I think 15 is ok.

src/scripts/dungeons/Dungeon.ts
jk13pclick: 😆 Yes, it was inteded to be OP, no more whining from Challenge Mode players. I figured you might want it changed, but thought leaving it high to start was ok.

src/scripts/dungeons/Dungeon.ts
jk13pclick: 15 or 20 works, but I think I would want to Max timescleared at some low number as well. 1.61% doesn't feel like a big number, but with a big amount of chests per dungeon (especially with no size reduction) it comes out to a fairly frequent rate per run.

jk13pclick:
Closing this. I think we have moved beyond the idea of a "Hard Dungeon Challenge", especially with respect to current loot plans.

PR #1730, opened 2021-12-24
Stones and Eggs everywhere

Ultima1990:
After the region they first show up in of course.

Did I miss any?


RedSparr0w:
Sorry, looks like there's some code conflicts that need fixing before merge.

src/scripts/towns/Town.ts
```diff
@@ -371,9 +371,6 @@ TownList['Cerulean Cave'] = new DungeonTown(
 const NewBarkTownShop = new TownShop([
     ItemList['Pokeball'],
 ]);
-const CherrygroveCityShop = new TownShop([
-    ItemList['Greatball'],
```
RedSparr0w, line 375: Are these still available elsewhere in the region?

RedSparr0w (approved):
Thanks for the PR! 👍

src/scripts/towns/Town.ts
RedSparr0w: Nvm, found em 👍

Ultima1990: Yes they are in Ecruteak.

PR #1731, opened 2021-12-24
Remove Master Balls from chests

Ultima1990:
As far as I'm concerned ONE free masterball per region is fine. But repeatable grindable master balls from chests? Nah. Too OP.


DMaurin:
But why ?

RedSparr0w (approved):

PR #1732, opened 2021-12-24
Add Aqua Magma quest in Hoenn

Ultima1990:
I am a benevolent dev, I take away Masterballs, I give back Masterball... ye just the one.

This was kind of weird, so I tried to create a standard that can be used for these questlines going forward.

The main purpose of the questline is to reward the Masterball. In addition it'll give the player some guidance if they are lost on what to do. In RSE the Masterball is found in the Aqua Hideout. But it's kind of weird to have a quest for the Aqua hideout and not one for the Magma hideout. So it should cover both. But also it's kind of weird to have a questline covering Aqua and Magma shenanigans that just ends at the Hideout when there are more shenanigans to be had at the Seafloor Cavern. But then it's weird to have the Masterball reward at the Hideout when there's another quest in the line after that. But also I don't want the questline to cover the entire region.

So here's the new standard I'm proposing for all the villain team questlines going forward:
1: Every dungeon that has a _named_ villain team bad guy as a boss should be part of the questline. Preferably this named bad guy will be mandatory, at least on the first clear.
    This means Mt. Moon will not be in the Kanto questline. They are just nameless grunts, and they aren't even bosses!
    It also means Petalburg Woods and Rusturf Tunnel are not in this questline, because the boss is just a nameless grunt.
2: Push the Masterball to the end of the questline.


src/scripts/quests/QuestLineHelper.ts
```diff
@@ -258,6 +258,37 @@ class QuestLineHelper {
         App.game.quests.questLines().push(RocketjohtoQuestLine);
     }
 
+    public static createAquaMagmahoennQuestLine() {
+        const AquaMagmahoennQuestLine = new QuestLine('Land vs Water', 'Put a stop to the schemes of Team Aqua and Team Magma');
+
+        const clearMtChimney = new CustomQuest(1, 0, 'Stop Team Magma at Mt. Chimney', () => App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex('Mt. Chimney')](), 0);
```
RedSparr0w, line 264: Just checking that you expect these quest to auto complete if someone has already defeated all these dungeons?

Ultima1990: Yes that is what I expect.

RedSparr0w (approved):
Nice work, thanks for the PR 👍

PR #1717, opened 2021-12-21
Observable fixup #2

jk13pclick:
Might want to consider adding an increment observable to the actual gainPokeballs method. Maybe want to split up observables for "Bought" and "Obtained"

PR #1733, opened 2021-12-24
Update AchievementHandler.ts

amative1:
Changed text for ball cheevos from Obtain to Purchase, due to some balls now being obtainable via other means.


apple096 (changes requested):
People seems to forget the Update function, so this change will break 100% of saves

amative1:
Interesting, as testing with multiple save files (both current and a few versions back) show no issues.

PixLSteam:
> People seems to forget the Update function, so this change will break 100% of saves

Wouldn't that only be relevant if the name of the achievements changed?

apple096:
> > People seems to forget the Update function, so this change will break 100% of saves
> 
> Wouldn't that only be relevant if the name of the achievements changed?

Oh this is a description change

jk13pclick:
What does not have an update function and could probably use an update function is #1717, but depending on how this PR goes I may just end up closing that one anyways

RedSparr0w:
Not sure if this is still relevant?
Does the achievement require purchasing them? or just simply obtaining them?

amative1:
Achievement is coded for purchase, which now that we have balls being obtainable in other ways it needed clarified, hence the PR.

RedSparr0w (approved):
Thanks for following up 👍 

PR #1735, opened 2021-12-26
Fix the DT Catch glitch

### LINKED ISSUE #1361: [BUG] I have Discovered that upon defeating a Pokémon, while catching it, you can switch to a different route, and it will then classify as catching a Pokémon from that route, making it give the Dungeon Tokens from that Route.
**Info**
 - OS: Unknown (I'm a moron)
 - Browser Microsoft Edge
 - Version Core i5 9th Gen

**Description:**
I have Discovered that upon defeating a Pokémon, while catching it, you can switch to a different route, and it will then classify as catching a Pokémon from that route, making it give the Dungeon Tokens from that Route.

**Steps to Reproduce:**

1. Go to 'Route 1'
1.5. Defeat encountered Pokémon
2. Click on 'route 23'

**Expected Behavior:**
Nothing, as I was just grinding on Route one, and then and wanted to go to Route 23, and I was still catching the pokemon, I caught it, and got 135 Dungeon Tokens.

**Screenshots:**

**Additional Information:**

jk13pclick:
Closes #1361. Should hopefully (🤞 ) fix the DT switch route glitch. Did a little bit of testing and it seems to be working fine.


apple096:
Why do I have a feeling this will backfire very hard in term of problems and opens a new can of worms

jk13pclick:
It might but I don't think it will? We could send it to Crew for testing just in case.

RedSparr0w (approved):
Looks good, thanks for fixing this 👍

PR #1736, opened 2021-12-26
Treasure Scanner Fix

### LINKED ISSUE #1309: [BUG] Conflict between Mine Items Quests / Treasure Scanner
**Info**
 - OS: Windows 10
 - Browser: Chrome
 - Version: `v0.7.9`

**Description:**
When collecting more than one item from the underground by using a Treasure Scanner of any level, the quest "Mine X items in the Underground" only registers one item being collected.

**Expected Behavior:**
This quest should count the bonus items received through the use of the Treasure Scanner, as the current form only counts each item once, despite the fact that multiple items have been gained.

jk13pclick:
Closes #1309. Tested with totalMineItems observable, should also work for the quest.


RedSparr0w (approved):

PR #1739, opened 2021-12-28
Migrate negative Pokemon to decimals

apple096:
I lied.
This PR migrates negative ID Pokemons to decimal.
There's also an update function for it


amative1 (changes requested):
New ID #s for Mega forms are the same as those for the clone forms. Is this intentional, or an oversight?

apple096:
Oversight, won't hurt anytime soon but it's good to have them changed soon-ish

amative1 (approved):

RedSparr0w:
I feel like we should also move the Pokémon in PokemonList to be next to the base Pokémon with the IDs, just so it's less likely we end up with Pokémon having the same IDs in the future?
_as what almost happened with the megas/clones_

RedSparr0w:
I've removed the Mega changes from this PR for now,
Only updates the Event Pokémon.
Moved the event Pokémon to be next to their base forms in PokemonList.
Updated the Pokémon & Shiny Pokemon image names.
Updated the Update code to make use of @Jaaslet code from #1833 so it updates the statistics too.

RedSparr0w (approved):
Thanks for the PR 👍

PR #1740, opened 2021-12-28
Refactor DungeonEvolution

apple096:
Too long for me. Refactored them to one line. Can make it for weather as well but weather is a lot more unique and there isn't really an use case right now


src/scripts/pokemons/PokemonList.ts
```diff
@@ -21957,6 +21972,13 @@ const maxEggCycles = Math.max(...pokemonList.map(p => p.eggCycles));
 // This needs to be initiallised before pokemonMap as some other things rely on it for data
 // Specifically Roamers not sure what else.
 pokemonList.forEach(p => {
+    p['evolution'].forEach(e => {
```
Jaaslet, line 21975: On the savescreen, i get the error "p.evolution is undefined"

src/scripts/pokemons/PokemonList.ts
```diff
@@ -8,6 +8,33 @@
 /// <reference path="../../declarations/enums/BerryType.d.ts"/>
 
 const pokemonBabyPrevolutionMap: { [name: string]: PokemonNameType } = {};
+/**
+ * Utility class
+ *
+ * Currently work for grouped dungeon evolution, but can easily refactored to work for Weather as well
+ */
+class GroupedDungeonEvolution extends Evolution {
```
Jaaslet, line 16: This should be in it's own class

Jaaslet (changes requested):
It currently does not start (might be because it's old, and new stuff has broken it). See the comments below for what does wrong.

PR #1741, opened 2021-12-28
Procedural loot generation for infinite Battle Frontier

apple096:
Yes, Battle Frontier is infinite now


RedSparr0w:
I'm not too sure about the random infinite loot, I feel like it should at least be seeded random (not sure what it should be seeded with though), or just pre-generated.
Also need a way for the upcoming loot to show while in the BF.

Jaaslet:
I will close this PR as it has been dead for a while.
And i agree with red that we need a smarter solution, if we want to do this in the future

PR #1742, opened 2021-12-28
[WIP] Sevii Island

apple096:
We all know the relationship between me and work-in-progress projects (party, dungeon reworks, other things). But this will be short, so I'm adding it.

Sevii Island will be locked behind Johto Indigo and it will contain Gen 2 mons


amative1:
Closing per #2083 and future second-half

PR #1757, opened 2022-01-06
Extend Pokémon League Johto down to Victory Road Johto

### LINKED ISSUE #1780: [SUGGESTION] Restoring Route 26 filter so that it cover the entire route
**Current Behavior**

The route 26 filter currently stop at the Victory Road Johto, but the route is still on the map after it.

**Improved Behavior**

Making the filter cover the route even after the Victory Road.

**Reason for change**

The route is still visible on the map, but is not covered by the filter which is weird.

**Detailed Description**

Making the filter of route 26 cover the entirety of the route, just like it does in Kanto where route 23 is also interupted by a Victory road dungeon, but still has it filter cover all the route. 
This difference in implementation is weird.

**Screenshots:**

Ultima1990:
The space between Victory Road and the Pokémon league is left open for a very specific (potential future rival battle related) reason, but people seem to dislike that the road is just empty and has no purpose. Extending the Pokémon League down onto the road might help a bit with that.

closes #1780


RedSparr0w (approved):

PR #962, opened 2020-10-18
Update Gym.ts

Ultima1990:
fixes #960
Also changes many gym Pokémon and quotes in Kanto to reflect Yellow version.
Also changes a couple of Pokémon and quotes in Johto to reflect Crystal version.
Also changes misc quotes and Pokémon all over the place to properly reflect... you got the deal.
Focus games for these changes are Yellow, Crystal, Emerald, Platinum.

Part of a larger Yellowfication project for Kanto.

PR #1759, opened 2022-01-07
Undo Yellowfications, back to FRLG.

Ultima1990:
Basically undoes #1008 and #962

Best be sure we want to go with FRLG before merging this. If we decide to stick with yellow I _will_ reopen #1589 and fight to yellowficationate Blue's team!

Enjoy getting Chancey in Cerulean Cave while you still can lol.


Ultima1990:
Nerfed Brock.

Because in this FRLG world the Nidorans (and by extension Nidorina and Nidorino) are unavailable before Brock, the player's attack at this stage is significantly reduced. From 167 (Pikachu starter, all available mons at 20, no Mew), to 121. They are poison types, so they don't contribute hugely to damage against Brock's rock types. Effective damage against ground/rock goes from 102 in yellow world to 90 in FRLG world. Doesn't look like a big difference, but it can't be ignored at this early stage of the game.

The intent of this Yellow to FRLG change is not to make the game harder, (except for the bits where the game was made unintentionally easier by the yellowfication (looking at you Sabrina)) so I have nerfed Geodude and Onix's health by 10%. Yes, from 102 to 90 is not exactly a 10% reduction. Spare me.

After Brock it pretty much evens out, the Nidorans live on route 3, so I don't think further nerfs need to be made.

Jaaslet (approved):
Looks good.
Looked though like half of the changes, and there where no mistakes there. And if there are any in the rest, it's not a huge problem and the playerbase will notice.
All pokemons are still obtainable, and the early game still seams beatable.
I approve. Thanks for the PR!

PR #1761, opened 2022-01-08
Have blimp, will travel

Ultima1990:
Players sometimes have trouble figuring out how to travel to the next region. Having something new appear on the map might subtly push them in the right direction.

When the player is able to travel to the next region a blimp appears on the map next to the home town. The blimp is owned by each region's professor because I thought that would be fun. Instead of adding the travel to next region button to the blimp I just have the blimp act as an extension to the home town, which conveniently already has the travel to next region button.

Yes, I know Juniper lives in Nuvema, not Aspertia, and I know Sycamore lives in Lumiose, not Vaniville, but screw it. Home town is where you travel from, so home town is where the blimp appears.

Thanks to RedSparrow for the sprite.


Ultima1990:
The blimp will now have a Pikachu on it if the player started with Pikachu.

This is currently achieved through the magic of having two overlays, one that activates when Pikachu is the starter, one that activates when it isn't.

Heppity:
> The blimp will now have a Pikachu on it if the player started with Pikachu.
> 
> This is currently achieved through the magic of having two overlays, one that activates when Pikachu is the starter, one that activates when it isn't.
Only question about it is when you unlock the blimp? Like in Hoenn or Johto?

amative1:
> Only question about it is when you unlock the blimp? Like in Hoenn or Johto?

It shows up when you can progress to the next area... for ALL areas (so first instance will be in Kanto progressing to Johto)

RedSparr0w (approved):

PR #1774, opened 2022-01-15
MovetoRoute logic for Hotkey

jk13pclick:
Uses route order instead of just player.Route() allowing for better regional progression with hotkeys. Could be improved with the routes at the beginning and end of the region iterating in a loop, but this is better for now


RedSparr0w (approved):

PR #1779, opened 2022-01-19
Add Alola Wanderers

Ultima1990:
Adds Cutiefly and Bounsweet as wanderers. Cutiefly because it's a li'l bug that likes flowers and Bounsweet because it is basically a berry.

Also Oricorio (Baile), Oricorio (Pom-pom), Oricorio (Pa'u) and Oricorio (Sensu) are now attracted by red, yellow, pink and purple berries respectively, as an extra nod to the different colored flowers in the meadows that they appear in.

None of these Pokémon and forms are farm exclusive, Oricorio will still appear in the meadows.


RedSparr0w (approved):

PR #1798, opened 2022-02-06
Add update functions for new observables/statistics

RedSparr0w:
Add update function for #1717 observables,
@jk13pclick could you please check this covers everything correctly.


jk13pclick (approved):
Looks good, tested and works great!

PR #1812, opened 2022-02-12
Misc Galar changes

RedSparr0w:
Change Hop to a Gym Leader instead of a 2nd Champion (otherwise message appears twice saying you are the Galar champion)
Split `Crown Tundra` and `Isle of Armor` into sub regions of Galar instead of their own regions.


scolpitts536:
Can you split galar into two sub regions one top and one bottom. For when I get to making that map.

RedSparr0w:
Added `Galar South` and `Galar North`

RedSparr0w:
Added 2 new requirements:
Subregion requirement (requires you to be in a specific subregion, currently used for Armor/Crown roamers)
Null requirement (always false, for features that are still under development)

scolpitts536:
> Added `Galar South` and `Galar North`

Thx

PR #1816, opened 2022-02-13
Block Resolution Cave

Ultima1990:
Adds a transparent overlay on top of Resolution Cave, blocking access to the dungeon in the event that Alola is released before #1615 is merged. To be manually removed if/when #1615 is merged.

DO NOT FORGET TO REMOVE.

I chose to do it this way instead of just adopting the new requirement from the UB PR. The new requirement would tell players they need certain UBs to access the dungeon, which would lead to questions on how to get those required UBs. This seems to me like the clearest way to communicate to players that this area is just unavailable for now.


src/components/AlolaSVG.html
```diff
@@ -1024,5 +1024,20 @@
         , width: 4, height: 3
         })
     %>
+
+    <!-- ------------------ -->
+    <!-- PLACES OF INTEREST -->
+    <!-- ------------------ -->
+
+    <!--REMOVE IF UB QUESTLINE GETS MERGED-->
```
RedSparr0w, line 1032:
```suggestion
    <!-- TODO: REMOVE IF UB QUESTLINE GETS MERGED -->
```

RedSparr0w (approved):
👍 Thanks for the PR!

PR #1817, opened 2022-02-13
Update QuestLineHelper.ts

amative1:
Removing unnecessary period (each hint already has one, so this is causing two periods at the end of every Viv hint).


amative1:
Added text fix to Kanto Berry Master (similar issue, different root cause)

RedSparr0w (approved):
Thanks for the PR! 👍

PR #1818, opened 2022-02-13
Increase steps of some totems

Ultima1990:
Evolved Pokémon get increased steps. Totems are technically not evolved, so they don't get these increased steps. Manually increased the steps of certain totems to that they match their non-totem counterparts.


RedSparr0w (approved):
Nice work, thanks for the PR!

PR #1824, opened 2022-02-15
Dungeon pokemons level adjustment

Symi001:
Dungeon pokemon now use the routeLevel function to get their level, it should significantly boost the xp of the pokemon in dungeons.


Symi001:
Followed the advice of jk. This gives a bit more exp even to older dungeons. Tested with several dungeons (in Kalos and Kanto) and it seemed alright, but I'll let you see if you think it's a bit too much exp :)

RedSparr0w:
Should we just be removing `Dungeon.level` entirely?

Doesn't really seem to be used elsewhere, especially if we can just use the route level as a default?

Symi001:
> Should we just be removing `Dungeon.level` entirely?
> 
> Doesn't really seem to be used elsewhere, especially if we can just use the route level as a default?

Should be alright now, I got rid of it, I still have a question, what do you think about it: should I create a function to get a dungeon level as it's already the same call in three places ?

RedSparr0w:
> Should be alright now, I got rid of it, I still have a question, what do you think about it: should I create a function to get a dungeon level as it's already the same call in three places ?

👍 Yeah that's probably a good idea, instead of us calling the same set of functions multiple times.

Symi001:
Should be all good now, sorry for the delay ^^"

RedSparr0w (approved):
Nice work! Thanks for the PR 👍

PR #1827, opened 2022-02-15
Make Wishiwashi (School) an evolution

Ultima1990:
Wishiwashi's ability, Schooling, transforms Wishiwashi (Solo) into Wishiwashi (School) if it's at or above level 20 and its HP is at or above 25%. In clicker this can be represented as a level 20 evolution.

Yes. This makes Wishiwashi (School) easier to get. But also this actually nerfs Wishiwashi (School). By making it an evolution the School form also gets increased egg steps, which it kind of needs. With its absurd stats and Solo's very low steps Wishiwashi (School) was looking like it was going to be an insane powerhouse, especially with max protein. Not Gyarados level, but close.

Also increased Totem Wishiwashi (School)'s steps to match the non-totem.

Gyarados reaches efficiency 9 in Alola btw. Thought someone'd be interested by that.


src/scripts/pokemons/PokemonList.ts
```diff
@@ -16379,7 +16380,7 @@ const pokemonList = createPokemonArray(
         'id': 746.02,
         'name': 'Totem Wishiwashi (School)',
         'type': [PokemonType.Water],
-        'eggCycles': 15,
+        'eggCycles': 23,
```
RedSparr0w, line 16383: Does this actually affect it's egg steps?

Ultima1990: Yes it's 920 up from 600.
WishiwashiSteps

RedSparr0w: Ah sorry, didn't realise this was for the totem variant 👍

RedSparr0w (approved):
Thanks for the PR! 👍

PR #1828, opened 2022-02-15
Make Type: Null and Poipole more expensive

Ultima1990:
Type: Null 10k.
Poipole 25k.

Not Meloetta (150k) or Furfrou (500k) level expensive. IMO such prices are for alternate forms. However, the 5k standard for pokemonitems is not that much at this point and Poipole is a freakin' UB, so I think these prices are good.


RedSparr0w (approved):
👍

PR #1829, opened 2022-02-16
Add Necrozma forms

Ultima1990:
Dawn Wings and Dusk Mane. For future use. Hadn't been added yet apparently.

Also corrected Necrozma (Ultra)'s type.


RedSparr0w (approved):
Nice catch! 👍

PR #1830, opened 2022-02-16
Prevent DNA items showing up in trades

Ultima1990:
Like this I think?


RedSparr0w (approved):
👍 Nice work, thanks for the PR!

PR #1833, opened 2022-02-18
Changed event pokemon id's to follow the new patten

Jaaslet:
_I don't know if you want to use this, but i hope it will be used, because it's something that annoys me a bit. If it does not fit the plan, or is too risky, just delete the pull request._

**Change:**
Event-pokemons no longer have a negative Id. For example: Flying pikachu will now have id 25.4 instead of -1.

**Reasons:**
- It seams like the new Id-patten is to have each alternative form, to have the same id as the normal form, just with a decimal-number.
- It kinda annoyed me that all the event-pokemons was in the start, when sorted pokemons by id

**Testing**
This is something, where data can be lost. So i have tested the following (and made sure it works):
- Players keep the pokemons after update
- Pokemons keeps the number of proteins
- Pokemons keep the number of breeding
- The conversion on update also affects pokemons in the breeding queue


Ultima1990:
#1739

Jaaslet:
> The update function in the other PR should transfer all the data (shiny/breeding/etc), so there's no functional difference.

As far as i can see, how many times they have been breed, is saved in saveData.statistics.pokemonHatched, and not saveData.party.caughtPokemon. Atleast the "Hatches: n" display in the day care menu did lose it's data for me, when i only updated saveData.party.caughtPokemon

PR #1834, opened 2022-02-18
Make Timerball go to 0 in Dungeons

jk13pclick:
Little fix, made a mistake with the implementation


src/scripts/pokeballs/Pokeballs.ts
```diff
@@ -46,7 +46,8 @@ class Pokeballs implements Feature {
                     return Math.min(15, Math.max(0, Math.pow(16, Math.pow(kills, 0.6) / 250) - 1));
                 }
                 if (App.game.gameState == GameConstants.GameState.dungeon) {
-                    return Math.min(15,100 / DungeonRunner.timeLeftPercentage());
+                    const x = 200;
+                    return Math.min(15,(x / DungeonRunner.timeLeftPercentage() - x / 100));
```
Aegyo, line 50: These numbers feel a little opaque to me
Could we have something like
```js
const maxBonus = 15;
const timeLeftPercentWhenMax = 11.7;
return Math.min(maxBonus, <insert math here that does what the variables claim>);
```

RedSparr0w (approved):
Thanks for the PR! 👍 
Looks good.

PR #1838, opened 2022-02-20
Change Sinnoh 224 requirement

Ultima1990:
Turns out that in the main series the exit from Victory Road that leads to 224 is blocked until the player enters the hall of fame.

Changed Sinnoh Route 224 requirement from clearing Victory Road to defeating the Sinnoh champion.


RedSparr0w (approved):
Awesome, thanks for the PR! 👍

PR #1839, opened 2022-02-20
Fixing the hotkey fix

jk13pclick:
Tested in every region, works well. Maybe some unforeseen stuff, but this gets rid of the oddities.


RedSparr0w (approved):
Works great!
Thanks for the PR 👍

PR #1846, opened 2022-02-26
Change the order of start menu

Ultima1990:
Items and Shards closer to the top should be more convenient for everyone as they are menus that (most) players will access frequently. Hopefully helps new players find the shards.
newstart


RedSparr0w (approved):

PR #1848, opened 2022-03-01
Fixes the DT Missingno and catchrate:catchrate bugs

jk13pclick:
Hopefully fixes the weird catch rate:catch rate bug and the Missingno glitch caused by the initial DT fix. Doesn't fix a slightly abusable trick where you can go to a Dungeon mid-catch and gain the dungeonDifficultyRoute amount of DT. Related, switching to a Town mid catch gives you route 1 (of that region) DT. There is probably a fix in there, but this is a good start.

### DIFF
--- a/src/scripts/worldmap/MapHelper.ts
+++ b/src/scripts/worldmap/MapHelper.ts
@@ -8,7 +8,7 @@ class MapHelper {
             return;
         }
         let genNewEnemy = false;
-        if (route != player.route()) {
+        if (route != Battle.route) {
             genNewEnemy = true;
         }
         if (this.accessToRoute(route, region)) {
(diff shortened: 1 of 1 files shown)


RedSparr0w (approved):
Looks good 👍 

This kind of creates a speen run strat:
Once catch animation appears, go to a town, then back to a route, skips the catch animation, still have the ability to capture the mons and gain dungeon tokens.
_not absolute most effective thing, but could shave off a bit of time_

And it fixes another bug too (related to doing the same thing):
If you are on a route, once the catch animation starts, go to a town, then back to a route before the animation finishes.
The Pokémon images and catch rates won't appear anymore, if you go back to a town, it shows an Unknown route.

PR #1851, opened 2022-03-05
Trimming searchtext in pokedex

Jaaslet:
This pull request removed whitespaces before and after the searchtext, when searching in the pokedex/breeding. This makes it less annoying, when you accidentally copy a whitespace, while copying a pokemon name.

Also some small code cleanup: 2 methods where marked as private, but used publicly, and one of the methods had an unused parameter


RedSparr0w (approved):
👍 Thanks for the PR!

PR #1855, opened 2022-03-07
Rename Totem Wishiwashi

Ultima1990:
Removed (School) from Totem Wishiwashi's name. There is no Solo form of the Totem Wishiwashi so the distinction is not really necessary. Also the name was just too long.


RedSparr0w (approved):
👍

PR #1858, opened 2022-03-10
Improvements to Melemele map

Ultima1990:
Made some changes to the visuals of the Ruins of Conflict, Verdant Cavern, Melemele Meadow area.


RedSparr0w (approved):
👍 Thanks for the PR!
Definitely makes the dungeons more visible

PR #1859, opened 2022-03-11
"Refresh quests" no longer shows on new game

Jaaslet:
The "You have a free quest refresh."-message now only shows if daily quests have been unlocked (after the tutorial)


RedSparr0w (changes requested):
👍 Looking good, thanks for the PR!

Just a couple of small changes:
Move the function to be a part of `App.game.quests` rather than the `QuestLineHelper` as it has more relevance there.
Update the function name to something like `isDailyQuestsUnlocked` just so it's more consistent with our current function names.

Jaaslet:
> Just a couple of small changes: Move the function to be a part of `App.game.quests` rather than the `QuestLineHelper` as it has more relevance there. Update the function name to something like `isDailyQuestsUnlocked` just so it's more consistent with our current function names.

Done!

RedSparr0w (approved):
Works great! Thanks for the PR 👍

PR #1861, opened 2022-03-11
Minty now have the correct textcolors

### LINKED ISSUE #1860: Minty Theme not showing diamond value [BUG]
- OS: Windows 10
 - Browser and version: I think any, I've tried on Chrome and on Edge and nothing changes.

Description: With the Minty theme on, you cannot see the mining diamond value you currently have. I think the problem is the font is the same color as the background (white). I mean, if you put the mouse on top, it shows up the litte window with the items you own, so the value is there, but you can't see it.

Steps to Reproduce:
1. Open Pokeclicker with the Minty theme set
2. Click on Underground, you'll see it

Expected Behavior:
I expected to see the diamond value just like with other themes

Jaaslet:
Fixes #1860


RedSparr0w:
I think the minty theme looks a bit odd with all the buttons text being dark:
image
image

Not sure if there's an easy css selector to just target `.btn` without any other `.btn-*`  classes,
or if we just apply one of the other button styles to the button like `btn-light` (which may look odd on dark themes), `btn-info` or similar,
or just not use a button element/class.

src/styles/themes.less
```diff
@@ -42,6 +42,17 @@
...
 
+.minty {
+    // text color in buttons are white, but expected to be black
+    .btn, .btn:hover {
+        color: #495057;
+    }
+    // text color in primary buttons gets overridden by the lines above
+    .btn-primary, .btn-primary:hover {
+        color: #fff;
+    }
```
RedSparr0w, line 53: Something like this should work:
_(I don't think the `:hover` part is needed for this change from testing, but I may have missed something)_
```suggestion
    // Button without color class text is light, but expected to be dark
    .btn:not([class*="btn-"]) {
        color: #495057;
    }
```

Jaaslet: Your solution solves the problem you pointed out. And the hover-part is not needed, with your solution.
I hope i have accepted it (still new to git)

RedSparr0w (approved):

Issue #1784, opened 2022-01-23
[SUGGESTION] More map customization


Derpychu:
I would like to see the ability to set more map colors, for things such as uncompleted acheivements

amative1:
Complete per #1864

PR #1864, opened 2022-03-12
More mapcolors: added for routes

Jaaslet:
Implements #1784
Default should work as before, but now you can change the colors for "missing shiny+achievement" and "missing achievement (but have all shiny)"

- [x] Added to settings
- [x] Added for routes
- [x] Added for dungeons
- [x] Added for gyms
- [x] Added for leagues


src/scripts/dungeons/DungeonRunner.ts
```diff
@@ -221,6 +221,17 @@ class DungeonRunner {
...
 
+    public static isAchievementsComplete(dungeon: Dungeon) {
+        let result = true;
+        AchievementHandler.achievementList.forEach(achievement => {
+            if (achievement.property instanceof ClearDungeonRequirement && achievement.property.dungeonIndex === GameConstants.getDungeonIndex(dungeon.name) && !achievement.isCompleted()) {
+                result = false;
+                return true;
+            }
+        });
+        return result;
```
RedSparr0w, line 232: Minor optimisations:
_Calculate the dungeon index outside of the loop, so we don't need to keep calling the same function._
_Use `.every` instead of `.forEach` so we stop as soon as we hit a false_
```suggestion
        // Calculate dungeon index
        const dungeonIndex = GameConstants.getDungeonIndex(dungeon.name);
        // Stop if we get a false
        AchievementHandler.achievementList.every(achievement => {
          return !(achievement.property instanceof ClearDungeonRequirement && achievement.property.dungeonIndex === dungeonIndex && !achievement.isCompleted());
        });
```
Could you apply a similar approach to the other achievement checks please.

RedSparr0w:
Looking good so far!
Could you also update the default colors so they are different from what we currently have set?
Maybe a light purple/pink for achievements?

I'm not entirely sure if it's worthwhile having the option for `!Achievement & !Shiny` or if we should just favor one over the other.
I think I'd rather favor `!Shiny` > `!Achievement` as it looks like the achievement checks could be quite costly in terms of performance as it has to loop over all of the achievements for every location.

Jaaslet:
> I'm not entirely sure if it's worthwhile having the option for `!Achievement & !Shiny` or if we should just favor one over the other.
> I think I'd rather favor `!Shiny` > `!Achievement` as it looks like the achievement checks could be quite costly in terms of performance as it has to loop over all of the achievements for every location.

Some are shiny-hunting, and some are achievement-hunting. These options opens up for both players. But i can remove the combined one, if you want.
About the performance: I could not feel any difference on my cheap computer, so i think it's fine. And with your other performance optimizations (which i will add later this week), it will be even better.

> Could you also update the default colors so they are different from what we currently have set?

I am really bad at colors. That is why i did not choose any colors. But if you can give me some hex-codes, i will gladly add them

src/scripts/dungeons/DungeonRunner.ts
Jaaslet: improvements added

Jaaslet:
I have done what i can, for this issue. So time to give a full explanation, for my solution:
I have added two new map colors: MissingAchievement and MissingShinyAndAchievement. The explanation for this, is it opens up for all the possibilities.
A way of solving it, would just be to add MissingAchievement, and then do so people could turn off the MissingShiny-color, if they care more about the achievements. The problem is that i am not good enough at CSS magic, to make that solution, and it will not work for people, who cares mosts about achievements, but also wants to see missing shines, when the achievement is completed.
How it currently works, is that if you set MissingShiny to transparent, it will just look completed, even if you are missing the achievement, and have set a mapcolor for that. This is the part i could not fix, with my CSS skills.

Another thing i want to say, is that i am bad at colors. But i have selected one, which i personally think looks fine.

The default settings will currently be the same as before, but it will show a new color, if you have the shiny, but are missing the achievement (where it before just was "completed"). But you can change the colors, if you want, to do something else.

RedSparr0w (approved):
Works great 🎉 

Thanks for the PR!

PR #1882, opened 2022-03-17
increasing size of Gem Icons

NiCeDiCe90:
Minor change.
I think they are super tiny atm. Should look better that way.

Worlds smallest PR :)


RedSparr0w (approved):
Definitely looks better 👍

Issue #1805, opened 2022-02-11
[SUGGESTION] Add categories to notifications/sounds


RedSparr0w:
**Current Behavior**
Uncategorised list of notifications
![image](https://user-images.githubusercontent.com/7288322/153536263-3cc2094c-407b-48ef-bba7-083ff7e782b0.png)

**Improved Behavior**
Sort different notifications and sounds into collapsible categories
![image](https://user-images.githubusercontent.com/7288322/153536451-331aec66-0ea8-47af-9135-c27409c1e68b.png)

**Reason for change**
With the recent addition of #1760 the notification list is getting quite large

**Detailed Description**
Would be good to have them categorised for easier control and at a glance what feature of the game it's related to.

amative1:
Complete per #1884

PR #1884, opened 2022-03-18
Notification settings are now grouped

Jaaslet:
Implements #1805
Problems, that might need to be fixed:

- [x] The HTML could be more compact/generated. I was having a hard time getting knockout to have a foreach inside a foreach
- [ ] The category titles changes place, when all categories are collapsed


Jaaslet:
Currently, the category titles jumps around, when you collapse/uncollapse the categories. I don't know enough about CSS to be able to fix it. And i don't know if it's a big enough problem, that it needs to be fixed. But if it is, i need help

RedSparr0w (approved):
Looks great!

Thanks for the PR! 👍 

Would be great if we could do the same for sounds, is this something you'd want to work on?
If not I can do it.

RedSparr0w:
Made some minor changes:
- Titles changed to buttons _(with small top margin to separate them)_
- Put each of the tables in a `div` so they shrink/grow correctly _(seems tables don't like how bootstrap grows/shrinks the collapsable elements)_

Jaaslet:
> Would be great if we could do the same for sounds, is this something you'd want to work on?

I will do it, at some point

PR #1885, opened 2022-03-18
Option to ignore poke lvl in DamageCalculator

Symi001:
Adding the option, and aligning the text to the left because it seemed off otherwise.


RedSparr0w (approved):

PR #1886, opened 2022-03-19
Dungeon Clear Tooltip

### LINKED ISSUE #1877: [SUGGESTION] Display a 'Times Run' counter on Dungeon Start screen
**Current Behavior**
Currently, the only way to see the number of times you've run a dungeon is via the achievement window, but this has a hard limit of 1,000

**Improved Behavior**
I'd like to be able to see the number of times I've run a dungeon on the 'Start dungeon' screen without having to dig through the achievements window to find a particular dungeon.

**Reason for change**
With the latest update causing some drops not to be available at all until a dungeon has been run 1,500 times, not knowing how many times you've run a dungeon when going beyond 1,000 (where the achievement caps) is an issue if you're hunting for a particular drop.  
Example:
{ loot: 'Liechi', weight: 0, requirement: new ClearDungeonRequirement(1500, GameConstants.getDungeonIndex('Cave of Origin')) },

Also, it'd just be user-friendly to not have to dig through the achievements menu for a particular dungeon to see how many times it's been run.

jk13pclick:
Closes #1877. Puts the clear total in an info button on the townView for Dungeon Towns. Feel free to make changes if necessary.


RedSparr0w:
Are there any other stats we could add to this?

Otherwise I feel like it should just be displayed rather than in a tooltip:
_kind of hard to see here_
image
_looks a bit odd on the left 🤷_
image
_hide the `?` after tutorial completed?_
image

jk13pclick:
> Are there any other stats we could add to this?
I was thinking maybe putting Loot in the tooltip could be a good idea, but of the images linked the third is probably the best. Whatever you think is best I can go for.

jk13pclick:
Most of the questions we get in the Discord/Reddit are about getting enough DT for VF anyways, tutorial feels fine to me. Removing it entirely is certainly much easier, we could try that and if we see a ton of questions like "How to beat VF, not enough DT" we can put it back.

RedSparr0w:
Sounds good 👍

jk13pclick:
Updated, also added Gym Clears to the Townview (locked behind Tutorial). there may be some unnecessary code in here, but I tested with the tutorial and it seems to work well.

RedSparr0w (approved):

PR #1887, opened 2022-03-19
Turn Mina's Trial into a dungeon

Ultima1990:
Mina's Trial is now a dungeon. Mina's Houseboat (as suggested by Vic).
Also removed Dhelmise from Poni Wilds and made it exclusive to the Mina's Boathouse dungeon, since in the main games it's only found inside the town.
Also also changed the Mina npc text to reflect the fact that it's no longer just a gym battle.
Also also also added credit for Leaf, for creating all the Totem sprites.


RedSparr0w (approved):

PR #1888, opened 2022-03-19
More requirements for Anville Town and Pledge Grove

Ultima1990:
Added extra requirements to Anville Town and Pledge Grove to prevent sequence breaking into these locations by using shiny code to obtain certain Pokémon early.

Keldeo lives in Moor of Icirrus, so Pledge Grove now requires one Moor of Icirrus clear in addition to owning Keldeo.
Meloetta (aria) starts roaming after the Unova Champion is defeated, so Anville Town now requires defeating the Unova Champion in addition to owning Meloetta (aria).


RedSparr0w (approved):

PR #1889, opened 2022-03-20
Hotkeys can now move between sub regions

Jaaslet:
This fixes a problem, where the hotkeys - and + acted weird, in areas with sub regions. It could move between sub regions, but the frontend did not follow, when the sub region changed.
The solution is to allow the hotkeys to move between sub regions, and change the map, when it happens


RedSparr0w (approved):
Nice work!

Thanks for the PR! 🎉

PR #1891, opened 2022-03-21
Professor oak now congratulates after each league

Jaaslet:
Professor oak now congratulates the player, after beating each champion, like he did after Kanto.
This fixes the problem with the Kanto message showing up after beating the Alola league.
The texts might need to be improved, but i did the best i could.


RedSparr0w (approved):

PR #1892, opened 2022-03-21
rebalancing QP on Shop Pokemon

NiCeDiCe90:
Dex required:
reduced Kanto (12k total -> 10k total) to reduce the QP wall a tiny bit
increased Johto -> 15k total (+50% of previous region total)
increased hoenn -> 22500 total
increased Sinnoh -> 33750 total
increased Unova -> 50625 total
Kalos -> non dex required available 
increased Alola -> 114k total
increased Galar -> 170k total

optional:
Meloetta -> 200k
Furfrou (Dandy) -> reduced to 250k (reduced because a lot more is required in previous and later regions. Still a lot higher than Meloetta because there is no other QP Pokémon in the Kalos region)

Not touching Eggs and Evo items. In the end it's still a bit cheaper than the current version if you take Furfrou and Meloetta into account. 
But that should already make QP a bit more important later in the game and slow down mid game progress.


RedSparr0w (approved):
👍 Looks good,

The Togepi and Beldum feel a little out of place _(I get why they are the way they are)_, just seems like such a big jump.

Issue #783, opened 2020-09-26
[Suggestion] Port from Knockout to modern framework


SpenserJ:
I've been playing Pokeclicker on and off since its inception, and have always felt that there were significant performance issues, especially as you reach the later regions. After doing some benchmarking, its clear that many of the issues are coming from the way in which Knockout updates the DOM. While some of the issues can be solved through optimization and careful thought about which templating functions are used, I feel that it would be significantly easier to port to a more modern (and maintained) framework that addresses some of these issues, like React.

Is there any interested from @RedSparr0w and @Ishadijcks in a port? I'd be happy to take on the majority of the work, but I don't want to create a large PR if it isn't in-line with the goals of the maintainers.

Ishadijcks:
Interesting proposal. There are a few things to consider. Note that I have limited experience with React, and decent with Vue.

#### Good things

- Better template support. There is a ton of duplicated code, especially in the modals.

- Better error detection. Knockout doesn't really report errors well.

- All of the controller code can be moved to components, which is nice.

- Better performance, I think.

#### Not so good things

- Adding a reactive framework complicates development work. For newcomers obviously, but also the developers need to learn it (if they haven't already). Knockout is easy enough to make changes to without understanding it completely.

- Having an "outsider" do it means the knowledge is not with the maintainers. Especially if framework experience is minimal. This can lead to some annoying situations

- It's a lot of work :')

As I'm no longer affiliated with this project, I'll end with some suggestions/recommendations:

- Port incrementally. Don't do it all at once, you'll never finish.

- Vue is pretty neat. I prefer it over React. I've been using it for another project. You can properly separate UI and Game logic. Take a look :) https://github.com/Ishadijcks/vue-incremental-game-template

SpenserJ:
Very good points. I'm on the opposite side with a massive amount of React experience, and intermediate Vue experience, and I'd probably add a few more positives:

* Larger community with a significant amount of documentation, examples, tools/libraries, and best practices
* Significantly better performance due to advancements and lessons from data binding and batching DOM updates

The negatives of a more complex framework and ensuring the maintainers have the appropriate knowledge is the main reason I opened the issue. I'll probably play around with porting it for fun and the challenge of having a hybrid Knockout/React (or Vue), but I'd like to invest any effort in whatever direction the maintainers prefer.

Aegyo:
[...]

PR #1895, opened 2022-03-22
Migrate LogBook modal rendering to React

Aegyo:
An example of how we could migrate to rendering with react, without having to rip out all the knockout code we have.
Related to #783

The ko-store library unfortunately didn't work out, my guess is because we are using a global ko object from an included script which didn't mesh with the ko the library imports. Luckily, it is licensed under MIT and the code is quite small, so I just copied what we need (useComputed, useForceUpdate, useMemoWithDisposer)

I've added two "hooks" based on redux -> `useGame` is sort of like redux's `useStore`, and `useSelector` is like the redux hook of the same name. These are used for accessing some piece of App.Game, without needing either a global variable or passing it down through every component manually. A "selector" in our context is just a function that grabs some observable (or computedObservable) when given a game object, and the finding and passing of a game object is handled automagically for you with `useSelector`.

I think with this particular modal, I would move all of the filter stuff into react state, since it is really only to do with this bit of the UI. I've left it as knockout for now just to show how we might do other things.

Other modals will be easy to absorb into the ReactApp component, but other pieces of the UI might be an issue since React will want to take control of all its children. We might be able to have multiple react bindings, one for each UI column, and have them gradually absorb their surroundings until they meet at the top of the DOM and can be merged together.


RedSparr0w:
I don't have any previous experience in React but from what I can see this seems to work great.

I assume we will only be able to convert modules to react components?
So something like the quests modal wouldn't work with this until we move/convert it to modules?

As far as I've seen React performance is a ton better than Knockout, so I think we should definitely move forwards with this.

RedSparr0w:
Also is it worth moving static content to react such as the FAQ modal?

Aegyo:
I think we can reactify non module code, it would just need a little extra work mocking the type on GameShim. I'll try that out with Quests tomorrow to get an idea of how difficult that might be.

Re: moving static content - it is probably worth doing so for anything that can be simplified with a shared component (eg, modal) or which we might want to have dynamic content in the future (translations?)

Aegyo:
Integrating the Sortable drag + drop functionality might also be a pain point, will have to look into that too

RedSparr0w:
> I think we can reactify non module code, it would just need a little extra work mocking the type on GameShim. I'll try that out with Quests tomorrow to get an idea of how difficult that might be.

I think we can probably just ignore non module code for now, I'm trying to slowly move it all over, there isn't a whole ton left, and anything that we move to react we can just move to modules at the same time (where possible)

Aegyo:
Closing as I've changed my mind on using react, see comment in the linked issue

PR #1897, opened 2022-03-22
Move some stuff to modules

RedSparr0w:
`Information` This could probably use a better name 🤷‍♂️
`TypeHelper`
`Upgrade`
`ExpUpgrade`
`OakItemType` (moved from `OakItems.OakItem`, wasn't sure how to handle it)
`OakItem`
`OakItems`
`BoughtOakItem`
`OakItemController`
`OakItemLoadouts`

#1153


src/modules/globals.ts
```diff
@@ -28,8 +30,8 @@ type TmpGameType = {
     wallet: Wallet,
     keyItems: any,
     badgeCase: BadgeCase,
-    oakItems: any,
-    oakItemLoadouts: any,
```
Aegyo, line 32: I knew this file existed somewhere... was trying to find it for that react pr instead of the GameShim I made...

Aegyo (approved):

PR #1900, opened 2022-03-23
Rename Totem Raticate and Totem Marowak

Ultima1990:
Same as #1855. There are no Kantonian versions of these Totems so the distinction is unnecessary and the name is very long.


RedSparr0w (approved):
Thanks for the PR 👍

PR #1901, opened 2022-03-23
Sound settings are now grouped

Jaaslet:
Grouped sounds settings, like i did with notifications in #1884


RedSparr0w (approved):
Looks great, Thanks for the PR! 👍

PR #1907, opened 2022-03-24
Fixed require complete pokedex challenge

Jaaslet:
The code based the "complete pokedex challenge" on the achievement for the next region, and not the current one, which made it impossible to progress, with that challenge active


RedSparr0w (approved):
Nice catch!

Would have been awkward if that went live 😅 

Thanks for the PR 🎉

PR #1916, opened 2022-03-28
Hide stones

Jaaslet:
This PR will hide stones, from the stone-tab in items, until the the region, the stone is first used, is unlocked.
Also now using GameConsts.StoneTypes in underground items, to have more connecting code


RedSparr0w (approved):

PR #1919, opened 2022-03-29
Showing more information if more evolutions are available, in evolution stone tab

Jaaslet:
I was playing around with Alola, and i got confused why pikachu did not evolve to Alola Ritchu, when i used a thunderstone.
It turns out if more evolutions are available, one random will be selected. So now i have added a helptext, which only shows up, if the pokemon has more than one stone evolution


Jaaslet:
> There's already an NPC that mentions this

Hm... I should start listening to the NPC's... I am just afraid that we will get some bug reports, because other people don't listen to the NPC's. It's weird that nothing happens, without any indication of it "failing", when you are used to stuff happening.
But Red can close the issue, if he thinks the NPC is enough

Qwertypop4:
Absolutely I agree the NPC will be ignored, so more could probably help, given as you already have the other form so literally nothing is there to tell you it evolved into the other form. Just wanted to point out that it already existed in case you were unaware

Jaaslet:
> Absolutely I agree the NPC will be ignored, so more could probably help, given as you already have the other form so literally nothing is there to tell you it evolved into the other form. Just wanted to point out that it already existed in case you were unaware

I was :) Thanks!
I might want to take that NPC to the farm, if this PR gets through. Will think about it

Jaaslet:
> I might want to take that NPC to the farm, if this PR gets through. Will think about it

I think he still gives some useful information, so i will keep him... (NPC "RoyalAvenueSpectator" for reference)

RedSparr0w (approved):

PR #1923, opened 2022-03-30
Upgrade and Dragon Scale now shown in johto

Jaaslet:
Fixed a mistake in #1916


Aegyo:
We could have all of these regions calculated from pokemonList on page load, maybe after this is merged to get a fix out quicker though.

Jaaslet:
Items now show up if they are owned, even if they are in a later region.
This is maybe a temp fix. The real fix might be to make sure people cannot get the evo-stones in earlier regions (depending on who you ask)

Aegyo (approved):

PR #1930, opened 2022-03-31
Codeimprovement: Automate evolution stone unlocked region code

Jaaslet:
I made an attempt on calculating which stones should be unlocked at which region, instead of setting them all automatic.
It turned out to be hard to code... I don't know if we think this change is safe (because the code is complicated to check), but here it is.
The first commit contains some cleanup of the initialize code, so that might be useful


src/scripts/pokemons/PokemonList.ts
```diff
@@ -21322,29 +21322,6 @@ const pokemonList = createPokemonArray(
...
-    } else {
-        (p as PokemonListData).evolutions?.forEach(evo => {
-            const poke = pokemonList.find(_p => _p.name == evo.getEvolvedPokemon());
-            poke.eggCycles = Math.min(maxEggCycles, Math.round(p.eggCycles * 1.5));
-        });
-    }
-    (p as PokemonListData).nativeRegion = (p as PokemonListData).nativeRegion || GameConstants.MaxIDPerRegion.findIndex(maxRegionID => maxRegionID >= Math.floor(p.id));
-    pokemonNameIndex[p.name.toLowerCase()] = p;
-});
-
```
RedSparr0w, line 21347: I'm not sure if this will affect `pokemonMap` or not, I'll try give this a test in the morning though.

Otherwise as far as I can tell everything looks good 👍

RedSparr0w (approved):

PR #1940, opened 2022-04-02
Held items

Jaaslet:
Status: Ready to be reviewed. A lot of the items cannot be obtained, but Ultima will add obtain methods, after the merge.
Added held items.
- You can max use 6 held items at a time
- Each item can be used by one pokemon at a time
- Held items "breaks" after used
- Held items can be found in the item bag
- I have added "Wonder Chest" as a tutorial item. It can be bough it Violet City
- I have added "Macho Brace" to boost EV gain
- I have added "Light Ball", which will be given as a reward for beating red, when it's merged with Temporary battles
- I have added a lot of the type specific bonus attack items. They are found in the Johto dungeons
- I have added Quick Powder and Thick Club (don't know how they will be rewarded yet)
- Leek is used to evolve Galarian Farfetch'd into Sirfetch'd (obtain method incoming)


Aegyo (approved):
Looks like it all works fine, I have a couple of questions though:

1. A couple of items have their unlockRegion as Region.final - any reason for that?
2. legendary/mythic dungeon loot is going to make it very difficult for late game players to get those items. Maybe we could have those items in shops when you reach whatever region starts debuffing those dungeon loots? Especially since they will break after one use.

kittenchilly:
Metal Coat, Protector, Dragon Scale, Electirizer, Magmarizer, Up-grade, Dubious Disc, Reaper Cloth, Deep Sea Tooth, Deep Sea Scale should all be converted into held items for evos.

Jaaslet:
> * A couple of items have their unlockRegion as Region.final - any reason for that?

They don't have an obtain method yet, so i don't want them to show up on the list. I am thinking we want to release what we have, just to get feedback, before we find obtain methods for the rest.

> * legendary/mythic dungeon loot is going to make it very difficult for late game players to get those items. Maybe we could have those items in shops when you reach whatever region starts debuffing those dungeon loots? Especially since they will break after one use.

I think we should add them to dungeons in later regions too.
I was kinda hoping to get jk to do it (he is better at dungeon loot), after merge, but we might want to add it on this PR?

PR #1947, opened 2022-04-03
Fixed bug with shiny stone icon in stores

Jaaslet:
Fixes a small bug. How to reproduce it:
- Have an evo-stone, where you have not caught any of the baseforms
(i used ice stone in alola)
- Go to a store, that sells that stone
- Notice that the store claims you have all the shines, you can get from that stone

This is not totally wrong... You have all 0! But it looks weird...
This change does so the pokeball does not show up, if there are no pokemons, who can use the stone


RedSparr0w (approved):
Works well! Thanks for the PR 🎉

PR #1953, opened 2022-04-05
Added arrows to travel between subregions

Jaaslet:
Tried to make it easier to travel between subregions in Alola, by adding arrows


RedSparr0w:
👍 Nice work, just a couple of small changes:

The arrows appear even if there isn't any sub regions available/unlocked.
Maybe put the buttons in a `button group` so there isn't any spacing between them:
https://getbootstrap.com/docs/4.5/components/button-group/

Jaaslet:
> The arrows appear even if there isn't any sub regions available/unlocked.
> Maybe put the buttons in a `button group` so there isn't any spacing between them

Moved the visibility, to the button div
I added the button group. It looks a lot better, but it might be hard to know what you are pressing on (at least with "Cerulean"-theme)

RedSparr0w:
>  I added the button group. It looks a lot better, but it might be hard to know what you are pressing on (at least with "Cerulean"-theme)

Made the center button a primary and the arrows a secondary.
A lot more visible now, although sometimes the arrows blend into the background a little depending on the theme, but not enough to not notice they are there:
image
image
image

RedSparr0w (approved):
Thanks for the PR! Works great 🎉

PR #1955, opened 2022-04-05
Added tooltip to lock on dungeon pokemon

Jaaslet:
New players don't know NPCs tell them how to unlock locked dungeon pokemons (i just found out).
I tried to add a vague helptext, on the locks, which makes new players start talking with the NPCs.
The text should reflect that not all locks have an NPC explaining it


RedSparr0w (approved):

PR #1965, opened 2022-04-07
indicate HP bonus #1917

### LINKED ISSUE #1917: [SUGGESTION] In dungeons, indicate increased boss HP due to collecting chests
**Current Behavior**
Bosses in dungeons have increased HP based on how many chests you've collected, but this information is not clearly communicated. 

**Improved Behavior**
Add a UI element that does demonstrate that you are facing a boss with increased HP, ideally because you have collected treasure chests.

**Reason for change**
Players who don't do research off the game, and miss the NPC that tells you, will not know this information unless they happen to notice and ask about it or intuit it. This makes dungeons significantly harder that any other part of the game, especially since collecting chests is encouraged by revealing the map. 

Speaking personally, I just reached Kalos and learned this information after complaining that dungeons are much harder than gyms, which are trivial. This is because I have been making the game much harder for myself by always collecting chests, causing me to grind much longer than was necessary to play.

**Detailed Description**
Provide some element that does clearly communicates this behavior. This could be shown in early-game dialogue that's difficult to ignore (but I don't like this solution, because if you learn it early it's much less relevant and you could forget it), or ideally, on the screen during boss fights themselves. Even some simple thing like 'x2 HP" or "+20% HP" would be enough information for users to wonder why that is and look it up.

**Screenshots:**
This is merely to demonstrate what I mean and isn't a suggestion as to what it should look like.
increasedhp

Note the text below the health bar

JapuDCret:
closes #1917

---

Dungeon screen without any opened chests
pokeclicker_hp-bonus-indicator_0-chests

Dungeon screen with 1 opened chest
pokeclicker_hp-bonus-indicator_1-chest


RedSparr0w:
Not a huge fan of having the percentage on the HP bar.
Maybe we could instead just show it on dungeon chest, before opening?
Something like this maybe:
image

JapuDCret:
I implemented now like this:
firefox_qtYbfmOQsm

RedSparr0w (approved):

Issue #1941, opened 2022-04-02
[SUGGESTION] Adjust circumstances where "random evolution" text appears in Evolution Stones menu


PolygonStars:
While lurking in the Pokéclicker Discord, I noticed a discussion surrounding the text added in preparation for Alola, currently reading "If the requirements for more than one evolution is fulfilled, a random evolution will be selected". I didn't participate in the discussion while it was happening, but since it was established that this text isn't a bug, and that random evolutions are something the developers are interested in adding, I wanted to throw my own thoughts in here.

**Current Behavior**
The text described above appears for every Evolution Stone item when selecting Pokémon who can evolve into multiple forms.

**Improved Behavior**
Rather than appearing beneath every Pokémon with multiple possible forms from using a given Evolution Item, have this text appear only beneath specific Pokémon.

**Reason for change**
While players who are aware of the day/night cycle's implementation on Pokéclicker will know that Eevee's evolution into Espeon and Umbreon is not random, there is a point to be made that those who aren't aware and don't look it up may think otherwise. While time-based evolutions do occur elsewhere in Pokéclicker (For instance, Riolu into Lucario during the day), they aren't available until after Johto is completed, and other time-based events seen in the main Pokémon series, such as time-based encounters, do not occur on Pokéclicker - As another Johto-based example, Hoothoot can be encountered regardless of whether it is currently day or night. This discrepancy between the main series and Pokéclicker, combined with the text referring to "random evolution[s]", could cause new players to think that there is no time of day system at all, especially given the fact that before players can obtain a Soothe Bell to evolve Eevee, they will have likely already noticed that time-based wild Pokémon encounters aren't implemented the way they are in the main series. Choosing to have the text only appear under Pokémon whose evolution is legitimately random would help mitigate some of that potential confusion, and make it clearer which Pokémon the random evolution aspect applies to.

**Detailed Description**
[...]

PR #1976, opened 2022-04-09
"multiple stone evo is random"-warning only shows up, when relevant

Jaaslet:
Implements #1941


RedSparr0w (approved):
Looks great, thanks for the PR! 👍

Issue #450, opened 2019-10-09
Add desktop notifications


Claygame:
have it so that the website would request the notification permission just like the previous version which enabled the game to show notification of egg hatch and shiny encounter etc

amative1:
Completed per #1977

PR #1977, opened 2022-04-09
Desktop notification

Jaaslet:
Implements #450

Gives each notification-setting a checkbox for desktop notifications. They are false as default, and can only be checked, if the user have accepted desktop notifications.
I have added settings for some more basic stuff, so they can be used for desktop notifications too.

When/if the PR is merged, we should ask the crew to check if all relevant notifications can show up as desktop notifications. Relevant = notifications, which can show up, while the user does not have the game open.

And yes, the notifications looks ugly (at least in firefox), but it seams like we don't have much choice for how they look.


src/modules/notifications/NotificationConstants.ts
```diff
@@ -36,34 +45,37 @@ const NotificationConstants = {
...
+            achievement_complete: new NotificationSetting('notification.achievement_complete', 'Achievement complete', true, false, NotificationWarnOnBlocked),
+            new_catch: new NotificationSetting('notification.new_catch', 'New Pokémon/shiny captured', true, false, NotificationWarnOnBlocked),
+            battle_frontier: new NotificationSetting('notification.battle_frontier', 'Battle Frontier', true, false, NotificationWarnOnBlocked),
         },
         Hatchery: {
-            ready_to_hatch: new BooleanSetting('notification.ready_to_hatch', 'Egg ready to hatch', true),
-            hatched: new BooleanSetting('notification.hatched', 'Egg hatched', true),
-            hatched_shiny: new BooleanSetting('notification.hatched_shiny', 'Egg hatched a shiny', true),
-            empty_queue: new BooleanSetting('empty_queue', 'Hatchery queue is empty', true),
+            ready_to_hatch: new NotificationSetting('notification.ready_to_hatch', 'Egg ready to hatch', true, false, NotificationWarnOnBlocked),
```
RedSparr0w, line 59: Is there a way we could not specify `, false, NotificationWarnOnBlocked` on every `NotificationSetting` as these values shouldn't ever be different.
```suggestion
            ready_to_hatch: new NotificationSetting('notification.ready_to_hatch', 'Egg ready to hatch', true),
```

src/modules/notifications/Notifier.ts
```diff
@@ -14,24 +13,36 @@ export default class Notifier {
         sound = null,
         setting = null,
         image = null,
+        htmlLessMessage = null,
```
RedSparr0w, line 16: Is this necessary?
Could we instead just use something like:
```js
const tempEl = document.createElement('div');
tempEl.innerHTML = message;
const strippedMessage = tempEl.innerText.replace(/  +/g, ' ');
```
Only thing that I could see we need to do is replace any `<br>` with `\n` as they already get converted to `<br/>` within the notifier.

src/modules/notifications/Notifier.ts
```diff
@@ -14,24 +13,36 @@ export default class Notifier {
...
             // If we have sounds enabled for this, play it now
             if (sound) {
                 sound.play();
             }
 
+            if (setting && setting.desktopNotification.value && Notification.permission === 'granted') {
+                const desktopNotification = new Notification(title, {
+                    body: htmlLessMessage ?? message,
+                    icon: image,
+                });
```
RedSparr0w, line 38: Should we also respect the sound setting here?
```suggestion
                const desktopNotification = new Notification(title, {
                    body: htmlLessMessage ?? message,
                    icon: image,
                    silent: (check if sound disabled)
                });
```
Maybe it should also be silent if they have game and desktop notifications on, as the in game sound should still play?

src/modules/notifications/NotificationConstants.ts
```diff
@@ -36,34 +45,37 @@ const NotificationConstants = {
...
+            underground_energy_full: new NotificationSetting('notification.underground_energy_full', 'Mining energy reached maximum capacity', true, false, NotificationWarnOnBlocked),
+            event_start_end: new NotificationSetting('notification.event_start_end', 'Event start/end information', true, false, NotificationWarnOnBlocked),
+            gym_won: new NotificationSetting('notification.gym_won', 'Gym leader defeated', true, false, NotificationWarnOnBlocked),
+            offline_earnings: new NotificationSetting('notification.offline_earnings', 'Offline earnings', true, false, NotificationWarnOnBlocked),
+            achievement_complete: new NotificationSetting('notification.achievement_complete', 'Achievement complete', true, false, NotificationWarnOnBlocked),
+            new_catch: new NotificationSetting('notification.new_catch', 'New Pokémon/shiny captured', true, false, NotificationWarnOnBlocked),
+            battle_frontier: new NotificationSetting('notification.battle_frontier', 'Battle Frontier', true, false, NotificationWarnOnBlocked),
```
RedSparr0w, line 56: Could we disable the option for these to be turned off in game, but still able to enable desktop notifications, we want these notifications to always be displayed.

src/modules/notifications/Notifier.ts
Jaaslet: I also remove the pokemoney symbol, somewhere. And i think there also where a span...

src/modules/notifications/NotificationConstants.ts
Jaaslet: I can remove the false, but the NotificationWarnOnBlocked, was to prevent circular references. It was the best solution i could find.

src/modules/notifications/Notifier.ts
RedSparr0w: It should automatically remove them with the code above and only take the text.

Jaaslet: Oh, i replaced the pokemoney-symbol, with some text (see the notification for offlineprogress). Should have said that in my first comment...
I think i will keep the htmlLessMessage, only for that case, and then use your code for all other cases? Is that fine?

src/modules/notifications/NotificationConstants.ts
Jaaslet: done

src/modules/notifications/Notifier.ts
Jaaslet: So, right now the sound does not care, if you have the notification on or not. I did not know if it was intended, or not, but i did not change that part.

Firefox does not have sounds on their notifications (so i did not notice it, during development) and Chromes sounds are waaaaaay worse than ours... So i just disabled sounds... If they have sound enabled, it will now play the in game, no matter how their notification-settings are set. I think it's the best solution

src/modules/notifications/Notifier.ts
Jaaslet: It turns out that the notification just removes html (at least in firefox). So i am now replacing <br> with \n.
I kept the htmlLessMessage-code for the offline message (it had the pokedollar-symbol)

src/modules/notifications/NotificationConstants.ts
Jaaslet: done!

Jaaslet:
I think i have handled all reviews now. Thanks for the feedback!

src/modules/notifications/NotificationConstants.ts
RedSparr0w: Think I figured it out 🎉 

```diff
-import NotificationSetting from '../settings/NotificationSetting';
+import type NotificationSetting from '../settings/NotificationSetting';
```

We just had to import the type as that's all we are using in that file, and that removes the circular dependency! 👍

RedSparr0w (approved):

PR #1979, opened 2022-04-10
Update Achievement.ts

amative1:
Fixed display/sorting bug that prevented completed cheevos from being hidden when completed (if list is already set to sort as "incomplete" only)


amative1:
Nope, didn't fix.

PR #1980, opened 2022-04-10
Flash subregion-button when new subregion unlocked

Jaaslet:
To make it easier for players to see, when they unlock a new subregion in Alola


src/scripts/worldmap/MapHelper.ts
```diff
@@ -285,6 +285,7 @@ class MapHelper {
             // Gain queue slots based on highest region
             App.game.breeding.gainQueueSlot(App.game.breeding.queueSlotsGainedFromRegion(player.highestRegion()));
             GameHelper.incrementObservable(player.highestRegion);
+            player.highestSubRegion = 0;
```
RedSparr0w, line 288: I think this needs to make use of the following, as we don't have a setter/getter bound for it?
```suggestion
            player.highestSubRegion(0);
```

src/index.html
```diff
@@ -371,7 +371,9 @@ <h2 class="pageItemTitle" style="display: block;">
                         <img width=30px src="" data-bind="attr: { src: Weather.image() }"/>
                     </button>
                     <div id="mapBody" class="card-body p-0 show">
-                        <div id="subregion-travel-buttons" class="btn-group mr-2" role="group" style="width: 100%;" data-bind="visible: SubRegions.getSubRegions(player.region).filter(sr => sr.unlocked()).length > 1">
+                        <div id="subregion-travel-buttons" class="btn-group mr-2" role="group" style="width: 100%;" data-bind="
+                            visible: SubRegions.getSubRegions(player.region).filter(sr => sr.unlocked()).length > 1,
+                            css: { 'slow-flash': player.highestRegion() == player.region && SubRegions.getSubRegions(player.region).filter(sr => sr.unlocked()).length > player.highestSubRegion() + 1 }">
```
RedSparr0w, line 376: Would checking against the players current subregion work with this?
```suggestion
                            css: { 'slow-flash': player.highestRegion() == player.region && player.subregion < player.highestSubRegion() }">
```

Jaaslet: > I think this needs to make use of the following, as we don't have a setter/getter bound for it?

Fixed

> Would checking against the players current subregion work with this?

Then it will only flash, when the player is at an earlier subregion, than they have visited

src/scripts/worldmap/MapHelper.ts
Jaaslet: > I think this needs to make use of the following, as we don't have a setter/getter bound for it?

fixed

src/index.html
RedSparr0w: Yeah ignore this sorry, not sure what I was thinking 😅

RedSparr0w (approved):

PR #1981, opened 2022-04-10
Fix consol errors

### LINKED ISSUE #1743: [BUG] Console error The play() request was interrupted by a call to pause().
**Info**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 86.0.4240.198

**Description**
Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().

**Screenshot**
Image Text

Jaaslet:
I got a lot of console-spam (in firefox), so i tried to remove some of the unwanted. Ended up removing two errors, from the log:

- The value was checked, in RangeSetting, before the minValue og maxValue was set. This led the check to always return false.
Was only a problem in the constructor, and not when the value was later overwritten with the saved value, so it was only an error in the console.

- #1743 Some code failed, because it did not have a catch... But it still works fine... So i gave it an empty catch, to make it happy.

closes #1743


RedSparr0w (approved):

PR #1982, opened 2022-04-10
Now attempting to save on window unload

Jaaslet:
I have seen a couple of people been confused that the game only saves each 10th sec.
My solution is to do a save, when the window is unloading, which should fix most cases.
The save-code seams to be quick enough, that we are not hanging on to the window long enough, for the user to notice.
Only people, i could see not liking this change, is people exploiting the delay, to farm shiny from evo-stones... I don't know if that is a problem...

### DIFF
--- a/src/scripts/Game.ts
+++ b/src/scripts/Game.ts
@@ -191,10 +191,15 @@ class Game {
         }
 
         this.interval = setInterval(() => !this.worker || !Settings.getSetting('useWebWorkerForGameTicks').value ? this.gameTick() : null, GameConstants.TICK_TIME);
+        window.onbeforeunload = () => {
+            player._lastSeen = Date.now();
+            Save.store(player);
+        };
     }
 
     stop() {
         clearTimeout(this.interval);
+        window.onbeforeunload = () => {};
     }
 
     gameTick() {


src/scripts/Game.ts
```diff
@@ -191,10 +191,15 @@ class Game {
         }
 
         this.interval = setInterval(() => !this.worker || !Settings.getSetting('useWebWorkerForGameTicks').value ? this.gameTick() : null, GameConstants.TICK_TIME);
+        window.onbeforeunload = () => {
+            player._lastSeen = Date.now();
+            Save.store(player);
```
JapuDCret, line 196: maybe externalize the saving of the player to a private function, since this code would now be duplicated

RedSparr0w:
Hmm, not sure if we consider soft resetting a feature or not,
I'll leave this open for a bit just in case anyone else has an opinion on it.

Aegyo:
You could still soft reset by loading a backup

src/scripts/Game.ts
Jaaslet: done

JapuDCret:
I think for most players, the saving in the background is more magic and to have this "feature" raises more eyebrows in that group.

If one were to use a reload to soft reset a couple of seconds back, than that person is probably very experienced. And if we want to cater to these people, maybe we could add a toggle in the settings to disable save-on-unload?

RedSparr0w (approved):

PR #1991, opened 2022-04-11
Adding a little secret

NiCeDiCe90:
I hope it's ok and I didn't miss anything.


RedSparr0w:
This would probably be better as a Valentines event?

NiCeDiCe90:
> This would probably be better as a Valentines event?

That was my original idea for it but I was told we don't want any more RL related Events. 
I don't mind changing it to an Event encounter if you think that would be the better way.

Ultima1990:
Can maybe grab something from this list https://bulbapedia.bulbagarden.net/wiki/Category:Festivals_and_holidays

JapuDCret:
the [Pokemon Flower Arrangement](https://m.bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_Flower_Arrangement) sounds appropriate

RedSparr0w:
👍 Yeah some sort of event, as we don't have any other non cannon encounters that aren't events currently.

NiCeDiCe90:
So I found this one among the "in lore" Festivals: https://bulbapedia.bulbagarden.net/wiki/Kids_Day

But its a little weak. 

According to the real life Kids day in Japan it's part of a chain of multiple holidays called "golden week"
My suggestion for an ingame event would now be the "golden week event" from April 29th to May 6th.
would that be ok for you?
Just need to decide how to add it? (Another...) Kanto roamer?

RedSparr0w:
Doesn't need to be a Kanto appearance IMO

NiCeDiCe90:
Not a little secret anymore.
It's now a little event.

RedSparr0w (approved):

PR #1992, opened 2022-04-11
Now going to the correct subregion when moving region

Jaaslet:
When you moved from Alola, it remembered which subregion you where in. So when you moved back (using the dock), it set you to a town in subregion 0, but the map was still in the same subregion, as before you left.
This PR is a quick fix. As long as the main-dock is in subregion 0, it will work. So it might be a problem in later regions.

The "correct" fix is to add subregions to each town, like i have done with routes, and then change the subregion, in the MapHelper.moveToTown()-method.


RedSparr0w (approved):

PR #1993, opened 2022-04-11
Players highest subregion visited, is now saved

Jaaslet:
The "flash on new subregion" resetted on page reload. Fixed that.


RedSparr0w (approved):
👍 Nice catch

PR #1996, opened 2022-04-13
Add ability to hide dungeon bosses from the encounter list

RedSparr0w:
Bosses will be hidden if `hide: true`
Will be shown if the boss has a requirement, that is currently met.
if `hide: true` and no requirement, the boss will always be hidden (but can still be encountered).

Added `SeededDateRand` class, which is always seeded with todays date.
Added `SeededDateRequirement` which checks against a provided function.

Also updated `Ruins of Alph` to include all 28 `Unown`, but only 1 variant will be available per day.
image

### DIFF
--- a/src/components/townView.html
+++ b/src/components/townView.html
@@ -133,7 +133,7 @@
           <!-- /ko -->
           <!--Display all available bosses in this dungeon-->
           <!-- ko foreach: player.town().dungeon.bossEncounterList -->
-          <li class="list-inline-item">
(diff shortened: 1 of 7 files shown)


Jaaslet:
Discord exclusive unowns are no longer exclusive. Is that intended?

RedSparr0w:
Yeah, figured the base forms should probably just be available.
Still left them as purchasable in case you wanna get them early though I guess 🤷

PR #2003, opened 2022-04-13
Fix weird rounding of some items in the store

Jaaslet:
Some pokemons (like "Type: null") costed 1 less. This was because of a float getting rounded down, instead of just rounded


RedSparr0w (approved):

PR #2010, opened 2022-04-14
Change Unowns

Ultima1990:
Removed E and ?! from Ruins of Alph. E because it'd be nice to keep a Solaceon exclusive. ?! because I think we should just keep them unavailable until they are eventually introduced in Sevii Islands. Which definitely will happen at some point. Probably.

Then I went ahead and did the Alph thing to Solaceon Ruins. Adding all the Unowns on a rotating basis except ?! for the same reason I mentioned above, but also F, H and P, keeping them as exclusive to Alph. Thereby returning the old exclusivity to both locations.

### DIFF
--- a/src/scripts/dungeons/Dungeon.ts
+++ b/src/scripts/dungeons/Dungeon.ts
@@ -991,115 +991,103 @@ dungeonList['Ruins of Alph'] = new Dungeon('Ruins of Alph',
     [
         new DungeonBossPokemon('Unown (A)', 280000, 14, {
             hide: true,
-            requirement: new SeededDateRequirement(() => SeededDateRand.intBetween(0, 27) == 0),
+            requirement: new SeededDateRequirement(() => SeededDateRand.intBetween(0, 24) == 0),
         }),
         new DungeonBossPokemon('Unown (B)', 280000, 14, {
             hide: true,
-            requirement: new SeededDateRequirement(() => SeededDateRand.intBetween(0, 27) == 1),
+            requirement: new SeededDateRequirement(() => SeededDateRand.intBetween(0, 24) == 1),
         }),
         new DungeonBossPokemon('Unown (C)', 280000, 14, {
             hide: true,
-            requirement: new SeededDateRequirement(() => SeededDateRand.intBetween(0, 27) == 2),
+            requirement: new SeededDateRequirement(() => SeededDateRand.intBetween(0, 24) == 2),
         }),
         new DungeonBossPokemon('Unown (D)', 280000, 14, {
             hide: true,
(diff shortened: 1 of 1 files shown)


Qwertypop4:
Are Alph and Solaceon on the same rotation, or will they each have a different Unown each day?

Ultima1990:
It's a date based rotation. I _think_ the rotations would be out of sync because there's a different amount of possible mons at each location. Which means there's different Unown at each location on most days.

src/scripts/dungeons/Dungeon.ts
```diff
@@ -2692,11 +2680,98 @@ dungeonList['Solaceon Ruins'] = new Dungeon('Solaceon Ruins',
...
+            requirement: new SeededDateRequirement(() => SeededDateRand.intBetween(0, 22) == 20),
+        }),
+        new DungeonBossPokemon('Unown (Y)', 4100000, 30, {
+            hide: true,
+            requirement: new SeededDateRequirement(() => SeededDateRand.intBetween(0, 22) == 21),
+        }),
+        new DungeonBossPokemon('Unown (Z)', 4100000, 30, {
+            hide: true,
+            requirement: new SeededDateRequirement(() => SeededDateRand.intBetween(0, 22) == 22),
+        }),
```
RedSparr0w, line 2774: I was originally going to add something like this to this dungeon too, but the main issue is that it will be the same encounters (for most days).
The only thing I could think that might help, is adjusting the order?
Would still make each dungeons encounters basically tied together, but would at least be a different Unown in each dungeon on each day.

Ultima1990: Actually I think my theory that the cycles are out of sync because there's different amounts of Unowns holds up. Like, if I test it today there's different Unowns.

RedSparr0w: It'll always be similar though,
Like if you set the numbers to 0 → 100 and it selected 50 today, then you set the numbers to 0 → 50 it will be 25.

Ultima1990: How 'bout now?

Jaaslet: We could make them use different seeds. Like the first one uses CurrentDate, while the other one uses CurrentDate + 3. Should give a random enough pattern, while still maintaining the changes-each-day-feature.
It is probably out side Ultima's coding skills, but i can look at it, in some days, unless Red wants to

Ultima1990: > It is probably out side Ultima's coding skills

Oi. Rude.

But true.

PR #2011, opened 2022-04-14
PikaClones in Pikachu Valley

Ultima1990:
Added all of the previous generation's "Pikachu clones" to the random encounters in Pikachu Valley, giving a bit more encounter diversity than just Pikachu and Pichu over and over.


RedSparr0w (approved):
Thanks for the PR, Looks good 👍

PR #2012, opened 2022-04-14
Galar gyms and dungeons

Qwertypop04:
First pass at balancing Galar gyms and dungeons


src/modules/enums/Badges.ts
```diff
@@ -110,18 +110,20 @@ enum BadgeEnums {
     'Galar_Water',
     'Galar_Fire',
     'Galar_Fighting',
+    'Galar_Ghost',
     'Galar_Fairy',
     'Galar_Rock',
+    'Galar_Ice',
```
RedSparr0w, line 116: I'm not sure how the extra badges will work, or if this is the best way to add them yet.

Qwertypop04: The badge case changes just made allow a region to have 10 badges without it looking weird. If we still want them to be a multiple of 4, we can always merge the badges later, Someone already made images for that.

RedSparr0w: Sorry, wasn't meaning about having 10 badges, just having 2 different gym leaders from 2 different games.

Qwertypop4: It is a little weird, but it's better than just randomly choosing one game, when everything else in pokeclicker either uses expanded versions or mixes both versions. And there's no expanded version for gen 8

Jaaslet:
The code is failing, because 'Trainer Bede' is now called 'Gym Leader Bede' in the GymList, but not in the TownList (might have been lost in the merge. Wyndon Stadium got moved from PokemonLeague.ts to TownList.ts)

RedSparr0w (approved):

PR #2013, opened 2022-04-14
Added new requirements

Jaaslet:
The QB-team wanted requirements for completing a quest line and completing a part of a quest line. Here they are.
They are currently not in use, but they will be used, in the QB-PR


RedSparr0w:
Do you think it's worth adding some checks to ensure the questline exist?
As I could definitely see requirements being added, but the name being spelt wrong 😅

src/modules/requirements/QuestCompletedRequirement.ts
```diff
@@ -0,0 +1,22 @@
+import { AchievementOption } from '../GameConstants';
+
+import Requirement from './Requirement';
+
+export default class QuestLineCompletedRequirement extends Requirement {
```
RedSparr0w, line 5: This class seems to have the same name as the below class, some possible alternatives:
```suggestion
export default class QuestLineStepCompletedRequirement extends Requirement {
```
```suggestion
export default class QuestLineCompletedStepRequirement extends Requirement {
```
```suggestion
export default class QuestLineQuestCompletedRequirement extends Requirement {
```
```suggestion
export default class QuestLineCompletedQuestRequirement extends Requirement {
```

Jaaslet:
> Do you think it's worth adding some checks to ensure the questline exist?
Just to give a better error message?
It should fail, and give a console error, so it will quicker be noticed under development, imo

src/modules/requirements/QuestCompletedRequirement.ts
Jaaslet: I was planning to call it "QuestCompletedRequirement", but my copy/paste-skills failed me...
But i like "QuestLineStepCompletedRequirement" a lot better, so i choose that

RedSparr0w (approved):

PR #2014, opened 2022-04-14
Badge case is now split up by region

Jaaslet:
Now we can have 10 badges in Galar, without it looking weird!


RedSparr0w (approved):

PR #2021, opened 2022-04-15
Galar item evolutions and Gigantamax

Qwertypop04:
Adds Galar evolution items and evolutions. Also fixes Galarian Zapdos and Galarian Articuno being the wrong way around in RoamingPokemonLists.ts
Will also add Gigantimax forms later


RedSparr0w:
Could you split this in to different PRs?
Still not sure how Gigantamax forms will be handled, and if they will be positive IDs.

Qwertypop04:
Yeah I'll split them tomorrow. Although, I though we were going for all alt forms being decimal?
In terms of the way of handling Gigantamax, I'm not sure, the evo item is just a placeholder really. I do want to do them in some way for initial Galar, because if I don't I'm pretty sure they'll literally never happen.

RedSparr0w:
>Yeah I'll split them tomorrow. 

👍 
> I though we were going for all alt forms being decimal?

All alt forms, but not necessarily Mega/Gigantamax yet, as otherwise they become dex entries and appear in the Frontier.

PR #2023, opened 2022-04-15
"Sell all" in underground no longer start fossils

Jaaslet:
Fixed a bug, which made the "Sell all"-button start breeding fossils


RedSparr0w (approved):

PR #2025, opened 2022-04-15
Town code rework

Jaaslet:
My attempt to rewrite the Town code.
The basic idea is to give the town a "content"-argument, which contains everything that is in that town. This will solve the following problems:
- Mega dungeons (for Ultima)
- Multiple gyms per town (for the Galar-duo)
- Add a "Dock"-button to all the docks in Alola


src/scripts/gym/Gym.ts
```diff
@@ -14,11 +15,46 @@ interface gymFlags {
...
+    }
+    public areaStatus(): areaStatus {
+        if (Gym.isUnlocked(this)) {
+            if (!App.game.badgeCase.hasBadge(this.badgeReward)) {
+                return areaStatus.unlockedUnfinished;
+            } else if (!Gym.isAchievementsComplete(this)) {
+                return areaStatus.missingAchievement;
+            }
+        }
+        return areaStatus.completed;
```
RedSparr0w, line 56: Should this be returning `locked` if it's not unlocked yet?

src/scripts/gym/Gym.ts
```diff
@@ -27,15 +63,18 @@ class Gym {
...
             quest = true,
             achievement = true,
-        }: gymFlags = {}
+        }: gymFlags = {},
+        buttonText = 'Gym'
     ) {
+        super(requirements);
         this.flags.quest = quest;
         this.flags.achievement = achievement;
+        this.buttonText = buttonText;
```
RedSparr0w, line 77: Reduce some duplication:
```suggestion
        public buttonText = 'Gym'
    ) {
        super(requirements);
        this.flags.quest = quest;
        this.flags.achievement = achievement;
```

src/scripts/pokemons/PokemonHelper.ts
```diff
@@ -200,13 +200,13 @@ class PokemonHelper {
...
-            if (town.shops && town.shops.find(s => s.items)) {
+            if (town.content.find(c => c instanceof Shop && c.items)) {
                 // If we only want to check up to a maximum region
                 const region = town.region;
                 if (maxRegion != GameConstants.Region.none && region > maxRegion) {
                     return false;
                 }
-                const hasPokemon = town.shops.find(s => s.items?.find(item => item.name == pokemonName));
+                const hasPokemon = town.content.find(c => c instanceof Shop && c.items?.find(item => item.name == pokemonName));
                 if (hasPokemon) {
```
RedSparr0w, line 210: Might be better to instead filter only shops first so on the second find, you aren't parsing through all the content again.
Something like:
```js
const townShops = town.content.filter(c => c instanceof Shop && c.items);
if (townShops.length) {
....
                const hasPokemon = townShops.find(c => c.items.find(item => item.name == pokemonName));

RedSparr0w (changes requested):
Looking good to me so far,

Are we able to change it so Gyms appear first though?

Also worth noting that the Leagues don't appear (and seems to mess up other towns after visiting), but I presume you are still working on that.

src/scripts/pokemons/PokemonHelper.ts
Jaaslet: Done

src/scripts/gym/Gym.ts
Jaaslet: Does not work :(

src/scripts/gym/Gym.ts
Jaaslet: If the town is unlocked, but the gym is not, i don't want the town to be greyed out, like you can't access it. That was atleast the idea

Jaaslet:
> Looking good to me so far,

Thanks!

> Are we able to change it so Gyms appear first though?

It should already work like that

> Also worth noting that the Leagues don't appear (and seems to mess up other towns after visiting), but I presume you are still working on that.

I did not know that i had broken that part. Should be fixed now

Jaaslet:
I declare this PR ready for review!
I have more stuff i wanted to change, but i don't have any good ideas right now how to do it... And i probably need a break from this code. And i fixed the 3 things, in the description of this PR.

It should work as before, with some small changes (hopefully improvements):
- Hotkeys can now do everything in a town
- Clears are now shown beside each button. Which shows up in leagues
- Dock-button added to all Alola dock-towns

When it's merged, i will explain Ultima and Galar-duo how to use their new tools.
And please don't merge it the same day as we release a new version. :) I want a day or two for the pokecrew to play around with it, to check i did not change something while copy/pasting

RedSparr0w (approved):
🎉 Looks good, thanks for the PR! 👍

PR #2026, opened 2022-04-16
Unown tidy-up

amative1:
This is a WIP attempt to condense the code that generates the Unown for the Ruins of Alph dungeon. It currently fails multiple times over, however. Any suggestions are much appreciated.


amative1:
Should be g2g now, also added Solaceon (with a different random list). FHP and E are exclusive to their respective dungeons again, ! and ? have been removed (saving them for an upcoming project).

RedSparr0w (approved):

PR #2031, opened 2022-04-16
Remove trainer from Iron Island

Ultima1990:
Galactic Grunts random trainer. Just an accidental duplicate of the boss.


RedSparr0w (approved):
👍

Issue #2036, opened 2022-04-17
[SUGGESTION] reduce dungeon entry price with achivement progression


PokemonMrT:
i think it would be a nice change to have the entry cost of dungeons slightly decrease with progression mastery i.e after 10 / 100 / 250 / 500 clears, it doesnt have to be a big decrease but maybe 5% per tier could be nice and would make the later dungeons feel less grindy.

**Current Behavior**
Fixed entry cost for dungeon

**Improved Behavior**
5% (at your discretion) reduction per achievement unlock

**Reason for change**
i think it would be a nice little QoL update for people grinding the tokens needed for late game

Jaaslet:
Will close this issue. See linked PR for more information

PR #2039, opened 2022-04-19
Dungeon Cost Reduction

griffin104:
Related Issue: https://github.com/pokeclicker/pokeclicker/issues/2036

Reduced the cost of dungeon by 5% for every multiple of 10 completions (10, 100, 1000, etc.)


Ultima1990:
Not a fan.

Jaaslet:
Thanks for your PR, but in consultation with other developers and code contributors we have come to the conclusion that the idea does not currently fit our concept for the game. We find it to be a nerf, that is not needed.
I will close the PR.

PR #2044, opened 2022-04-19
Change Mina's Houseboat to be water environment.

Ultima1990:
Wouldn't want to imply her boat has a leak but boats are just tied to water and this certainly beats implying that Mina has a grassy field inside her boat.


RedSparr0w (approved):
👍

PR #2045, opened 2022-04-19
Adds a Tapu Village NPC for Crabominable

Ultima1990:
Or, as I like to call it, the bromance between a construction worker and his Crabominable.


RedSparr0w (approved):
👍

PR #2050, opened 2022-04-20
Added hotkeys to shipmodal

Jaaslet:
When you are in the shipmodal, you can now press the buttons, using the numbered keys


RedSparr0w (approved):

PR #2059, opened 2022-04-21
Hotkeys now ignore hidden buttons

Jaaslet:
You can reproduce the bug by:
- Having reached Alola
- Galar should not be unlocked
- Go to any starting town
- Press key [Non-NPC-buttons-length] + 1
This opens the "go to galar"-menu, before the PR


RedSparr0w (approved):

PR #2061, opened 2022-04-22
Pokemonlist under evostone now have bigger buttons

Jaaslet:
Some of the pokemons (like Alolan Meowth) had too long names, to fit on one line, on the button, when you have selected an evostone.
There where room to make the buttons bigger, so now it looks better (i think).


RedSparr0w (approved):

For the future though, if you make a mostly visual change, could you please also provide a screenshot.

PR #2064, opened 2022-04-22
Changed default value of berry time

Symi001:
By default, the berries are no longer in "next stage" display but in "ripe death" one.
Tested by playing a new file, value was correct in the settings, and it was correctly applied to the farm


RedSparr0w (approved):
👍

PR #2070, opened 2022-04-23
Mt. Chimney Megadungeon

Ultima1990:
Bit stuck on this one. Probably gonna need some kind of update function because I renamed the Mt. Chimney dungeon to Mt. Chimney Crater and then created a whole new town named Mt. Chimney. Apparently that means stuff gets confused.


RedSparr0w (approved):

Issue #2052, opened 2022-04-21
[SUGGESTION] move Sell all items button in underground


Claygame:
**Current Behavior**
Sell all item button in underground is right next to the Close button in underground in all tabs of underground

**Improved Behavior**
move the button to only "Treasures" tab and move it far away from Close button

**Reason for change**
that button can break the whole progress of someone who is following a chain but pressed the button accidently while trying to press the Close button. It would also be extremely annoying to lose all your underground items just because u misclicked a button that is placed poorly.

amative1:
Completed per #2074

PR #2074, opened 2022-04-24
Moved the underground "sell all" button to the tab

Jaaslet:
Solves #2052. The button is now at the bottom of the treasure tab.
image


Claygame:
can you make the button location/size something that cant be pressed accidentally? this huge button right above the Close window button looks easy accident and a devastating one for someone's progress

Jaaslet:
> can you make the button location/size something that cant be pressed accidentally? this huge button right above the Close window button looks easy accident and a devastating one for someone's progress

It's in the bottom of the treasure tab. So you first need to scroll all the way to the bottom.
And the butyon has a confirmation-window, so it should easy to avoid selling all.
Let me know if you still don't think it's enough

Claygame:
> It's in the bottom of the treasure tab. So you first need to scroll all the way to the bottom. And the butyon has a confirmation-window, so it should easy to avoid selling all. Let me know if you still don't think it's enough

ah, with a confirmation box it is good. Thank you for the change ^_^

RedSparr0w (approved):

PR #2065, opened 2022-04-22
One time battles (rival battles)

Jaaslet:
This PR will add One time battles as town buttons. It can later be added as icons on the map, by someone else.
It also moves the ???-battle in Aether Paradise, from a gym to a one time battle
This PR does not cover Ultima's wish that the pokemons should changed, based on requirements (like the starter). I am planning to do a small rework, at some point, which changes how the champion is coded, so all gym leaders, champions and one time battles, can change pokemons, based on different requirements.

PR #2075, opened 2022-04-25
Meltan!

amative1:
Framework for Meltan quest.


amative1:
Blocked by #2065

amative1:
Need to rebase, too many conflicts

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -1968,11 +1968,181 @@ class QuestLineHelper {
         App.game.quests.questLines().push(hoopaDayPikabluQuestLine);
     }
 
+    public static createMeltanQuestLine() {
```
Jaaslet, line 1971: The questlines needs to be moved up to the alola questlines

Jaaslet (changes requested):
Looks good. Have some comments:
- I have fixed some small errors
- The merge conflicts are too big for me to handle them on github. It's after Jessica moved pokemonList to moduels. After that, the indexes in the update functions needs to be moved up once
- The temp battles are missing images. Or you at least need to point it to the correct image, if they already exists as dungeon trainers or something

src/scripts/quests/QuestLineHelper.ts
amative1: Merge conflicts resolved, eslint passing.

amative1: Questlines moved into Alola section. Not sure what you mean by the indexes in the update functions. And idk if we have images for any of the temp battles yet (at the very least the two grunts, and maybe Faba? We should have the Leaders.)

amative1: (ah, I see, completely missed those Update lines... not that I would've figured out what needed fixing)

Jaaslet (approved):
I am going to approve this now. We are still missing some credits (i think?), but we can do that in an other PR, when we figure it out. I just want this merged :)
Looks good. Thanks for the PR

PR #2081, opened 2022-04-26
Fixed flutes multiplier

Jaaslet:
Before this PR, flutes started at 0% bonus, and gave 2% bonus at 100% achievement bonus. Now it starts at 2%.
This follows everything else that uses achievement bonus


RedSparr0w (approved):

PR #2083, opened 2022-04-26
The first 3 Sevii Islands

Ultima1990:
This is a very basic implementation of One Island, Two Island and Three Island as a subregion of Kanto, requiring completion of at least some of its content, one route and one dungeon, between the Cinnabar and Viridian gyms, and requiring completion of one more route and dungeon if the pokédex requirement is active because Moltres has moved to Mt. Ember.

The main issue is that by implementing it this way, with four new routes, the routehealth of route 29 in Johto is pushed up from 45710 to 65651. Basically this affects balance across the entire game. So maybe this is not the way to do this.


lytium:
Why not release them separate from Kanto ? (And so, after alola or Galar.) 
So, it doesnt fuck up the balance. 

To make them worth to do, we can add pokemon from the island of the anime. (Crystal onix, pink pokemon) or even Alpha pokemon of Gen1/2 pokemon, like in Alola.

Just a random idea. ~

Ultima1990:
Because that's simply not where they belong.

RedSparr0w:
Before we merge something like this, we definitely need to add something to the `Route` class so these routes don't affect future regions.
Maybe just manually assigned HP would work, although would become a bit labour intensive.
I'm thinking some flag that excludes it from the current calculations, but you can also add a number for where it should be in terms of HP:
```js
{
  // Excluded from the calculation formula for main progression routes
  calculate: false,
  // Would have the same HP as the current 60th route in the game
  // Would override the main calculation if not excluded from the calculations
  // Required if excluded from main calculation
  // Auto excluded if this option is specified?
  hp: 60
  // Able to specify a fixed HP amount, doesn't get calculated, basically same functionality as above
  fixedHP: 175000
}
```

Ultima1990:
In this version of a Sevii islands One, Two and Three implementation, where they are actually part of the Kanto region progression, the thing that I mentioned as the "main issue" is definitely something you'd expect. The question is, is that something we want?

In potential future optional subregions that are not part of region progression, yes, we'd want to do that thing with the Route class.

Ultima1990:
I thought of one more thing that needs to be done before this is ready:
Mew needs to be restricted to only roam in mainland Kanto.

src/modules/routes/Routes.ts
```diff
@@ -36,10 +42,21 @@ export default class Routes {
...
+            for (let i = this.regionRoutes.findIndex((routeData) => routeData.region === region && routeData.number === route) - 1; i >= 0; i--) {
+                if (!this.regionRoutes[i].ignoreRouteInCalculations) {
+                    return i + 1;
+                }
+                if (i === 0) {
+                    throw new Error('Not implemented for ignoreRouteInCalculations = true on first region route');
+                }
+            }
+        }
+        return filteredRegionRoutes.findIndex((routeData) => routeData.region === region && routeData.number === route) + 1;
```
RedSparr0w, line 60: Could we please get some code comments added here, as it's a bit complicated what's happening

src/scripts/Battle.ts
```diff
@@ -108,7 +108,7 @@ class Battle {
      */
     public static generateNewEnemy() {
         this.counter = 0;
-        this.enemyPokemon(PokemonFactory.generateWildPokemon(player.route(), player.region));
+        this.enemyPokemon(PokemonFactory.generateWildPokemon(player.route(), player.region, SubRegions.getSubRegionById(player.region, player.subregion)));
```
RedSparr0w, line 111: Should we be storing the current SubRegion object somewhere so we won't need to call the `SubRegions.getSubRegionById` every time we generate a Pokémon?
Similar to how we store `player.town`

src/scripts/pokemons/RoamingPokemonList.ts
```diff
@@ -4,31 +4,39 @@
 ///<reference path="../../declarations/requirements/OneFromManyRequirement.d.ts"/>
 
 class RoamingPokemonList {
-    public static list: Partial<Record<GameConstants.Region, Array<RoamingPokemon>>> = {};
-    public static increasedChanceRoute: Array<KnockoutObservable<RegionRoute>> = new Array(GameHelper.enumLength(GameConstants.Region)).fill(0).map((route, region) => ko.observable(null));
+    public static list: Partial<Record<GameConstants.Region, Array<Array<RoamingPokemon>>>> = {};
+    // TODO: Right now increasedChanceRoute limits our subRegionGroups to 3.
+    // This is hardcoded because we cannot get number of subregiongroups at this point in the code.
+    public static increasedChanceRoute: Array<Array<KnockoutObservable<RegionRoute>>> = new Array(GameHelper.enumLength(GameConstants.Region))
+        .fill(0).map(() => new Array(3).fill(0).map(() => ko.observable(undefined)));
```
RedSparr0w, line 11: Would it be worth limiting it to `4` now rather than later, as Alola will eventually be 4 sub regions, and chances are this piece of code will be forgotten when splitting it.

src/scripts/pokemons/RoamingPokemonList.ts
```diff
@@ -38,80 +46,60 @@ class RoamingPokemonList {
...
+                const selectedRoute = SeededRand.fromArray(routes);
+                route(selectedRoute);
+            });
         });
     }
 }
 
 // Kanto
-RoamingPokemonList.add(GameConstants.Region.kanto, new RoamingPokemon('Mew'));
+RoamingPokemonList.add(GameConstants.Region.kanto, 0, new RoamingPokemon('Mew'));
```
RedSparr0w, line 61: Would be better to use our `KantoSubregions` variable in this file instead of magic numbers.

RedSparr0w: Actually it's probably worth moving all the individual subregion contants to `GameConstants` as an array and handle it simillarly to `RegionDungeons`.

That way we could also remove the `TODO` above about not knowing the max subregions count.

src/scripts/gym/GymList.ts
```diff
@@ -114,7 +117,7 @@ GymList['Viridian City'] = new Gym(
     6000,
     'Ha! That was a truly intense fight! You have won! As proof, here is the Earth Badge!',
     [
-        new GymBadgeRequirement(BadgeEnums.Volcano),
+        new QuestLineCompletedRequirement('Bill\'s Errand'),
```
RedSparr0w, line 120: Does this need to be locked behind Sevii?

src/scripts/specialEvents/SpecialEvents.ts
```diff
@@ -45,11 +45,11 @@ class SpecialEvents implements Feature {
 SpecialEvents.newEvent('Lunar New Year', 'Encounter Fancy Pattern Vivillon for a limited time roaming Kalos.',
     // Start
     new Date(new Date().getFullYear(), 0, 24, 1), () => {
-        RoamingPokemonList.add(GameConstants.Region.kalos, new RoamingPokemon('Vivillon (Fancy)'));
+        RoamingPokemonList.add(GameConstants.Region.kalos, 0, new RoamingPokemon('Vivillon (Fancy)'));
```
RedSparr0w, line 48: Same thing in this file for the magic numbers

src/scripts/towns/TownList.ts
```diff
@@ -190,12 +191,49 @@ const SaffronBreeder = new NPC('Breeder', [
 
 const FuchsiaKantoRoamerNPC = new RoamerNPC('Youngster Wendy', [
     'There\'s been some recent sightings of roaming Pokémon on {ROUTE_NAME}!',
-], GameConstants.Region.kanto);
+], GameConstants.Region.kanto, 0);
```
RedSparr0w, line 194: Same thing for this file and the magic numbers

src/scripts/gym/GymList.ts
Ultima1990: Yes the intent is that this Sevii map becomes a required part of the Kanto experience. Which is also why I think it's a good place to put the client content.

The other Sevii maps, once added, should be fully optional for players that don't care about getting all the forms.

Ultima1990: It's also reflected in the NPC dialogue during the questline:
"I can see that your friend is eager to get back to Kanto and challenge the Pokémon League."

src/modules/GameConstants.ts
```diff
@@ -1200,3 +1202,50 @@ export enum DayOfWeek {
...
+export const RoamerGroups = [
+    RoamerGroupsKanto,
+    RoamerGroupsJohto,
+    RoamerGroupsHoenn,
+    RoamerGroupsSinnoh,
+    RoamerGroupsUnova,
+    RoamerGroupsKalos,
+    RoamerGroupsAlola,
+    RoamerGroupsGalar,
+];
```
RedSparr0w, line 1251: Not sure how I feel about the naming of this..
Why not just have these as subregion, as it's essentially the same thing, but makes sense when it's used in other contexts too.

RedSparr0w: Will also mean we don't need to change multiple places whenever a new subregion is added.
As these values are currently mostly duplicated here `src/modules/subRegion/SubRegions.ts`.

Jaaslet: Galar is split into north and south in the galar PR. The first Galar roamer group will cover both north and south subregion.
And Alola has 3 subregions, but only one roamer group.
I am not sure how i should use subregions here?

src/scripts/Player.ts
```diff
@@ -45,6 +46,10 @@ class Player {
         }
         this._region = ko.observable(savedPlayer._region);
         this._subregion = ko.observable(savedPlayer._subregion || 0);
+        this.subregionObject = ko.observable(SubRegions.getSubRegionById(this._region(), this._subregion()));
+        this._region.subscribe(r => this.subregionObject(SubRegions.getSubRegionById(r, this._subregion())));
+        this._subregion.subscribe(sr => this.subregionObject(SubRegions.getSubRegionById(this._region(), sr)));
+
```
Aegyo, line 52: You can use a pureComputed to automatically subscribe to these

```suggestion
        this.subregionObject = ko.pureComputed(() => SubRegions.getSubRegionById(this._region(), this._subregion()));
```

Jaaslet: I have committed your suggestion, and it still works!

src/modules/GameConstants.ts
Jaaslet: The gameconstant enums are now the subregions, and roamergroups are now defined in the roamerlist

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -134,6 +134,37 @@ class QuestLineHelper {
         App.game.quests.questLines().push(undergroundQuestLine);
     }
 
+    public static createBillSeviiQuestLine() {
+        const billSeviiQuestLine = new QuestLine('Bill\'s Errand', 'Bill has asked you to journey to the Sevii islands with him to set up a digital connection to mainland Kanto.');
```
CorgiOnNeptune, line 138: Missing capital on Islands :)

Ultima1990: I fix!

Jaaslet (approved):
Crew has played through it, and we have solved their comments.
Both Red and Jessica have looked through most of the code (including everything i wrote) and i have looked through what was made since, and it looks good.
This PR is ready to be merged!

PR #2090, opened 2022-04-28
Added a tooltip to sort direction and fixed sortdirections

Jaaslet:
People where confused


RedSparr0w:
We will also need to adjust the actual sort directions with this:
Specifically `Attack`, `Base Attack`, `Breeding Effeciency`:
image
image
image
image
image
image

It's also probably worth adding the tooltip to other places we have sorting:
- Main hatchery sort
- Items menu (protein)
- Pokémon List
Think that might be all the places.. 🤔

Jaaslet:
Done and done.
Held items also uses sorting, but i will add that to the held item PR, after this have been merged

RedSparr0w (approved):

PR #2091, opened 2022-04-28
Gives the player a starter for each region

Jaaslet:
Now the player can select a starter, for each region (requested by Ultima).
I could not come up with a fun text, so i just made a basic one. If someone has any ideas, feel free to share it, and i might use it.
If the player have not selected a starter (either they refreshed when the window was open, or they are an old player), it will teleport the player to the newest region and give them the window again. If other than player.highestRegion are lacking starters, the first one will be selected as the starter (but not given to the player). It will not make much sense to let the players choose a starter, if they most likely already have it. It does not change much, which starter they have selected.
image


Ultima1990:
Just a quick check, is this gonna allow the Pikachu option by closing the modal several times in every region? Probably wouldn't want that...

Ultima1990:
Oh. OH! An idea has jumped into my brain. If we want to keep the sneaky 4th option thing going in every region, we can do the Capachus. And if you don't pick the Capachu you have to wait till Pikachu Valley to get it.

RedSparr0w:
I haven't had a chance to test this yet, but just a couple questions.
What happens for player who are already in newer regions, will they be asked to pick a starter for each of their previous regions on update?
What happens if you refresh the page while the new starter modal is shown, will it be shown when the load the game again?

Jaaslet:
> Just a quick check, is this gonna allow the Pikachu option by closing the modal several times in every region? Probably wouldn't want that...

I made a new window and new code for the other regions. The other one is a part of the tutorial, so it was different enough, that i did not want to reuse the code

Jaaslet:
> What happens for player who are already in newer regions, will they be asked to pick a starter for each of their previous regions on update? 

They don't get to pick a starter, and we pretend like they picked the first one in each region.

> What happens if you refresh the page while the new starter modal is shown, will it be shown when the load the game again?

Yeah... i got lazy... you don't get a starter, and we pretend like you picked the first. I was hoping no one refreshes their game here.
I can try to figure out a way to fix that. It might be easier to program, if we also want the window to show up for old players, for each region

Jaaslet:
> If we want to keep the sneaky 4th option thing going in every region, we can do the Capachus. And if you don't pick the Capachu you have to wait till Pikachu Valley to get it.

I liked playing though Pikachu Valley, so i don't like to ruin that experiance for players, who looked at the wiki for the optimal way to play

Jaaslet:
I have now fixed the refresh problem (see description for how).
The secret cap pikachu will not be a part of this PR (if we even want it).
So now it should be ready for review

src/scripts/gym/Champion.ts
```diff
@@ -18,22 +18,22 @@ class Champion extends Gym {
         super(leaderName, town, basePokemon, badgeReward, moneyReward, rewardMessage, requirements, rewardFunction);
     }
 
-    public setPokemon(starter: GameConstants.Starter) {
+    public setPokemon(starter: number) {
         this.pokemons = [...this.basePokemon];
         switch (starter) {
-            case GameConstants.Starter.Bulbasaur: {
```
RedSparr0w, line 24: Are these still used anywhere now? or should they be removed from GameConstants too?

RedSparr0w (changes requested):
At the moment, you can just click straight off the starter modal and it will disappear, we should probably disable this or make it re-appear until they choose a starter.

Jaaslet:
> At the moment, you can just click straight off the starter modal and it will disappear, we should probably disable this or make it re-appear until they choose a starter.

This comment is fixed. The other comment is waiting for a discord discussion

RedSparr0w (approved):

PR #2095, opened 2022-04-29
Discord pokemons can no longer be unlocked in earlier regions

Jaaslet:
If you for example try to unlock an unknown in kanto, it will now give you an error. 
I think people will hate this change, but it feels correct (and makes it a tiny bit more fun to unlock a new region)


amative1:
Can it be coded to "accept" the Unown code, but it just doesn't provide any attack until Johto is reached?

Jaaslet:
> Can it be coded to "accept" the Unown code, but it just doesn't provide any attack until Johto is reached?

That would be a lot of work, but yes.
What are the reasons for us wanting it to work like that?

amative1:
I'm just thinking about the after-support, when we get a bunch of questions about why the code people bought doesn't work

RedSparr0w (approved):

PR #2096, opened 2022-04-29
Underground Changes

jk13pclick:
It's happening. Task list and Nerfs/Buffs to follow.

### NERF(s)
- [ ] **Daily Deals currently restricted to `25` per day per trade:** Need to change this to scale as per dev-chat recommendation
- [ ] **Rebalance the Dungeon Loot:** This is a big one, doesn't need to happen entirely in this PR. Ideally, it should be quicker to get diamonds from the Underground than in Dungeons. Improving the active yield from the Underground should help - if we need to alter how Diamonds from Dungeon Loot works that is okay too. Should probably be packaged with the larger Loot Rebalance V2

### BUFF(s)
- [ ] **Rebalance the energy gain upgrades:** Make these easier to get initially?
- [ ] **Hammer/Chisel Upgrade:** The current underground meta is bombing until you reveal an item, and then chiseling until you get the item. Restores are cheap, so you can do this pretty indefinitely. A Hammer/Chisel upgrade (more tiles damaged/wider area affected) could change this meta back towards actively mining.
- [ ] **Weight the Underground Items:** Diamond Items need to show up more. Generally, the order should go Diamond Items > Fossils > Stones/Plates
- [ ] **Maybe change the value of some items:** Diamond Items probably should be worth more, especially with Hatchery Helpers that consume them.


Chronicar:
Have you thought of making plates give a random number of diamonds when you "sell" them? Maybe 1-3 diamonds since they are the most prevalent item in the underground.

jk13pclick:
Going to close this. Seems like Shards have changed the Underground landscape, and I don't have motivation to pick this PR up again at the moment. Maybe I'll revive it some day if necessary.

PR #2100, opened 2022-04-29
Category Sort Fixup

jk13pclick:
When selected, category sorts all Pokémon with a Category to the end of the list by default. This inverts it so they appear first.


RedSparr0w (approved):

PR #2102, opened 2022-04-29
Proposed rebalance

Ultima1990:
For Dungeons I've mostly moved some HP away from the bosses and into the random encounters.

For gyms the first 2 have received a buff, and every single gym after that has received a nerf. In Diantha's case over 50%.

All of this to give the region a somewhat smoother curve towards Alola, and to bring the region more in line with the scaling of its route difficulty, so that route difficulty doesn't become utterly irrelevant compared to gyms and dungeons. The same way I balanced Alola.


Ultima1990:
I've made a change to how I'm dealing with dungeon health.

Everything here comes from the following two ideas:
Route difficulty should not become irrelevant.
Gyms and the Pokémon league should be the big road bumps. Bigger than dungeons. Exceptions can be made for optional and post-league dungeons.

With this PR gyms in Kalos have a total health pool of 10x the most recent routehealth. This means we expect players to clear ten times the most recent route's routehealth in 30 seconds.
Before my latest push dungeon random encounters had x2 the most recent routehealth and bosses had a total health pool of 10x the most recent routehealth. This means the expectation here was that the player will find the boss and clear ten times the most recent routehealth in 60 seconds. Which means dungeons are way harder than dungeons, especially in later regions and especially with trainers. Because essentially, in a world where we want dungeons and gyms to be about equal difficulty, if a player can clear 10x in 30 seconds, then they can clear 20x in 60 seconds. 10x of those 20x is the boss. Which means the other 10x is random encounters. Random encounters are 2x. So we assume 5 random encounters on average (ignoring travel time, empty tiles, treasures). This is just not realistic, especially later on with very big dungeons and trainers.

So now I've changed dungeons to x1,5 the most recent routehealth for random encounters and a total health pool of 7,5x for bosses. I do want dungeon random encounters to be harder than routemons, and I also want to keep a big x5 difference between random encounters and bosses. Bosses should be beefy compared to the small fry. 20x in 60 seconds, 7,5 of that is the boss, 12,5 for random encounters, allowing for 8,333(etc) average random encounters. Which I think is more reasonable.

PR #2110, opened 2022-05-01
Reduce save file size

RedSparr0w:
This PR removes some of the useless information from save files.

For a basic comparison of before/after:

Before | After | Notes
:--: | :--: | :--
33.7kb | 25.7kb | Brand new save, no starter
86.5kb | 49.6kb | Save progress finished Kanto
231kb | 111kb | Save progress most of the way through Sinnoh
369kb | 173kb | Save progress finished Alola

Statistics:
- Removes any object keys that have a value of 0

Hatchery Helpers/Farm Hands:
- Removed any keys that contained values which never change

Farm:
- Mutations only had 1 key, it's now just an array of true/false

Party Pokémon:
- Removed any keys where the value was still the default value
- Bound save keys to enum values as these keys are repeated for every caught Pokémon

Achievements:
- No longer tracking completed status, calculate it on game start.

Player:
- Removed unused keys
- Remove default values from some of our keys


Jaaslet:
I have tested the stuff mentioned in the description, taken a quick overview, to see that everything else is saved as normal and scanned the code. I did not find anything

PR #2113, opened 2022-05-02
Add image credit for Kalos images

Ultima1990:
All from #1190. Couldn't find the source for the Sightseer (female) and our Clemont is slightly different.

Also removed Team Flare Admins and Team Flare Aliana because they came from a deviantart user that explicitly prohibits republication on their about page. https://www.deviantart.com/muzyun/about


RedSparr0w (approved):

PR #2117, opened 2022-05-03
Replace all Pokémon League sprites

Ultima1990:
Except the Kanto one. Cuz that one's already good.
Gave all the other Pokémon Leagues proper Pokémon League buildings. Instead of using that Weather Institute thing.

Even did the Alola one. Even though that map is going to go the way of the dodo eventually. Cuz why not.


RedSparr0w (approved):

PR #2126, opened 2022-05-07
Option to hide shiny Pokémon in the protein menu

FredisonP:
Could be useful for people who want to prioritize non-shinies for breeding.


Jaaslet (approved):
It works as it should and the code looks good.
Right now the options are centered, which means the checkboxes are not aligned. I am not sure, but i think it looks too weird, if the options are left-aligned, so i think i prefer it as it is right now.
Current look:
image


With left align:
image

PR #2141, opened 2022-05-10
Shows sub region name

### LINKED ISSUE #2129: [SUGGESTION] Change Alola's "Subregional Travel" Text to Island Name
**Current Behavior**

The island navigation in Alola always says "Subregional Travel".

**Improved Behavior**

The island navigation in Alola says the name of the island e.g., Melemele Island.

**Reason for change**

It's easier to identify where I am. I know players will eventually get used to knowing where things are by seeing the islands long enough, but knowing what the islands look like and their names are different things, especially if discussion around islands involving their proper names are used. I know clicking on Subregional Travel opens a popup that specifies the island names, which is nice, but I think it would be helpful to see the island name instead of "Subregional Travel" text.

**Detailed Description**

"Subregional Travel" sounds good to make new players aware of the functionality, I guess. However, the text doesn't make me know what it does without clicking it. And I think the gameplay does a good enough job at highlighting to the player that the next island is available by blinking it so I don't think the text is needed. "Island Travel" would make more sense to me than "Subregional Travel", but I digress.

**Screenshots:**

Jaaslet:
Fixes #2129 (and a small code cleanup)

image


RedSparr0w (approved):

PR #2142, opened 2022-05-10
Now clarifies hatchery helpers cost is per hatch

### LINKED ISSUE #2140: [SUGGESTION] Add cost scale to Hatchery Helpers
**Current Behavior**

The Hatchery Helpers table displays the cost.

**Improved Behavior**

The Hatchery Helpers table displays the cost and the unit of measurement (or scale). Similar to how Farm Helpers uses "/hour", Hatchery Helpers should have a "/hatch".

**Reason for change**

It's not explicitly obvious in-game what the cost unit for Hatchery Helpers are. The other helpers are Farm Helpers and those are billed per hour. I know the wiki clarifies this but having it in-game would be helpful.

**Detailed Description**

**Screenshots:**

Something like this
10-090855_jHTPAfB5wH

Jaaslet:
Sorta fixes #2140. I tried a couple of different solutions, and this was the one i liked the most.

image


RedSparr0w (approved):

PR #2146, opened 2022-05-10
EXP Rebalance

jk13pclick:
EXP is way too easy to get, especially in later regions. This is my attempt to rebalance the curve. See the linked Desmos graphs for comparisons.

Wild Level Rebalance - https://www.desmos.com/calculator/8lwxeki7us
EXP Yield Rebalance - https://www.desmos.com/calculator/gdjw6zokji


RedSparr0w:
I haven't tested that this works as expected, but the graphs look good to me, kinda buffs early game a bit, but definitely nerfs later game.

jk13pclick:
> kinda buffs early game a bit, but definitely nerfs later game.

I felt that the buff to early game was okay, especially with the huge nerf to late game. I haven't played through early game, but my assumption is that it won't be balance breaking. 

I left this as a Draft mainly because it is experimental and may need some more stuff to make it feel good - if you feel like it could be worth sending to crew as-is that is fine by me.

Jaaslet (approved):
It seams to work as the graph shows. At least exp is actually something you notice in Alola now.
As decided by the devel-team, the graph looks good, so let's release it as see what happens.
Thanks for the PR!

PR #2148, opened 2022-05-11
Areas with quests are now dark green

### LINKED ISSUE #1017: [SUGGESTION] Add highlight to quest locations

Jaaslet:
Closes #1017
image
image
image

Currently it only works for daily quests. I looked into doing it for questlines. Questlines currently are either pointing towards new locations (which will be orange) or are meant to be a search, So we will only implement it for daily quests (for now)


Jaaslet:
I tried to make the current location lighter (suggestion from Red), but it got too close to the shiny color. And i think it's a bigger problem if current location looks like the shiny color, than it looks like the quest color.

RedSparr0w (approved):

PR #2164, opened 2022-05-14
Some temporary battle fixes

Ultima1990:
Adds Ultra Wormhole sprite. Also used for Ultra Megalopolis. Very ugly. Anyone is free to make something better.
Turned Ultra Megalopolis into a one time battle. Removed Elite_UltraNecrozma badge.
Changed Eater of Light quest to target the TemporaryBattles instead of the Gyms.


Someonealive-QN:
Like this? 
86 Sem Título_20220514112541
 Ins't mine but the person who made them said its free to use 
Screenshot_20220514-111344
Here

Ultima1990:
> Like this? 86 Sem Título_20220514112541 Ins't mine but the person who made them said its free to use Screenshot_20220514-111344 Here

Link?

Someonealive-QN:
Here:
https://www.pokecommunity.com/showthread.php?p=9849231

src/scripts/Update.ts
```diff
@@ -716,6 +716,12 @@ class Update implements Saveable {
             // Remove the Elite_Nihilego Gym, now a temporary battle instead of a gym
             saveData.statistics.gymsDefeated.splice(84, 1);
             saveData.badgeCase.splice(84, 1);
+            // Change Ultra Megalopolis to a Temporary Battle
+            saveData.statistics.temporaryBattleDefeated = new Array<number>();
```
Jaaslet, line 720: This line needs to be removed. It's overwriting the Nihilego temp battle update

src/scripts/Update.ts
```diff
@@ -716,6 +716,12 @@ class Update implements Saveable {
...
+            saveData.statistics.temporaryBattleDefeated[1] = saveData.statistics.gymsDefeated[89];
```
Jaaslet, line 721: Because Nihilego have been removed earlier in the update function, this gym now has index 88 and not index 89

Jaaslet (changes requested):
You also need to remove 'Elite_UltraNecrozma' from badges.ts for it to work.
I have reviewed the code, tested the update function and tested it as a lategame player.
I have not yet tested if the quest still works

Jaaslet (approved):
Looks good.
The update function now works and i have now also played though the quest as a new Alola player. It works just as before.
So i have no more comments

PR #2165, opened 2022-05-14
Bulletin board (quest hub)

Jaaslet:
We have a lot of plans for quests, and a lot of them are optional.
When players start complaining that there are too many quests in their quest-window, we can just merge this PR, and it will be fixed!

The idea is each questline will have an "optional" parameter. If that is set, the quest will end up on the bulletin board, instead of in the quest-box. Then the players can accept it, when they are ready for more spam in their main window!
The bulletin board can be accessed at all professors lab.

image

To tester: There is a board set up in the home town of Kalos. 'The Great Vivillon Hunt!' and 'The new kid' are currently the only optional quests. I will try to keep this PR updated, when other quests gets merged. To test it, you either need to unlock one of those quests, or just run the following command, to reset the quest:
`App.game.quests.getQuestLine('The Great Vivillon Hunt!').state(QuestLineState.onBulletinBoard)`
If you want to see more than one quest, you can add all quests to the board by using the command:
`App.game.quests.questLines().forEach(q => q.state(QuestLineState.onBulletinBoard))`
(Note that non-optional quests should not be able to end up on the board, unless you are using a command like this).

How to add more quests to the board: QuestLine-class now have two extra parameters. The first new needs to be true, and then the next will decide what board it will show up on. Default is All boards. Example is:
`new QuestLine('The Great Vivillon Hunt!', 'Discover the beauty of Vivillon', true, GameConstants.BulletinBoards.Kalos);`

How to add more boards: Add an other element to GameConstants.BulletinBoards. Add a new BulletinBoard-element to a town, as a townContent. Example:
`TownList['Vaniville Town'] = new Town(
    'Vaniville Town',
    GameConstants.Region.kalos,
    [VanivilleTownShop, new BulletinBoard(GameConstants.BulletinBoards.Kalos)],
    {
        requirements: [new GymBadgeRequirement(BadgeEnums.Elite_UnovaChampion)],
    }
);`
Use the newly added GameConstants.BulletinBoards in both the new quests and as argument in the town content element.


RedSparr0w:
Could we remove the custom font, as this will add a lot more bandwidth to the server.

Have the images been minified with something like https://tinypng.com/?
If they already have, are we able to find a repeating corkboard pattern instead of the 3 separate images as this will be a lot smaller in size.

Could we also instead just make use of bootstrap cards instead of the paper/pin?
https://getbootstrap.com/docs/4.5/components/card/
This way it will follow whatever current theme the user is using and work with any screen size without blurring the image,
Could always add a pin image/svg to the top of the card if needed.

Jaaslet:
I have removed the corkboard and funny font. I have added Ash's quests as optional.
It's ready for review again.

src/scripts/quests/QuestLine.ts
```diff
@@ -61,7 +65,11 @@ class QuestLine {
         this.quests.push(quest);
     }
 
-    beginQuest(index = 0, initial?: number) {
+    beginQuest(index = 0, initial?: number, force = false) {
+        if (index == 0 && this.optional && !force) {
+            this.state(QuestLineState.onBulletinBoard);
+            return;
+        }
```
RedSparr0w, line 72: I'm a little confused as to what this does that it wasn't doing before?

RedSparr0w (changes requested):
Made some minor changes to the look of the quests within the board:
image

Let me know if it's okay, otherwise we can revert them.

Could we also make it so quests stay on the board once accepted, but not yet completed?
Could we also have something shown when there are no current quests available:
image

src/scripts/quests/QuestLine.ts
Jaaslet: Sets the state to onBulletinBoard when started. Does that answer your question?

RedSparr0w: I feel like we should just be using requirements for that, rather than updating it's state?

RedSparr0w (approved):

PR #2166, opened 2022-05-14
(feat): Electron'ize pokeclicker

alaborderie:
Just a quick PR to add an Electron build.

Feel free to update it as required, I just setup the minimal configuration to be able to build a Linux + Windows electron app as non-dev friends requested it.

You can also close it if you don't want to add an electron build. A few friends requested it so I think some other people might want a standalone app too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules/
 .idea/
 build/
+dist/
 *.db
 config.js
 src/declarations/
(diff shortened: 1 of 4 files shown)


Aegyo:
An electron wrapper is being maintained on another repo, with auto updating and discord integration, though it only has a windows build being generated so far:

https://github.com/RedSparr0w/Pokeclicker-desktop/releases/latest

alaborderie:
Oh! I had no idea, thanks for the info! :)

PR #2169, opened 2022-05-15
Remove Oricorio from route 6

Ultima1990:
All the Oricorio forms live in their own meadow, one on each island. All of them, except Oricorio Pa'u, which just lives on route 6. This is true in the main series too. An Oricorio just living on a random route is lame, but luckily this makes Pa'u a prime candidate to become farm exclusive. Everybody loves farm exclusives!

Featuring the fourth (and possibly final) appearance of Actor Meredith.


RedSparr0w (approved):

PR #2173, opened 2022-05-16
Relic Passage and Volcarona change

Ultima1990:
In the main series Relic Passage connects Castelia City, Driftveil City and Relic Castle. The access point in Castelia City, in the Castelia Sewers, is actually the least interesting one. You can only go to a very small area. The trip from Driftveil to Castelia via the passage is one-way, because of ledges. Similarly the access point to Relic Castle is only accessible if you enter from Driftveil. In fact, this entrance into Relic Castle from the Passage is the only way to get into Volcarona's room. You can't get anywhere interesting if you enter from the Castelia side.

And yet, in our game, Relic Passage unlocks after clearing Castelia Sewers and then doesn't actually do anything interesting. Time for change.

Relic Passage now unlocks after clearing route 5 and owning the Jet badge. Same requirement as Driftveil City itself. Additionally, Volcarona in Relic Castle now requires clearing Relic Passage at least once. Balance of Relic Castle, Lostlorn Forest and Relic Passage has been shuffled around to reflect this change in dungeon order.

Additionally I have given Relic Passage a new look on the map, to properly reflect what it is: an underground passage that goes under the river, instead of just a rectangle sitting on top of the water.

Originally a part of my rebalance draft, but I decided that it deserves its own PR.


RedSparr0w (approved):
Thanks for the PR! 👍

PR #2177, opened 2022-05-16
Achievement tracker is now bound to Holo Caster

Jaaslet:
New players did not notice that achievement tracker showed up, because it was at a very random time. To make this clear, i have bound it to a key item.
We (discord chat) came to the conclusion that Holo Caster was the best key item for the job, so i have used that.

image


src/modules/keyItems/KeyItems.ts
```diff
@@ -81,10 +81,11 @@ export default class KeyItems implements Feature {
             new KeyItem(KeyItemType.Super_rod, 'The best fishing rod for catching wild water Pokémon',
                 () => App.game.statistics.routeKills[Region.kanto][12]() >= ROUTE_KILLS_NEEDED, undefined, undefined, 'Super Rod'),
             // TODO obtain somewhere at the start
```
RedSparr0w, line 83:
```suggestion
```

RedSparr0w (approved):

PR #2181, opened 2022-05-16
Adding Checkpoints to BF

Symi001:
At the moment, checkpoints are the last stage you've beaten, no rounding to 10 or 100 (to prevent exploit at least for the 100).
Tests: - Tested going in & out of BF repeatedly with either method
- Tested end of BF run.


Jaaslet (changes requested):
- It works great! I have tested finished run, leaving a run midway and reloading the page in the middle of a run.
- The leave warning is a bit dramatic, now that leaving is not that bad. I don't know if we still want it (maybe ask on Discord), but it at least needs to reflect that the player don't just lose everything
- I don't think "checkpoint" is the correct word anymore, now that you return where you left and not at some... checkpoint, i guess?
You can maybe call the button "Continue from stage N" and call it something like "lastCurrentStage" in the code?

PR #2186, opened 2022-05-18
Add overworld Oak to remind the player of dex requirement

Ultima1990:
Adds an overworld Professor Oak sprite on the map next to Pallet Town as an extension of Pallet Town to attract the player's attention to the existing Professor Oak NPC that will tell them about the dex requirement to move on to the next region.

This Professor Oak sprite will only exist when Blue is defeated but the player is currently unable to travel, and will not reappear again after Johto travel is unlocked. I'll assume this is a lesson that needs to be learned only once and the player doesn't need to be reminded of the dex requirement in every region.

image


silverwolf2000:
Some people will want to click on it.
Maybe we need something in this case?
A tooltip? A conversation?

Ultima1990:
> 

> Adds an overworld Professor Oak sprite on the map next to Pallet Town **_as an extension of Pallet Town_**

This means that clicking Oak on the map will open up Pallet Town.

> to attract the player's attention to the existing Professor Oak NPC

This means that there already is a conversation.

Jaaslet (approved):
Looks good.
The features works as described in the PR and the code looks fine

PR #2187, opened 2022-05-18
Use player settings for the Safari movement controls

yannhodiesne:
Use the player settings to control the character movements inside the Safari zone
This change allows players on other keyboard layouts to move using their configured keys instead of WASD


RedSparr0w (approved):

PR #2200, opened 2022-05-20
Pay to start events early

### LINKED ISSUE #1421: [SUGGESTION] Event pokemons without reliance to calendar
**Current Behavior**
Event pokemons only appear when event is hosted per calendar time

**Improved Behavior**
Adding event master NPC to later regions that could trigger any of existing events on for 24hours, at cost of QP or some other currency; with hefty price tag as this is meant for completionists as way to fill pokedex without idly waiting months 

**Reason for change**
Current event pokemons have nasty thing on them: miss relevant time and you have to wait for one year to fill your pokedex

Jaaslet:
The idea is that you should be able to start an event early. This will cost more, the further away the event is.
The prices should reflect that it's a feature for the late game people, who have completed every thing else, but missed an event/completed everything in less than a year.
Closes #1421

- [ ] Wait for trades to be more flexible (we want the event calendar to cost a bit of every currency). I have promised to make this PR one day
- [x] Update the event calendar key item to reflect the new usage
- [x] Rewrite how we handle events. I want event pokemons to be a default part of our code, but with a "isEventActive"-requirement. Inject pokemons into other code is harder to handle
- [x] Update event calendar menu item, to reflect which events are ongoing and how much time is left
- [x] Make sure events can be started from the menu item. Current price is 500 QP/day remaining to the start of the event. Should last for 12 hours
- [ ] New events should not be able to be activated early
- [ ] Make sure all events are readded to the new code
- [ ] Find a way to do notifications
- [ ] Let crew test it!


Jaaslet:
This PR ended up too big and complicated. I will try again, in smaller PRs, starting with #4150

PR #2202, opened 2022-05-20
Fixed wrong name of given shiny

Symi001:
Changed the pokemon name so that it is the one from the rounded ID and not the base (with alternate forms) ID.
Tests: code works fine on two different occasions, message was the correct one.


RedSparr0w (approved):

Issue #2143, opened 2022-05-10
[SUGGESTION] Small Dungeon Improvements


locodude0001:
**Current Behavior**

Issue 1: Dungeons begin the timer as soon as you pay for it.

Issue 2: Dungeon map is purely for navigation.
**Improved Behavior**

Issue 1: I'd like to see the timer only start after the first click in the dungeon. 

Issue 2: I'd like for clicks in the dungeon space to count as clicks against wild Pokemon.

**Reason for change**

Issue 1: The dungeons have been getting larger, meaning they've started taking up more screen space. The dungeon space doesn't appear until after I've paid for the dungeon. I play on a laptop, and usually the bottom of the dungeon is offscreen so I need to scroll down and find the perfect place where I can see the dungeon and also click the Pokemon/Chest/Start the boss. I lose 4 seconds because the dungeon is bigger, so I'd like for dungeon timers to not begin until after you've made the first move. A little prep time, especially since I can't see how big the dungeon is until it's already started.

Issue 2: I'm not so worried about this to be honest, I'd much prefer the first thing get added. This is really just a bonus thought. It'd just be nice to not have to whip the mouse up and down so much trying to defeat Pokemon. 
**Detailed Description**

Issue 1: Change is simple. Dungeon timer doesn't start until after you've made the first move in the dungeon.

Issue 2: May be a little harder to implement, again I don't care so much for this. Maybe make it so that clicks in the same dungeon tile your character is in counts as click attacks toward the enemy Pokemon. I'd still want to click the confirm buttons for chests and bosses.

CorgiOnNeptune:
The timer could be a decent QoL change, especially for people playing on mobile devices or smaller screens. 

As for issue 2 you can use keyboard arrows to move around the dungeon and space bar to open chests/start boss. Though if you’re playing mobile I guess that doesnt help.

Aegyo:
Issue 1 was resolved in the last release, and Issue 2 I don't agree with the proposed solution. Are you able to just zoom out for these big dungeons?

Farboo171:
Part 1 resolved, part 2 too difficult for too little impact. Closing.

PR #2208, opened 2022-05-21
Wait until player moves to start the dungeon timer

zarunet:
As suggested by locodude0001 in #2143.

Turned out that `DungeonMap` already tracked this, which saved me half of the work :p

Also sets `DungeonRunner.timeLeftPercentage` to 100 in the initializer, so that the timer bar is properly filled when the dungeon begins. This was not a problem before because the timer ticked instantly and refreshed the value.

Test clip:

https://user-images.githubusercontent.com/43345608/169644564-0215160b-e316-40a9-bab1-f9b509b2e4c0.mp4


RedSparr0w (approved):

PR #2223, opened 2022-05-23
Caught/shiny in pokedex is now a select

### LINKED ISSUE #2199: [BUG] Pokedex auto enabling caught and uncaught filters
**Info**
 - Seems to be mobile and desktop, possibly any browser

**Description:**

Multiple reports in discord of pokedex showing nothing, turns out both caught and uncaught filters are enabled and user doesn't have any idea how it happened 

**Steps to Reproduce:**

Unknown, please comment if you have a consistent way to make this happen

**Expected Behavior:**

These settings shouldn't turn on unless the user does it manually, but also we probably shouldn't allow mutually exclusive settings enabled at the same time (because then no pokemon are shown)
Caught/uncaught could be a single setting with three possible States (the third being "show both")

Jaaslet:
Fixes #2199 (will now always default to "all")

image


RedSparr0w:
This looks good, although I was thinking maybe it should just be a dropdown?:
```
All
Uncaught
Caught
Caught Not Shiny
Caught Shiny
```
As even this way people can still end up with impossible options such as `uncaught/shiny` or `uncaught/not shiny`

Jaaslet:
> This looks good, although I was thinking maybe it should just be a dropdown?

It is now a dropdown (updated picture in the description)

RedSparr0w (approved):

PR #2231, opened 2022-05-23
#2040 - Allow achievement sorting

### LINKED ISSUE #2040: [SUGGESTION] Achievement sort options
I want to sort achievements by percentage completed, for example to filter to Route Kill and check where I'm close to getting the next achievement, or similair for Gym and Dungeon Clear.

**Current Behavior**

No sorting is available, only filtering

**Improved Behavior**

Choose to sort as well, by percentage left or done, or by amount of bonus you can recieve, or by absolute amount to next level (so 9/10 can be shown as the top one, instead of 910/1000 appearing earlier).

**Reason for change**

This allows players to easily check what routes, dungeons or gyms should be cleared to improve the bonus percentage

**Detailed Description**

**Screenshots:**

MatthieuMesnager:
* Added the ability to sort achievements by specific properties. 
* Modified UI according to recent changes about achievement sorting.

Closes #2040 

Currently I've only added 2 options for sorting but we can easily add more if we want to.


src/modules/achievements/AchievementSortOptions.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    // true if the default sort direction should be descending
+    invert?: boolean;
+};
+
+export const AchievementSortOptionConfigs: Record<AchievementSortOptions, AchievementSortOptionConfig> = {
+    [AchievementSortOptions.default]: {
+        text: 'Default',
+        getValue: (a) => a.bonusWeight,
+    },
```
RedSparr0w, line 21: This doesn't seem to be our default sorting option.

Could you change this to be how they are currently sorted and add Bonus % as its own option.

RedSparr0w (approved):

PR #2232, opened 2022-05-24
Johto + Sinnoh dungeon updates

zyx-xyzzy:
- Gives the special Unown A encounter at Ruins of Alph the same HP/level as the usual Unown bosses there rather than a copy of Celebi's, so it isn't a random difficulty spike.

- Generic multibattle trainers in Sinnoh (found mainly in Stark Mountain and Iron Island) no longer get to have 5- or 6-mon teams, because that's a level of roadblock that should be rare—outside of bosses or other particularly notable foes, at least—rather than getting spammed across a dungeon. 
Relatedly, Stark Mountain's Mars and Jupiter have been combined into a single trainer à la their Spear Pillar encounter, because having them be objectively weaker than the majority of the regular trainers there was just sad.


Ultima1990:
I disagree with reducing the number of mons on the double battle teams. Just nerf their health or something.
I also disagree with combining Mars and Jupiter into a double battle. Just buff their health or something.

Fine with the Unown change.

zyx-xyzzy:
> 

Changing the team size affects their effective strength at all stages of gameplay, whereas changing the HP only matters if the player either has a weak enough boosted-Pokémon Attack to not one-shot (i.e., pretty much only while mid-Sinnoh) or has a strong enough Click Attack to defeat them before the first mon-attack fires (which isn't really feasible for Sinnoh trainers). Do you have any particular reason for disagreeing with this, besides just being contrary?

And in case you're about to cite it not matching the canon teams, I'd love to hear a good explanation for why changing canon is considered acceptable for as minor a reason as it being "lame" (https://github.com/pokeclicker/pokeclicker/pull/2169) but not for game-balance purposes like this!

amative1:
We're sorry, but we (the developers and contributors) do not agree with the proposed balance changes. The Unown change is fine, but everything else cannot be accepted at the current time. If you wish to see changes to these trainers, please open an issue so we can discuss it before going further.

PR #2236, opened 2022-05-24
added a setting to disable offline progress

### LINKED ISSUE #2234: [SUGGESTION] Add an option to disable offline progress
**Current Behavior**
When coming back to the game after offline time, you get money according to the time you went offline 

**Improved Behavior**
Add a way to opt out of the offline progress

**Reason for change**
I've played for 2h, gone offline for 1 week, gained money equivalent to 80000% of my current balance (2 000P -> 1 600 000P)
for now only money is awarded afaik, but I don't want to go from "need to grind some more for better pokeball" to "bought 10 000 pokeball, still rich"

**Detailed Description**
Add a checkbox in the settings to disable offline progress

**Screenshots:**
image
image

jonatjano:
Application of the suggestion I mentioned in #2234

It adds a setting in the setting tab "other" to enable/disable offline progress, this option on `true` by default meaning offline progress is enabled

image

Closes #2234


Jaaslet:
(have not tested)
Can we call it "disable" instead, to reflect what we think is the recommended/default?

jk13pclick:
> It can easily be done, should I do it or wait for input by other people ?

It should be done before this is approved, @Jaaslet's suggestion is good - do it.

Jaaslet (approved):
We don't understand why anyone would do this to them self, but it's a pretty harmless PR, so if it makes you happy, i can't see why not. Hopefully other people will find it useful too.
I will merge this. Thanks for the PR

PR #2239, opened 2022-05-25
Fix damage calculator not updating when "ignore level" checked

Aegyo:
Not sure if an issue was ever made for this, but to see the bug on development:
1. Check "ignore level" in damage calculator
2. Maybe also check the breeding one too, I didn't test without that
3. Hatch an egg
4. Don't catch any new pokemon while doing this
5. Total damage in the calculator doesn't increase (but should have because we hatched something)


RedSparr0w (approved):

PR #2243, opened 2022-05-25
fix links on the save selector component

LoickMunoz:
Clicking on the links on the save selector screen does not open a new tab. This fixes it.


RedSparr0w (approved):

PR #2256, opened 2022-05-26
Add weather selector in damage calculator

Aegyo:
Adds weather selector + some html changes to make the dropdowns break onto two lines before there isn't enough space.
Screenshot_2022-05-26_05-01-41
Screenshot_2022-05-26_05-02-02


RedSparr0w (approved):

PR #2298, opened 2022-05-28
Hide type bars in damage calculator with 0 damage

Aegyo:
When you have no damage in some type, there is a bigger gap in the damage calculator type bars because of this 0-width bar still existing with some margin.

Before:
Screenshot_2022-05-28_07-23-54

After:
Screenshot_2022-05-28_07-26-13


RedSparr0w (approved):
Much nicer, thanks for the PR! 👍

PR #2303, opened 2022-05-28
#1904 - Allow berry farm locked plots

### LINKED ISSUE #1904: [SUGGESTION] Locked Plots on the farm
I'd like to have the option, to lock certain spaces in the garden. This "locked" section should be manually applied/removed from a section. 

**Current Behavior**
No locked spaces / plots

**Improved Behavior**
Certain spaces in the garden, which cannot accidentally be removed (harvest all / misclick..)

**Reason for change**
Simply a QoL Improvement, specially for the late-game berries, which require days to grow. It is a pain to accidentally harvest them and then to replant them again, waiting (worst case) for another day. A common use for example is the petaya-berry: If you have a field and want to remove everything *except* the berry, there is no option. 

**Detailed Description**
I would love to see an option (maybe shift + left-click) to lock a field, so it isn't interacted with (no new mulch, no shovel, no harvest all and no manual harvest - until the field is unlocked again (possibly via shift + left-click again). A locked tile could be indicated by a small lock-sign or simply graying it out for a more visual approach.

**Screenshots:**
None

Might also be kind of relatable to #1902

MatthieuMesnager:
Closes #1904 

Allow players to 'safe-lock' a specific plot by using `shift + left-click` in order to prevent shovel, harvest, and new mulch on this plot.

* **Added**: new `_isSafeLocked` property on the `Plot.js` class.
* **Modified**: `FarmController.ts` to be able to handle clicks & shift + clicked actions.
* **Modified**: `Farming.ts` methods (`shovel()`, `harvest()` & `canMulch()`) now take the new plot `_isSafeLocked` property into account.
* **Added**: new lock-sign icon on `farmModal.html` to notify players whether the plot is 'safe-locked' or not.

I was not sure of how to call the new property on `Plot.ts` because there is already a property named `_isUnlocked` so I thought that it would have been a bit weird to have `_isUnlocked` and `_isLocked` properties since they are not related.

What do you think ?

Another point, I've added a simple lock-sign svg icon from [Heroicons](https://heroicons.com/) but maybe it would be better to add a custom 'pokemon style' lock-sign `.png` ?


Aegyo:
Hi, it looks like you have included work from a different PR and a bunch of other things accidentally. 

Could you instead create a fresh branch from our `develop`, and only add the changes needed for the locked plots?

<hr />

I don't have any problem with the _isSafeLocked name, but if you are still looking for something else, maybe _isProtected

Aegyo:
Actually it doesn't look like there are other actual code changes included, but there are a bunch of other commits mentioned here? Maybe the rebase you did has reordered events and git is confused...

MatthieuMesnager:
Yeah, it looks like there has been an issue with my repository 😅 I'll clean up and create a new pull request correcly.

Well, if you are okay with the _isSafeLocked, I'll stick up with that then.

PR #2305, opened 2022-05-28
change(farm): Allow berry farm locked plots

### LINKED ISSUE #1904: [SUGGESTION] Locked Plots on the farm

MatthieuMesnager:
Closes #1904 

Allow players to 'safe-lock' a specific plot by using shift + left-click in order to prevent shovel, harvest, and new mulch on this plot.

* **Added**: new `_isSafeLocked `property on the `Plot.js` class.
* **Modified**: `FarmController.ts` to be able to handle clicks & shift + clicked actions.
* **Modified**: `Farming.ts` methods (`shovel()`, `harvest()` & `canMulch()`) now take the new plot `_isSafeLocked` property into account.
* **Added**: new lock-sign icon on `farmModal.html` to notify players whether the plot is 'safe-locked' or not.
* **Added**: new style for the lock-sign icon in` farm.less`

Cleaned-up pull request from #2303


src/components/farmModal.html
```diff
@@ -151,6 +151,11 @@
                                         <div class="plotImage">
                                             <img style="width: 100%" src="assets/images/farm/soil.png"
                                                 data-bind="class: FarmController.calcMulchClass($data)">
+                                            <!-- ko if: $data.isSafeLocked -->
+                                            <svg class="plotLockSignIcon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="#fff">
+                                                <path fill-rule="evenodd" d="M5 9V7a5 5 0 0110 0v2a2 2 0 012 2v5a2 2 0 01-2 2H5a2 2 0 01-2-2v-5a2 2 0 012-2zm8-2v2H7V7a3 3 0 016 0z" clip-rule="evenodd" />
+                                            </svg>
```
Jaaslet, line 157: All our other SVGs are stored in a file. Can we do the same for the lock?

src/scripts/farming/FarmController.ts
```diff
@@ -80,25 +80,42 @@ class FarmController {
...
         const plot: Plot = App.game.farming.plotList[index];
+
+        if (event.shiftKey) {
+            this.handleShiftClickActions(plot, index);
+        } else {
+            this.handleClickActions(plot, index);
+        }
+    }
+
+    private static handleShiftClickActions(plot: Plot, index: number) {
```
Jaaslet, line 93: I think it would be better if the method name described what it did, instead of when it should be used.

src/scripts/farming/Plot.ts
```diff
@@ -9,6 +9,7 @@ class Plot implements Saveable {
     };
 
     _isUnlocked: KnockoutObservable<boolean>;
+    _isSafeLocked: KnockoutObservable<boolean>;
```
Jaaslet, line 12: We should save this value, by adding it to the toJson-method

Jaaslet (changes requested):
It looks good and works! But I still have some comments:
Right now it's a hidden feature. Can we add it either to the help-tab or to the hotkeys tab in settings?

Jaaslet:
Closing this, as it was finished in PR #2799

PR #2319, opened 2022-05-29
Allow dungeons to have content

Qwertypop04:
This should allow dungeonstowns to have content, similar to normal towns, such as shops, gyms and temporary battles.
The intention is to be used for temporary battles, due to some rival fights taking place in dungeons, but could allow for other things in future as well


src/scripts/towns/Town.ts
```diff
@@ -57,8 +57,8 @@ class Town {
 class DungeonTown extends Town {
     dungeon: Dungeon
 
-    constructor(name: string, region: GameConstants.Region, requirements: (Requirement | OneFromManyRequirement)[] = []) {
-        super(name, region, [], { requirements });
+    constructor(name: string, region: GameConstants.Region, content:  TownContent[] = [], requirements: (Requirement | OneFromManyRequirement)[] = []) {
```
Jaaslet, line 60: If you move the content-parameter to the end of the list, you won't need all the changes in TownList.ts

Qwertypop04: Ah, thanks, that makes it a lot cleaner

Jaaslet (approved):
It works. Thanks for the PR

PR #2324, opened 2022-05-29
New NPCs

Ultima1990:
Two new Kanto tutorial NPCs. One in Pewter to explain gems. Pewter because it has the museum. Makes sense to me. One in Saffron to explain breeding, because that's close to the Daycare on route 5.

Both are locked behind the Earth badge, which is when the player gains access to gems and also their Pokémon can reach level 100 and so breeding is unlocked. It makes no sense to me to have NPCs that explain things that the player does not currently have access to, hence locking them. Yes, this might mean players will not find them when they need them, I intend to have tutorial quests later on that will lead the player to go and talk to these NPCs, but that tech is currently locked inside another PR.
The only other improvement I can think of for these guys is maybe having the gems NPC not show up if the player has the no gems challenge active.

The third and final NPC is Riley in Canalave to explain the Riolu evolution. Not a lot of text there. He's a man of few words I guess.

### DIFF
--- a/src/scripts/towns/TownList.ts
+++ b/src/scripts/towns/TownList.ts
@@ -468,6 +468,26 @@ const AzaleaOldMan = new NPC('Wise Old Man', [
     'It says that the mythical Pokémon Celebi will appear before anyone who has proven they are a Champion Pokémon Trainer.',
 ]);
 
+const EcruteakBill = new NPC('Bill', [
+    'I traveled here all the way from Goldenrod to buy a Soothe Bell for my Eevee.',
+    'When I use a Soothe Bell on Eevee it can evolve into Espeon or Umbreon. It depends on the time. Espeon during the day, Umbreon at night.',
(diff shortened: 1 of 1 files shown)


Jaaslet (changes requested):
On Discord we agreed that it would make sense if 'Kimono Girl' shows up after beating Burning Tower.
Other than that, it looks great. Hopefully it will remove a bit of the spam in the bugs channel

Jaaslet (approved):
You changed it, before i left my review.
It works, so i approve the PR

Jaaslet (changes requested):
More code added after i approved? :O
Then i need to review again!
The two new NPCs looks great, but we might want to move both of them up the list. They both seams more important than the other NPCs in the town they are in, so i feel like they should be the top NPC

Jaaslet (approved):
We don't want to change the order of the NPCs anyway, so i approve this PR

PR #2328, opened 2022-05-30
WIP: Challenge starter

DNA666777:
Design in progress
Challenge Modal : 
image
Prof Oak : 
image
After Selecting :
image


Jaaslet:
I will close this PR.
It looks like it's not going to be finished, and even if it would be finished, we are not going to merged it.
Monotype will be hard to maintain. We need to make sure all types are playable at all times.
Custom starter is most of the time a way to make the game easier, and not just a fun challenge.

PR #2336, opened 2022-05-30
Adding Lapras to Union Cave on Fridays

### LINKED ISSUE #2301: [SUGGESTION] Add Lapras to Union Cave
**Current Behavior**
There is no Lapras in Union Cave.

**Improved Behavior**
There is now a Lapras in Union Cave.

**Reason for change**
There is a Lapras in Union Cave on fridays in the original games.

**Detailed Description**
Add Lapras as a boss encounter on fridays after earning the Fog Badge.

Nydaleclya:
Adding a new Requirment for Weekdays.
Adding Lapras to Union Cave on Fridays, locked behind the Fog badge as in the games.

Closes #2301


Ultima1990:
I'm lovin' it. No clue about the implementation, someone else will have to weigh in on that, but I very much like the idea.

CorgiOnNeptune:
I'd suggest maybe changing "WeekdayRequirement" to "DayOfWeekRequirement". Since a 'weekday' typically refers to just Monday->Friday. But otherwise looks good to me! There are a few other canon appearances and temporary battles that could possibly implement this requirement as well down the line.

Jaaslet:
(Have not tested yet)
We need an NPC to explain how it works. Else we are going to get a ton of questions as soon as it's released

Aegyo:
Could you make the requirement take a DayOfWeek instead of a generic function? As it is now, it is looks like a CustomRequirement with opinionated variable names.

Also it doesn't look like it will update when the day actually changes, to do this you want a ko.observable somewhere (maybe a static property of DayOfWeekRequirement for now?), initialised with the current current day (or weekday number), and you can update it in the same place as where we give a new quest refresh and generate new daily deals (I think this is in App or Game). Then your "what day is it?" function should use that observable instead of a new Date

Ultima1990:
I don't think we'll need an NPC if the Lapras is hidden when it's unavailable. Then it'll just be a fun little easter egg.

Jaaslet (changes requested):
I did a full review!
Overall, it looks cool and works great. Thanks!
I have some small changes:

PR #2337, opened 2022-05-30
Unown (A) Health Fix

Nydaleclya:
The static Unown (A) had much more health than the daily ones.
Now they are the same.


Jaaslet:
This is already a part of PR  #2232. I will close this PR.
But thanks :)

PR #2338, opened 2022-05-31
Add filtering to protein menu

### LINKED ISSUE #1300: [SUGGESTION] Add filter to protein usage menu
**Current Behavior**
The game currently only has a sort dropdown. 

**Improved Behavior**
Ideally the protein usage menu would have both the sort and filter functionalities, similar to the hatchery. 

**Reason for change**
This will make it much easier to apply proteins to much more specific groups of pokemon - e.g. non-shinies of the newest region. 

**Detailed Description**
Expand the protein usage menu to have both a sort and a filter functionality. They could be minimizable, like in the hatchery menu. The filter menu could also be a pop-out, which would be inconsistent with the hatchery, but still okay. 

**Screenshots:**

shallotmama:
Does what the title says :)

I generalized the filtering a bit based on what already existed in the breeding code, then updated the breeding filtering to use it and added it to the protein page.

Closes #1300


Jaaslet (changes requested):
(I have not given i a full review yet)
It looks good. But if we want more stuff in that window, we need to make it bigger.
The dropdowns are too small, and the pokemon list is getting smaller.
If we can make the window wider and move some of the stuff up to the same line, i think it would look a lot better.

Jaaslet (changes requested):
It looks better now.
Any chance you know how to remove the vertical scrollbar? It takes up a row of the grid.
I did not find any correct way of doing it, but if you can't either, we will just throw a `overflow-x: hidden;` on the element.
I think the problem is that the search-row somehow is too wide

shallotmama:
I assumed you meant the horizontal bar. Turns out the `.form-row` class has a negative margin on the horizontals for some reason, and that was poking it outside the bounds of the parent. Setting the margins to 0 fixed it.

Jaaslet (approved):
> I assumed you meant the horizontal bar.

Yeah i did.
It looks great now! Thanks for the PR

PR #2341, opened 2022-05-31
Being able to sort and filter pokedex like hatchery

Rohydre:
Suggested on discord, I decided to work on it.

Problems comes when dealing with uncaught/unseen pokemon. Dealt with it by sorting them at the end when value can't be found


Jaaslet:
Heads up. The build check is failing, so it won't be reviewed, before the check succeeds

Jaaslet:
I will close this PR as it seems dead. The building never succeeded, and it has many conflicts now.
Feel free to reopen it, if you want to finish it

Issue #934, opened 2020-10-13
[BUG] Labeling/Spelling/General wording issues


HLXII:
The usage of Pokemon vs Pok**é**mon is not consistent. 

Some example locations:
1. The new Pokeball selector modal (Main header and description)
2. Hatchery Tab name.
3. Sorting method name ("Pokemon #")
4. Underground Help section
5. Pokeball shop entry.
6. Damage calculator modal
7. Statistics Modal entries
8. Couple of towns/Dungeons (Pokemon Tower, Pokemon Mansion, the Leagues).

Not sure if anyone wants to go through the effort to change all of these (I'm not sure if I recorded all of them). In addition, changing the dungeon names would also mess with how we store the save data, which doesn't sound fun.

We also use Bruno2 for the name of Bruno in Johto. I saw at least one person confused about that. 

In general, the item names are a bit off.
1. Most item names have their second word not capitalized. This is due to how we obtain the name by just replacing underscores with spaces.
2. The Restores don't even have spaces.

Cleaned up List:
- [x] Pokemon/Pokeball references
- [x] Item names
- [x] Updating names for Towns (Pokemon Tower, Pokemon Mansion, Pokemon Day Care, Pokemon Leagues)
- [ ] Updating certain trainer battles (Bruno2)

PR #2345, opened 2022-05-31
No more numbers

jk13pclick:
Removes numbers from Gym Leaders with multiple instances on the front-end. Back-end stays the same. Hopefully I caught everything - achievements also seem to update correctly.

Closes a task on #934


Qwertypop04:
Would it be possible for this also to be applied to temporary battles? Rivals are terrible for appearing again and again and again...

PR #2348, opened 2022-06-01
Auto update profile card in game with new badge numbers, etc

### LINKED ISSUE #2347: [BUG] Menu -> Profile not updating while in-game, needs reload
**Info**
 - Client

**Description:**
The information showed in the trainer card in menu -> profile does not get updated as one progresses through the game. To make it show the current information one needs to reload the game.

**Steps to Reproduce:**
1. Open save.
2. Go catch a unique mon or earn a new badge.
3. Open menu -> profile.
4. Compare and realize it's not updated.

**Expected Behavior:**
Have the trainer card show the updated information without having to reload the game.

**Screenshots:**
PB7-H
That badge number should be 33 since I beat Tate & Liza's Gym yet it remains at 32, the value it had when I opened the file. If I were to reload it would show the updated number.
badges
(Second image is just to show my badge count).
Also if I were to reload the file I'm aware the main screen will also show me the right count.

Aegyo:
Fixes #2347

<hr>

Throttled the time played to one update per minute because we don't really need this updating once per second just for that. Didn't seem like there was an easy option to have the observables attacked to just their bit of the html, without changing a bunch of stuff affecting the save selector screen anyway.


Jaaslet (approved):
The code looks good, and the trainer card now gets updated when it should.
Thanks for the PR!

PR #2360, opened 2022-06-02
Fix typo in proteins

Rohydre:
Just a minor thing that have basically no impact except when we use our IDE search tool


Jaaslet (approved):
Only thing i found is that if people have the protein display set, it will reset when the game updates.
But it's a too small thing, that i think it's worth creating an update function for.
Approved! Thanks for the PR.

PR #2361, opened 2022-06-02
Save data can now be copied into the clipboard

Jaaslet:
A workaround for Apple not allowing people to download our saves on phones.
Hope it works. I don't have an Apple phone it test it with, but it works in Firefox.
image


Jaaslet:
jk13 has tested this workaround and successfully gotten a save from Chrome on iphone, to chrome on Mac.

RedSparr0w (approved):

PR #2366, opened 2022-06-03
Fixed missing rename in a requirement

### LINKED ISSUE #2384: [BUG] Cannot unlock Pryce's Gym in Mahogany Town
**Info**
 - Version: commit 32c252bb3b372c9654048982c18cd0b276222a68
 - Browser: Firefox

**Description:**
Price's Gym does not unlock event after completing everything else up to this point in Johto.
When clicking on it, the alert message says "You don't have access yet. undefined needs to be completed."

**Expected Behavior:**
The gym should unlock after completing Team Rocket's Hideout, but it is not the case.

**Additional Information:**
It seems the problem comes from #2242 which changed the dungeon name "Team Rockets Hideout" to "Team Rocket's Hideout" without updating the requirement in [src/scripts/gym/GymList.ts](https://github.com/pokeclicker/pokeclicker/blob/32c252bb3b372c9654048982c18cd0b276222a68/src/scripts/gym/GymList.ts#L311).

Jaaslet:
The 7th badge was broken after the grammer-fixes. This PR fixes that problem.
Also fixes #2384


RedSparr0w (approved):
👍 Thanks for the PR!

PR #2368, opened 2022-06-03
Buff Diglett's Cave

Ultima1990:
Diglett's Cave is still balanced like a route 2 dungeon. Both in HP and cost. It should be harder than Mt. Moon but not as hard as Rock Tunnel.


Symi001:
I guess there won't be a way of putting it in between in terms of road ? Meaning it'll yield way less exp ?

Ultima1990:
The dungeon is now marked as route 11 rather than route 2. If that has any effect on dungeon exp then Diglett's Cave exp gain also got buffed. Tho TBH I have no idea how that works.

Symi001:
If it's marked as route 11 it'll also make the exp change, perfect, sry for the bother ^^

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2375, opened 2022-06-03
added slow pulse to locations with quest

shallotmama:
I find it EXTREMELY difficult to see quest objective locations with the default color, especially when it's a cave on some grass. For example, find Chargestone Cave here (it's on the left half of the pic).

image

This adds the option to add a slow flash to map objectives that have an active quest.


Jaaslet:
Thanks for the PR, but the develop team don't think it fits with our vision for the game.
We rather want to fix the problem by changing the default map-colors.
If you think your are up for the task of changing the default colors, feel free to submit a PR. To be honest, i am hoping for someone with some understanding of colors to solve it, which is why I have not fix it myself yet.
I will close this PR.

PR #2380, opened 2022-06-05
Added CaptureSpecificPokemonQuest

Jaaslet:
Including a "ignore breeding"-parameter.
Needed for UB-PR, maybe other things too. Will use it one day to clean up the Vivillon quest line


src/scripts/quests/questTypes/CaptureSpecificPokemonQuest.ts
```diff
@@ -0,0 +1,29 @@
...
+        if (typeof reward != 'number') {
+            this.customReward = reward;
+        }
+        this.pokemon = pokemonMap[pokemonName];
+        this.customDescription = description;
+        this.focus = ko.pureComputed(() => App.game.statistics.pokemonCaptured[this.pokemon.id]() - (includeBreeding ? 0 : App.game.statistics.pokemonHatched[this.pokemon.id]()));
+    }
+
+    get description(): string {
+        return this.customDescription ?? `Capture ${this.pokemon.name} Pokémon.`;
```
RedSparr0w, line 20: Could we also make use of the `capturesNeeded` variable here (in case it is more than 1)?

RedSparr0w (changes requested):
PR looks good,
Could you please also update the questlines where we require capturing a specific Pokémon to make use of this too?

Jaaslet:
Should have fixed both of your comments now. The Vivillon hunt was the only one i could find, which could use the new quest.
I also removed 'Pokemon' from the description, because it seemed redundant

Aegyo (approved):

PR #2383, opened 2022-06-05
Fix missing achievements

### LINKED ISSUE #2280: [BUG] Achievement Modal showing some completed achievements as incomplete
**Info**
 - Windows 10
 - Chrome
 - 9.5

**Description:**
4 achiviements that reach 100% completion rate still apears as no completed:

**Steps to Reproduce:**
I don't sure how this happened, i'm pretty sure that when the game updates to 9.4 works fine, but at some point in that version they became incomplete

**Screenshots:**
image

Jaaslet:
Closes #2280
So what i think went wrong is that observables don't work well, while inside a lambda-function. So our filter-function never got triggered.
So i have made a observable for how many oakitems are max level and how many plots are unlocked. And after using those for the achievement, it works as it should.
It's an ugly solution, but i think it's correct.


src/scripts/achievements/AchievementHandler.ts
```diff
@@ -213,7 +213,6 @@ class AchievementHandler {
         AchievementHandler.addAchievement('Is that how I use this?', 'Level 1 Oak Item to the maximum level', new MaxLevelOakItemRequirement(1), 0.05, GameConstants.Region.none, () => !challenges.list.disableOakItems.active());
         AchievementHandler.addAchievement('I\'ve got my hands full with all of these', 'Level 3 Oak Items to the maximum level', new MaxLevelOakItemRequirement(3), 0.1, GameConstants.Region.none, () => !challenges.list.disableOakItems.active());
         AchievementHandler.addAchievement('Professor Oak is the best!', 'Level 8 Oak Items to the maximum level', new MaxLevelOakItemRequirement(8), 0.14, GameConstants.Region.none, () => !challenges.list.disableOakItems.active());
-        AchievementHandler.addAchievement('Prepared for anything!', 'Level 11 Oak Items to the maximum level', new MaxLevelOakItemRequirement(11), 0.18, GameConstants.Region.none, () => !challenges.list.disableOakItems.active() && challenges.list.disableClickAttack.active());
```
RedSparr0w, line 216: This achievement is required for the no click challenge.

Jaaslet: Now i understand! I could not figure out why we had that achievement...
But i have readded it now

RedSparr0w (approved):

PR #2387, opened 2022-06-05
Re: Re: Relic Passage

Ultima1990:
After numerous complaints about people unable to find Relic Passage I've decided to redesign the Relic Passage redesign. Instead of placing it under the river I've now given it the Diglett's Cave treatment. One entrance on one side, one on the other side.

Coincidentally this also makes room in the river for... other things. How convenient for me!


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2388, opened 2022-06-05
New Weather Institute map design

Ultima1990:
New map sprite creation by NiCeDiCe.

The new Weather Institute sprite can't be made smaller without losing too much detail, so something had to give. So yeah. Instead of carving out a bit of the mountain I decided to move something else instead. All of 119, Fortree City, 120 and the farm have been moved over by one tile. 118 is now one tile longer, 121 and 123 are one tile shorter.


Jaaslet (approved):
Looks much better now. No more dungeon-looking building!
Thanks for the PR

PR #2391, opened 2022-06-05
fix(dungeon): Unown A hp

amative1:
#2337 closed and deleted, recreated to fix Unown A stats.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2392, opened 2022-06-06
Plot.harvestAmount() Fixup

### LINKED ISSUE #2382: [BUG] Discrepancy between number of harvested berries and number counted towards quest
**Info**
* OS: Mac OS 10.15
* Browser: Firefox
* Version: v0.9.5

**Description:**

When using a combination of effects to increase berry harvest, it's possible that a quest to "harvest X berries" counts one more berry than actually harvested.
In my case I harvested a plant for 20 berries, but got 21 counted towards the quest.

This likely happens when the harvest number after the boost is applied is above .5, the quest is rounding up but the harvest down.

**Steps to Reproduce:**

For simplicity's sake I'll just share my exact case. But I expect the same thing to happen with other berries and setups, as long as the resulting harvest boost falls in the right range.

* Have quest "Harvest X lum berries at the farm"
* Setup a square of 8 Passho berries surrounding a lum berry plant in the farm
* Once all this is ripe, since the Passho berries' effect is boosted by the Lum berry, they boost the harvest by 10.33
* Use Rich Mulch on Lum Berry
* Harvest Lum Berry
* You should receive 20 lum berries, but the quest will count 21.

**Expected Behavior:**

The number of berries counted towards the quest should be the same as the number of harvested berries

**Screenshots:**

https://user-images.githubusercontent.com/54149968/172049380-8eb822e3-62b7-4261-90b8-f69f72d09a75.mov

So sorry about the beginning of the video, the first seconds got lost to oblivion. In the first few frames it's still possible to make out that the quest sits at 21 berries but just in case here's a screenshot from when I noticed the issue.
Screenshot 2022-06-04 at 22 38 32

As shown in the video, after harvest the number is 42, while my actual number of berries only went from 16 to 36.

**Additional Information:**

I did the same setup 5 times total to work towards the quest, and each time got 21 berry added to the quest count, ending up at 105. So, this is likely not an issue of the quest storing a decimal number, as if it was the case, I would have ended at 20.66*5 = ~103 berries.
The 6th time I didn't use the mulch so with the multiplier being only 10.33, got 10 berries, and only 10 were added to the quest count, as expected.
lum-berry-quest

jk13pclick:
Closes #2382. Hopefully should track with the berriesHarvested observable correctly now.


Jaaslet (approved):
It looks correct and works as described. Thanks for the PR!

PR #2400, opened 2022-06-07
I have just freed the mega evolution.

invisiblemanofpoke:
Can you respect this please ? It's for all fan of pokemon.


NiCeDiCe90:
They are still not obtainable with your PR. The only change would be, that they appear in the Battle Frontier.

There are plans to add them but is has to be more special.

Jaaslet:
They won't event appear in the battle frontier, because the id is negative

invisiblemanofpoke:
it's just to show everyone that you are moving forward to add the mega evolutions. If people see that the developers are talking about it then they will probably be satisfied. That's how I see it, I could be wrong. particularly since it does not change the game

Qwertypop04:
We _aren't_ moving forward to add mega evolutions at the moment though. We have no good enough ideas on how to implement them

Jaaslet:
> We aren't moving forward to add mega evolutions at the moment though. We have no good enough ideas on how to implement them

Not totally correct... We know how we want to implement Megas, but it requires the Held items PR to be merged first.
And we have ideas about how to obtain *some* megas. The plan is to release those megas, when they are ready and then add the rest, when we get ideas for how to obtain them.

invisiblemanofpoke:
okay okay no problem

PR #2404, opened 2022-06-07
Add save reminder notification

RedSparr0w:
This PR adds a Save Reminder notification:
image

The time between reminders can be adjusted in the settings:
_(with 12 hours being the default)_
image

You will only be notified if you haven't download a save or been reminded with that time frame.

### DIFF
--- a/src/modules/saveReminder/SaveReminder.ts
+++ b/src/modules/saveReminder/SaveReminder.ts
@@ -28,15 +28,16 @@ export default class SaveReminder implements Saveable {
         let timeSinceSaveReminder = App.game.statistics.secondsPlayed() - Math.max(App.game.saveReminder.lastDownloaded(), App.game.saveReminder.lastReminder());
         // Adjust to measure against ms
         timeSinceSaveReminder *= 1000;
-
-        if (timeSinceSaveReminder >= +Settings.getSetting('saveReminder').value) {
+        const saveReminderInterval = +Settings.getSetting('saveReminder').value;
+        if (timeSinceSaveReminder >= saveReminderInterval) {
             // Show reminder notification
             Notifier.notify({
                 title: 'Save Reminder',
-                message: `It has been ${GameConstants.formatTimeShortWords(+Settings.getSetting('saveReminder').value)} since your last save download, Would you like to download a backup now?
+                message: `It has been ${GameConstants.formatTimeShortWords(saveReminderInterval)} since your last save download, Would you like to download a backup now?
 
                 <button class="btn btn-block btn-success" onclick="Save.download()" data-dismiss="toast">Download Save</button>`,
-                timeout: 3 * GameConstants.HOUR,
+                // Timeout either the reminder interval or 3 hours, whichever is lower
+                timeout: Math.min(saveReminderInterval, 3 * GameConstants.HOUR),
             });
             App.game.saveReminder.lastReminder(App.game.statistics.secondsPlayed());
         }
--- a/src/modules/saveReminder/SaveReminder.ts
+++ b/src/modules/saveReminder/SaveReminder.ts
@@ -1,0 +1,55 @@
+import {
+    Observable as KnockoutObservable,
+} from 'knockout';
+import { Saveable } from '../DataStore/common/Saveable';
+import * as GameConstants from '../GameConstants';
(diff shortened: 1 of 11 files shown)


src/modules/saveReminder/SaveReminder.ts
```diff
@@ -0,0 +1,60 @@
+import {
+    Observable as KnockoutObservable,
+} from 'knockout';
+import { Saveable } from '../DataStore/common/Saveable';
+import * as GameConstants from '../GameConstants';
+import Notifier from '../notifications/Notifier';
+import Settings from '../settings';
+
+export default class SaveReminder implements Saveable {
+    public static counter = 0;
```
Jaaslet, line 10: counter looks unused

src/modules/saveReminder/SaveReminder.ts
```diff
@@ -0,0 +1,60 @@
...
+        timeSinceSaveReminder *= 1000;
+
+        if (timeSinceSaveReminder >= +Settings.getSetting('saveReminder').value) {
+            // Show reminder notification
+            Notifier.notify({
+                title: 'Save Reminder',
+                message: `It has been ${GameConstants.formatTimeShortWords(+Settings.getSetting('saveReminder').value)} since your last save download, Would you like to download a backup now?
+
+                <button class="btn btn-block btn-success" onclick="Save.download()" data-dismiss="toast">Download Save</button>`,
+                timeout: 3 * GameConstants.HOUR,
```
Jaaslet, line 39: Because the lowest value is 1 hour, the player can end up with 2 popups on their screen, which some might find annoying

src/modules/saveReminder/SaveReminder.ts
```diff
@@ -0,0 +1,60 @@
...
+
+        let timeSinceSaveReminder = App.game.statistics.secondsPlayed() - Math.max(App.game.saveReminder.lastDownloaded(), App.game.saveReminder.lastReminder());
+        // Adjust to measure against ms
+        timeSinceSaveReminder *= 1000;
+
+        if (timeSinceSaveReminder >= +Settings.getSetting('saveReminder').value) {
+            // Show reminder notification
+            Notifier.notify({
+                title: 'Save Reminder',
+                message: `It has been ${GameConstants.formatTimeShortWords(+Settings.getSetting('saveReminder').value)} since your last save download, Would you like to download a backup now?
```
Jaaslet, line 36: Should we add a "This interval can be changed in the settings"? It seams like something a lot of people want to customize, but i don't think it's super intuitive that it's something you can change, so we might get some questions about it.
But i am really not sure, if i think this is a good idea or not...

Jaaslet (changes requested):
Looks good, and works great! I just have some small comments:

src/modules/saveReminder/SaveReminder.ts
RedSparr0w: The counter is incremented within `Game.ts`, which runs the `tick` function every 5 minutes.

src/modules/saveReminder/SaveReminder.ts
RedSparr0w: I think it would just add too much clutter, and don't think every bit of information needs to be spoon fed to people, because even if we did, we would still get people coming to ask about blue coin.

src/modules/saveReminder/SaveReminder.ts
RedSparr0w: Good point 👍
Changed it to be whichever is lower (the interval or 3 hours)

src/modules/saveReminder/SaveReminder.ts
Jaaslet: Yes, but we don't use the value anywhere, right?

RedSparr0w: The value is also used in `Game.ts`
```ts
        SaveReminder.counter += GameConstants.TICK_TIME;
        if (SaveReminder.counter >= 5 * GameConstants.MINUTE) {
```

Jaaslet: Missed that.
Great! Will look at the other change later today

Jaaslet (approved):
Looks good. Thanks for the PR 👍

PR #2405, opened 2022-06-08
Add currency lost/spent animation

CorgiOnNeptune:
This adds an option to display a currency lost animation. The main utilization of this is to better visualize currency spent on Helpers, and for people who struggle with reading, currency spent on auto-battling gyms. 

Tried to implement it into the existing animateCurrency function, but couldn't figure it out, so I just made a new function replicating it. Could probably be implemented in a cleaner way within UI.ts.

### DIFF
--- a/src/components/settingsModal.html
+++ b/src/components/settingsModal.html
@@ -36,6 +36,7 @@
                                 <tr data-bind="template: { name: 'MultipleChoiceSettingTemplate', data: Settings.getSetting('hideHatchery')}"></tr>
                                 <tr data-bind="template: { name: 'MultipleChoiceSettingTemplate', data: Settings.getSetting('farmDisplay')}"></tr>
                                 <tr data-bind="template: { name: 'BooleanSettingTemplate', data: Settings.getSetting('showCurrencyGainedAnimation')}"></tr>
+                                <tr data-bind="template: { name: 'BooleanSettingTemplate', data: Settings.getSetting('showCurrencyLostAnimation')}"></tr>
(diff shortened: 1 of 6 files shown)


RedSparr0w:
I think this should definitely be a part of the existing `animateCurrency` as then we only need to make changes to the 1 file if the animations change etc.

In terms of making the changes, you probably just need to add a check if the amount is positive/negative then return based on setting.
Only show the `+` if positive amount.
Move the currency downwards if negative?

CorgiOnNeptune:
From what I can see, amount is only viewed as a positive integer through the wallets methods so I might have to work around it another way. But I'll take a longer look at it.

RedSparr0w:
Something like this should work in the loseAmount method:
```
        animateCurrency({ ...amount, amount: -amount.amount });
```

CorgiOnNeptune:
Still not quite what you were looking for.. But closer.. I couldn't get your suggestion to work. It didn't give me any errors, but I also couldn't get any results from it. 

Also, it should probably be just the one setting for both animations rather than having options for each, ya?

RedSparr0w:
Gave it a minor refactor to simplify things a little, could you check it still performs the way you expected?

> Also, it should probably be just the one setting for both animations rather than having options for each, ya?

Either works for me, just whichever you think works best?

CorgiOnNeptune:
Thanks for cleaning it up. The code looks much better and is working as intended.

For the settings, I suppose it doesn't hurt to leave both options there.

I think I can see what you did differently to make the `animateCurrency({ ...amount, amount: -amount.amount });` work when I couldn't, so thanks again.

PR #2415, opened 2022-06-09
Add glow to the quest marker

TractoBoy:
Hello, 

A recurring complaint I've seen in a discord server playing Pokeclicker was the difficulty to locate quests spots, especially dungeons as they often "fuse" with the green of the background image. 

I've had the idea to add a simple "glowing" effect to the class "questAtLocation", which catch the eye a bit.
I tried to not overdo it. 

Here is a small sample :

https://user-images.githubusercontent.com/4422077/172954892-97ae8235-e528-4a99-a84d-8f809654209d.mp4

Thanks for reading !

### DIFF
--- a/src/styles/map.less
+++ b/src/styles/map.less
@@ -1,7 +1,5 @@
 @import "variables.less";
 
-
-
 #map {
   @keyframes faintLight {
     0% { filter:brightness(1) }
--- a/src/styles/map.less
+++ b/src/styles/map.less
@@ -1,6 +1,14 @@
 @import "variables.less";
 
+
+
 #map {
+  @keyframes faintLight {
+    0% { filter:brightness(1) }
+    50% { filter:brightness(2) }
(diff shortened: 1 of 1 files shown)


CorgiOnNeptune:
Devs will have to compare the PRs, but somebody else already made a PR for this under #2375 .

TractoBoy:
Ah, didn't find it under the Issues tab, I forgot to check PRs, sorry.

Looks like we can maybe "merge" both our implementations, using a filter to produce the glow/pulse reduce the footprint of the css/js involved in their PR.

Jaaslet:
Thanks for the PR, but the develop team don't think it fits with our vision for the game.
We rather want to fix the problem by changing the default map-colors.
If you think your are up for the task of changing the default colors, feel free to submit a PR. To be honest, i am hoping for someone with some understanding of colors to solve it, which is why I have not fix it myself yet.
I will close this PR.

PR #2416, opened 2022-06-09
[SUGGESTION] Add "Breeding efficiency (Current region)" (or Max Region)

TractoBoy:
**Current Behavior**

Currently, there is only an option for "Breeding Efficiency". However, if the player has the challenge "Attack Debuff" active, the Breeding Efficiency will vary, sometimes greatly, from the generic formula.

**Improved Behavior**

I've implemented a "Breeding Efficiency (Current Region)" for players who wants to know what's the best to break through a bottleneck at the current region (can't beat Elite/League, not enough damage to OS mons in routes/dungeons, etc).

This feature could be changed/extended with a "highest region" filter/sorting order. 

I understand this is a very "niche" use case, but maybe people are interested.

**Screenshots:**

Example of sorting & displaying breeding efficiency for current region (Alola) : 
image

Same, without accounting for regional debuff : 
image


Aegyo:
This has a build failure, could you fix it please? The line too long error can probably be autofixed with `npm run eslint-fix`

TractoBoy:
Autofix wouldn't work, just report the error and leave it here. 

I've exploded the line into several, each holding a part of the equation.

Jaaslet:
#2576 implements a more in dept version of this PR, so i will close this one. Sorry, but thanks for your work :)

PR #2418, opened 2022-06-10
Battle Frontier Logbook

### LINKED ISSUE #2289: [SUGGESTION] BF notification changes, add entry to logbook
**Current Behavior**

Currently, the BF notifications don't show up in the logbook.  Also, the BF notifications stick around for a lot longer than normal notifications for some reason.

**Improved Behavior**

I'd like BF notifications to show up in the logbook and the BF notifications to have the normal notification timeout.  A setting to turn off the notification entirely would also be nice to have.

**Reason for change**

If the BF notification, such as the 'you reached x stage, and earned y amount of BF tokens' notification is in the logbook, then there's no reason for the BF notification to have such a longer notification time than any normal notification.  It'd also be easier to just come back to the game some time after BF ends and find out how much BF tokens you earned / what stage you got to.  Also, we have a logbook.  Not using it to log something like this when we do use it to log quest completion and such just feels silly.  

As for turning the notification off - I don't feel it's necessary for myself, but given that a number of players want all notifications to be possible to be disabled, it'd be a welcomed addition.

**Detailed Description**

**Screenshots:**

jk13pclick:
Closes #2289. Probably a good addition, although I could see this done without a new category.

### DIFF
--- a/src/modules/logbook/LogBookTypes.ts
+++ b/src/modules/logbook/LogBookTypes.ts
@@ -40,4 +40,8 @@ export const LogBookTypes: Record<string, LogBookType> = {
         display: 'dark',
         label: 'OTHER',
     },
+    FRONTIER: {
+        display: 'success',
+        label: 'FRONTIER',
+    },
 };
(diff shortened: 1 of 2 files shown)


jk13pclick:
Closing, #2419 does this.

PR #2419, opened 2022-06-10
Notif and Log Updates (And some more)

### LINKED ISSUE #2167: [SUGGESTION] Some QoL notification settings
**Current Behavior**

The following notification bubbles cannot be disabled:
1. "You have successfully completed the dungeon"
2. "You dig deeper..."

**Improved Behavior**

A notification setting to disable them.

**Reason for change**

Allow the player to have notification bubbles for things they need/want to pay attention to, outside of stating the obvious.

1. This is okay the first few times when someone is new to the game and the confirmation of "oh, this is considered completing the dungeon" is nice, but after that point, it feels unnecessary. Dungeons require manual play so I know when I've completed it and I don't need a notification bubble to reassure me. Especially when I'm grinding out the 500 run achievement, this can be very noisy when the boss is near the entrance a few times in a row.
2. Similar to the above reason, I can tell when I'm done with a new layer because I'm right there, seeing it happen. It's not like there is a helper that's digging up items for me and that notification may serve a purpose of informing me that a layer is done. Even then, the usefulness of the notification is questionable because layers don't really matter compared to the items.

**Detailed Description**

**Screenshots:**

### LINKED ISSUE #2178: [SUGGESTION]  Highlight the rares drops in dungeon
I was wondering if Highlighting the rares drops in dungeon would be a good idea ? I'm sorry if thats was already asked, i'm not familiar at all with github so i didn't found my request anywhere or i didn't searched very well.

### LINKED ISSUE #2180: [SUGGESTION] Underground Tunnel Notifications
### **Current Behavior:**

There doesn't seems to be any setting to enable/disable notifications related to the Underground Tunnel

### **Improved Behavior:**

Having the ability to toggle off "You found item X" or Treasure Scanner "You found an extra Y" or "You dig deeper..."

### **Reason for change:**

Notifications can get a bit spammy at late game

### **Detailed Description:**

Include a section in the settings to disable notifications related to the Underground Tunnel

### **Screenshots:**

spammy_notifications

### LINKED ISSUE #2289: [SUGGESTION] BF notification changes, add entry to logbook

CorgiOnNeptune:
Some stuff I was working on and have a list of more I want to add before this PR is complete.
@jk13pclick got ahead of me on this by already adding a couple of the BF stuff, I'll change my logbook entry to match his PR. 

_New_
- Leave Dungeon confirmation dialog option
- New berry unlocked notification and log
- Log and improved notification when you can't afford Helpers/Flutes anymore.
- BF log upon completion and milestones
- Add (duplicate) note to logbook for duplicate shiny catches/hatches
- Log for roamers

_Notification Settings_
- 'You dig deeper'
- Dungeon Completed
- Quest Completed

_Fixes_
- misc minor text fixes
- Fix logbook sometimes increasing all rows height if one row has overflow (Can screenshot if need)
- Fix/add log tags for some shiny encounters

Will add the new sounds in a separate PR. 

closes #2180
closes #2178 
closes #2167 
closes #2289


jk13pclick:
Also closes #2289.

CorgiOnNeptune:
There was a couple other related notif/log things I wanted to add to this still, but I'll mark as ready, since what's in it should be good to go. I'll put the other things into smaller PRs later since this is already quite hefty in terms of file changes and will be easier to merge sooner rather than later.

src/scripts/breeding/HatcheryHelper.ts
```diff
@@ -75,13 +75,23 @@ class HatcheryHelper {
...
 
+    // String for currency in Notifications and Logs
+    currencyString(string: string): any {
+        if (GameConstants.humanifyString(GameConstants.Currency[this.cost.currency]) != 'money') {
+            string = (`${GameConstants.camelCaseToString(GameConstants.Currency[this.cost.currency])}s`);
+        } else {
+            string = ('Pokédollars');
+        }
+        return string;
+    }
```
Aegyo, line 86: Why does this have a string input, it looks like it will always be overwritten?
Also this feels like it should be a switch-case instead, but I assume eslint complains if we do that without enough cases...

CorgiOnNeptune: Does the change I just pushed look better? Doesn't look like any eslint complaints about the single switch case.

Aegyo (approved):
Changes look good to me, only thing we might want to add are new sound files, and possibly an update function for the notifications that have been moved.

CorgiOnNeptune:
> Changes look good to me, only thing we might want to add are new sound files, and possibly an update function for the notifications that have been moved.

I can look at both of those. Any specific guidelines I should follow, or things I should avoid for the sounds? I was just going to hop into my DAW and mess around with some stuff.

Aegyo:
I don't think we have any guidelines for sounds, I would just post them in the discord first to confirm :+1:

src/modules/logbook/LogBookTypes.ts
```diff
@@ -24,6 +24,10 @@ export const LogBookTypes: Record<string, LogBookType> = {
         display: 'primary',
         label: 'FOUND',
     },
+    ROAMER: {
+        display: 'info',
+        label: 'ROAMER',
+    },
```
CorgiOnNeptune, line 30: This might be too many Logbook filters. Will probably just put the roamer log under a different filter before this is finished.

CorgiOnNeptune:
I'm not sure how notification settings are stored, but it seems like the ones that have changed category don't need an Update function to transfer the player's setting over. Only if I changed the actual name of any of the settings, which in this case I didn't. So I just need to finish making the new sounds then this should be good to go.

Jaaslet (changes requested):
I think the leave dungeon confirmation box should be something you turn on in the settings. It seems like something only advance players uses, when they start spamclicking the dungeons, to grind them, so it will just be an annoyance for new players.

Jaaslet:
I am guessing the sounds are not added yet? Can you mark your PRs as drafts, while you are still working on them?
And maybe split PRs up. It's kinda hard to review so many things at the same time. And more PRs also leads to a higher number in the credits ;)

CorgiOnNeptune:
Ya, this one ended up bulkier than I originally intended. Will definitely have stuff more split up in the future when the changes aren’t dependant on one another. 

This was a draft for a while, but I marked it ready since I was gonna space out some of the other additions I wanted, like the sounds, so it could get the merge of all the small file changes out of the way. 

I’m pretty busy the next few days, so I can get that setting for the leave dungeon confirm later today and this should be good to merge to minimize future conflicts. Then I can just make the sounds a separate PR later in the week when I get the time.

Jaaslet (approved):
Looks good. Thanks for the PR!

Issue #2156, opened 2022-05-13
[SUGGESTION] Visual improvements for gym battles


Kays:
**Current Behavior**

1. Gym battles do not show the gym leader's image in the background.
2. Elite Four & Champion battles use the same background.

**Improved Behavior**

1. Show the gym leader's image in the background when fighting them.
2. Change the background to a theme that best matches the Elite Four's type and Champion fights. I'm aware that most Champions do not have singular types so it can be tricky. Maybe we can leave those alone or if there can be a new image drawn up for those fights.

On the note of 2, I did a quick review of the gym battles and I think these can be improved, using the existing images, but if you want to change them to new images to better match the fights, I'm good with that!
**Hoenn**
* Sootopolis gym should use water background instead of Ice. Juan's Pokémons are primarily Water types.

** Unova**
_(confession - I skipped Unova both in the anime and the games so I really don't know if the current images make sense with the way things work in Unova. If so, I apologize; feel free to dismiss my suggestion(s). I don't know if this game aims to match them, but my suggestions are based on the gym's type.)_
* Virbank gym should use a different background, but I'm not sure which. Maybe ghost one? It currently uses a background image that's commonly associated with Electric-types when it's a Poison-type gym.
* Castelia's gym should use a forest background for Bug-type.
* Driftveil's gym should use a cave background for Ground-type.
* Mistralton's gym should use an open field background for Flying-type.

**Kalos**
* Santalune's gym should use a forest background.
* Cyllage's gym should use a cave background.
* Coumarine's gym should use a grass background instead of water. Despite the city's name, Ramos mains Grass-types and none of them are dual Grass/Water.
* Anistar's gym should use a ghost background. I don't know if it's the best but at least it's consistent and better than the current one.

**Alola**
The are not many gyms in Alola and they all use the same background. I suppose that's fine since all the fights tend to be outside, but if we want to match up with types, they can changed. I'll defer.

**Reason for change**

1. Just to give the fights more oomph. Fighting them and only seeing the Pokémon feels like a random encounter rather than a gym/trainer battle.
2. The field background image idea is not a new concept and it has been used in gyms since Kanto. It just so happens that Elite Four and Champion fights always use the same background. I know they the fights are in a big building and not outdoors, but even in the video games, the field change to match the Elite Four's dominant Pokémon type. Champion, I'm not sure. Drake, Wallace, and Iris use a dominant type. We can go off of the Champion's ace Pokémon as one way to do it.

**Detailed Description**

[...]

PR #2422, opened 2022-06-11
Gym Leader Images and Background Update

### LINKED ISSUE #1238: [SUGGESTION] Add Gym Leader Images to Gym View
In #1219, I've added some commented-out code in `gymView.html`

```HTML
        <div>
            <!-- TODO: HLXII - Add after all Gym Leader images are formatted correctly
            <div style="position:absolute; left:65%;">
                <img height=80px data-bind="attr:{ src: GymBattle.gym.imagePath }"/>
            </div>
            -->
            <img class="enemy" data-bind="attr:{ src: PokemonHelper.getImage(GymBattle.enemyPokemon(), GymBattle.enemyPokemon().shiny) }" src=""/>
        </div>
```

This will display the Gym Leader sprite during the Gym battle, much like how the Trainer images will appear during Trainer battles. The main issue with this is that currently some of the Gym Leaders do not have good images to display in the view.

I'm currently assuming the Gym Leader images are 80x80 pixels. If this is the case, they will be displayed correctly:
image

If they are smaller, or aren't pixel art at all, it looks bad:

image
Looks slightly blurry due to the image being 64x64.

image
lol.

To fix this, we need to update the Gym Leader images to be 80x80 pixel art for every Gym Leader.

jk13pclick:
Closes #1238 and part of #2156. We should probably decide what to do about more backgrounds before this is merged


RedSparr0w:
Just double checking, have you run all the images through something like https://tinypng.com/ to minimise the file sizes?

zyx-xyzzy:
It looks like Jagged Pass and Po Town have no background at all set in the current state of the PR? (they were removed from Mansion, but not readded anywhere else)

jk13pclick:
> It looks like Jagged Pass and Po Town have no background at all set in the current state of the PR? (they were removed from Mansion, but not readded anywhere else)

Po Town was a mistake, I wanted to give Jagged Pass the default route background.

jk13pclick:
Closed by #2860 and #2897, so closing this.

PR #2425, opened 2022-06-11
Move Hoenn Farm SVG

Ultima1990:
Totally forgot to move the farm SVG over one tile when I moved over that whole area on the map to make room for Weather Institute.

Oops.


Jaaslet (approved):
The farm is now the correct place.
Thanks for the PR

PR #2426, opened 2022-06-11
Normal -> Neutral

Ultima1990:
Turns Normal type effectiveness into Neutral type effectiveness. Because Normal is itself a type. So that might be confusing.


Jaaslet (approved):
I like the change. And after doing a quick google search, it does not look like there is an official word for it.
I clicked around, and stuff seamed to work. It turns out that we are not referencing this many places, which is why the PR is smaller than i expected.
Approved. Thanks for the PR!

PR #2427, opened 2022-06-11
Night time Gliscor and Weavile

Ultima1990:
Because it's canon.

With NPC, because we don't want any more of that Riolu confusion.


Jaaslet (approved):
Looks good. It works as described, and it's canon, so might as well.
And the NPC is good.
Thankss for the PR

PR #2429, opened 2022-06-11
Fixed eggqueue stuck with hatchery helpers

Jaaslet:
Consistent way of reproducing it (don't know if there are other ways):
- Set filters, so you can fill up the queue with all pokemons in that filter
- Start hatchery helpers
- Fill up the queue
- Wait for a helper to take a new pokemon, and then wait for the 4th slot to be hatched


RedSparr0w (approved):

PR #2433, opened 2022-06-12
feat(hatchery): allow eggs and fossils in queue

### LINKED ISSUE #1362: [SUGGESTION] Allow eggs and fossils to be placed in the hatchery queue
**Current Behavior**
The hatchery queue doesn't allow for eggs and fossils to be placed inside

**Improved Behavior**
I would like to see the hatchery queue being able to take eggs and fossils

**Reason for change**
We have to clear out the queue if we want to add an egg or a fossil

**Detailed Description**
instead of placing the eggs/fossils directly into the hatchery I would like to see it being able to put into the queue aswell

**Screenshots:**

image

shallotmama:
Adds the ability to put eggs and fossils in the queue.

Closes #1362


shallotmama:
image

Jaaslet:
(Have not tested it yet. Just took a quick look)
The breeding window still has both the "full queue" and "no empty slots" warning. I don't think there are any situations anymore, that only cares about empty slots? So we should remove that.
And the implementation looks cool! It's how i would had made it, if i ever found time to do it.

Jaaslet (changes requested):
HatcheryQueueEntry.ts seams a bit overcomplicated, which makes the code hard to understand (and hard for me to review the PR).
Can it be changed to an abstract class, with a child-class for each of the 3 types?
Then it can also contain stuff like how to find the image and how to add/remove it from queue, to make the html look nicer.

shallotmama:
The three types are all string literal union types pulled from enums, so they don't make sense as their own classes. I did a little cleanup though, and I think it's a lot clearer now.

The main complication was that `EggItemType` and the image files use the format of `<type>_egg` while everything else just uses the type name. I don't know why I was doing the weird conversions instead of just doing it the way it is now. 😆

src/modules/pokemons/PokemonNameType.ts
```diff
@@ -4,1205 +4,1209 @@ To update this type when adding new pokemon:
...
-    | 'Squirtle (clone)'
-    | 'Wartortle (clone)'
-    | 'Blastoise (clone)'
-    | 'Mega Venusaur'
-    | 'Mega Charizard X'
-    | 'Mega Gyarados'
-    | 'Mega Gardevoir'
-    | 'Mega Rayquaza'
-    | '???';
+export const pokemonNameList = ['Bulbasaur',
```
shallotmama, line 19: I changed this to allow for a safe type guard on `PokemonNameType` using the pattern found here: https://stackoverflow.com/questions/36836011/checking-validity-of-string-literal-union-type-at-runtime

If this is an unwanted change, let me know, and I can go back to using the unsafe type guard (which just checked if it wasn't an `EggNameType` or `FossilNameType`).

Aegyo:
This PR has gotten a bit stale and collected a bunch of conflicts (our fault, sorry about that), I'm going to close this one and if you are still interested in doing this you can redo/rebase your current work over the latest develop and open a new PR.

PR #2435, opened 2022-06-13
Erika/Sabrina gym rebalance

zyx-xyzzy:
Currently the progression of total gym pokémon health in the middle of Kanto goes 37,000 (Surge) -> 105,550 (Erika) -> 77,040 (Sabrina) -> 137,210 (Koga). This PR takes the lazy way of fixing this uneven curve by just swapping them: now Erika has 77,040 health and Sabrina has 105,550, distributed among their mons in approx. the same ratios as the previous setup had. Prob not perfect balance-wise, ofc, but hopefully should still at least be an improvement?


zyx-xyzzy:
> #1759 already does this.

From the looks of it that PR has been waiting in limbo for months and thus presumably won't come out for months still; there's nothing wrong with adding this as a temporary fix in the time between now and then.

Ultima1990:
> presumably won't come out for months still

You'd be surprised.

Qwertypop04:
As an additional note, we (the developers and contributors) have decided we should be in charge of all future balance changes. If you believe balance changes are warranted, please open an issue instead of a PR, and suggest the nature of the change you believe should be made instead of making the changes yourself. This allows for proper discussion of the matter, and then for a decision to be made.

zyx-xyzzy:
> As an additional note, we (the developers and contributors) have decided we should be in charge of all future balance changes. If you believe balance changes are warranted, please open an issue instead of a PR, and suggest the nature of the change you believe should be made instead of making the changes yourself. This allows for proper discussion of the matter, and then for a decision to be made.

Alright, fair enough. That said, could you please make information on development policies like this more readily available to the public in the future? Now I just look like a moron and/or an entitled asshole when it could've been avoided if I had known; would be nice to reduce the odds of smth similar happening to other contributors going forward.

PR #2442, opened 2022-06-13
Team Rocket in Kanto

Ultima1990:
Featuring 2 trainer dungeons, one questline, and one free master ball.


Jaaslet (changes requested):
It looks good, and works fine.
But it's a bit weird that half of the city is a dungeon. It's hard to hit the city, and you get confused first time you press the dungeon.
Maybe there should not be any overworld layer for the two new dungeons, and you can only enter them from the city?

Ultima1990:
> It looks good, and works fine. But it's a bit weird that half of the city is a dungeon. It's hard to hit the city, and you get confused first time you press the dungeon. Maybe there should not be any overworld layer for the two new dungeons, and you can only enter them from the city?

If we don't have them on the map we get the old Team Rocket Hideout situation again. Where people can't find the dungeon. Also there's still more saffron on the map compared to Silph Co. than there is Lavender on the map compared to Pokémon Tower. I'd say it's most similar to Goldenrod/Radio Tower. Which apparently works well.

IMO this is fine.

PR #2448, opened 2022-06-14
Magikarp jump quest

Jaaslet:
I have made a quest, which is basicly a reference to the mobile game Magikarp Jump. It's stupid and fun (or at least i tried to).
You can start it by going to route 3 pokemon center (kanto) after you reached Unova.
It's not meant to be an easy quest. It requires harvesting 2 different berries, which you need to mutate to get. But a lot of the stuff are meant to be a joke (like the temp battle only having 20 hp).
In the end, you get a Magikarp (spoilers!).

- [x] Write the questline
- [x] Wait for Sevii to be merged
- [ ] DefeatSpecificPokemonQuest
- [x] Get someone to make an alternative form to Magicarp
- [ ] Connect all the quests and NPCs
- [ ] Show it to the crew, and see if they like it


Jaaslet:
I don't really like my idea anymore, so it will be hard for me to turn it into something fun.
Bye bye PR

PR #2458, opened 2022-06-15
Some small Kanto map changes

Ultima1990:
Moved Daycare over one tile, no longer overlaps with route 5.
Made 24 longer, 25 shorter.
Moved Victory Road to the end of 23, 23 is now one continuous piece.


Jaaslet (approved):
Yay! More route 5!
Thanks for the PR

Issue #2346, opened 2022-06-01
Visual Issue on Profile [BUG] 


Shinesprite41:
**Info**
 - OS: Windows 10
 - Browser Chrome & Firefox
 - Version 0.9.5

**Description:**
Very small visual issue but when you go into Menu->Profile, the version number is covered by the challenge ribbons. This is not the case on the main screen, and is only in profile. 

**Steps to Reproduce:**

1. Go to Menu
2. Click on Profile
3. See Error

**Expected Behavior:**
The version number shouldn't be covered. 

**Screenshots:**
Unknown Browser (Eat Pant)
![image](https://user-images.githubusercontent.com/73353569/171306050-63ec4010-3f4e-4b6b-82b2-58d2ef5d6698.png)
Firefox
![image](https://user-images.githubusercontent.com/73353569/171306089-fa64a449-37e2-4900-9702-3ae220591224.png)
Chrome
![image](https://user-images.githubusercontent.com/73353569/171306195-761c6aa1-87bd-4fba-af11-491fcd54b7f0.png)

**Additional Information:**
Jessica is cool

Jaaslet:
#1926 made the icons smaller, so this should not be a problem anymore

PR #2459, opened 2022-06-15
Visual Issue on Profile [BUG]

joaosantos99:
Fix for issue #2346  . Added a media query to bigger screen with a adjusted margin on the ribbons.


Jaaslet:
Thanks for the PR, but this problem was already solved as a part of #1926, so i will close this.
And if we get enough challenges that this becomes a problem again, i think the correct solution is to make the profile window bigger, so the trainer cards are consistent in size.

PR #2462, opened 2022-06-16
Observable Fixup + Loot Notification

jk13pclick:
This PR kind of snowballed into a bunch of changes. See list below:

* Protein Obtained/Purchased observable split
* Some code-cleanup of the ``DungeonRunner.gainLoot`` function
* Loot notification specifications (PR #2419 and some more)
     * Images for loot in the notification bar
     * Different colored notification bar for different weight items
     * Notification setting for weights >= 2 and weights < 2


CorgiOnNeptune:
Oh nice! I was going to work on some of this, but haven't had time since I started my PR. Yours looks better than what I had started regardless! But exact same idea I had with regards to rarity and the item images in the Notifs.

src/scripts/Update.ts
```diff
@@ -730,6 +730,16 @@ class Update implements Saveable {
                 lastReminder: saveData.statistics.secondsPlayed,
                 lastDownloaded: saveData.statistics.secondsPlayed,
             };
+
+            // Just incase statistics is not set
+            saveData.statistics = saveData.statistics || {};
+
+            // Rename from the old statistic name
```
Jaaslet, line 737: This comment looks like a copy/paste mistake

src/scripts/Update.ts
```diff
@@ -730,6 +730,16 @@ class Update implements Saveable {
...
+            saveData.statistics = {
+                ...saveData.statistics,
+                totalProteinsPurchased: saveData.statistics.totalProteinsObtained || 0,
```
Jaaslet, line 740: It's a bit weird that they are just set to the same value... It's confusing for new players... But i have no idea how else to do it...

src/modules/notifications/NotificationConstants.ts
```diff
@@ -57,7 +57,8 @@ const NotificationConstants = {
         },
         Items: {
             route_item_found: new NotificationSetting('notification.route_item_found', 'Item found during route battle', true),
-            dungeon_item_found: new NotificationSetting('notification.dungeon_item_found', 'Item found in dungeon chest', true),
+            common_dungeon_item_found: new NotificationSetting('notification.common_dungeon_item_found', 'Item found in dungeon chest', true),
+            rare_dungeon_item_found: new NotificationSetting('notification.rare_dungeon_item_found', 'Item found in dungeon chest', true),
```
Jaaslet, line 61: They currently have the same title.
Also, maybe we should have an update function for them, so people get to keep their setup?

Jaaslet (changes requested):
It's a bit hard to test a PR with so many different things in the same PR, but i think i tested everything...
Overall it looks good. I still have some comments :)

src/modules/notifications/NotificationConstants.ts
jk13pclick: Changed title, unsure how to write the update function

Jaaslet (approved):
I have added the update function. The rest looks good.
Thanks for the PR!

PR #2464, opened 2022-06-16
Add AZ battle

Ultima1990:
Initially I wanted to make this a quest. But it's just one battle. Not really worth a grand questline. So instead I'm just adding this as a post-kaloschamp battle that the player will have to find on their own.


Jaaslet (approved):
New pokemon!
It looks good, and the update function seams to work fine.
Thanks for the PR

PR #2465, opened 2022-06-16
Fixing a comment instruction.

Nydaleclya:
So that no one messes up by blindly following instructions.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2467, opened 2022-06-16
Shop updates

Ultima1990:
All shops in towns now sell balls. Pokéballs, Greatballs and Ultraballs depending on where they would unlock in the main games.
Most shops now sell restores, _after_ they are first introduced.
Small Restores appear alongside Pokéballs.
Medium Restores now appear alongside Greatballs.
Large Restores now appear alongside Ultraballs.
Some exceptions. Medium and Large Restores don't appear in Kanto. Medium Restore appears in Violet City even though there are no Greatballs.
All home towns no longer have shops.
All other towns that didn't have shops now have shops. For example you can't buy balls in New Bark anymore, but you can now buy some in Cherrygrove. You can survive clearing one route without a shop surely?
Some towns still don't have the balls and restores treatment. Because they aren't actually towns. Places like Pal Park, Aether Paradise.

Created a consistent item order. After balls and restores it's battle items, eggs, stones, pokemonitems, other.
For eggs it's grass, fire, water, electric, fighting, dragon.
For stones it's leaf, fire, water, thunder, moon, trade. After that sorted by region and then by the dex number of the first mon it would be used on.

Also added Moon Stone to Cianwood because Johto didn't have any Moon Stone for sale for some reason.
Also moved Kalos' Upgrade and Prism Scale to Cyllage because that town just had nothing for some reason.

### DIFF
--- a/src/scripts/towns/TownList.ts
+++ b/src/scripts/towns/TownList.ts
@@ -46,6 +46,12 @@ const pokeLeagueShop = () => new Shop([
 ]);
 
 //Kanto Shops
+const ViridianCityShop = new Shop([
+    ItemList['Pokeball'],
+    ItemList['xAttack'],
+    ItemList['xClick'],
+    ItemList['Dungeon_ticket'],
+]);
 const PewterCityShop = new Shop([
     ItemList['Pokeball'],
     ItemList['Token_collector'],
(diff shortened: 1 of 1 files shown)


Jaaslet (approved):
It looks good. I agree with the change.
I took a quick look, and it looks correct. But this is again an issue where the players will see if there are any mistakes, and it will not be a big problem

PR #2468, opened 2022-06-17
Even better shopping

Ultima1990:
Changed my mind on the restores. They don't need to be available in every single town. That's a bit overkill. So now there's one shop for each kind of restore in every region + the department store. So 2 shops per restore total. Except in Kanto, where the only available type of restore is the Small Restore.

Battle items should be available in every store however. And so now they are.
Spot the pattern.
There is a pattern.
Also not actually every store, because some regions just have so many towns that I couldn't be arsed. To be more specific, post-league towns don't have battle items.

Also I swapped the items in Jubilife and Oldale. Because it's become a thing that the first QP item available in a region is the Mystery Egg. So it's now Mystery Egg in Jubilife, Moon Stone and Sun Stone in Oldale.

### DIFF
--- a/src/scripts/towns/TownList.ts
+++ b/src/scripts/towns/TownList.ts
@@ -2418,8 +2418,8 @@ const SnowbelleCityShop = new Shop([
     ItemList['Greatball'],
     ItemList['Ultraball'],
     ItemList['Protector'],
-    ItemList['Reaper_cloth'],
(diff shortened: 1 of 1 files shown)


Jaaslet (approved):
Looks good. The Wiki-people are going to have a blast updating stuff, when this update comes out!
Thanks for the PR

PR #2471, opened 2022-06-18
Move Battle Item Tooltips to Left

### LINKED ISSUE #2470: [SUGGESTION] Allow disabling of Battle Item tooltips or have tooltip appear above battle item
**Current Behavior**

Currently, hovering your mouse over a battle item causes a tooltip to pop up explaining what that battle item does.

**Improved Behavior**

Either allow us to disable that tooltip or cause the tooltip to appear above the battle item rather than below it

**Reason for change**

When the tooltip appears - and it appears while clicking the battle item - it blocks the time remaining from being shown for that battle item.  This causes issues if I'm trying to keep a battle item active for only a certain amount of time - like setting xAttack to be active for the next 3 hours while I do a BF run, and not longer than that.  It's annoying having to move my mouse away from clicking the battle item to see how long it's set to be active for.

**Detailed Description**

**Screenshots:**
battle_items

CorgiOnNeptune:
Moves the hover tooltip on battle items/flutes container to left. This way the tooltip doesn't cover up either the timer or the item quantity while hovering over/using it. 

Closes #2470

### DIFF
--- a/src/components/battleItemContainer.html
+++ b/src/components/battleItemContainer.html
@@ -40,7 +40,7 @@
                             tooltip: {
                               title: ItemList[$data].displayName + '<br/>' + ItemList[$data].description,
                               trigger: 'hover',
-                              placement:'bottom',
+                              placement:'left',
                               html: true
                             }">
                 <img src="" class="clickable" data-bind="attr: { src: ItemList[$data].image }" width="26px">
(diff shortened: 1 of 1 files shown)


Jaaslet (approved):
Looks better now. Thanks for the PR

PR #2472, opened 2022-06-18
Add new NPC

Ultima1990:
New NPC in Lavender Town that points out the FAQ button in the Start Menu.


CorgiOnNeptune:
Could maybe do with an update to the content of the FAQ alongside this. Since there are still lines like this in the FAQ:

faqss

Ultima1990:
The berries one is also in the Discord version of the FAQ and also is still true. So I'm gonna leave that one.
But I'm updating a bunch of other FAQ stuff.

Jaaslet (approved):
I like the new NPC, and the information in the FAQ now looks correct.
Thanks for the PR

PR #2475, opened 2022-06-18
New New Mauville image

Ultima1990:
Certainly gives a better impression of what the place is than that tiny little square entrance room.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2476, opened 2022-06-18
QuestLineHelper Cleanup

Ultima1990:
Creates some order in the unorganized chaos of QuestLineHelper.ts. Will definitely be useful in future when we potentially get many more questlines for obtaining legendaries and alternate forms and such stuff. It's gonna be like a bajillion questlines.

QuestLines are now ordered by region, within regions they are ordered by when they would be obtained, and there's a bonus Event "region" at the bottom. With comments for where each region starts.


Baldrani (approved):

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -134,34 +243,145 @@ class QuestLineHelper {
...
+        minasTrialAlolaQuestLine.addQuest(clearKahunaNanu);
+
+        const clearMinasHouseboat = new CustomQuest(1, 0, 'Complete the Trial! Clear Mina\'s Houseboat in Seafolk Village.', () => App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex('Mina\'s Houseboat')]());
+        minasTrialAlolaQuestLine.addQuest(clearMinasHouseboat);
+
+        App.game.quests.questLines().push(minasTrialAlolaQuestLine);
     }
 
+    //Event QuestLines
     public static createVivillonQuestLine() {
```
Jaaslet, line 385: VivillonQuestLine is a Kalos questline, not an event questline.
The Vivillon event is just a roamer

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -3,6 +3,7 @@
  */
 class QuestLineHelper {
 
+    //Kanto QuestLines
```
Jaaslet, line 6: Space between // and Kanto, to follow what we are doing other places

Jaaslet (changes requested):
It works fine. At least the couple of quests i looked at. And looking at the code, there should not be any problems doing it like this.
I have a couple of things:

src/scripts/quests/QuestLineHelper.ts
Ultima1990: KK I fix

src/scripts/quests/QuestLineHelper.ts
Ultima1990: Oopsie poopsie. I fix.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2477, opened 2022-06-19
Item Magnet -> Dowsing Machine

Ultima1990:
To free up the Magnet for other uses.

With bonus NPC for the funny.

Hoping for some hero to bring in a higher quality Dowsing Machine sprite.


src/scripts/Update.ts
```diff
@@ -746,6 +746,11 @@ class Update implements Saveable {
 
             // Add AZ TemporaryBattle
             saveData.statistics.temporaryBattleDefeated = Update.moveIndex(saveData.statistics.temporaryBattleDefeated, 0);
+
+            // Give the players Dowsing Machines in place of Item Magnets
+            playerData._itemList.Dowsing_machine = playerData._itemList.Item_magnet;
+            delete playerData._itemList.Item_magnet;
+            delete localStorage.mine;
```
Jaaslet, line 753: I think localStorage.mine was something else, that should not have been copied. At least it works fine without.

Jaaslet (changes requested):
You are missing a battle frontier reward. Else it works. I have one more small comment:

src/scripts/Update.ts
Ultima1990: I fix!

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2478, opened 2022-06-19
Poison Barb -> Rocky Helmet

Ultima1990:
To free up Poison Barb for other uses.

Probably never gonna have contact damage be a thing anyway. Because the player never takes damage.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2480, opened 2022-06-20
Fire environment

Qwertypop04:
Adds the 'Fire' environment, along with a background


zyx-xyzzy:
Looks like Watchtower Ruins was removed from the Cave environment without being added anywhere else; is that intentional?

Qwertypop04:
> Looks like Watchtower Ruins was removed from the Cave environment without being added anywhere else; is that intentional?

Not really, but I took this from the Galar PR, where it's no longer a dungeon, so it shouldn't matter

Jaaslet (approved):
I moved the background a bit, so now i think it looks better.
image
Other than that, it looks good. Thanks for the PR!

PR #2482, opened 2022-06-20
Questline not starting fix

Nydaleclya:
Land vs. Water now starting after clearing Watson.

Also is seems that any savefile that cleared Watson before v0.8.14 did **not** get this quest correctly.
They did get "Land vs Water" not "Land vs. Water".

### DIFF
--- a/src/scripts/gym/GymList.ts
+++ b/src/scripts/gym/GymList.ts
@@ -454,7 +454,7 @@ GymList['Mauville City'] = new Gym(
     'Wahahahah! Fine, I lost! You ended up giving me a thrill! Take this Badge!',
     [new GymBadgeRequirement(BadgeEnums.Knuckle)],
     () => {
-        App.game.quests.getQuestLine('Land vs Water').beginQuest();
+        App.game.quests.getQuestLine('Land vs. Water').beginQuest();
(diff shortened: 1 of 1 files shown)


Ultima1990:
> Also is seems that any savefile that cleared Watson before v0.8.14 did **not** get this quest correctly. They did get "Land vs Water" not "Land vs. Water".

You'd think so but that change to the title happened with #2242 which has been merged but has not been pushed to live yet.
image

Jaaslet:
> You'd think so but that change to the title happened with #2242 which has been merged but has not been pushed to live yet.

We still need this PR, because else, if someone is going from version 0.8.13 to 0.9.6, they will encounter this bug

Ultima1990:
Oh yes absolutely we do this is definitely a correction that needs to happen and I didn't mean to imply that we don't need this. That one line was just wrong so I had to nitpick that.

Jaaslet:
Fair.
And i just looked at it again. This PR only fixes it for new players, but we need to fix Update.ts line 543, to make sure old players, who have had a break, gets it too

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2484, opened 2022-06-21
Dowsing machine's update function no longer throws away effect time left

Jaaslet:
There was a bug in the update function, which replaced Item Magnet with Dowsing machine, which threw away how long time was left of the effect. This PR fixes that update function


RedSparr0w (approved):
👍 Thanks for the PR!

PR #2487, opened 2022-06-22
Dynamic Background fixes

zyx-xyzzy:
Implements some minor fixes to the dynamic background system, as discussed on Discord a couple months back:

- Kyogre, Phione, and Manaphy are no longer classified as "flying" since they... don't actually fly, lol. Reshiram, Tornadus, and Mismagius are now "flying" to match their counterparts and pre-evo, respectively. Tranquill, Mandibuzz, and Yveltal are now "flying" in accordance with their depiction in their dynamic-background sprites and main-game 3D models.

- Added missing images for non-shiny Decidueye, non-shiny Incineroar, and shiny Bounsweet (thanks Kita!). The Bounsweet one replaces a weird placeholder image that was in that slot previously.

- Removed all alt-form images, as it's been stated that their presence there was not intended.

- Updated image filenames of a bunch of Alolan mons to account for the gaps in numbering caused by deleted alt-forms (as well as a couple of pre-existing gaps).


Jaaslet:
(Have not tested yet, just gave it a quick look).
Kita should be added to Credits.ts

zyx-xyzzy:
> (Have not tested yet, just gave it a quick look). Kita should be added to Credits.ts

Done. (Does anyone know who made the rest of the Gen 5+ dynamic background sprites, btw? It seems like that info should be listed in Credits.ts too, since they seem to be apparently original to Pokéclicker, but it isn't on there atm.)

Jaaslet (approved):
It still works. Looked at a couple of them, and it looks correct. Great that you removed unused files!
There is no way that i am going to check every one of them, but it should not be a problem if any of them are wrong. And if it is a problem, people will tell us.
Thanks for the PR! Great work!

PR #2500, opened 2022-06-25
Disable bomb/survey if layer loading

### LINKED ISSUE #2497: [BUG] Bombing with Hotkey while digging deeper bombs the old layer, wasting energy
**Info**
 - OS: [e.g. Windows 10] Ubuntu 20.04
 - Browser [e.g. Chrome, Safari, Firefox] [alexislours] Pokeclicker Client
 - Version [e.g. `v1.0.0`] v1.2.0.deb

**Description:**

The bombing hotkey is not disabled while the game is 'digging deeper' in the underground.  Thus, it is possible to use the hotkey to bomb an already fully cleared layer, wasting mining energy.

**Steps to Reproduce:**

1. Go to Underground
2. Mine a layer completely
3. Click the Bomb hotkey just as the game begins digging deeper
4. Notice that the bomb went off on the cleared layer, as the new layer hasn't been bombed, and some tiles (depending on how cleared the previously cleared layer was) could be seen being removed by the bomb on the old layer.

**Expected Behavior:**

I'd expect that if the bomb hotkey was used during the 'digging deeper' it would be applied to the new layer, or the hotkey would be disabled like the button is while the digging deeper animation occurs.  

**Screenshots:**

**Additional Information:**

The timing on this bug is a bit finnicky, as I have gotten the bomb hotkey to bomb a new layer before you can push the bomb button again, but if you push the bomb hotkey during most of the digging deeper animation, it will bomb the old layer.  It's obviously not a serious issue, as you can just not bomb during that animation, but it is a bug so I figured I'd report it.

CorgiOnNeptune:
Closes #2497


src/scripts/underground/Mine.ts
```diff
@@ -154,6 +154,11 @@ class Mine {
     }
 
     public static survey() {
+        // Disable survey while loading new layer
+        if (this.loadingNewLayer == true) {
```
Aegyo, line 158:
```suggestion
        if (this.loadingNewLayer) {
```

src/scripts/underground/Mine.ts
```diff
@@ -256,6 +261,11 @@ class Mine {
     }
 
     public static bomb() {
+        // Disable bomb while loading new layer
+        if (this.loadingNewLayer == true) {
```
Aegyo, line 265:
```suggestion
        if (this.loadingNewLayer) {
```

RedSparr0w (approved):

PR #2501, opened 2022-06-25
Tyrogue can now be unlocked in Mt. Mortar again

Jaaslet:
After #2242 Tyrogue can never be unlocked in Mt. Mortar. This PR fixes that


RedSparr0w (approved):
👍

PR #2511, opened 2022-06-26
Fix Survival Area shop

Ultima1990:
This seems to have been causing at least some of the weird Sold Out issues.


Jaaslet (approved):
It looks like it fixed the bug. Thanks for the PR!

PR #2516, opened 2022-06-26
Fix bug when reloading on keyitem gained

### LINKED ISSUE #2513: [BUG] Possible to skip KeyItem reward function (Affects Pokerus)
**Description:**

Reward function is given on modal close button, and we don't open it a second time, so if you manage to escape the modal without clicking that button you miss out on the reward.

**Steps to Reproduce:**

1. Load a pre 0.9.6 save that has beaten distortion world
2. Refresh the page after seeing pokerus modal opened
3. See no pokerus on starter and no way to get it since the keyitem is already obtained 

**Expected Behavior:**

Should not be possible to avoid the reward function

Jaaslet:
Fixes #2513
The solution looks a bit weird, but the "rewards" from the tutorial keyitems needs to show up on close, for it to work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "pokeclicker",
-  "version": "0.9.6",
+  "version": "0.9.7",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "pokeclicker",
-      "version": "0.9.5",
+      "version": "0.9.7",
       "license": "ISC",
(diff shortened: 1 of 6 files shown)


RedSparr0w:
Not entirely sure about this solution, but can't really think of a better one atm either 😅 

Could you also add something to the update function to apply the Pokérus to the players starter if not already applied?

RedSparr0w (approved):

PR #2517, opened 2022-06-26
Fix Floette Eternal stats

Ultima1990:
Turns out this Flabble has different stats than other Flabbles. Cuz it's like 3k years old I guess?


Jaaslet (approved):
The new stats matches. Thanks for the PR!

PR #2529, opened 2022-06-27
Now displaying EV bonus on hover over

Jaaslet:
Now people can more easily see the pokemons EVs AND they can actually see the bonus it gives!
I am hoping this solution is enough, because we don't want the pokedex to be more cluttered, but let's release this with the hotpatch, and if people don't like it, we can change it in the EVs fixes patch.
image

I want to also add the bonus in the pokemon statistics window, but the bonus is not an observable, and i don't feel like doing the EVs code rewrite on this side of the hotpatch.


RedSparr0w (approved):

PR #2530, opened 2022-06-27
Removed None filter in BreedingFilters

Symi001:
None was appearing twice as it turns out it needs to be lower case in the test to filter it out :)


Jaaslet (changes requested):
It seams like we don't really use the Region.None anymore for pokemons. It's currently only used for 3 unobtainable pokemons. We might as well just remove both, as it's a useless filter.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2531, opened 2022-06-27
EXP Nerf Fix

jk13pclick:
EXP formula was incorrect, this should fix it at least temporarily


Aegyo (approved):
Matches the formula that was intended, according to the desmos links on #2146 👍 

Can you link the new desmos graph here, with corrected old exp formulae?

PR #2533, opened 2022-06-28
Enumerable Pokerus System

jk13pclick:
Rough draft, works but needs some more work

__Features:__
- Enumerable Pokerus states (good for images and for new spread system)
- New spread system:
   - A Pokémon with _at least one shared type_ (not including ``None``) with a ``Contagious`` Pokémon can contract Pokérus 
   - ``Contagious`` Pokémon must be _not ready to hatch_ to spread the virus
   - Pokémon that are being maintained by a ``Hatchery Helper`` cannot spread __or__ contract the virus.
   - Pokémon must first become ``Infected``, then hatch once to become ``Contagious``


src/scripts/Update.ts
```diff
@@ -800,6 +800,16 @@ class Update implements Saveable {
 
         '0.9.8': ({ playerData, saveData }) => {
             saveData.oakItemLoadouts = saveData.oakItemLoadouts.map((list, index) => ({ name: `Loadout ${index + 1}`, loadout: list }));
+            //UNTESTED POKERUS UPDATE FUNCTION - MAKE SURE THIS WORKS!!!!
```
Jaaslet, line 803: This is not a comment i want to merge :)
You can bump the version by searching for '0.9.7' and replace it in all non-doc locations (should be 3). Be aware that it will not trigger at build itself, so you need to rebuild manually, or change something else.
And i don't think it works. The value is still "true/false" on my save, after the update has been applied.

Jaaslet (changes requested):
The update function does not work (see comment below), which makes it hard to test.
I have two comments, and will give it a full review, when the update method works :)

src/scripts/Update.ts
Aegyo: It probably doesn't work because we replace the property names other than id with a number in the save data, to save space.
Looks like pokerus should be accessed with `p[8]`. probably shouldn't bother using the enum from PartyPokemon in case it changes in the future.

src/modules/GameConstants.ts
```diff
@@ -1200,3 +1200,10 @@ export enum DayOfWeek {
     'Friday',
     'Saturday',
 }
+
+export enum Pokerus {
+    'Uninfected' = 0,
+    'Infected',
+    'Contagious',
+    'Cured',
+}
```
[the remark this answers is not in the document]
jk13pclick: Done

src/scripts/Update.ts
```diff
@@ -800,6 +800,16 @@ class Update implements Saveable {
...
+            saveData.party.caughtPokemon.forEach(p => {
+                if (p[8]) {
+                    delete p[8];
+                    p[8] = GameConstants.Pokerus.Contagious;
+                } else {
+                    delete p[8];
+                    p[8] = GameConstants.Pokerus.None;
```
lenormandSeb, line 810: Maybe ternary is more readable ? 
```suggestion
              delete p[8] ;
              p[8] = (p[8]) ? Game Constants.Pokerus.Contagious : GameConstant.Pokerus.None;
```

PixLSteam: If you do that you definitely shouldn't delete the value before you read it.

PixLSteam: Either way, I believe we usually avoid using external constants or enums inside update scripts because they might change in the future.

jk13pclick: Made it ternary and removed constants, thanks! 👍

src/scripts/Update.ts
jk13pclick: Done

src/scripts/party/Party.ts
```diff
@@ -174,7 +174,7 @@ class Party implements Feature {
 
     public calculateEVAttackBonus(pokemon: PartyPokemon): number {
         const EVs = this.getEffortValues(pokemon)();
-        if (!pokemon.pokerus) {
+        if (pokemon.pokerus < GameConstants.Pokerus.Contagious) {
```
RedSparr0w, line 177: I'm not sure I get this part, shouldn't they be able to get EVs as soon as they are infected, even if not contagious?

jk13pclick: We could make it that way. I don't think it matters very much. A Pokémon is ``Infected`` only for a short period of time before they hatch, when they then become contagious. You could theoretically catch that Pokémon in that time when it is in the hatchery and ``Infected`` (but not ``Contagious``), but I figured it was better this way.

RedSparr0w: 👍 Ah I see, works fine this way, I was just mistaken in how it all worked 😅

Jaaslet (changes requested):
I have played around with it, and it seems to work as described (mostly)! Got some notes:
- "Cured" is not implemented yet.
- I moved the icon in the breeding display a bit, so it looks better.
- You are not consistent in using `>=` and `>` when writing `.pokerus >= GameConstants.Pokerus.Contagious)`. I think `>=` is more readable, else you need to remember what comes after in the enum, to understand what unlocks the stuff the code is handling.

src/scripts/party/evolutions/Evolution.ts
```diff
@@ -35,8 +35,12 @@ abstract class Evolution {
 
         const shiny = PokemonFactory.generateShiny(GameConstants.SHINY_CHANCE_STONE);
         App.game.party.gainPokemonById(PokemonHelper.getPokemonByName(evolvedPokemon).id, shiny, true);
-        const EPYield = (shiny ? GameConstants.STONE_EP_YIELD * 2 : GameConstants.STONE_EP_YIELD);
-        GameHelper.incrementObservable(App.game.statistics.effortPoints[PokemonHelper.getPokemonByName(evolvedPokemon).id], EPYield);
+
+        if (App.game.party.getPokemon(PokemonHelper.getPokemonByName(evolvedPokemon).id).pokerus >= GameConstants.Pokerus.Contagious) {
+            const EPYield = (shiny ? GameConstants.STONE_EP_YIELD * 2 : GameConstants.STONE_EP_YIELD);
```
Nydaleclya, line 40:
```suggestion
            const EPYield = (shiny ? GameConstans.SHINY_EP_YIELD : 1) * GameConstants.STONE_EP_YIELD;
```

Jaaslet (approved):
Looks good!
I have fixed the update function, so it also checks for cured.
I have done some small tests, and everything seems like they work as they should. But i will be moving EVs from statistics to PartyPokemon before next patch, so i am going to make a more in dept test there anyways.
Thanks for the PR!

PR #2538, opened 2022-06-28
Added shiny dex required challenge

Jaaslet:
I have added shiny dex required challenge.
Yes, it's stupidly hard. No, i don't expect most challenge runners to want to use it, but it's easy and fun to add, and some people will like it.
I am also splitting up challenges into categories, to make it easier for people to talk about. Right now, it's "recommended", "hard" and "nightmare".


src/modules/Credits.ts
```diff
@@ -220,6 +220,7 @@ export const SpriteCredits: Credit[] = [
         name: 'NiCeDiCe',
         link: 'https://github.com/NiCeDiCe90',
         resources: [
+            'Require Complete Shiny Pokedex Challenge Ribbin',
```
Aegyo, line 223:
```suggestion
            'Require Complete Shiny Pokedex Challenge Ribbon',
```

Jaaslet: Thanks!

src/scripts/worldmap/MapHelper.ts
```diff
@@ -218,8 +218,11 @@ class MapHelper {
         }
 
         const challengeActive = App.game.challenges.list.requireCompletePokedex.active();
+        const shinyChallengeActive = App.game.challenges.list.requireCompleteShinyPokedex.active();
         const nextStartingTownUnlocked = TownList[GameConstants.StartingTowns[player.highestRegion() + 1]]?.isUnlocked() ?? false;
-        const fullDex = AchievementHandler.findByName(`${GameConstants.camelCaseToString(GameConstants.Region[player.highestRegion()])} Master`).isCompleted();
+        const fullDex = AchievementHandler.findByName(
+            `${GameConstants.camelCaseToString(GameConstants.Region[player.highestRegion()])} ${(shinyChallengeActive ? 'Shiny' : '')} Master`
```
Aegyo, line 224: This and the ProfNPC one are broken, because `shinyChallengeActive = false` means we have two spaces in this string where we had one before, so the achievement can't be found.

Causes the game to crash for me with this when loading:
```
Uncaught (in promise) TypeError: Unable to process binding "css: function(){return { flash:MapHelper.ableToTravel()} }"
Message: AchievementHandler.findByName(...) is undefined
```

Jaaslet: Yeah, that was not smart. I have fixed it now, so now you can talk with Oak again

Jaaslet:
I will close this PR. We have discussed it in the dev chats, and we don't think it will fit the game well. Some reasons:
- Few people will use it
- Some people will get annoyed that AC is stupidly hard
- Hunting for shinies will be the only challenge, as you are super overbreed, when you reach Johto

PR #2539, opened 2022-06-28
Real evolution challenge

Jaaslet:
Don't you think it's weird that you evolve your Machop to a Machoke but for some reason, you still own a Machop? Boy do i have a challenge for you!
Each time you evolve a pokemon, you lose the base form. Makes the start more fun! You get to catch to many new pokemons!

The plan is to transfer stuff like shiny-status, EVs and breeding on evolution.


src/modules/challenges/Challenges.ts
```diff
@@ -16,6 +16,7 @@ export default class Challenges implements Saveable {
         disableGems: new Challenge('No Gem', 'Disables the usage of Gems for increasing damage multipliers'),
         disableProteins: new Challenge('No Protein', 'Disables the usage of Proteins'),
         slowEVs: new Challenge('Slow EVs', 'Gain EVs 10x slower'),
+        realEvolutions: new Challenge('Real evolutions', 'Your Pokémons go away, when they evolve'),
```
RedSparr0w, line 19: Pokémon is also the plural of Pokémon.
```suggestion
        realEvolutions: new Challenge('Real evolutions', 'Your Pokémon go away, when they evolve'),
```

src/scripts/party/Party.ts
```diff
@@ -79,6 +79,10 @@ class Party implements Feature {
         this._caughtPokemon.push(pokemon);
     }
 
+    public removePokemon(name: PokemonNameType) {
+        this._caughtPokemon.remove(p => p.name == name);
+    }
+
```
RedSparr0w, line 85: Probably best to keep the method naming similar to the gain Pokémon methods:
```
removePokemonById(id: number)
removePokemon(pokemon: PartyPokemon)
removePokemonByName(name: PokemonNameType)
```

RedSparr0w (commented):
Cool idea!
Just a couple minor comments.

Jaaslet:
It should work now. I have even thrown the ribbon through tinypng.com (it's magic. I have no idea how it works). Ready for review :)

src/scripts/party/evolutions/Evolution.ts
```diff
@@ -41,8 +42,22 @@ abstract class Evolution {
...
+            evolvedPartyPokemon.effortPoints = basePartyPokemon.effortPoints;
+            evolvedPartyPokemon.pokerus = basePartyPokemon.pokerus;
+            evolvedPartyPokemon.shiny = evolvedPartyPokemon.shiny || basePartyPokemon.shiny;
+            evolvedPartyPokemon.attackBonusAmount = basePartyPokemon.attackBonusAmount;
+            evolvedPartyPokemon.attackBonusPercent = basePartyPokemon.attackBonusPercent;
+            evolvedPartyPokemon.proteinsUsed = basePartyPokemon.proteinsUsed;
+            App.game.party.removePokemonByName(this.basePokemon);
+            const basePokemonStatistics = App.game.statistics.pokemonCaptured[PokemonHelper.getPokemonByName(this.basePokemon).id];
+            App.game.statistics.pokemonCaptured[evolvedPartyPokemon.id](basePokemonStatistics());
+            basePokemonStatistics(0);
```
Aegyo, line 57: Why do the capture statistics move? Also these don't move the gender specific statistics so you can have Total: 0, male: 1 on the basePokemon

src/scripts/party/evolutions/Evolution.ts
```diff
@@ -41,8 +42,22 @@ abstract class Evolution {
...
-        // EVs
         const evolvedPartyPokemon = App.game.party.getPokemonByName(evolvedPokemon);
+        if (newPokemon && App.game.challenges.list.realEvolutions.active()) {
+            const basePartyPokemon = App.game.party.getPokemon(PokemonHelper.getPokemonByName(this.basePokemon).id);
+            evolvedPartyPokemon.effortPoints = basePartyPokemon.effortPoints;
+            evolvedPartyPokemon.pokerus = basePartyPokemon.pokerus;
+            evolvedPartyPokemon.shiny = evolvedPartyPokemon.shiny || basePartyPokemon.shiny;
+            evolvedPartyPokemon.attackBonusAmount = basePartyPokemon.attackBonusAmount;
+            evolvedPartyPokemon.attackBonusPercent = basePartyPokemon.attackBonusPercent;
+            evolvedPartyPokemon.proteinsUsed = basePartyPokemon.proteinsUsed;
```
Aegyo, line 53: Makes me nervous we may miss something if a new property gets added, but I'm not sure of a better solution.
RIP basePartyPokemon.heldItem

Aegyo (changes requested):
Getting a game crash when Rattata evolves:
1. Rattata has two level evolutions: a proper one at 20, and a fake one for Valencia at 20000
2. When Rattata levels up to 20, we look through its level evolutions and for each one attempt to evolve it, if the evo is satisfied. Part of that satisfied check tries to get the pokemon from the party and check the level.
3. Rattata's first evo passes, we gain Raticate and lose Rattata
4. Then it continues onto the Valencia evo, since we were in the middle of that loop, tries to check if it is satisfied, but we no longer have a Rattata to check and the game crashes :(
Possible solution: Evolution.isSatisfied first checks if we even have the basePokemon at all, before going on to do everything else

src/scripts/party/evolutions/Evolution.ts
Jaaslet: Now moving held items. I have added a comment in party, so coders are aware of it

src/scripts/party/evolutions/Evolution.ts
Jaaslet: I really can't remember... There was something that used the statistics to see if the pokemon was caught, but that has been removed since, it looks like. At least i can't find it in the code.
I will revert it and hope for the best. should be easy to clean up, if the players find out if i missed something

Jaaslet:
I fixed stuff!

Aegyo (approved):

EatPant2nd:
I have a question with this...
What happens with people that have Dex Off, get the starter, evolve it, never get the pre-evos and then make their way to Sinnoh? Will they skip Pokérus entirely? I assume they will, no? That doesn't sound correct and I understand that it's a very specific scenario but I can see people only wanting to keep the last stage evolutions for X reasons.

Jaaslet:
> I have a question with this...
> What happens with people that have Dex Off, get the starter, evolve it, never get the pre-evos and then make their way to Sinnoh? Will they skip Pokérus entirely? I assume they will, no? That doesn't sound correct and I understand that it's a very specific scenario but I can see people only wanting to keep the last stage evolutions for X reasons.

It might crash the game... Can you please not try this?

PR #2541, opened 2022-06-28
Fix protein searchbar on reopen

Jaaslet:
- Open the protein window
- Write something in the searchbar
- Close the window
- Open it again
- The searchbar is empty, but the filter still applies
There is an if around the modal, so everything is reset, each time it closes down. In this PR, the value is now reapplied on reopen.

### DIFF
--- a/src/components/pokemonSelectorModal.html
+++ b/src/components/pokemonSelectorModal.html
@@ -62,8 +62,7 @@
             <div class="form-group col-md-6 col-6 p-0">
                 <label for="protein-search">Search</label>
                 <input id="protein-search" autocomplete="off" class="form-control" placeholder="search here"
-                    oninput="ProteinFilters.search.value(this.value)"
-                    data-bind="value: ProteinFilters.search.value()"
+                    data-bind="textInput: ProteinFilters.search.value"
                 />
             </div>
             <div class="form-group col-md-3 col-3 p-0">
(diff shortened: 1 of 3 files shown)


src/components/pokemonSelectorModal.html
```diff
@@ -61,7 +61,10 @@ <h5 class="modal-title text-light">Select a Pokémon</h5>&nbsp;
           <div class="form-row m-0" id="protein-filter">
             <div class="form-group col-md-6 col-6 p-0">
                 <label for="protein-search">Search</label>
-                <input id="protein-search" autocomplete="off" class="form-control" oninput="ProteinFilters.search.value(new RegExp(`(${this.value})`, 'i'))" placeholder="search here"/>
+                <input id="protein-search" autocomplete="off" class="form-control" placeholder="search here"
+                    oninput="ProteinFilters.search.value(this.value)"
+                    data-bind="value: ProteinFilters.search.value()"
+                />
```
Aegyo, line 67: I'm pretty sure you can do this with the `textInput` binding
```suggestion
                <input id="protein-search" autocomplete="off" class="form-control" placeholder="search here"
                    data-bind="textInput: ProteinFilters.search.value()"
                />
```

Jaaslet: I committed your suggestion, and it still works! And it looks cleaner :)

Aegyo (approved):

PR #2542, opened 2022-06-28
Make pokedex 'Hide shiny image' display the underlying observable state

Aegyo:
Ties the checkbox to the state of the observable, instead of updating the observable on change.
Should prevent any weirdness with it getting out of sync and doing the opposite of what it claims (there have been a few reports in discord)


RedSparr0w (approved):

PR #2543, opened 2022-06-29
Fix Battle Items on small screens

### LINKED ISSUE #2520: [BUG] Unable to use the dungeon token on mobile
**Info**
 - OS: [Android 11]
 - Browser [Chrome mobile]
 - Version  `v.0.9.6`]

**Description:**
< The Dungeon Collector token item is unclickable -->

**Steps to Reproduce:**

1. Launch game
2. Click on Dungeon Token Collector
3. See that you're unable to click it

**Expected Behavior:**
We can use it as in previous versions

CorgiOnNeptune:
Just changes the tooltip placement back to bottom for screens under 768px width. Should fix any issues mobile players are having with using the Token Collector and Red Flute.
Closes #2520

### DIFF
--- a/src/components/battleItemContainer.html
+++ b/src/components/battleItemContainer.html
@@ -40,7 +40,7 @@
                             tooltip: {
                               title: ItemList[$data].displayName + '<br/>' + ItemList[$data].description,
                               trigger: 'hover',
-                              placement:'left',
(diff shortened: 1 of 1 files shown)


Jaaslet (approved):
It fixes the problem, and the solution looks fine.
Thanks for the PR

PR #2545, opened 2022-06-29
No longer giving out infinit free fighting eggs

Jaaslet:
Dungeon clear rewards was given on each clear, which does not work well, when we are trying to give a fighting egg on the first clear.
The mt. moon dungeon had it's own check in the reward function, but i think it's cleaner, if we just have the check for all rewards. Right now, it's fighting eggs, fossils and gym badges, all which should only be given on first clear.

The argument against this solution, is that we might want to give an reward for grinding a dungeon, in the future. But i think that we always want dungeon loot for that. So if the clear reward can give more than one reward, code contributors might think that we think it's a good idea.
And if we find a good use for it, it's a 2 liner change.

### DIFF
--- a/src/scripts/dungeons/DungeonRunner.ts
+++ b/src/scripts/dungeons/DungeonRunner.ts
@@ -222,10 +222,11 @@ class DungeonRunner {
     public static dungeonWon() {
         if (!DungeonRunner.dungeonFinished()) {
             DungeonRunner.dungeonFinished(true);
+            if (!App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]()) {
+                DungeonRunner.dungeon.rewardFunction();
+            }
             GameHelper.incrementObservable(App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]);
             MapHelper.moveToTown(DungeonRunner.dungeon.name);
-            DungeonRunner.dungeon.rewardFunction();
-            // TODO award loot with a special screen
             Notifier.notify({
                 message: 'You have successfully completed the dungeon',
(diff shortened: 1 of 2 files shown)


src/scripts/dungeons/DungeonRunner.ts
```diff
@@ -222,10 +222,11 @@ class DungeonRunner {
     public static dungeonWon() {
         if (!DungeonRunner.dungeonFinished()) {
             DungeonRunner.dungeonFinished(true);
+            if (!App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]()) {
+                DungeonRunner.dungeon.rewardFunction();
+            }
             GameHelper.incrementObservable(App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]);
             MapHelper.moveToTown(DungeonRunner.dungeon.name);
-            DungeonRunner.dungeon.rewardFunction();
-            // TODO award loot with a special screen
```
Ultima1990, line 228: Maybe keep the TODO? I can see us maybe wanting to still do that at some point.

Jaaslet: Then it should be an issue imo. The code is not the correct place for suggestions :)

Aegyo (approved):

RedSparr0w:
Looks good to me 👍 
It's probably a good idea to update any other dungeons which also check for first clear in this PR too?

Jaaslet:
> Looks good to me 👍 It's probably a good idea to update any other dungeons which also check for first clear in this PR too?

I think mt. Moon was the only one? I could at least not find any other

RedSparr0w:
Must be thinking of gyms 😅

PR #2546, opened 2022-06-29
Add Meltan to the game

Alekemon:
Meltan: Roaming encounter in Kanto when Alola is beaten
Melmetal: Catch 400 Meltan quest


amative1:
We already have a Meltan PR in progress, with more of a questline (that was discussed and planned out). See #2075.

PR #2547, opened 2022-06-30
Prevent directly sorting caughtPokemon

Aegyo:
... and fix the hatchery helper code which was doing it.

Sorting the caughtPokemon list directly causes lag because so many parts of the ui are subscribed to it.


Aegyo:
The ReadOnlyArray type is probably better for this on second thought. I didn't use it since we still want to push new pokemon in, but that should all be happening on the observableArray _caughtPokemon instead anyway.

RedSparr0w:
Good point, should the `setter` also be removed?

Aegyo:
Probably, I can't think of any reason to keep it

PR #2548, opened 2022-06-30
eslint dot notation

RedSparr0w:
Update eslint rules for dot notation.


Jaaslet (approved):
It works and fixes our problem with weird uses of Enums. Let's try it, and see how fun it will be to work with!
Thanks for the PR

PR #2550, opened 2022-06-30
eslint rule no-sparse-arrays

RedSparr0w:
This should stop us adding extra `,` within arrays by accident.

For example shops, even though we have the typing of `Item` it still allows empty items in the array, which broke the shops in `0.9.6`.


Jaaslet (approved):
It works! Thanks for the PR

PR #2552, opened 2022-06-30
Add player sprite to current location

Dragonchitos:
The player sprite from profile indicates the current position. It works for routes, town, dungeon and overlay related to dungeon (like distortion world). 
The green background that indicated the position is removed, so we can see the state of the current location (route or dungeon) more easily.
I didn't test for Alola because I don't have a save in Alola yet.


src/templates/mapOverlay.ejs
```diff
@@ -25,6 +25,9 @@ Parameters:
     - visible : Required : String
         A knockout data-bind for visibility when true
 
+    - moveToTown : Optional : String
```
Jaaslet, line 28: Not really a "moveToTown". That kinda indicates a clickaction. Let's call it something like "connectedToTown" or something

src/templates/mapDungeon.ejs
```diff
@@ -15,14 +15,26 @@ Parameters:
     - y : Required : Number
         y axis position of the dungeon icon.
 
+    - notMain : Optional : Boolean
```
Jaaslet, line 18: Let's call it "isExtension" to match our comments in the SVGs

src/components/AlolaSVG.html
```diff
@@ -1165,7 +1164,7 @@
         <%- include('templates/mapTown',
             { name: "Iki Town"
             , x: 31, y: 16
-            , width: 1, height: 7
+            , width: 1, height: 7, notMain: true
```
Jaaslet, line 1167: Let's move all the notMain to a new line. You are renaming them anyways :)

Jaaslet (changes requested):
I LOVE IT! It looks cool. But i still have some comments for you :)
- Some places, i am seeing double
  - MeleMele sea
  - Relic Passage
  - Dark cave
  - Diglett's Cave
- Mega dungeons (like Mt. Chimney) does not show the icon, when you enter it

src/styles/variables.less
```diff
@@ -11,9 +11,8 @@
 
 // Map
 @locked: #000000;
-@currentPlace: #55ff00;
 @incomplete: #ff9100;
-@questAtLocation: #34BF45;
+@questAtLocation: #ff0033;
```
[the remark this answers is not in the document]
Dragonchitos: I don't know how to do that. Where should i put the function ?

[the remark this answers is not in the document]
Jaaslet: Update.ts

src/scripts/worldmap/MapHelper.ts
```diff
@@ -114,11 +111,21 @@ class MapHelper {
         return cls;
     }
 
-    public static calculateTownCssClass(townName: string): string {
-        // Check if we are currently at this location
+    public static calculateRouteIsCurrentLocation(route: number, region: GameConstants.Region): boolean {
+        if (player.route() == route && player.region == region) {
```
silverwolf2000, line 115:
```suggestion
        return (player.route() == route && player.region == region)
```

Dragonchitos:
So : 
 - I did all the rename and move things
 - I add the update function
 - I think I fix Melemele sea, not sure for this one
 - I fix "mega dungeon", which are actually town with several dungeons related without own icon

For Relic Passage, Dark cave and Diglett's Cave : they have two separated icons, so I think it's weird to have the player icon only in one. The player may be confused to click on dungeon and not see the player icon appear, but see it in another place. But I know it's also weird to see two player icons, so I don't know which solution is best.

Jaaslet (approved):
I have fixed the update function and changed the quest default color to the old current location color.
The rest looks good. Thanks for the PR!

Jaaslet:
> For Relic Passage, Dark cave and Diglett's Cave : they have two separated icons, so I think it's weird to have the player icon only in one. The player may be confused to click on dungeon and not see the player icon appear, but see it in another place. But I know it's also weird to see two player icons, so I don't know which solution is best.

I think two icons is the best solution, so we are keeping that one. We can always change it later, if someone has a good argument.

PR #2553, opened 2022-06-30
Fix 0.5 harvest berries not dropping 1 on wither

Aegyo:
A change to the harvestAmount code stopped these berries dropping anything on wither.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2555, opened 2022-07-01
Fixup Pokémon location checks

RedSparr0w:
This fixes the location checks.

_Currently only used by the bot AFAIK, possibly in the future will be added to the pokédex or something_

Battle Frontier milestones (rewards handling was changed)
Added Discord Shop encounters
Add Dungeon Chest (mimic) encounters


Jaaslet (approved):
I have tested the function by using the method on different pokemons in the browser console. It works fine there, so i think it should work with the bot.
Thanks for the PR!

PR #2556, opened 2022-07-01
Bind Party Pokémon observables to be specifically a number/boolean

RedSparr0w:
This PR binds the observables used in `PartyPokemon` to specifically be a number or boolean, so they can't become something else.


lenormandSeb (commented):
If i understand `extend` add a default value, for a shiny pokemon, this default is a boolean ? If it that boolean are always `true` or `false`, not null 😊

Aegyo:
You are mistaken about what extend does, it modifies the observable behavior with predefined extender functions.

src/scripts/party/PartyPokemon.ts
```diff
@@ -53,16 +53,16 @@ class PartyPokemon implements Saveable {
...
+        this.proteinsUsed = ko.observable(proteinsUsed).extend({ numeric: 0 });
+        this.effortPoints = ko.observable(effortPoints).extend({ numeric: 0 });
+        this._breeding = ko.observable(breeding).extend({ boolean: null });
+        this._shiny = ko.observable(shiny).extend({ boolean: null });
+        this._level = ko.observable(1).extend({ numeric: 0 });
+        this._attackBonusPercent = ko.observable(attackBonusPercent).extend({ numeric: 0 });
+        this._attackBonusAmount = ko.observable(attackBonusAmount).extend({ numeric: 0 });
+        this._attack = ko.observable(this.calculateAttack()).extend({ numeric: 0 });
+        this._category = ko.observable(category).extend({ numeric: 0 });
+        this._pokerus = ko.observable(pokerus).extend({ boolean: null });
```
Aegyo, line 65: This one is not going to be boolean with the pokerus enum PR, do you want to leave it out of this one?

src/scripts/party/PartyPokemon.ts
```diff
@@ -53,16 +53,16 @@ class PartyPokemon implements Saveable {
...
-        this._category = ko.observable(category);
-        this._pokerus = ko.observable(pokerus);
+        this.proteinsUsed = ko.observable(proteinsUsed).extend({ numeric: 0 });
+        this.effortPoints = ko.observable(effortPoints).extend({ numeric: 0 });
+        this._breeding = ko.observable(breeding).extend({ boolean: null });
+        this._shiny = ko.observable(shiny).extend({ boolean: null });
+        this._level = ko.observable(1).extend({ numeric: 0 });
+        this._attackBonusPercent = ko.observable(attackBonusPercent).extend({ numeric: 0 });
+        this._attackBonusAmount = ko.observable(attackBonusAmount).extend({ numeric: 0 });
+        this._attack = ko.observable(this.calculateAttack()).extend({ numeric: 0 });
```
Aegyo, line 63: This property doesn't look like we update or use it anywhere, unless I failed searching somehow.

Aegyo (approved):

src/scripts/party/PartyPokemon.ts
Jaaslet: I think we still want to keep it, so we remember to change it when we get a merged conflict

src/scripts/party/PartyPokemon.ts
RedSparr0w: Only place It seems to be used is `Party.ts → calculateOnePokemonAttack` which doesn't seem right? maybe?
_i'm probably just misinterpreting what it's doing_
```ts
        const pAttack = useBaseAttack ? pokemon.baseAttack : (ignoreLevel ? pokemon.calculateAttack(ignoreLevel) : (pokemon.attack * (this.calculateEVAttackBonus(pokemon))));
```
Also `Egg.ts → hatch`
```ts
            // Recalculate current attack
            partyPokemon.attack = partyPokemon.calculateAttack();
```
And lastly `BreedingController.ts → getDisplayValue` (and the statistics modal)
```ts
            case 'attack': return `Attack: ${pokemon.attack.toLocaleString('en-US')}`;
```
It looks like we could also use it for the sorting instead of recalculating the attack each time?
```diff
    [SortOptions.attack]: {
        text: 'Attack',
-        getValue: (p) => p.calculateAttack(),
+        getValue: (p) => p.attack,
    },
```

src/scripts/party/PartyPokemon.ts
RedSparr0w: I don't think it will create a conflict, because nothing has changed on this line for the other PR.
I'll remove it and create a commit on the other PR to update it.
```suggestion
        this._pokerus = ko.observable(pokerus);
```

src/scripts/party/PartyPokemon.ts
Aegyo: Ah I think I tried searching with a bracket because it is an observable, forgetting we have getters to avoid doing that...

PR #2559, opened 2022-07-01
Update Route Health to stay within variance

Nydaleclya:
With the current formula for Route Health more than 16% of Routes have at least one Pokemon which does not stay within the +/- 10% variance.
In some cases leading to a variance of 25%.
image

This PR tries to keep every Pokemon within this +/- 10% variance as far as rounding to Integers allows.

For more numbers check [this spreadsheet](https://docs.google.com/spreadsheets/d/1EXYK5f51Lsqkpe92OC8Hk_fBU6zI9DGcf1wN9x3tLJ0/edit?usp=sharing)

### DIFF
--- a/src/scripts/pokemons/PokemonFactory.ts
+++ b/src/scripts/pokemons/PokemonFactory.ts
@@ -23,15 +23,18 @@ class PokemonFactory {
         }
         const basePokemon = PokemonHelper.getPokemonByName(name);
         const id = basePokemon.id;
-        const routeAvgHp = (region, route) => {
+        const routeHpModifier = (region, route, pokemonHP = basePokemon.hitpoints) => {
             const poke = [...new Set(Object.values(Routes.getRoute(region, route).pokemon).flat().map(p => p.pokemon ?? p).flat())];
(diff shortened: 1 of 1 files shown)


Ultima1990:
Remind me, do we have a stated goal of 10% variance anywhere?

As far as I know the 10% refers to the amount of health that gets removed, modified, and put back.

Nydaleclya:
In the current formula you remove 10% and then add `basePokemon.hitpoints / routeAvgHp(region, route) * 10%` back on.
There are cases where `basePokemon.hitpoints / routeAvgHp(region, route) * 10%` adds a total of 35% back on.
Which I think is not the way it was intended according to the comment above the formula.

Ultima1990:
I don't know exactly what the comment is about. Maybe it just needs to be rewritten.
I do know what the intent of the formula is. Which is to simulate the main series base stat HP differences between mons while keeping routehealth scaling mostly intact. Some mons have significantly more health than others. I'm willing to bet that the outliers on 209 and 210 are Chansey, which just happens to be a mon that is known in the main series for its absurdly large health pool (tho not as large as Blissey of course). IMO having that represented in Clicker is kinda neat.

PR #2563, opened 2022-07-02
Added pokerus filter to breeding window

### LINKED ISSUE #2506: [SUGGESTION]  Add pokérus as a filter in hatchery
**Current Behavior**
It's hard to find which pokemon don't have pokérus.

**Improved Behavior**
Make it possible to filter pokemon by if they have pokérus or not.

**Reason for change**
It would make it easier to make sure all your pokemon have pokérus.

Jaaslet:
image

I did not add infected to the filter, as it should never show up in the hatchery.
I made the text search smaller, to not add a new line to the filter. The filter was getting too big. If we never want to add more filters, we might need to rethink how the filters should look... And then we can make the box bigger again.

Closes #2506


src/components/breedingModal.html
```diff
@@ -94,6 +94,14 @@ <h5 class="modal-title" id="breedingModalLabel">Day Care</h5>
                                     <!-- /ko -->
                                 </select>
                             </div>
+                            <div class="form-group col-md-3 col-6">
+                                <label>Pokerus Status</label>
```
Aegyo, line 98:
```suggestion
                                <label>Pokérus Status</label>
```

Aegyo (changes requested):
Minor text gripe, otherwise perfectly fine :+1:

Aegyo:
Do we want a Pokérus sort option too?

Jaaslet:
> Do we want a Pokérus sort option too?

No idea. Personally, i can't see a use case, when people have the filter, so my plan is to leave it out, until someone gives me a good reason

Qwertypop04:
> No idea. Personally, i can't see a use case, when people have the filter, so my plan is to leave it out, until someone gives me a good reason

Would a sort for number of EVs be more useful, maybe?

Jaaslet:
> Would a sort for number of EVs be more useful, maybe?

Yeah, that would make more sense. Yeah, i will add it tomorrow. A sort option does not take up any UI space, so might as well

CorgiOnNeptune:
A Display Value for EVs would also be a useful addition.

Jaaslet:
I have added sorting option and display now. So it should be ready for review again.

Aegyo (approved):
#2574 will need a change updating how evs are calculated for the display and sort option, but everything seems to work here :+1:

Jaaslet:
> #2574 will need a change updating how evs are calculated for the display and sort option, but everything seems to work here 👍

I will then merge this one, so i can fix it in the other PR

PR #2564, opened 2022-07-02
Fix loading pre 0.9.6 saves

Aegyo:
Trying to load a save before effortPoints were added currently fails.


src/scripts/Update.ts
```diff
@@ -800,13 +800,13 @@ class Update implements Saveable {
         },
 
         '0.9.8': ({ playerData, saveData, settingsData }) => {
-            saveData.oakItemLoadouts = saveData.oakItemLoadouts.map((list, index) => ({ name: `Loadout ${index + 1}`, loadout: list }));
+            saveData.oakItemLoadouts = saveData.oakItemLoadouts?.map((list, index) => ({ name: `Loadout ${index + 1}`, loadout: list }));
```
RedSparr0w, line 803: Should we add a default value? so `oakItemLoadouts` exist in any future update functions after this that need to modify it too?
```suggestion
            saveData.oakItemLoadouts = saveData.oakItemLoadouts?.map((list, index) => ({ name: `Loadout ${index + 1}`, loadout: list })) || [];
```

src/scripts/Update.ts
```diff
@@ -800,13 +800,13 @@ class Update implements Saveable {
...
 
             // Fix pokerus
             saveData.party.caughtPokemon.forEach(p => {
                 let status = (p[8]) ? 2 : 0;
                 const requiredForCured = saveData.challenges.list.slowEVs ? 5000 : 500;
-                if (saveData.statistics.effortPoints[p.id] >= requiredForCured) {
+                if (saveData.statistics.effortPoints?.[p.id] >= requiredForCured) {
```
RedSparr0w, line 809: And same thing for this? (although not sure if it's a `{}` or `[]`)
```suggestion
                saveData.statistics.effortPoints = saveData.statistics.effortPoints || {};
                if (saveData.statistics.effortPoints.[p.id] >= requiredForCured) {
```

Aegyo: I don't think a default is needed here since we are moving effortpoints onto the pokemon anyway

RedSparr0w (approved):

PR #2565, opened 2022-07-02
Fix pokerus display for uncaught pokemon eggs

Aegyo:
`visible` binding on these shop/fossil egg pokerus image doesn't prevent knockout from evaluating the html/code, resulting in console errors when we try to access properties on undefined.
Changed to `if` binding, and replaced some `party.getPokemon` with just accessing the partyPokemon from the egg since it should already be there.


RedSparr0w (approved):

PR #2566, opened 2022-07-02
More efficient party pokemon lookup

Aegyo:
This one doesn't have a noticeable performance improvement, just something in the code that was bothering me.


RedSparr0w (approved):

PR #2567, opened 2022-07-02
Rewrite Lapras NPC

Ultima1990:
This rewrites the text of the Lapras NPC in Azalea.
Also it renames the hiker to one that canonically appears in Union Cave (and is a boss in our version).
Don't know why he never heard it himself. Maybe he just doesn't go that deep. Gotta stay out of those optional areas. Stay near the surface so that he can be a boss for unsuspecting players :)


RedSparr0w (approved):

PR #2568, opened 2022-07-02
More comma separators

CorgiOnNeptune:
Everybody gets a comma.


Jaaslet (changes requested):
You have added 3 `GameConstants.camelCaseToString()` that i don't really see a use for. All the underground items names works fine (as far as i can see).
But other than that, it looks good (numbers easier to ready! yay!).

CorgiOnNeptune:
Those just "fix" capitalization for the 'egg' items in Day Care. And the gem values for plates in underground, but that one doesn't really matter either way I guess.

Jaaslet (approved):
Oh, i did not notice that "egg" was not capitalized in the hatchery. Then it's a good change.
Thanks for the PR!

PR #2569, opened 2022-07-02
Add Cyrus battle in Celestic Town

Ultima1990:
Adds a Cyrus battle to Celestic Town. Unlocks after clearing Pastoria Gym and is now required to proceed to routes 218 and 219. Now Celestic Town doesn't just pointlessly unlock and do nothing after clearing Pastoria! Also this would function as a more logical point to unlock Surf, which might just be useful later.

Also added this Cyrus battle to the Galactic questline. Might need update function?


Jaaslet (approved):
I tested with a save, which was halfway through the questline, and the behavior seems to match the one people with a completed questline gets. The questline is pushed back one, but autocomplets again, because you have already done the thing.
No comments other than that. Thanks for the PR!

PR #2572, opened 2022-07-03
Player icon on map ignore mouse

RedSparr0w:
This PR makes it so that the player icon ignores mouse events.
This way when you hover the player icon it will still show the towns name etc.
image


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2574, opened 2022-07-03
Move EVs from statistics

Jaaslet:
EVs should be saved on PartyPokemon and not in statistics. This should also fix the problem people have with EVs multiplier not showing up everywhere we want it to


Aegyo (changes requested):
Couple of comments, works fine otherwise from what I can see :+1:

src/scripts/Update.ts
```diff
@@ -842,6 +842,11 @@ class Update implements Saveable {
                 delete settingsData['notification.dungeon_item_found'];
                 delete settingsData['notification.dungeon_item_found.desktop'];
             }
+
+            // Moved EVs from statistics
+            saveData.party.caughtPokemon.forEach(p => {
+                p[9] = saveData.statistics.effortPoints[p.id];
```
Aegyo, line 848: For saves before effortPoints was added:
```suggestion
                p[9] = saveData.statistics.effortPoints?.[p.id] || 0;
```

Jaaslet:
I have merged this with the filters and fixed it. I have resolved all your comments. It's ready for review again :)

Aegyo (approved):

PR #2576, opened 2022-07-03
Add regional debuff setting to hatchery modal

pjeanjean:
This PR adds a menu to the Hatchery modal to enable/disable regional attack debuff.
After discussing it on the Discord server, I separated it from the Filters and Sort menus, and moved Display Value alongside it since it was technically not a filter.
Right now the new menu is called "Settings", but the name is not final.
image

Setting a region leads to a few changes in the modal:
  - Some values are affected by the debuff: Attack, Attack Bonus, Breeding Efficiency and Steps per Attack Bonus
  - Sorting with either Attack or Breeding Efficiency takes into account the debuff (the order should be the same as the Protein wiki page)
  - Helpers using either Attack or Breeding Efficiency for sorting are impacted

It should be mergeable without any issue, but I'm still open to suggestions and/or improvements.

image


src/scripts/breeding/BreedingController.ts
```diff
@@ -203,16 +203,26 @@ class BreedingController {
...
         }
     }
+
+    // Applied regional debuff
+    public static regionalAttackDebuff = ko.observable(-1);
+
+    public static calculateRegionalMultiplier(pokemon: PartyPokemon): number {
+        if (BreedingController.regionalAttackDebuff() > -1 && PokemonHelper.calcNativeRegion(pokemon.name) !== BreedingController.regionalAttackDebuff()) {
+            return App.game.party.getRegionAttackMultiplier();
+        }
```
Jaaslet, line 225: I think this method is sharing some code with App.game.party.getRegionAttackMultiplier. At least i think we can just call that directly, instead of calling this method

Jaaslet (changes requested):
It looks good and works!
We have discussed it a bit in the developer group, and we think sort should be a part of the settings. Then we can also make sort smaller, to fit it on one line (more room for pokemons).
I also think Filters should be on top.

src/scripts/breeding/BreedingController.ts
pjeanjean: I do need to filter Pokemon based on the selected region.
The difference here is that `getRegionAttackMultiplier` takes only the highest region as parameter, but not the current one.

So, let's say my highest reached region is Kalos, and I want to look at the Breeding Efficiency for the Kanto region.
Gyarados should be displayed without any debuff, while every mon from Johto -> Kalos should have the debuff applied.
That's something I can't do currently with only `getRegionAttackMultiplier`, but maybe I could improve it to take an optional mon as a parameter.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2577, opened 2022-07-03
Bigger Sinnoh map

Ultima1990:
A Sinnoh map that is twice as wide and twice as high. So 4x the size. And you know what they say about 4x the size. Right, Todd Howard?
This is a thing because it allows us to create a more detailed map, and also because it simply gives us more space on the west side of Coronet, which is currently very cramped.

Alongside this I'd like to have a feature that allows the player to zoom in on this map and scroll/click-drag the map around inside the Town Map modal. This thing doesn't really work without that.

This is a proof of concept. If this works we can actually have all of Alola on one map!

TODO:
- [ ] More detailing and decorating
- [ ] Zoom and scroll/drag function

**NOTE**: The big map is currently made fully visible in the town map by changing the map viewBox size in index.html. _This affects all maps!_ This is a temporary thing done to be able to see the entire map and test.


Ultima1990:
Gonna need a better way to do the player icon on map thing for Distortion world. Currently it just appears in both places.

PR #2583, opened 2022-07-04
Gym pokemon requirement

Jaaslet:
Now gym pokemons can have requirements. This means gyms, temp battles and dungeon trainers can all set a team, based on some requirements.
Also added StarterRequirement.
Also turned the Champion class into a gymFlag, now that we don't need that logic anymore


src/modules/requirements/StarterRequirement.ts
```diff
@@ -0,0 +1,19 @@
...
+        this.region = region;
+    }
+
+    public getProgress() {
+        return player.regionStarters[this.region]();
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    public hint(): string {
+        return 'Wrong starter.';
```
RedSparr0w, line 17: Could we change this to something like `Requires ${starter} to be chosen as your ${region} starter Pokémon`.
I know it won't likely be shown anywhere, but may end up being used by the bot for displaying some requirements later.

src/scripts/gym/GymList.ts
```diff
@@ -195,36 +195,32 @@ GymList['Elite Lance'] = new Gym(
...
 // Kanto Champion
-GymList['Champion Blue'] = new Champion(
+GymList['Champion Blue'] = new Gym(
     'Blue',
     'Champion Blue',
     [
         new GymPokemon('Pidgeot', 52340, 59),
         new GymPokemon('Alakazam', 56320, 57),
         new GymPokemon('Rhydon', 58340, 59),
+        new GymPokemon('Exeggutor', 57520, 59, new StarterRequirement(GameConstants.Region.kanto, 0)),
```
RedSparr0w, line 205: If we are going to be using these around, I think we should probably create something in `GameConstants` for each of the starters rather than just having magic numbers.

RedSparr0w: Even if we just end up having something like:
```ts
declare enum starters {
  none = -1,
  grass = 0,
  fire = 1,
  water = 2,
  special = 3,
}
```

src/scripts/trainers/Trainer.ts
```diff
@@ -2,19 +2,25 @@
...
 
     constructor(
         public trainerClass: string,
-        public team: GymPokemon[],
+        team: GymPokemon[],
         name?: string,
         public subTrainerClass?: string
     ) {
         this.name = name ? `${trainerClass} ${name}` : trainerClass;
+        this.team = team;
```
RedSparr0w, line 14: Does this part need to be changed?
As far as I can tell, it seems to be doing the same thing in long form?

Jaaslet: I thought local parameters defaulted to private (just looked it up, it does not). But you can do the same public-in-constructor trick, just with private. I think i have made this mistake a couple of places... Oh well, i learned something now, and i will fix the others, if i come by them.
But yeah i want it to be private, because i don't think anyone should ever use the team, without it being filtered by requirements first

src/modules/requirements/StarterRequirement.ts
Aegyo: Marked resolved but doesn't look like there were changes to resolve it?

src/scripts/temporaryBattle/TemporaryBattle.ts
```diff
@@ -66,4 +66,8 @@ class TemporaryBattle extends TownContent {
         }
         this.completeRequirements = completeRequirements;
     }
+
+    public getPokemonList() {
```
Aegyo, line 70: Same reasoning as with gym pokemons
```suggestion
    get pokemons(): GymPokemon[] {
```

src/modules/requirements/StarterRequirement.ts
Jaaslet: I don't know what happened. Stuff goes wrong when you spend more than 4 months on a PR. But should be fixed now

src/scripts/trainers/Trainer.ts
```diff
@@ -17,4 +16,8 @@ class Trainer {
         return `assets/images/trainers/${imageName}.png`;
     }
 
+    public getTeam() {
```
Aegyo, line 19: Same reasoning as gym pokemons
```suggestion
    get team(): GymPokemon[] {
```

Aegyo (changes requested):
A MissingNo bug + a couple of style suggestions

Aegyo (approved):

PR #2587, opened 2022-07-05
[Quests] Add region name for uncommon route names

### LINKED ISSUE #2584: [BUG] Undella bay quest missing region
**Info**
 - OS: Ubuntu 22.04
 - Browser: Chrome v103.0.5060.53 64bit
 - Version: v0.9.7

**Description:**
The `Defeat N Pokémon on Undella Bay.` quest is missing the region.

**Steps to Reproduce:**
1. Open Quest list

**Expected Behavior:**
Quests on a certain area usually display which region that is in, this on seems to be missing it (Unova).

**Screenshots:**
image

**Additional Information:**

pjeanjean:
Some routes don't include the name of the region, e.g. Desert Resort and Undella Bay.
This resulted in incomplete descriptions for defeat quests on this routes.

Fixes #2584


Jaaslet (changes requested):
I think the code should be moved into Routes.getName().
That method is used 3 places: quests, achievements and in the battle window. I think the first two wants your change, but not the 3rd one. So Routes.getName() needs a parameter, to turn it off for the battle window.
But other than that, i think this change will clear up some confusion for players.

pjeanjean:
Should be good now.

Jaaslet (changes requested):
Turns out achievements does not use the requirement text, but defines it's own test in the constructor. So right now, AchievementHandler.ts uses `route.routeName` instead of `Routes.getName()`. But the change now affects requirements, like when you are clicking on a locked route, so that is cool.
I also changed `includeRegionName` to `alwaysIncludeRegionName`.

pjeanjean:
@Jaaslet I noticed, but I thought you actually meant the requirements.
Do you want me to change the achievements too?

@lenormandSeb I'm not sure how to name the parameter to make sense in this case, `alwaysIncludeRegionName` is understandable, but `dontAlwaysIncludeRegionName` not so much.

Jaaslet:
> Do you want me to change the achievements too?

I believe that if people gets confused by the quests, they will be confused by the achievements too. But i might be wrong

dragonkin550:
I don't think that it's necessary to do it for achievements. We have region filters for those already

Jaaslet:
> Now I'm also considering adding the region name to Gyms and Dungeons achievements, for consistency.
> Wdyt? Too much?

We talked about it in the dev team, and we like the current change, but we might need to change it later. We might want to add stuff like gen 2 Kanto, which will make this confusing. But until then, we like the change.
This means that there is no reason to spend more time on it, unless you want to :)
But if you feel like doing it, we might as well add it there too.

PR #2588, opened 2022-07-05
[Hatchery] Fix category offset in Firefox

pjeanjean:
This fixes a display issue in Firefox where category indicators have a slight vertical offset.
image
image


pjeanjean:
It should be more consistent now.
I slightly modified the alignment of the Shiny indicator too since it was lower than the category indicator.

Also, I just noticed that all the names on Firefox are one pixel lower than they should (also applies to Pokedex).
I'll try to find a fix for that too so don't merge just yet please ;)

pjeanjean:
Ok, forget that, it's practically every text in the UI that has this 1px offset.
Firefox just sucks.

Jaaslet (approved):
Firefox is great! But other than that, it looks good. Thanks for the PR :)

PR #2589, opened 2022-07-05
Fix category dropdown bar colors not updating on delete

### LINKED ISSUE #2571: [BUG] Problem with hatchery categories color
Something with the recent addition of the "Categories" button to the hatchery broke how categories are labelled. I created some groups to help organize how I'm breeding, and when I deleted and re-colored one of the groups (via the categories menu, 1st screenshot) it didn't properly update the drop-down menu for each Pokémon (2nd screenshot).
Screen Shot 2022-07-02 at 5 58 08 PM
Screen Shot 2022-07-02 at 5 58 17 PM

The group I deleted was the first group after "None", and was originally colored green, so it looks like the colors in the 2nd screenshot are maintaining the original colors, rather than updating to match the new groups.

CorgiOnNeptune:
Fixes **part of** #2571 

This fixes the category dropdown bars colors not updating properly when you delete a preceding category. 

The color on the buttons still aren’t updating properly, but somebody else will have to take a look at that as I was unable to find a solution for the buttons.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2590, opened 2022-07-06
Fix bug that prevented catching of new pokemons

Jaaslet:
After my EVs rewrite, the catching of new pokemons sometimes bugged out. I have no idea what triggers it, but i caught like 10 new pokemons, before i saw the bug. But it looks like it should happen all the time...


Ultima1990 (approved):
It works.

PR #2592, opened 2022-07-06
Tutorial overhaul

Ultima1990:
Changes to the tutorial quests.
Now requires talking to 2 NPCs:
Mom, who recommends talking to NPCs. An important lesson for players I think.
Old Man, who explains the ball selector.
Also now requires catching 5 Pidgey. The player now has to prove that they understand the ball selector. RIP monotype.

No more How Blu!


src/scripts/quests/QuestLineHelper.ts
```diff
@@ -23,22 +23,76 @@ class QuestLineHelper {
...
+                        }, {
+                            once: true,
+                        });
+                    });
+                }, {
+                    once: true,
+                });
+            });
+        };
+        const talkToOldMan = new TalkToNPCQuest(ViridianCityOldMan2, 'Talk to the Old Man in Viridian City to learn about catching. Talk to the Old Man again if you need a reminder.', OldManReward);
```
Jaaslet, line 90: "Talk to the Old Man again if you need a reminder."
It seems confusing that it's a part of the quest text. It kinda looks like it's relevant for the quest. And the old man is already saying something like that.

Jaaslet (changes requested):
It looks good. I did not find a way to break the tutorial. I just have one small comment.

xslk:
What if, to save monotype challenges, you change the quest to a 2 part. 

Instead of catch 5 Pidgeys he just ask to catch 5 pokemon. You return to him and then he says something like "did you noticed your dungeon tokens increasing after each catch? Oh you didn't? Catch 5 more pokemon and pay attention to the number this time."

Ultima1990:
> What if, to save monotype challenges, you change the quest to a 2 part.
> 
> Instead of catch 5 Pidgeys he just ask to catch 5 pokemon. You return to him and then he says something like "did you noticed your dungeon tokens increasing after each catch? Oh you didn't? Catch 5 more pokemon and pay attention to the number this time."

No can do. The catch 5 Pidgey quest is intended to force the player to use the ball selector and set a ball to Caught. If they can just catch any 5 mons it's possible that they won't need to do this.

Jaaslet (approved):
It works now, and i think it's more helpful. Let's release it, and see if it stops some of the questions.
Thanks for the PR

PR #2593, opened 2022-07-07
Fix BF break from Sevii

CorgiOnNeptune:
Simple fix for TypeError breaking Battle Frontier.


Jaaslet (approved):
It works again! Thanks for the PR

PR #2596, opened 2022-07-07
Stop BG Lycanroc from flying

CorgiOnNeptune:
Not all dogs go to heaven!

Added Cutiefly, Ribombee, and Magearna to the flying list in Lycanroc's place.

Hard to test, but I'm sure someone will notice if one of the new ones look funny in the air.


Jaaslet (approved):
The dog does not have wings and the rest does.
Thanks for the PR!

PR #2598, opened 2022-07-08
Add gem count indicator in flute description

Dragonchitos:
I added the number of gems that the player has in the hover description of the flutes to avoid having to open the gems menu each time the timer is low.


zyx-xyzzy:
You can already see the gem counts by hovering over the remaining time (rather than over the flute icon), so this isn't necessary.

Jaaslet (changes requested):
I like the change. I think it's useful.
Can we add a `.toLocaleString('en-US')` to the numbers, so they are easier to read?

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2600, opened 2022-07-08
Add quests to talk to Breeder and Gem Scientist NPCs

Ultima1990:
To ensure that players find these NPCs and learn this information. Also gives some QP to help getting over that Kanto QP wall. Learning is so rewarding!


Jaaslet (approved):
I talked to the people. There where nice. Other people should do the same.
Thanks for the PR!

PR #2608, opened 2022-07-09
feature(farm): Freeze Mulch

Aegyo:
Adds a new mulch for freezing berry growth, at the cost of no aura emitted (and the farm points to buy it).
Should make some of the mutations with fast growing berries a little easier to manage without constant attention.

~The mulch bag image is just a copy of Boost Mulch, I don't have the skills to create a proper variant :sweat_smile:~
Mulch bag image has been provided by eugene171 on discord :+1:


Jaaslet (changes requested):
It works. Code looks good too. And i think we have ended up with what we agree on, in the dev-chat.
Just one bug: Petaya does still work when frozen (because it's not an aura backend).
And i moved the credits, so they are sorted alphabetic again

Jaaslet (approved):
It works now!
Thanks for the PR

Issue #2056, opened 2022-04-21
[SUGGESTION] Confirmation before applying mulch when using 'mulch all', mulch multiplier button static size


dragonkin550:
**Current Behavior**
Currently, mulch all is a pretty big button right next to the much smaller mulch amount selector button.  The mulch amount selector changes in size based on the current mulch application amount number.  That makes it easy to click, say, the right edge of 10x a few times to get back to 1x, and then, when the amount goes to 'all' instead of 'x1', accidentally hit 'mulch all'

**Improved Behavior**
I'd like the mulch multiplier button to be a constant, fixed size, so that it's less likely for accidents to happen.
I'd also like a confirmation when using 'mulch all', though perhaps the confirmation of mulch all should only be applied in some scenarios - like when the multiplier is 100x or more - as I could see using 'mulch all' with 1x on purpose a few times or even 10x, and it not really being nearly as painful as with wasting 100x mulch or more on all plots of the farm.

**Reason for change**
I just blew 3k mulch by accidentally hitting mulch all when clicking the mulch amount button.  It was at 10x, and I was clicking the right side of the button four times quickly to get back to 1x.  Instead of getting back to 1x, what happened was that the multiplier went to 100x, 1000x, all, and then my mouse button was over the mulch all, I clicked it without realizing, and all the mulch I had was applied to my entire farm.

dragonkin550:
closing this as fixed by #2609

PR #2609, opened 2022-07-10
Fix mulch multiplier and add confirmation on mulch

lenormandSeb:
Add quick fix to provide miss click on mulch.

Add confirmation for mulch all on click

based on this issue : https://github.com/pokeclicker/pokeclicker/issues/2056


CorgiOnNeptune:
Could you make it a setting? ‘Confirm before mulching all plots’

src/scripts/farming/FarmController.ts
```diff
@@ -105,7 +105,20 @@ class FarmController {
...
 
     public static mulchAll() {
-        App.game.farming.mulchAll(FarmController.selectedMulch(), this.getAmount());
+        if (!Settings.getSetting('confirmBeformeMulchingAllPlots').observableValue()) {
+            return App.game.farming.mulchAll(FarmController.selectedMulch(), this.getAmount());
+        }
+
+        Notifier.confirm({
+            title: 'Farm Mulch',
+            message: `Are you sure to mulch ${this.getAmount()} of ${MulchType[FarmController.selectedMulch()].replace('_', ' ')} ?`,
```
Jaaslet, line 114: This is not correct english :) My suggestion is:
Title: "Mulch all"
Message: "Are you sure you want to use <amount> <mulch type> on all plots?"
And there is no space between the mulch type and the questionmark

src/components/farmModal.html
```diff
@@ -93,7 +93,7 @@
                                     </ul>
                                     <div class="justify-content-center py-1">
                                         <div class="btn-group btn-block">
-                                            <button type="button" class="btn btn-sm btn-secondary"
+                                            <button type="button" class="btn btn-sm btn-secondary mulchMultiplier"
```
Jaaslet, line 96: Standard bootstrap would be to use either `col-3` or `col-4` as class to set the width, instead of creating a new class, only for adding width

Jaaslet (changes requested):
It looks good. The button is much nicer now, and easier to hit.
I am happy that the confirmation box is hidden behind a setting, which is off by default, as it's mainly used by lategame players, and we don't want new players to see it as a popup simulator :)
I have two comments:

src/scripts/farming/FarmController.ts
lenormandSeb: Me and english, a great story of love ... almost (more thunder than sunny :smile: )
[34e7529](https://github.com/pokeclicker/pokeclicker/pull/2609/commits/34e7529edc3889768da1afe17b2419d8232ffc3b)

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2611, opened 2022-07-11
Battle Frontier timer adjust

DataCrusade:
Changed the timer bar in Battle Frontier to turn red ("danger") when there is 10 seconds left (1/3rd of the time left) instead of 20 seconds left (2/3rd of the time left).
This is more intuitive and consistent with the proportions on dungeon timers.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2615, opened 2022-07-12
Attack rebalance 2

apple096:
The sudden sequel to #586 from around 2 years ago

Motive : 
 - Fix the underlying issues with the current formula
   + Shedinja (10 => 19)
   + Deoxys Normal has higher attack than Deoxys Attack (144 => 272, 122 => 279)
   + Attack imbalance (Game-wide)
- Change the current meta
  + Hard nerfing the baby Protein meta (33-50% nerf)
  + Hard nerfing Gyarados 
  + Slight nerf to Heatran 
  + Hard nerfing Clefable 
  + Hard buffing all Legendaries/Pseudo


Jaaslet:
I love that Machop has a breeding efficiency of 590,000, but the rest must be buggy. Gyarados having 56,000,000,000,000,000? That is even more broken than before!
My plan was to look at the changes, so i could write the start of a discussion. But i don't think this is what you where trying to do

Jaaslet:
PR is abandoned by the creator, so i will close it

PR #2618, opened 2022-07-12
Multiple quest quest

Jaaslet:
Want to be able to have two quest active at a time, in a questline? Now you can! Requested for Galar (and maybe other stuff).
Saving for questlines is weird. Found a solution, but it's super ugly. Will update it, if i find a better one, but not sure if i can...


src/components/questDisplay.html
```diff
@@ -69,6 +70,22 @@
...
+                                data-bind="attr:{ style: 'width:' + ($data.progress() * 100) + '%'}"
+                                aria-valuemin="0" aria-valuemax="100">
+                                <span data-bind="text: ko.pureComputed(function(){return $data.progressText()})">0/1000</span>
+                            </div>
+                        </div>
+                    </td>
+                </tr>
+                <!-- /ko -->
+                <!-- /ko -->
+                <!-- /ko -->
```
RedSparr0w, line 88: Could we increase the nested indentation here just so it's easier to see what's exactly happening.

src/scripts/quests/Quests.ts
```diff
@@ -305,7 +305,12 @@ class Quests implements Saveable {
                 if (ql) {
                     ql.state(questLine.state);
                     if (questLine.state == QuestLineState.started) {
-                        ql.resumeAt(questLine.quest, questLine.initial);
+                        ql.resumeAt(questLine.quest, questLine?.initial instanceof Number ? questLine.initial : 0);
```
RedSparr0w, line 308: I don't think we need the `questLine?.initial` as `questLine` should definitely exist.
The following should also work if `initial` isn't a number:
```suggestion
                        ql.resumeAt(questLine.quest, +questLine.initial || 0);
```

RedSparr0w (approved):
As far as I can tell, it all looks like it should work fine, left a couple of comments, unsure how to actually test it though 😅 
Thanks for the PR! 👍

Jaaslet:
I have fixed your two comments.
The way i tested it, was to replace the 2nd and 3rd `tutorial.addQuest();` with one `tutorial.addQuest(new MultipleQuestsQuest([captureOne, routeTwo], 'test'));`

RedSparr0w:
Gave it a quick test, it doesn't seem to show what the quest are, only the progress bars, is this intended?
image

Jaaslet:
> it doesn't seem to show what the quest are, only the progress bars, is this intended?

Yes it is. Ultima don't want quests to fill up to much of the screen, so the description should explain all the quests, to make it more compact.

> It also doesn't seem to give the individual quests rewards, Only the final reward provided to the MultipleQuestsQuest.

In my testing, it looked like you got the reward for each individual quest, when you completed that part of the quest. But i don't think it will ever be relevant, as it's weird to use the reward function on a part of a quest

RedSparr0w:
> > it doesn't seem to show what the quest are, only the progress bars, is this intended?
> 
> Yes it is. Ultima don't want quests to fill up to much of the screen, so the description should explain all the quests, to make it more compact.

I feel like it should only not have a description if one isn't specified for the individual quest.
For example if the quest was to catch 1 of each of the Kanto legendary birds, it would show like this:
image
And you wouldn't know which step had been completed.
Compared with something like:
_(ignore the fact they are all the same quest 😅)_
image

----------------------

> > It also doesn't seem to give the individual quests rewards, Only the final reward provided to the MultipleQuestsQuest.
> 
> In my testing, it looked like you got the reward for each individual quest, when you completed that part of the quest. But i don't think it will ever be relevant, as it's weird to use the reward function on a part of a quest

When I tried using this:
_(removed the 2 quest from adding on their own)_
`tutorial.addQuest(new MultipleQuestsQuest([captureOne, routeTwo], 'Just a test'));`
I didn't recieve any rewards for the parts that completed, unless I manually specified the total within the `MultipleQuestsQuest`:
`tutorial.addQuest(new MultipleQuestsQuest([captureOne, routeTwo], 'Just a test', 40));`
image

PR #2620, opened 2022-07-12
Protein rebalance

apple096:
Rebalance of protein, follow-up of #2615 

Changes : Protein buffs now scales with base stats and egg steps

Formula : 
image

Calculator : https://www.desmos.com/calculator/teouicfyqz

Q : Isn't this just Calcium :  

 - Possibly yes, but my stance is that there shouldn't be a static boost similar to Protein in the game. Currently, 30 proteins result in effectively adding 120 base stat to your Pokemon, which makes some Pokemon a lot more powerful than it is (Magikarp from 10 attack to 130 attack, making it one of the strongest Pokemon in the game)
 - And, even a small amount of Protein cap similar to the Calcium issues (10-15) will still result in a baby meta, which isn't good

Q : Why is the formula so weird :  

 - The plan for this is to move toward a legendary meta, and you need to scale semi-quadratically to make them better than Pokemons like Gyarados, even post-#2615


apple096:
What this doesn't include :  
 - An update function to migrate old protein to new protein : Too controversial, haven't decided yet

Jaaslet:
Gyarados is still in top 2. And it looks like every pokemon has been nerfed hard. I am not sure what the goal is?

DataCrusade:
It's definitely suppose to go with #2615 so similar/same goals as that. Mostly killing the buff baby meta and making proteins stronger on stronger mons

Jaaslet:
PR is abandoned by the creator, so i will close it

PR #2623, opened 2022-07-13
Pokémon seen but not caught being more noticeable in the Pokédex

Crobat4:
One of RedSparr0w requests, this will make Pokémon seen but not caught more noticeable (currently, they show full color for them).

Here's an example:
seen-not-caught

Never did a PR so tell me if something went wrong.


Jaaslet:
We need write access to the branch (it will ask you, when you create the PR). Else we won't be able to do stuff like merging into the branch, so we can merge this, or change stuff, if we have small changes we want to do, before we merge it.

Aegyo (approved):
Looks good :+1:
Thanks for the PR

PR #2624, opened 2022-07-13
Blank Plate from Legends Arceus

DataCrusade:
Adds the Blank Plate (normal type plate) from Pokemon: Legends Arceus to the Underground


Jaaslet:
We don't think the game needs a normal plate, and we might want to use it for something fun in Hisui, when we one day add that, so we won't be accepting this PR.
But thanks for the work anyways :)

PR #2630, opened 2022-07-14
Setting to increase game width

Crobat4:
This PR adds a setting to increase the max-width of the game container to 95%.

Here's how it looks:
game-width


RedSparr0w:
Looking good,
I believe we can just toggle the `container` and `container-fluid` classes on the `#game` div to get the same sort of effect?

Crobat4:
Yeah, you are right, not much difference.
I'll change it then.

RedSparr0w (approved):
Looks good, thanks for the PR!  👍

PR #2631, opened 2022-07-14
Shards rebalance + New Ball Types

jk13pclick:
Balance and introducing the ball types we all forgot about. The new system is mostly untested, and I didn't do the comparison-sheet as with the last rebalance. Feel free to do testing/comparison work.


Jaaslet (approved):
Code looks good. There are a lot of changes in Dungeon.ts, and i looked through a big chunk of them, and could not see anything that look like it must be wrong.
If anything is too broken, we will find out.
The code looks good, and the new weight system makes much more sense.
Great to see some fun loot in Alola and Thanks for the PR!

PR #2632, opened 2022-07-14
Changes on evolution methods for some pokemon

Crobat4:
This PR makes some changes in the evolution methods of some Pokémon to be more faithful to the main games as well as adds regional day/night level as new methods.

Affected Pokémon:
* Pikachu (No more 50/50 while player is in Alola)
* Exeggcute (No more 50/50 while player is in Alola)
* Koffing
* Mime Jr.
* Cubone (Marowak outside Alola, A. Marowak in Alola at night)
* Sliggoo (Thunderstorm added)
* Alolan Rattata (Night only)
* Tyrunt (Day only)
* Amaura (Night only)
* Yungoos (Day only)
* Charjabug (Sinnoh - Kalos dungeons with special magnetic field added)
* Fomantis (Day only)

I can add Castform as well to work like Cherubi but that depends if you want that.

Some console tests.
Before (player in Alola):
pikachu-before
After (first in Alola, second in Sinnoh):
pikachu-after2


DataCrusade:
Perhaps change Spiritomb to be a trade of Odd Keystones for it instead of Quest Points, or maybe change it to cost diamonds instead (as Odd Keystones can be sold for diamonds)

RedSparr0w (approved):
Works great, Thanks for the PR! 👍 

I'm not sure if we want Castform weather based evolutions or not tbh, @Ultima1990 thoughts?

PR #2636, opened 2022-07-14
Allow 5 columns when using extended game width

### LINKED ISSUE #2171: [SUGGESTION] More UI Column Space For Wide Screen Monitors
**Current Behavior**
The UI has currently has one column on either side of the main map column.

**Improved Behavior**
Increase the number of side columns to accommodate for wide screen monitors.

**Reason for change**
It's really awkward to have a large amount of screen space with no actual use in the game. I have to scroll down when there is plenty of side space.

**Detailed Description**
Even having just one more column on either side (totaling four side columns, x2 on each side) would alleviate this.

CURRENT
image

SUGGESTED (sorry for bad paint skills!)
image

### LINKED ISSUE #2633: [SUGGESTION] New setting for 5 columns in full width mode
**Current Behavior**
3 sortable columns:
image

**Improved Behavior**
5 sortable columns:
image

**Reason for change**
With the extra space some stuff just looks a bit too stretched _(especially on ultrawide monitors)_.

**Detailed Description**
Add a new setting to allow sorting into 5 columns when using the full width setting.
Would also need to figure out the best way to move modules back to the visible columns when the setting is disabled.

RedSparr0w:
Allows users to have 5 columns for game components when using the full width display mode.
This won't affect smaller displays, needs to be >= 992px wide for it to work at all.

Not sure about the wording for the setting.
`5 Columns (requires increased game width)`

Using 3 columns, Ultrawide, Full width:
image

Using 5 columns, Ultrawide, Full width:
image

Using 3 columns, 1920 x 1080, Full width:
image

Using 5 columns, 1920 x 1080, Full width:
image

closes #2633
closes #2171


Jaaslet (changes requested):
It looks like it work. I tried to move some stuff around and change the settings, and i don't think i broke anything.
My suggestion for the setting is `5 Columns (requires the setting "increased game width"):`.
If we want to get fancy, we can disable the 5 columns setting, if the increased game width setting is not set

RedSparr0w:
Or what about renaming the `increase game width` to something like `Game display style` and give it a drop down with
- Standard (3 column)
- Full width (3 column)
- Full width (5 column)

Jaaslet:
That sounds even better

RedSparr0w:
Updated to a drop down:
image

Also modified the `MultipleChoiceSettingTemplate` to not require `:` in the display name, so it's the same as other settings.

Jaaslet (approved):
Looks good. Thanks for the PR

Issue #2643, opened 2022-07-14
[SUGGESTION] graphical overhaul of the pokedex


lenormandSeb:
**Current Behavior**
Currently pokemon information is not quickly accessible, you have to either hover over it for some info or click on it to get them all

**Improved Behavior**
Have the list of pokemons directly in the form of a list (no alternate form) , and have all the information directly in the same modal

**Detailed Description**
If no pokemon is selected, the list takes the whole modal.
If one of the pokemon is selected, the info is on the right as well as the possibility of seeing the alternative forms of this one.
The non-chromatic and chromatic form.
Info in table

**Screenshots:**

Now

![Capture d’écran du 2022-07-14 14-50-35](https://user-images.githubusercontent.com/58883370/178991047-8ad33942-33aa-4847-b6f6-55ac10d0c75b.png)

New

![Capture d’écran du 2022-07-14 15-19-30](https://user-images.githubusercontent.com/58883370/178991853-5708c97f-0c31-48e4-b54e-c95b6265ab16.png)

![Capture d’écran du 2022-07-14 15-17-53](https://user-images.githubusercontent.com/58883370/178991596-aa9e58d0-7329-434e-972e-82cd6911bd8e.png)

![Capture d’écran du 2022-07-14 14-50-53](https://user-images.githubusercontent.com/58883370/178991100-f8be9887-526b-42d5-a1ae-d60455d95626.png)

![Capture d’écran du 2022-07-14 14-51-12](https://user-images.githubusercontent.com/58883370/178991129-db54e86c-badc-4ff8-b951-41bca010296e.png)

![Capture d’écran du 2022-07-14 15-16-56](https://user-images.githubusercontent.com/58883370/178991420-2161854a-d4d8-4b58-8427-ca95c8cd0b5a.png)

Qwertypop04:
This would make people think every alt form is available. To use the example in your screenshots, only Arceus (normal) is available,   yet this looks like all of them are

lenormandSeb:
I'm agree with you, code and visual need some adjustment . but you get the idea

PR #2645, opened 2022-07-14
Feature/more readable pokedex

lenormandSeb:
Based on this issue : https://github.com/pokeclicker/pokeclicker/issues/2643

I remove unobtainable form of some pokémon like Arceus on the screenshot, and the big "show alternate form" button.

I've test with an 0.9.5 and 0.9.8 save without problems, I hope that you will appreciate it.

PS: my english is still bad, you can blame google translate :smile: 

(some screen)

Capture d’écran du 2022-07-14 23-49-10
Capture d’écran du 2022-07-14 23-49-31
Capture d’écran du 2022-07-14 23-50-05
Capture d’écran du 2022-07-14 23-50-25

### DIFF
--- a/src/components/pokedex.html
+++ b/src/components/pokedex.html
@@ -1,126 +1,260 @@
 <div class="modal noselect" id="pokedexModal" tabindex="-1" role="dialog"
 aria-labelledby="pokedexModalLabel">
-<div class="modal-dialog modal-dialog-scrollable modal-lg" role="document">
-   <div class="modal-content">
-       <div class="modal-header">
-           <h4 style="margin-bottom: 0">Pokédex</h4>
-           <button style="float:right;" class="btn btn-secondary" type="button" data-toggle="collapse"
-                   data-target="#pokedex-filter-container" aria-expanded="false" aria-controls="pokedex-filter-container">
-               Filters
-           </button>
-       </div>
-       <div class="modal-body">
-            <div class="collapse" id="pokedex-filter-container">
-                <div class="container-fluid" style="border-bottom: 1px solid #eceeef">
-                    <div class="form-row text-left">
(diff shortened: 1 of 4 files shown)


NiCeDiCe90:
I don't really see the improvement of your Dex rework. :/

You see less Pokémon in the Overview (only 5) and the information being spread over the entire width makes is less readable actually. The current dex is more compact and I can see the same information. But way more pokemon at once.

The Labels for typing work pretty well for people who are not that familar with the typings but I think the background colors in the current dex look way better.

I can see the idea behind removing the alternate forms to make it less cluttered with walls of pikachu/unown/vivillon or a bazillion of Alcremie in the future. But since they are all individual Pokémon in your team, they may also deserve their own dex entry.

Not having them also gives wrong information about the alternate forms too.
Bulbasaur (Rose) for example is not grass/poison. It's a grass/fairy type. I assume you see that if you click on the form but as a player I would assume here that they are all the same type.

If you have a lot of forms you also have to scroll a lot to see all the statistics and stuff. Takes some time to get information about your Unown (Z) in your Dex. In the current dex you can find it way faster and you have the popup that is consistently the same size for all Pokémon.

Jaaslet:
The dev team agrees with @NiCeDiCe90.
I will close this PR. Sorry. But thanks for trying.

lenormandSeb:
Don't worry, I understand :smile:  there's a lot of change and my approach was going in the wrong direction.

I take note of the point of view of Nice and the majority of the players, by extension that of the devs too.

I will continue to issue smaller and more digestible suggestions.

And thank you for your feedback. :+1:

Jaaslet:
Bigger PRs are fine. Just make sure they are accepted, before you start :)

PR #2646, opened 2022-07-15
EV Berry

jk13pclick:
Also does some code rework of gaining EP. Needed for easier addition of EP from other methods.


src/scripts/party/Party.ts
```diff
@@ -178,6 +178,27 @@ class Party implements Feature {
...
     }
 
+    public gainEffortPoints(pokemon: PartyPokemon, shiny: boolean, number = GameConstants.BASE_EP_YIELD): number {
+        let EPNum = number;
+
+        if (pokemon.pokerus < GameConstants.Pokerus.Contagious) {
+            return 0;
+        }
+
+        if (App.game.multiplier.getBonus('ev') > 1) {
```
Jaaslet, line 188: I don't think this is needed, but i am not sure. It does not make sense that it can be below one, and if it can, it's because we have added a weird debuff, and then it won't work anyways.

Jaaslet (changes requested):
I think it now makes sense with some effort points inflation. Currently, the berries only has an effect, if you have enough to give a multiplier above 2.
My suggestion is to multiple all the effort points values (including how many required for a EV) by 100. Note that this will require an update function, but i think it will be fairly simple. Like: `saveData.party.forEach((p) => p.effortPoints *= 100);` (i have not tested if the syntax is correct)

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2647, opened 2022-07-15
Browser sync - Disable sync'd clicks, scroll etc

RedSparr0w:
This disables Browser Sync having clicks and scrolling sync'd between tabs/devices enabled by default.


Jaaslet (approved):
I really like this change! Thanks for the PR

PR #2648, opened 2022-07-15
Increase width of map legend popover

Crobat4:
This PR will increase width of the map legend popover to make it a bit more readable (due the long text "Uncaught shiny Pokémon and Missing Achievement")

This is how it looks:
map-legend


RedSparr0w (approved):

Still looks good on mobile too:
image

PR #2654, opened 2022-07-16
Good Hex Maniac sprite everywhere

Ultima1990:
For a while now we've been using two Hex Maniac sprites. The crappy gen 3 one in Hoenn and a way better custom one by Kyle-Dove (already credited) in Kalos. There's really no point in having two different sprites for the same trainer class, unless it's a real difference like male/female, so with this PR we now use the good custom one for all Hex Maniacs, and the crappy gen 3 sprite is removed.


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #2613, opened 2022-07-11
[BUG] 3rd stage pokemon with baby forms do not give the baby form when bred.


DataCrusade:
**Description:**
Breeding a pokemon like Azumarill, Raichu, etc who are 3rd stage pokemon while you have yet to unlock the 1st stage/baby form of them, does not give you them when you hatch the 3rd stage's egg. The 2nd stage pokemon in the same line however, will give you the baby form when bred.

**Steps to Reproduce:**
- Breed a 3rd stage mon with a baby form while not having obtained their baby form yet
- Hatch the egg

**Expected Behavior:**
Expected behavior is both 2nd and 3rd stage mons should unlock the baby form if not yet obtained when bred. This is more intuitive and consistent.

**Additional Info:**
The current behavior seems to be caused by supposedly old code and/or technical limitations according to devs and code contributors I talked to. Although creative liberties are taken in other aspects of the game, this is not one of them and is just an unintentional error.

PR #2656, opened 2022-07-16
add last evolution as breedable pokemon

lenormandSeb:
closes https://github.com/pokeclicker/pokeclicker/issues/2613

adding loop for second stade to add the third as breedable pokemon

How i test :
- Remove the baby is allready have
- Breed the last evolution
- Baby Hatched too

testing with Magby, Elekid and event Snorlax (because text are different) evolution line


DataCrusade:
I think lastStade should be lastStage, I assume stade is a typo and meant to be stage

lenormandSeb:
@DataCrusade Good look, thanks :+1: :smile:

Jaaslet (approved):
It only accounts for a chain of 3 pokemons, but currently, there are no chains above 3 pokemons, so this should be fine. We can always change it later, if they give us longer chains.
Thanks for the PR

PR #2657, opened 2022-07-16
Use all battle items

### LINKED ISSUE #2581: [SUGGESTION] A way to use all timed items at once & one other thing
I'd love for there to be a button that allows you to use all timed items at once next to the increment button OR opposite side so people don't click it by accident, I usually have all items running all the time. Gets tedious clicking them each individually when refreshing them. Maybe the button could let you choose which items to use when it's clicked to make it even better. (Could be greyed out when out of items & only light up when you have more than one type of timed item)

One other thing, maybe adding a 1px stock black border on the text on notification pop ups so they're even more readable. They kind of are hard to read with my f.lux filter on. At least the achieve ones anyway.

Dragonchitos:
I added new buttons to use battle items and flutes with one click (closes #2581)
The "BI" button just uses all items according to the multiplier, even if some of them are empty.
The "Flute" button activates all non-activated flutes with a right click, and deactivates all activated flutes with a left click. 
I wasn't sure which file to use for the functions, so I will move them if needed.

image

### DIFF
--- a/src/components/battleItemContainer.html
+++ b/src/components/battleItemContainer.html
@@ -71,9 +71,9 @@
     <div data-bind="visible: App.game.badgeCase.hasBadge(BadgeEnums.Heat)">
         <button type="button" class="btn btn-sm btn-primary"
                 style="position: absolute; left: 0px; width: auto; height: 41px"
-                onclick="ItemHandler.activateAllFlutes()"
-                oncontextmenu="ItemHandler.deactivateAllFlutes(); return false"
-                title="Enable all Flutes"
+                onclick="ItemHandler.useAllFlutes(true)"
+                oncontextmenu="ItemHandler.useAllFlutes(false); return false"
+                title="Activate/Deactivate all Flutes"
                 data-toggle="popover"
                 data-html="true"
                 data-content="Left/right click to deactivate/activate all Flutes at once"
--- a/src/components/battleItemContainer.html
+++ b/src/components/battleItemContainer.html
@@ -71,23 +71,14 @@
     <div data-bind="visible: App.game.badgeCase.hasBadge(BadgeEnums.Heat)">
         <button type="button" class="btn btn-sm btn-primary"
                 style="position: absolute; left: 0px; width: auto; height: 41px"
-                onclick="ItemHandler.useAllFlutes('enable')"
+                onclick="ItemHandler.activateAllFlutes()"
+                oncontextmenu="ItemHandler.desactivateAllFlutes(); return false"
                 title="Enable all Flutes"
                 data-toggle="popover"
                 data-html="true"
-                data-content="Click to enable all Flutes at once"
+                data-content="Left/right click to desactivate/activate all Flutes at once"
                 data-placement="left"
-                data-trigger="hover">Enable all
-        </button>
-        <button type="button" class="btn btn-sm btn-danger"
-                style="position: absolute; right: 0px; width: auto; height: 41px;"
-                onclick="ItemHandler.useAllFlutes('disable')"
-                title="Disable all Flutes"
-                data-toggle="popover"
-                data-html="true"
-                data-content="Click to disable all Flutes at once"
-                data-placement="right"
-                data-trigger="hover">Disable all
+                data-trigger="hover">Activate all
         </button>
         <div class="card-header p-0" data-toggle="collapse" href="#fluteItemContainerBody">
             <span>Flutes</span>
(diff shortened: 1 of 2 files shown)


src/scripts/items/ItemHandler.ts
```diff
@@ -23,6 +23,34 @@ class ItemHandler {
...
+
+    public static deactivateAllFlutes() {
+        GameHelper.chunk(6, Object.keys(ItemList).filter(i => ItemList[i].constructor.name == 'FluteItem')).forEach((item,index) => {
+            item.forEach((x) => {
+                if (FluteEffectRunner.isActive(GameConstants.FluteItemType[x])()) {
+                    ItemHandler.useItem(x);
+                }
+            });
+        });
+    }
```
lenormandSeb, line 52: can be fuse on one function with boolean as parameter, and I would have personally put "item" of the foreach in the plural and replaced "x" by item in the singular.

but deactiveAllFlutes really deactive ? the useItem make me doubts :thinking:

Dragonchitos: Yes it really deactivate, it confused me too at first. I will make the other corrections you mentioned.

src/components/battleItemContainer.html
```diff
@@ -1,4 +1,14 @@
 <div id="battleItemContainer" class="card sortable border-secondary mb-3" data-bind="visible: !(Settings.getSetting('hideChallengeRelatedModules').observableValue() && App.game.challenges.list.disableBattleItems.active())">
+    <button type="button" class="btn btn-sm btn-primary"
+            style="position: absolute; left: 0px; top: 0px; width: auto; height: 41px;"
+            onclick="ItemHandler.useAllBattleItems()"
+            title="Using all Battle Items"
+            data-toggle="popover"
+            data-html="true"
+            data-content="Click to use all items at the same time"
```
Jaaslet, line 8: For consistency, we should use the same tooltip, as we are using everywhere else (like the battle items)

src/components/battleItemContainer.html
```diff
@@ -58,8 +68,21 @@
...
-      <span>Flutes</span>
+    <div data-bind="visible: App.game.badgeCase.hasBadge(BadgeEnums.Heat)">
+        <button type="button" class="btn btn-sm btn-primary"
+                style="position: absolute; left: 0px; width: auto; height: 41px"
+                onclick="ItemHandler.useAllFlutes(true)"
+                oncontextmenu="ItemHandler.useAllFlutes(false); return false"
+                title="Activate/Deactivate all Flutes"
+                data-toggle="popover"
+                data-html="true"
+                data-content="Left/right click to deactivate/activate all Flutes at once"
```
Jaaslet, line 79: I don't like that we use right-click as a part of the gameplay. People are used to right-click not doing anything in browser games, other than giving you a weird menu, and we have not used it so far, in this game.
Instead i would suggest it turned into "Deactivate all", when all unlocked flutes was activated

src/scripts/items/ItemHandler.ts
```diff
@@ -23,6 +23,26 @@ class ItemHandler {
         return result == undefined ? true : result;
     }
 
+    public static useAllBattleItems() {
+        GameHelper.chunk(6, Object.keys(ItemList).filter(i => ItemList[i].constructor.name == 'BattleItem')).forEach((item,index) => {
```
Jaaslet, line 27: I don't think we need the `GameHelper.chunk()`. The filter should leave us with only the battle items. And this will make it harder to add more stuff (which is probs more relevant for flutes).

Jaaslet (changes requested):
I think it looks good. I had no idea where you would place that button, but you surprised me. It looks great! And it also works.
When you change UI, it would be great if you threw a screenshot into the PR description.
I have some changes for you :)

src/components/battleItemContainer.html
Dragonchitos: Right click is already used to decrease the multiplier of battle items, so it seems logical to me to use this function. And with right click you can deactivate all the flutes even if they are not all activated.

Jaaslet: Oh it is? You learn something new each day!
But it's still not intuitive. What if we do a mix? Right click always deactivates. Left click deactivates if all of them are active. Else it will activate.

Jaaslet:
This PR seems to be dead. I will close it. It can be reopened, if the creator comes back to it

PR #2658, opened 2022-07-16
Multiple quests quest fix

Jaaslet:
Fixes the last thing for multiple quests:
- If the subquest has a description, it's now shown in italic.
- If a subquest has a reward, it will be ignored. Saving became too weird to code, if we wanted it, and i don't think we ever do want it. And if we do, i will fix it then. But now it does nothing, instead of saying "you can claim this quest"


RedSparr0w (approved):

Issue #2601, opened 2022-07-08
[SUGGESTION] Change the Mystery Egg Key Item into something else


DataCrusade:
**Current Behavior**
Right now, when you unlock the hatchery, it says you obtained a Mystery Egg

**Improved Behavior**
I propose this is changed to something else, ex: Hatchery Pass, Hatchery License, Breeding License, something like that, perhaps.
Alternate potential solution: Give the player a hatch-able Mystery Egg item at the same time like the ones sold in Pewter City.

**Reason for change**
Currently, it's confusing. As a new player, I saw you can buy Mystery Eggs (An actual hatch-able egg that shares the same name as the key item). I thought I was given one of those for free to go with unlocking the hatchery, and I thought the game had bugged out when I couldn't see any eggs to put into the hatcery.

**Detailed Description**
See Improved Behavior. Would just need a sprite change and a rename to the key item, and the message when you unlock the hatchery would need to be updated

**Screenshots:**
![image](https://user-images.githubusercontent.com/36621129/178057625-9b635b5f-276e-4b81-a6a0-a7d853fd0270.png)
![image](https://user-images.githubusercontent.com/36621129/178057665-2419a964-50b5-483e-acbd-faf566555dbc.png)
Pictures of the 2 different Mystery Eggs

Dragonchitos:
I also thought of that when I started! Maybe we can use the Egg Incubator of Pokemon GO instead ?

DataCrusade:
> I also thought of that when I started! Maybe we can use the Egg Incubator of Pokemon GO instead ?

That's a much better idea for a key item, and a nice reference to PoGo too

Farboo171:
I'll get started on this, the plan is to call it the Egg Incubator and use the orange PoGo incubator as the image

PR #2660, opened 2022-07-16
Egg Incubator replacing Mystery Egg Key Item

Farboo171:
Name and image of the Mystery Egg key item updated per Issue #2601 to remove confusion with the Mystery Egg egg type.


DataCrusade:
The image should probably be a sprite to match the rest of the key items although I just noticed that the Pokerus Virus key item isn't a sprite either. I can make a sprite for this tomorrow if you'd like.

Ultima1990:
I do agree that renaming one of the Mystery Egg items is a good idea to avoid confusion. However, I'm not sure Egg Incubator is the way to go, since that is a thing you carry around and hatch eggs in while hatching in our game happens at the hatchery, which is a static location.

Farboo171:
> The image should probably be a sprite to match the rest of the key items although I just noticed that the Pokerus Virus key item isn't a sprite either. I can make a sprite for this tomorrow if you'd like.

Sure! I'm not sure how to sprite-ify images, so that's probably a good idea to be in keeping with the theme.

Farboo171:
I'll look into why this is messing up, and can float some other options besides incubator, sorry for the mess.

PR #2661, opened 2022-07-17
Shard Fixes

jk13pclick:
Blocks shards from the left side of daily deals, also removes them from the treasure list (for selling purposes)


Jaaslet (changes requested):
If we merge this, we will no longer have a list of shards obtained. So we either need to find a new way of showing that, or just leave them in the treasure list for now. We can still do so they can't be sold.
Other than that, it looks good.

Jaaslet (approved):
I removed the button instead of disabling it. Now everything looks good.
Thanks for the PR

PR #2662, opened 2022-07-17
Pimp my biker gang

Ultima1990:
Added an extra quest for the Sevii questline to defeat the 3 biker goons, rather than just having the quest for the biker gang leader say you need to beat the goons to get to the leader.
Updated questlinestepcompletedrequirements to account for this extra quest.
Changed Biker Goon display names to Biker Goon, instead of Biker Goon 1, Biker Goon 2, Biker Goon 3.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2665, opened 2022-07-17
Fixed missing apostrophe

DataCrusade:
Added a missing apostrophe to Cynthia's Grandmother (was missing from the name)


Jaaslet (approved):
Looks better now. Thanks for the PR

PR #2666, opened 2022-07-17
Temp battles can now reuse images

Jaaslet:
We no longer have 6 different Ash images


src/scripts/temporaryBattle/TemporaryBattle.ts
```diff
@@ -36,6 +37,10 @@ class TemporaryBattle extends TownContent {
     public getTown() {
         return this.parent ?? TownList[this.optionalArgs.returnTown] ?? TownList[GameConstants.DockTowns[player.region]];
     }
+    public getImage() {
+        const imageName = this.optionalArgs?.imageName ?? this.name;
```
RedSparr0w, line 41: Should this just default to the `displayName` as that seems to follow all the image names so far _(as far as I can tell)_

Jaaslet: It's weird imo to use display name to find a file, because... The display name should not be limited (and filenames are limited).
But yeah, right now it will make sense... I will think about it

RedSparr0w: Sorry, I don't mean remove the `imageName` param, but have `displayName` as the default instead of `name`.

Jaaslet: I know. Will keep it, so we can always make it work. But it still feels weird

Jaaslet: I think i don't like using displayName. Yes, it will make sense with the current temp battles, but it feel too weird. And displayName is not always set, so we need something with two defaults.

RedSparr0w (approved):

PR #2667, opened 2022-07-17
Quests in QuestLineRequirements are now cached

Jaaslet:
Refactor of some code, to save some lookups, each time the requirement is called


RedSparr0w (approved):
Works great for me, thanks for the PR! 👍

PR #2668, opened 2022-07-17
Added snorlax as temp battle

Jaaslet:
Snorlax is now an overworld sprite temp battle. Kanto has two. For each to go away, you need to beat it, and capture any of them.

- [x] Add snorlax
- [x] Add CaptureSpecificPokemonRequirement
- [x] Make sure wild temp battles can be caught
- [x] Make sure snorlax shows up on map before it's unlocked


RedSparr0w (approved):

RedSparr0w:
Just a couple of things before merging:

Are the Snorlax supposed to be a 0% catch rate with base Poké Balls?
Are they not supposed to disappear once defeated?
Is the only way to access route 16 by defeating the Snorlax, can you not leave Snorlax undefeated and access it from route 17? _(been awhile since I've played RBY)_
Are we able to grey/darken the images until they are able to be challenged?

Jaaslet:
> Are the Snorlax supposed to be a 0% catch rate with base Poké Balls?

No, that is a bug. I should look at that, before we merge.

> Are they not supposed to disappear once defeated?

They are suppose to disappear when defeated AND caught. Else the pokedex might not be possible

> Is the only way to access route 16 by defeating the Snorlax, can you not leave Snorlax undefeated and access it from route 17? _(been awhile since I've played RBY)_

I just followed the old requirements. Basicly just moved everything over once. But i am not sure it's correct.

> Are we able to grey/darken the images until they are able to be challenged?

I can look into that

Jaaslet:
Ready for review.
You have fixed the requirements. I have now updated it to the newest version, done some cleanup and added the locked mapoverlay.
I have no idea why, but the catchrate works now... Might have happened in some other PR, or during some of my cleanup... No idea

RedSparr0w (approved):

Issue #2663, opened 2022-07-17
[SUGGESTION] easy alternate forms filter on the pokedex


cacahueteone:
**Current Behavior**
 thers no entry for the alternated forms of pokemons and neither is there a filter for those

**Improved Behavior**
 id like a filter option to activate or deactivate the alternated forms sorta like the rare hold item option so it apears when needed

**Reason for change**
 it gets anoying to trak wich forms u habe for pokemons like unown an floette

**Detailed Description**
 its just a button on and off just under hide shiny image to be able to see wich forms u habe or not it would help a lot!

**Screenshots:**

![sugestion](https://user-images.githubusercontent.com/109468618/179400457-5777c0c3-6d7a-42f6-8cd3-f2967063b273.png)

PR #2672, opened 2022-07-17
Feature/add hide box alternate

lenormandSeb:
Issue : https://github.com/pokeclicker/pokeclicker/issues/2663

A little code for add another filter (for the walls of alternate form)


Jaaslet (approved):
Looks good. I was a bit confused, because the linked issue also wants us to show uncaught alternative forms (i think), which is not something we are interested in. But this PR does not implement that.
I changed "form" to "forms" on the label text.
Thanks for the PR

PR #2673, opened 2022-07-17
Breeding Modal Help Text for Pokérus

jk13pclick:
Might add some text elsewhere, could add rates to FAQ. This might be enough though.


Jaaslet (changes requested):
I am not sure i think it fits the breeding modal help. EVs has nothing to do with the breeding modal. Only how you spread the rus. I think this text should be in FAQ.
And the EVs NPC is outdated. Still talking about the starter is the only one spreading it.
But the text looks good. Maybe remove the part about the wiki, as i feel like the game should work independently of the wiki

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2678, opened 2022-07-18
Added Furfrou (star)

Jaaslet:
Tater gave me this amazing idea to add a furfrou as a shard trade... So i did it...
Found in Parfum Palace. Numbers might be a bit high, but it's a lategame grindy pokemon (like the rest). And if it's too high, we will nerf it in a later patch, and look like heroes.
Also turned Parfum Palace into a town. There was no reason for it to be a dungeon (did not add anything), and there are no encounters in the game.


RedSparr0w (approved):

Just double checking, `Furfrou` is still obtainable outside of `Parfum Palace`?

Qwertypop04:
> Just double checking, `Furfrou` is still obtainable outside of `Parfum Palace`?

Yep, it's on Kalos Route 5

CorgiOnNeptune:
The update function needs to be moved to 0.9.10, right?

RedSparr0w:
> Yep, it's on Kalos Route 5

Awesome, thanks 👍 

> The update function needs to be moved to 0.9.10, right?

Nice catch 👍

RedSparr0w:
Sorry, 1 last thing, should the town be changed to a Pokémon Center on the map rather than how it is at the moment?
image

PR #2679, opened 2022-07-18
Rose Evolutions

NiCeDiCe90:
Adds Rose Evolutions.
But they don't just evolve, they need to be caught (Or use another Event feature)

ToDo
- [ ] Add way to obtain them (Golden Week Event)


Ultima1990:
You'll need to manually change egg cycles if you want them to match the base form counterparts.

RedSparr0w:
> You'll need to manually change egg cycles if you want them to match the base form counterparts.

I think it would be better to just add them as impossible evolutions instead of manually adjusting the egg steps similar to Phione/Manaphy:
```ts
        'evolutions': [new LevelEvolution('Phione', 'Manaphy', 69420)],
```

NiCeDiCe90:
Totally forgot about that. 

The reason why they don't evolve is that we want them to be new content for the next Event.
If they just evolve they would be available as soon as they get merged.

PR #2681, opened 2022-07-19
Repeatball EV Effect

jk13pclick:
Buffs Repeat Ball for EVs. Could either remove the catch rate bonus, or nerf it.


Jaaslet (commented):
5 is a high number. I am not sure it's too high, as i don't know how hard repeat balls are to get... But that is my only concern. Else it looks good

Jaaslet (approved):
No one thought 5 was too high, and two thought it was a good number. So yeah, no more comments.
Thanks for the PR

PR #2682, opened 2022-07-19
Hide Pokeball in Statistics Modal if Pokemon is seen, but uncaught

CorgiOnNeptune:
Before:
before

After:
after


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2683, opened 2022-07-19
Slightly 'kinder' Ash Quest text

CorgiOnNeptune:
Maybe we were being too rough on the kid..


Jaaslet (approved):
These texts are more family friendly. At least no one complained about them.
The story is still fun. Thanks for the PR

PR #2684, opened 2022-07-19
Update Update.ts

RedSparr0w:
This PR groups the `v0.9.8` Update functions by what they are updating than by PR.


Jaaslet (approved):
It looks like everything is still updated as it should. Thanks for the PR

PR #2686, opened 2022-07-19
Fixed Galactic Boss Cyrus for old saves

Jaaslet:
If you load an old save, which is past Sinnoh but have not completed the Sinnoh questline (which is also any save before 0.9.0, which have completed Sinnoh), you can defeat Galactic Boss Cyrus before you get there in the questline. Which will lock you out of ever completing the questline.
The solution is to not unlock the boss before you each that point in the questline.


RedSparr0w (approved):

PR #2687, opened 2022-07-19
Force thousand separator in Pokemon List

CorgiOnNeptune:
Just forces 'en-US' in PokemonList attack numbers string for consistency with other numbers.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2689, opened 2022-07-19
Rewrite Plasma questline

Ultima1990:
Make use of new tech to put temporarybattles and npc conversations in the questline.

WIP

I still need to balance every temporarybattle after Opelucid Gym, I want to add a Colress battle in the Frigate dungeon and maybe a Ghetsis battle against Kyurem forms in the Chasm dungeon. Also I need to fix the visible requirement for the Plasma Frigate dungeon overlay and have a way for the Plasma Assault town to not disappear while the temporarybattles are undefeated, since the last temporarybattle in the Assault is now the requirement for Seaside Cave.
The player is present in a couple of the temporary frigate towns when they disappear. Hopefully I can find a way to kick the player out of the town when that happens.


Ultima1990:
Still to do:
Write update function.
Balance Colress and Ghetsis.
Lots of testing.

BTW. For those not in the know. This includes the removal of the Plasma Assault dungeon. Love me!

A useful save for testing.
[Unova Save.txt](https://github.com/pokeclicker/pokeclicker/files/9148631/Unova.Save.txt)

Ultima1990:
Wanna know something creepy?

Currently the Plasma Frigate consists of 2 overlays. One which disappears after 1 clear, and one which appears after the Plasma questline is completed to allow for replay. Today I experimented with combining those overlays into 1 with an OR function. And what I saw was horrifying.

image

That's me, having just cleared Giant Chasm. But look closer. What's that on Plasma Frigate?
Turns out any time a town overlay disappears while the player is in there, that town overlay abducts a permanent copy of the player avatar! I just didn't notice because the 2nd version of the Frigate is a different overlay! But now, by bringing back the same overlay, I have finally discovered the truth. I don't know what horrible purpose the overlay has for this copy...

This has been happening all along (on the dev branch where we have player avatars on the map)! And it must have been happening with the Plasma Assault dungeon overlay too! The overlays must have a small army of player avatars by now...

Ultima1990:
This is ready.

It completely overhauls the Team Plasma questline. Starring:
> 23 quests.
> 6 new NPCs, 5 of which must be talked to during the questline.
> 20 temporary battles.
> The removal of the Team Plasma Assault dungeon.
The temporary battles are added for everyone, even if you complete the old version of the questlines these battles are active and many of them are required to access areas that you previously had access to. That's just how requirements work.
Requirements are set up in such a way that it all works for new players, players that update partway through the questline and players that have already completed the old version. In all cases the temporary battles must be completed _in order_, starting in Virbank City and ending at Giant Chasm. Players that have completed the questline before do not get to play this new version on that save, but by making them complete the temporary battles in order they still get the general gist of the new version.

If you had already completed the questline before and are trying to go to a place you had access to before but is now locked, it is almost certainly locked behind one of the temporary battles. As mentioned before, they must be completed in order. So if the place says it's locked behind Zinzolin 1, and you can't find any Zinzolin 1 anywhere, that means Zinzolin 1 is himself locked behind another temporary battle earlier in the chain. Just look for the orange town and go from there.

You get the DNA Splicers at the end of the questline as usual. Even if you had completed the questline before, defeating the final temporary battle in the chain does show you a popup saying that you received the Splicers. Don't worry, this doesn't mean you got a second set, nothing went wrong.

I _would like_ to lock Castelia City behind one of the temporary battles, but with how docktowns are currently set up that causes a whole big mess. If someone eventually changes how they work I'll just change the requirement then. For now this is done.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2694, opened 2022-07-20
Rivals

Qwertypop04:
Added rival battles for Kanto through Alola, as well as a few other temporary battles. Also, added Silvally forms to PokemonList for Gladion fight, removed Wally from Victory Road Hoenn to replace with temp battle.


src/scripts/temporaryBattle/TemporaryBattleList.ts
```diff
@@ -1,6 +1,93 @@
...
 
 //Kanto Temporary Battles
+TemporaryBattleList['Blue 1'] = new TemporaryBattle(
+    'Blue 1',
+    [
+        new GymPokemon('Pidgey', 1040, 9),
+        new GymPokemon('Charmander', 1678, 9, new StarterRequirement(GameConstants.Region.kanto, GameConstants.Starter.Grass)),
+        new GymPokemon('Squirtle', 1678, 9, new StarterRequirement(GameConstants.Region.kanto, GameConstants.Starter.Fire)),
+        new GymPokemon('Bulbasaur', 1678, 9, new StarterRequirement(GameConstants.Region.kanto, GameConstants.Starter.Water)),
+        new GymPokemon('Bulbasaur', 1678, 9, new StarterRequirement(GameConstants.Region.kanto, GameConstants.Starter.Special)),
```
Aegyo, line 11: Shouldn't this be Eevee?

Qwertypop04: No. We use the same team as if Squirtle had been picked, if Pikachu was picked. Maybe that should be changed in future, but it's how it is for the Champion battle, at least.

PR #2696, opened 2022-07-20
Evolution Stones -> Evolution Items

DataCrusade:
Updates some text to not imply that every evolution item is a stone


Jaaslet (changes requested):
I think it will make it less confusing. It looks good, but i have some comments:
- "Fan Club Chairman" NPC needs text update
- Two lines in the FAQ needs to be updated: EV modifier and obtain method
- It's still called EvoStone in the code. Might turn the PR into a monster, if you start cleaning that up, so only if you feel like it

DataCrusade:
Not gonna touch how it's named in the code but adjusted the other things you mentioned

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2697, opened 2022-07-20
Update locked pokemon dungeon text

DataCrusade:
Adjusts the text when you don't have access to a pokemon in a dungeon to suggest more than just talking to NPCs, as the way to unlock them in most cases is dungeon completion, story progression, or catching certain mons


Jaaslet (approved):
I want this in the next update! Thanks for fixing my text!

PR #2699, opened 2022-07-20
Angry kitty for Mars

Ultima1990:
Also a slight team adjustment for Jupiter to make her consistent with the other Galactic Commanders at the lakes.

Sincerest apologies for your broken immersion.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2706, opened 2022-07-22
Fix Pokémon having Pokérus early and gaining EP early

RedSparr0w:
This PR should fix Pokémon gaining Pokérus early (before key item gained).
Should also remove any EP gained on Pokémon that do not currently have Pokérus.
Also fixed up previous update function.


Jaaslet (approved):
It looks like it fixes the bug. Thanks for the PR!

PR #2708, opened 2022-07-22
Update Pokérus icons

RedSparr0w:
This PR updates the Pokérus icons for more consistent sizing:

Before | Now
:---: | :---:
image | image
image | image
image | image
image | image

Also updated FAQ to use `×` instead of `x`:
image

**NOTE:** I'm fine if we want to keep the old icons, but I think they need to be made more consistent in terms of how they're displayed.


Jaaslet (approved):
Looks good. And the vote has decided that we like these better.
Thanks for the PR

PR #2711, opened 2022-07-22
Fix Protein Achievement

CorgiOnNeptune:
Fixes protein achievements not counting purchases.


Jaaslet:
Will close this, as #2713 is a more correct solution

PR #2713, opened 2022-07-22
Increment obtained when purchasing Proteins

RedSparr0w:
Only the Purchased observable is currently being incremented when purchasing Proteins, but the Obtained observable should also be updating.


Jaaslet (approved):
Looks good. Thanks for the PR

Cramoisi-San:
May i ask, are the proteins you already bought counts, or are they lost ?

Qwertypop04:
> May i ask, are the proteins you already bought counts, or are they lost ?

They count

PR #2716, opened 2022-07-22
Volcarona NPC

Ultima1990:
Added NPC for Volcarona in Relic Castle. Should hopefully explain that Relic Passage needs to be cleared.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2720, opened 2022-07-22
Survey now shows shards

Jaaslet:


src/scripts/underground/Mine.ts
```diff
@@ -190,6 +190,10 @@ class Mine {
             } else {
                 switch (reward.valueType) {
                     case 'Diamond': {
+                        if (reward.isShard()) {
+                            res.shards++;
+                            break;
+                        }
```
RedSparr0w, line 196: Not really related to this PR, but shouldn't the `valueType` for shards not be `Diamond`?
Would rather us sort that out first before we engrain the shards more into the `Diamond` type.

RedSparr0w (approved):
Actually going to merge as is, as the whole of the underground items probably needs a bit of a refactor.
Thanks for the PR! 👍

PR #2726, opened 2022-07-22
Tooltips for ball name in Poké Ball Selectors

CorgiOnNeptune:
Adds tooltips to the Poké Ball selectors to display the name of the ball when hovering over it's image.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2728, opened 2022-07-23
Underground tools tooltips

### LINKED ISSUE #2314: [BUG] Bombs loses efficiency on partially cleared layers
**Info**
 - OS: Windows 10
 - Browser Firefox
 - Version 0.9.5

**Description:**
Bombs seem to remove rock layers from already cleared tiles, resulting in a loss of bomb efficiency when an underground layer is partly cleared

**Steps to Reproduce:**

1. Go underground
2. Clear most of an underground layer (the more you clear, the easier it is to see the loss of bomb efficiency)
3. Use a bomb

**Expected Behavior:**
10 (or more depending on your bomb efficiency) tiles should have 2 layers removed

**Screenshots:**
Before using a bomb : 

before bomb
m. -->

After using a bomb. I have upgraded bomb efficiency once so it should have removed two layers from twelve tiles, but there seem to be only eight tiles affected (I may have missed one, but I don't think I missed four)
 
afterbomb

**Additional Information:**
I don't know if this is the expected behavior of bombs, but I don't recall it being explained anywhere, and even the wiki doesn't show this information.
Furthermore, bombs are already weak as is due to their randomness that make it so they often land on useless (uncleared) tiles once an underground layer is partly cleared, so having then also land on already cleared tiles makes them quite un-fun to use

CorgiOnNeptune:
Closes #2314 with clarifying tooltips.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2729, opened 2022-07-23
Resized Hoenn and Kalos badges

DataCrusade:
Hoenn and Kalos badges were smaller in comparison to all the other badges, which was bugging me.

Also put them through tinypng. Sourced from bulbapedia which appears to be the same place the original versions were from.


Jaaslet (approved):
Looks better. Thanks for the PR

PR #2730, opened 2022-07-23
Resized gen 6 fossil icons

DataCrusade:
Jaw and Sail fossil icon were smaller than the other fossil icons, so I fixed that.
Also deleted what seemed to be unnecessary copies of these files that were in the wrong folder


Jaaslet (approved):
Looks good. Thank you for the PR

PR #2732, opened 2022-07-23
New Battle Frontier rewards

DataCrusade:
Added in some new rewards to fill out the Battle Frontier more. More rewards/milestones patches the gaps between rewards in the later stages and makes progression feel more satisfying and rewarding, even if the rewards aren't crazy.

Also still leaves lots of room for future rewards to be added.


Jaaslet (approved):
The code looks good and stable. It even fixes the problem we have reach time we change the rewards (i think).
I have tried to set up the different weird states by changing my save in the console, and all of them worked. The rewards have been accepted by the dev team.
Thanks for the PR

PR #2740, opened 2022-07-23
Add Galactic Questline start to 0.9.10 update function

Ultima1990:
A while ago there was a bug with questlines starting. Used to not be a big deal. They missed a questline. Oh well. But it does mean that on some saves the Sinnoh questline never started.

Which is now a big deal, because now there are routes which are locked behind a temporarybattle which is locked behind progress in that quest line. Oh dear.

I _think_ the solution to this is simply repeating the 0.9.0 update function. Which, in the next update, should start the questline for players that accidentally skipped it during the quest starting bug. And shouldn't start it for players that have already completed it or have it in progress.


src/scripts/Update.ts
```diff
@@ -1008,6 +1008,11 @@ class Update implements Saveable {
             renamePokemon(saveData, 'Minior (Red-core)', 'Minior (Red Core)');
             renamePokemon(saveData, 'Minior (Violet-core)', 'Minior (Violet Core)');
             renamePokemon(saveData, 'Minior (Yellow-core)', 'Minior (Yellow Core)');
+
+            // Start Galactic questline if player has Coal Badge already
+            if (saveData.badgeCase[40]) {
```
RedSparr0w, line 1013: You will also need to add a check for if the player has already started this questline as it's not a new questline.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2741, opened 2022-07-23
Update Giovanni image

Ultima1990:
A proper square Giovanni image so that the boss icon in the dungeon isn't so tiny.


RedSparr0w (approved):

PR #2748, opened 2022-07-24
Fixup `0.9.9` update function

RedSparr0w:
Was checking for `0` beforehand which wouldn't exist due to being the default value so the check wasn't firing.
This will only affect saves that went from `0.9.7` → `0.9.8` → `0.9.9` as this update function was only meant to fix the `0.9.8` update function for saves already on that version.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2749, opened 2022-07-25
Fixup questlines loading

RedSparr0w:
Fixes questlines skipping the first quests if they already have 1+ of the focus value

### DIFF
--- a/src/scripts/quests/Quests.ts
+++ b/src/scripts/quests/Quests.ts
@@ -305,7 +305,7 @@ class Quests implements Saveable {
                 if (ql) {
                     ql.state(questLine.state);
                     if (questLine.state == QuestLineState.started) {
-                        ql.resumeAt(questLine.quest, +questLine.initial || 0);
+                        ql.resumeAt(questLine.quest, questLine.initial);
                         if (ql.curQuestObject() instanceof MultipleQuestsQuest) {
                             ql.curQuestObject().quests.forEach((q, i) => {
                                 q.initial(questLine?.initial[i] ?? 0);


src/scripts/quests/Quests.ts
```diff
@@ -305,7 +305,7 @@ class Quests implements Saveable {
                 if (ql) {
                     ql.state(questLine.state);
                     if (questLine.state == QuestLineState.started) {
-                        ql.resumeAt(questLine.quest, +questLine.initial || 0);
+                        ql.resumeAt(questLine.quest, questLine.initial);
                         if (ql.curQuestObject() instanceof MultipleQuestsQuest) {
                             ql.curQuestObject().quests.forEach((q, i) => {
                                 q.initial(questLine?.initial[i] ?? 0);
```
RedSparr0w, line 311: Will probably also need to do something on this line for the multi quests, but haven't looked into those yet.

Jaaslet: We don't need to do anything for multiple quests, i think. Nothing has changed for them, since they where reviewed and merged.

Jaaslet (approved):
Looks good. Thanks for the PR

src/scripts/quests/Quests.ts
RedSparr0w: The problem is this only occurs when we manually start a questline in `Update.ts` because we don't know the initial value.
So it wouldn't have been tested at that time (which is the same time this issue was introduced)
AFAIK if we pass `0` to the initial function, the quest will complete if you've previously defeated the required amount, so chances are the same bug will happen on the multi quest if we manually start those too (and they are the first step).

PR #2750, opened 2022-07-25
Underground refactor

RedSparr0w:
Refactoring some of the Underground code, mostly items.
Still need to figure out some things:
- [ ] `EvolutionItemRequirement` _(requires a bunch of module stuff moved)_
- [ ] Move mine inventory to Underground or something 🤷 maybe


RedSparr0w:
Think this should be fine for review now.
I wasn't able to move `EvolutionItemRequirement` yet, as it ends up being a massive amount of other stuff that also needs to be moved at the same time, so I've just made all the stones appear once you reach Hoenn (other than the sun stone) for now.

src/modules/underground/UndergroundItems.ts
```diff
@@ -0,0 +1,128 @@
...
+UndergroundItems.addItem(new UndergroundEvolutionItem('Moon Stone', 304, [[0, 1, 1, 1], [1, 1, 1, 0]], 1, StoneType.Moon_stone));
+// TODO: Replace these requirements with StoneUnlockedRequirement once moved to modules
+UndergroundItems.addItem(new UndergroundEvolutionItem('Sun Stone', 305, [[0, 1, 0], [1, 1, 1], [1, 1, 1]], 1, StoneType.Sun_stone, new MaxRegionRequirement(Region.johto)));
+UndergroundItems.addItem(new UndergroundEvolutionItem('Shiny Stone', 306, [[0, 1, 1], [1, 1, 1], [1, 1, 1]], 1, StoneType.Shiny_stone, new MaxRegionRequirement(Region.hoenn)));
+UndergroundItems.addItem(new UndergroundEvolutionItem('Dusk Stone', 307, [[1, 1, 1], [1, 1, 1], [1, 1, 0]], 1, StoneType.Dusk_stone, new MaxRegionRequirement(Region.hoenn)));
+UndergroundItems.addItem(new UndergroundEvolutionItem('Dawn Stone', 308, [[1, 1, 1], [1, 1, 1], [1, 1, 1]], 1, StoneType.Dawn_stone, new MaxRegionRequirement(Region.hoenn)));
+UndergroundItems.addItem(new UndergroundEvolutionItem('Ice Stone', 309, [[1, 1, 1], [1, 1, 1]], 1, StoneType.Ice_stone, new MaxRegionRequirement(Region.hoenn)));
```
Jaaslet, line 108: I am not sure these are correct? I think Ice stone is Alola, while Shiny+dusk+dawn is Sinnoh, right?

Jaaslet: And i think all the stones are missing an UndergroundItemValueType?

Jaaslet (changes requested):
I just updated my save, and this was the survey result (pressed the survey button, after i loaded the save):
image
I think evo stones counts as plates. I have no idea why one of the shards is bugged...

src/modules/underground/UndergroundItems.ts
RedSparr0w: Fixed 👍

Jaaslet (approved):
The bug i found with the shard not counting as a shard, is not new. For some reason, Survey does not handle already uncovered items very well. But i might look at it in an other PR.
But i will merge this, and ask the crew to look it through. I can't find any more bugs atleast.
Thanks for the PR

PR #2757, opened 2022-07-25
Fix dungeon counters

### LINKED ISSUE #2753: [BUG] Incorrect number of chests/enemies in dungeons
**Info**
 - OS: Windows 10
 - Browser Chrome
 - Version v0.9.9

**Description:**
Dungeons do not always have the correct number of chests/enemies

**Steps to Reproduce:**
1. Go to any dungeon bigger than 5x5
2. Proceed normally
3. Occasionally, the correct number of chests or enemies will not spawn

**Expected Behavior:**
The dungeons to have the correct number of chests/enemies

**Screenshots:**

wrong enemies
wrong chests

Jaaslet:
Closes  #2753
Code is weird, so chests and enemies can be placed on the entrance, which led to there being less enemies/chests than expected.
The solution is to count how many there are, instead of calculating it.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -356,11 +356,11 @@
                 <div class="card-header p-0">
                     <div style="position: absolute; left: 2px">
                         <img title="Chests opened in current dungeon" src="assets/images/dungeons/chest.png" height="25px" />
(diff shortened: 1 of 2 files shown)


RedSparr0w (approved):

Issue #2746, opened 2022-07-24
[BUG] Pokémon in hatchery disappeared since last update


Kyuryia:
**Info**
 - OS: Windows 11
 - Browser: Chrome
 - Version: v0.9.9

**Description:**
The slots in hatchery were empty when I logged in after several days without logging in, the hatchery queue wasn't affected

**Steps to Reproduce:**
1. Log in
2. See the hatchery slots empty but the hatchery queue full
3. Breed fossils or new Pokémon
4. Your Pokémon previously in hatchery disappeared

**Expected Behavior:**
The Pokémon in hatchery should be visible and hatch

**Screenshots:**
![image](https://user-images.githubusercontent.com/109911921/180654117-461e11d2-4e19-4386-9d93-72521ace3637.png)
![image](https://user-images.githubusercontent.com/109911921/180654168-5df4cb23-fccf-4f3c-89bc-37cd7086a27f.png)

**Additional information**
As you can see, I've captured Tornadus but it's nowhere in my game, neither in the hatchery nor in the Pokémon list, it disappeared

JordiMontblanc:
I have the same bug but also with Cobalion.

PR #2758, opened 2022-07-25
Disappearing egg bug

Jaaslet:
An attempt for fixing #2746.
I am not able to reproduce the bug (and no one so far have been able to tell me how), so i can only guess...
Only thing i could find in the code, is that we have two different ways of calculating if an egg is hatched or not. And if some rounding error does so only the first is true, the pokemon will be lost.
But yeah, the only way of testing it, is to release it and see if people still loses it.

### DIFF
--- a/src/scripts/breeding/HatcheryHelper.ts
+++ b/src/scripts/breeding/HatcheryHelper.ts
@@ -188,7 +188,7 @@ class HatcheryHelpers {
             egg.addSteps(steps, multiplier);
 
             // Check if the egg is ready to hatch
-            if (egg.progress() >= 100 || egg.isNone()) {
+            if (egg.canHatch()) {
                 egg.hatch(helper.attackEfficiency(), true);
                 this.hatchery.eggList[index](new Egg());
 
--- a/src/scripts/breeding/Breeding.ts
+++ b/src/scripts/breeding/Breeding.ts
@@ -211,7 +211,7 @@ class Breeding implements Feature {
                 partyPokemon.calculatePokerus();
             }
             egg.addSteps(amount, this.multiplier);
-            if (this.queueList().length && egg.progress() >= 100) {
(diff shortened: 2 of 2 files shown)


Jaaslet:
I think i also need to write an update function, to fix it for players where the bug has happened... But i can do that tomorrow

src/scripts/breeding/HatcheryHelper.ts
```diff
@@ -188,7 +188,7 @@ class HatcheryHelpers {
             egg.addSteps(steps, multiplier);
 
             // Check if the egg is ready to hatch
-            if (egg.progress() >= 100 || egg.isNone()) {
+            if (egg.canHatch()) {
```
RedSparr0w, line 191: We should move the code for checking if this slot is empty below this block maybe,
otherwise if the filter is ever empty the helpers won't continue breeding.

Jaaslet: I can't see that code anywhere?

Jaaslet:
I have now added some code, so the problem will be fixed on reload. It would be great to find the bug, but now we have this code, until we find the bug

src/scripts/breeding/HatcheryHelper.ts
RedSparr0w: You removed the check for ` || egg.isNone()`,
Which tris to hatch the egg (which shouldn't do anything),
Then replaces the egg.

Jaaslet: egg.canHatch() has the same check, right?

RedSparr0w: `egg.canHatch()` does the opposite, checks that the egg slot isn't empty.

RedSparr0w (approved):

PR #2765, opened 2022-07-26
Big Alola map

Ultima1990:
Same as the Sinnoh one. 2x high, 2x wide.


Jaaslet:
We have taken your Poni map, and turned it into a normal sized map (it looks good, thanks!). It's currently in an other PR.
People are currently working on the other Alola maps, so when someone has time to do the big map project, we need to start over anyways, to use all the changes. I will close this PR.

PR #2766, opened 2022-07-26
Show Poké balls if you have a quantity of them

RedSparr0w:
This is mostly for the Lure ball, which is awarded in Johto dungeons, but isn't shown in the list/unlocked until Hoenn.

If you have any, it will now show it as an option in the Poké ball selector list.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2775, opened 2022-07-28
Fix Evo Item Shiny Log

CorgiOnNeptune:
Getting a shiny through evo item always says '(duplicate)' in the logbook. Order of operations is important...
Also improved the notification.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2781, opened 2022-07-29
Fixed Nestball catchBonus function

PixLSteam:
This fixes a TypeError that occurs when Nestballs are being used. It only occurs when the player's highest region is one where the routes are not numbered regularly.


Jaaslet (approved):
After your change, maxRoute is 189 instead of 188 (with current endgame). I can't see why it has changed, but 189 is the correct number. And the bug is gone!
Thanks for the PR

PR #2786, opened 2022-07-30
Add Erika NPC for Togepi event

Ultima1990:
The quest text implies that there is one. So now there is one.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2790, opened 2022-08-01
Some fixes to the underground, after the lastest refactor

Jaaslet:
- Got a console error, when a evo stone was discovered

This seems to fix all bugs so far. But will keep an eye out

### DIFF
--- a/src/scripts/underground/Underground.ts
+++ b/src/scripts/underground/Underground.ts
@@ -174,7 +174,7 @@ class Underground implements Feature {
         const item = UndergroundItems.getById(id);
 
         if (item.valueType == UndergroundItemValueType.EvolutionItem) {
-            const evostone: EvolutionStone = (ItemList[item.valueType] as EvolutionStone);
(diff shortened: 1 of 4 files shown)


src/scripts/underground/Underground.ts
```diff
@@ -174,7 +174,7 @@ class Underground implements Feature {
         const item = UndergroundItems.getById(id);
 
         if (item.valueType == UndergroundItemValueType.EvolutionItem) {
-            const evostone: EvolutionStone = (ItemList[item.valueType] as EvolutionStone);
+            const evostone: EvolutionStone = (ItemList[item.itemListName] as EvolutionStone);
```
RedSparr0w, line 177: Instead of adding a new variable, couldn't we just do something like this instead?
```suggestion
            const evostone: EvolutionStone = (ItemList[StoneType[item.type]] as EvolutionStone);
```

Jaaslet: It looks like it works! Have cleaned up my solution, so ready for review again

RedSparr0w (approved):
Nice work, Thanks for the PR! 👍

PR #2792, opened 2022-08-01
gainPokemonEgg breeding property on success

PixLSteam:
Breeding::gainPokemonEgg now only sets breeding property on success.
Please note that I have not been able to test this change yet.


RedSparr0w (approved):

PR #2793, opened 2022-08-01
Underground survey changes

RedSparr0w:
Hide `Diamond Value` if it is `0`.
Hide Survey tooltip when clicked.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2794, opened 2022-08-01
Hide save selector challenge mode tooltips when clicked

RedSparr0w:
Currently if you click a save on the challenge icon, the tooltip stays visible.
image

This should now be fixed.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2796, opened 2022-08-02
Evo stones img are now showing up in daily deals

Jaaslet:
If an evo stone was in a daily deal, the image did not show up (Leafstone is a deal today, for testing)


RedSparr0w (approved):

PR #2799, opened 2022-08-02
Finish PR #2305 (feature): Lock Plots in the Farm

CorgiOnNeptune:
Jaaslet requested somebody to finish PR https://github.com/pokeclicker/pokeclicker/pull/2305 since it was mostly finished, but untouched by the original contributor for 2 months after Jaaslet requested a couple of small changes.

Ready for review again with Jaaslet's requested changes finished.

In addition to the Shift + Click, I also added a toggle-able button/hotkey to make the feature more visible to players and added information on the feature to the Help box.


Jaaslet (changes requested):
The lock-button does not toggle. It only can only activate the feature, but not disable it. I would expect it to have the same behavior as the hotkey.
Other than that, it looks good.

CorgiOnNeptune:
The button should function as you expected now.

Jaaslet (approved):
No more comments. Thanks for the PR

PR #2803, opened 2022-08-03
Fix evo stone amount in daily deals

### LINKED ISSUE #2801: [BUG] Daily Deal Menu Not Showing Correct Inventory Count on Leaf Stone (And Possibly More)
**Info**
 - OS: Windows 10
 - Browser Chrome
 - Version 0.9.10

**Description:**
When looking at the daily deals available in the underground, the number that represents my inventory when looking at leaf stones for today says I have 0, but I actually have 83. All other items I have seen are normal.

**Steps to Reproduce:**
1. Go to Underground
2. Click on Daily Deals
3. See error
(I do not know if this is a problem for anyone else, I have not seen other posts about it, and this is the first time it has happened to me, right after this update came out)

**Screenshots:**
Deals

Items

(My SS are looking odd on my end before posting, so I apologize if they do not appear)

**Additional Information:**
Doing the trade (this specific trade being 3 Sky Plates for 1 Leaf Stone) does not change the Inventory number, it remains 0. I have not been able to check if it happens to other evolution items as there is not a trade for them today.

Jaaslet:
Fixes #2801


RedSparr0w (approved):

PR #2808, opened 2022-08-03
fix(hatchery): Hatchery Pokérus status not updating

### LINKED ISSUE #2802: [BUG] Pokerus not show properly on the fourth slot of hatchery and on refresh.
**Info**
 - OS: Windows 10
 - Browser Chrome
 - Version 104.0.5112.81

**Description:**
When load the game none of eggs show the pokerus
image
When at least one egg hatches, then only 3 shows te pokerus (i have all pokemons infected)
image
image
image
image

Also notes that now the old pokerus symbol is used instead the one that shows that this pokemon is contagios

**Steps to Reproduce:**
Reload the game, then hatch one egg

**Expected Behavior:**
all eggs with pokerus shows properly

RedSparr0w:
This PR should fix the Pokérus icon not appearing when it should have

closes #2802


Jaaslet:
The 4th slot works fine now, but the icons are still missing on load. Don't know if you intended to fix both in this PR?
And also: closes #2802

RedSparr0w:
Not sure there is a good way to have it work on load, will have to look into it a bit more.

RedSparr0w:
I've made `partyPokemon` observable, which seems to fix both problems, not sure if it's the best solution, but it works 😅

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2809, opened 2022-08-04
Add "Fire All" button to the hatchery modal.

RegisCoaxans:
The title says it all.
It avoids scrolling in the helpers' tab and seeking out those that one hired in order to fire them each.
A red button appears in the footer once the helpers tab is unlocked.
The button only gets enabled when at least one of the helpers is hired and allows the player to fire them all at once.


Jaaslet (changes requested):
It currently shows up on all hatchery tabs. And because it only says "fire all", there is nothing really connecting it to the helpers, which makes it confusing.

RegisCoaxans:
That is true.
Do I hide it so it only shows up when the helpers tab is active or do I just replace the text to « Fire all helpers » ?

Jaaslet:
I would prefere it to only show up in the helpers tab, but both works

Chronicar:
I would prefer for hired helpers to just move to the top of the list, it avoids scrolling and you can fire just one or two at a time instead of all without having to scroll.

RegisCoaxans:
I have first thought about this and come to the conclusion that the field of hatchery helpers / farm hands was meant to be static thus having the employees' cards moving up and down through the tab as we click on them could make it more confusing.

Jaaslet (approved):
Looks good. Thanks for the PR

Issue #2810, opened 2022-08-04
[SUGGESTION] Pokerus Filter in Pokedex


laparel:
Hoping for the dev team to add a pokerus status filter in pokedex to separate: None, Contagious, and Cured. Similar to the hatchery filter.

**Current Behavior**
Currently, there's no way to filter out pokerus status in pokedex. There's a filter in the hatchery but one can't click on pokemons in the hatchery to view detailed info of specific pokemons. 

**Improved Behavior**
With a pokerus filter in the pokedex, one would be able to quickly filter out pokemon pokerus status and view details of specific pokemons.

**Reason for change**
Quality of life improvement.

Example: I want to cure all my pokemons. So I filter contagious pokemons in my pokedex. Then quickly check the EV status of magikarp, who still need 2 more EVs/captures to be cured. I go to a water route and work to cure my magikarp.

PR #2814, opened 2022-08-04
Feature/add pokerus filter

lenormandSeb:
From issue : https://github.com/pokeclicker/pokeclicker/issues/2810

With some adjustment : 
 - "None" became "Uninfected" and "Cured" became "Resistant" on UI
 - Hatchery and Pokedex for pokerus filter are visually identical, functionally too (almost)


src/components/pokedex.html
```diff
@@ -56,6 +56,17 @@ <h4 style="margin-bottom: 0">Pokédex</h4>
...
                         </div>
+
+                        <div class="form-group col-md-6 col-6">
+                            <label for="pokedex-filter-pokerus-status">Pokerus Status</label>
+                            <select id="pokedex-filter-pokerus-status" autocomplete="off" class="custom-select" onchange="PokedexHelper.updateList()">
+                                <option value="all" selected>All</option>
+                                <option value="none">Uninfected</option>
+                                <option value="infected">Infected</option>
+                                <option value="contagious">Contagious</option>
+                                <option value="cured">Resistant</option>
```
RedSparr0w, line 67: IMO we shouldn't do it this way, It just makes everything more confusing having it named differently in different places.
This also makes it harder if we change the names in the future as we have to remember this specific area exist.
If you want the names to be different, it should be updated in the enum so we can have it adjusted everywhere.

lenormandSeb: I update the enum for Pokerus and made a change where `Pokerus.None` & `Pokerus.Cured` are use to match with the newest reference.

Jaaslet (changes requested):
Look at how the filter is added for the breeding modal. This will save some code, and make it look more clean. And make it easier for us, if we for some reason want to rename stuff again later.

lenormandSeb:
@Jaaslet 
Since filters are used by both modals, wouldn't it be better, either in this PR or in a future one, to make a "GlobalFilter" or just "Filter" instead of "BreedingFilter" with the possibility of overriding some filter values?
I am thinking, for example, of the shiny, which are not the same thing in the pokedex or the hatchery but would avoid having several ways of filtering.

In addition, to be connected with the Breeding modal, that of the pokedex calls a setting called 'breedingPokerusFilter', which is confusing in a page that concerns the pokedex.

What do you think ?

Jaaslet:
I am not quite sure how this will look in code, but i think i agree.
It should not be called anything with "breeding" if it's used other places, but "global" might not be needed.
But if you are reusing the same settings object for it, you might have problems with not being able to change one, without the other

Jaaslet:
The way you have currently coded it, the filters are shared. So there is no way of changing the pokedex filters, without changing what pokemons the hatchery helpers are breeding.
So for this to work, they need each a different setting object, and FilterController needs to not be static.
And i see a problem with FilterController taking a partyPokemon as parameter, as not all pokemons in the pokedex are caught.
It might be better to revert the latest changes, and just keep them separated. I think there are enough differences

lenormandSeb:
I drop the useless commit, and take note from your comment.

I'm going to take a step back for something simpler and more flexible for the filters between the pokedex and the hatchery.

Jaaslet (approved):
I have cleaned up the code, so now it looks good. Thanks for the PR

PR #2816, opened 2022-08-06
Farm buttons cleanup

Jaaslet:
The new lock button is now hidden on other tabs.
Cleaned up the code a bit for the fire all button


RedSparr0w (approved):

PR #2822, opened 2022-08-08
Legendary Beasts Questlines

Ultima1990:
Quests for the legendary beasts. Makes use of the bulletin boards to start the quests.

Legendary Beasts quest becomes available after defeating Morty. Short and sweet.
1: Talk to Eusine in Ecruteak. (Burned Tower is now locked behind this quest)
2: Clear Burned Tower.
3: Talk to Pokéfan Derek. (Roamers are now locked behind this quest)
4: Catch the beasts.
Breeding counts as catching for Raikou and Entei. Because re-catching roamers is too big of a PITA for veteran players.

Suicune is a different story. It is no longer a roamer. It must be _caught_. Breeding does not count.
A second questline becomes available at the bulletin board after completing the Pokéfan Derek quest. This questline follows the HGSS chain of events for Suicune, except each encounter is a battle in clicker, while in the main games Suicune just runs away and no battle takes place.
1: Fight Suicune
2: Talk to Eusine
3: Fight Eusine
4: Fight Suicune
5: Talk to Eusine (Suicune moves to Kanto at this point and can only be found there if you are champion of Johto)
etc etc etc.
It ends with 2 Tempbattles at route 25, with a Eusine convo inbetween. The second of those battles is supposed to be the player's chance to catch it. Which means this PR is currently waiting for Snorlax.

Yes, this makes Suicune a post-champ mon. Yes, this means that Ho-oh is now also a post-champ mon.


Jaaslet (approved):
I have finished the PR, cleaned up some code and played through it.
Thanks for the PR

Ultima1990:
The "true" on the CaptureSpecificPokemonQuest for Entei and Raikou means that breeding works? Making players re-catch roamers is too big of a PITA, even for my standards. Breeding should work for them.

Jaaslet:
Yes it does. You can breed your way to the roamers, as an old player

PR #2824, opened 2022-08-08
Galar Route and Dungeon changes

Qwertypop04:
Switched Fighting egg exclusive from Falinks to Clobbopus. Also removed Drakloak, Dragapult and Grapploct from routes and dungeons and made fossil pokemon require themselves to be caught before appearing in routes or dungeons. Also removed unused image I missed


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2830, opened 2022-08-10
Celebi Questline

Ultima1990:
Starts at Kanto bulletin board. Yes. Kanto. To try and fit it to the GS Ball anime timeline.
Available after completing Bill's Errand.
Does require becoming champ of Johto before completion.
Mostly NPC convos, some tempbattles, requires catching the dungeon Celebi at the end as usual, incorporates some of the ingame HGSS Celebi events, with a new surprise about halfway through the questline so even long time players that already have Celebi will want to complete this one.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2832, opened 2022-08-11
Fix Slowpoke Well

Ultima1990:
Also moves some stuff on tiles_FRLG over by one pixel.


RedSparr0w:
This seems to have removed all the shadows from the tile set too, was this intentional?

Ultima1990:
> This seems to have removed all the shadows from the tile set too, was this intentional?

NGL I have no idea what shadows you're talking about. Had a closer look, didn't see anything in the before or after.

If it's there it's extremely subtle. So subtle that I can't spot it.
All I did was move the stuff over. If doing that removed those shadows then I can't make this tileset edit with my software (GIMP). IDK why, it just deleted the shadows apparently.

RedSparr0w:
Here's some of the shadows (shadowed left, no shadow right):
image
image
image

I just use https://pixlr.com/e/ for editing them, seems to keep the shadows.
Pushed a commit that keeps the shadows, can you just double check I did it right 😅

Ultima1990:
Looks good. Had a quick peek at Johto.tmx. All still lookin' good.

I mean, I didn't go over the whole map with a magnifying lens but Slowpoke Well still looks good at least.

Jaaslet (approved):
It looks better now, i think? Thanks for the PR

PR #2833, opened 2022-08-11
Make all wild mon tempbattles use Transparent.png

Ultima1990:
This is futureproofing. Once we inevitably update tempbattles to have the trainer on-screen during the battle, just like dungeon trainers, the result with these mons would be that they are on the battlefield twice. Once as the mon, once as the trainer. So better to have the "trainer" version of the wild mon be transparent (can't have no image at all).

Also made Ultra Megalopolis use the same image as Ultra Wormhole. Saves data.


Jaaslet:
Closing this PR as we need a better solution. We need the images for the defeated message.
I am planning to add an optional parameter for this, when i implement the feature

PR #2839, opened 2022-08-11
Allow more than one shard trader in each town

Jaaslet:
Was a problem in Stow-on-side in Galar.
This also allows shard traders to be named something other than the town, in the code


Aegyo (approved):
:+1:

PR #2841, opened 2022-08-11
Adjusted Sonia's name

DataCrusade:
Unless I'm missing something, there wasn't any need to include Sonia's last name and it  just confused me /shrug

### DIFF
--- a/src/scripts/towns/TownList.ts
+++ b/src/scripts/towns/TownList.ts
@@ -3638,7 +3638,7 @@ const TrainStationGuy = new NPC('Train Station Guy', [
     'One is an island paradise, and the other a freezing wasteland.',
     'I\m sure if you go to these places you\'ll find many unique and powerful Pokémon!',
 ]);
-const SouthGalarRoamerNPC = new RoamerNPC('Sonia', [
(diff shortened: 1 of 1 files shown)


Ultima1990:
Shouldn't it be Professor [Last name] then? It's the same with Oak. Named Samuel Oak, known as Professor Oak.

DataCrusade:
> Shouldn't it be Professor [Last name] then? It's the same with Oak. Named Samuel Oak, known as Professor Oak.

Her grandmother is Professor Magnolia, so she can't also be Professor Magnolia

Qwertypop04:
> Shouldn't it be Professor [Last name] then? It's the same with Oak. Named Samuel Oak, known as Professor Oak.

That would be her grandmother. There are two Professors Magnolia

Qwertypop04:
> Professor Magnolia Jr.?

She is referred to as Professor Sonia in game iirc. So that makes most sense even if Professor Magnolia didn't exist.

Ultima1990:
Ah. If it's Professor Sonia in-game then that's fine.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2843, opened 2022-08-11
Adjust Bulletin Board Text Body

CorgiOnNeptune:
Decreased font size a little and scroll bar will now show if text overflows.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2846, opened 2022-08-12
More consistent underground sort

### LINKED ISSUE #2815: [BUG] Treasure sorting in the Underderground
**Info**
 - Windows 10
 - Client
 - Version 0.9.10

**Description:**
Shards get sorted inbetween Fossils in the Value sorting option and one can slightly affect the order the appear

**Steps to Reproduce:**
1. Go to Underground
2. open the treasure tab
3. Click on either sorting option Amount or Item
4. Return to sorting by Value
5. Shards and Fossils are now lumbed together. All items, in their groups, are now sorted in the order they appeared in the other sorting option selected before.

**Expected Behavior:**
Shards and Fossils should be seperated.
All Items should have a set order in which they appear in and when sorting the items by increasing/ decreasing value they should invert the order in which they appear.

**Screenshots:**
Sorting by Amount (increasing)
image

Sorting by Value (increasing) afterwards
image

Sorting by Name (Starting with A)
image

Sorting by Value (increasing) afterwards
image

Jaaslet:
Fixes #2815
When we sort the underground items, i have added a secondary sort for id.
I tried to add a default sort for id, but that is too much work for what i have time for right now.

### DIFF
--- a/src/scripts/underground/Underground.ts
+++ b/src/scripts/underground/Underground.ts
@@ -314,14 +314,22 @@ class Underground implements Feature {
         }
 
         player.mineInventory.sort((a, b) => {
+            let result = 0;
             switch (prop) {
                 case 'Amount':
-                    return (a.amount() - b.amount()) * this.sortDirection;
+                    result = (a.amount() - b.amount()) * this.sortDirection;
+                    break;
                 case 'Value':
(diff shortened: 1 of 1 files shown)


RedSparr0w (approved):

PR #2851, opened 2022-08-13
Add number of pokemons shown in protein modal

Hawkinou:
Hello, I just added the number of result on the proteine modal
image


Jaaslet:
I think it would be more correct to call it "Showing n Pokemon" like in the pokedex

Hawkinou:
image

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2859, opened 2022-08-15
2nd attempt at fixing the multiquest problem

Jaaslet:
I did some more testing with the multiquest problem, and found out my solution did not fix everything. More specificily, it did not fix old saves. This solution works with this save at least:

[bugged_multi_quest.txt](https://github.com/pokeclicker/pokeclicker/files/9338720/bugged_multi_quest.txt)


RedSparr0w (approved):

PR #2862, opened 2022-08-16
Secrets of the Jungle Questline

Qwertypop04:
Adds a questline for Zarude. Zarude is now locked behind part of this questline. Flowering Celebi is obtained at the end of the questline, and finishing the questline opens a shop where Zarude (Dada) can be bought.

Required before merging:
- [x] Make Celebi temporary battle catchable (requires #2668)


Jaaslet (approved):
I have fixed some small things, and now it looks good.
Thanks for the PR

PR #2872, opened 2022-08-18
GalarOnly Stone evos

CorgiOnNeptune:
Add GalarOnly Stone Evos for Pokemon that normally evolve in a specific location (Glaceon, Leafeon, Magnezone and Vikavolt), as per canon SW/SH.


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #2774, opened 2022-07-28
[Quality of Life] Adding a checkbox field "Disable right click", in the game's parameters


ValentinBailleulGH:
**Current Behavior**
Right clicks are doable without condition (as usual)

**Improved Behavior**
Parameters -> checkbox to enable/disable right click pop ups

**Reason for change**
QOL improvement : possibility for touchpad (PC) players to disable right clicking (in order to avoid missclicks)

**Detailed Description**
Hi !
I am a PC player who uses the touchpad a lot. I am very often confronted with the fact that if I click with two fingers to improve my number of clicks per second, I end up clicking at the same time with my two fingers, and it triggers a right click (since it is a property of the touchpad). Then I unintentionally click on one of the buttons of the contextual menu opened when right-clicking ; sometimes it'll quit the game, refresh the page... In fact, the context menu popup itself is annoying.

Therefore, every day I manually add `oncontextmenu="return false;"` in the html of the game :
`<body class="...".... oncontextmenu="return false;">`

Do you think it would be possible to choose directly in the game settings to be able to disable the right click (without having to mess with the code) ?

Thanks for your attention !
Regards
Niasuf#1670 if ever needed

**Screenshots:**
I'm sure you know what right clicks do.

NiCeDiCe90:
You could use the client instead. There shouldn't be any rightclick popups.

ValentinBailleulGH:
❤️ thanks for all!

PR #2875, opened 2022-08-18
Added setting : disable right click in battle view.

RegisCoaxans:
As requested in [#2774](https://github.com/pokeclicker/pokeclicker/issues/2774), this PR adds the ability to deactivate the right click menu in the battle interface. This will help players on tactile device not trigger that menu anymore when they touch their screen.


Jaaslet (changes requested):
It looks good, but i have some comments:
- Why is it only in the battle view? I can't see why people want leftclick to show that menu, other places
- I think the setting should default to on. I feel like it's only the developers, who needs it off
- I think it fits under the "display" menu

RegisCoaxans:
I had made it work in battle view because players could want to right click on images and display them raw on a new tab. Plus, some browsers or devices allow for screenshots via that context menu.
On second thought, I guess that those circumstances are rare enough.

Is everything better, then ?

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2877, opened 2022-08-20
Small questline adjustments (Unfinished Business)

### LINKED ISSUE #2880: Professor Oak appearing twice [BUG]
**Info**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 0.9.12

**Description:**
After I accepted Professor Oak's quest from the new Bulletin Board, his name appears twice on the right side of the town, once before Mom and once after. The first Professor Oak gives you the "onwards to Johto" message. The second is the quest. 
This could be intended behaviour, but it didn't seem so to me at first glance.

**Steps to Reproduce:**
1. Go to Pallet Town in Kanto.
2. Click on the Bulletin Board.
3. Accept Professor Oak's quest.
4. Close Bulletin Board so that you can talk to Professor Oak. You will see what I described.

**Expected Behavior:**
Professor Oak's name only appears once in Pallet Town whether you have accepted the quest or not.

**Screenshots:**
prof oak 2

CorgiOnNeptune:
Professor Oak's lab was experimenting too much and the clone was starting to freak people out.. The copy has been... removed.

I also added an extra couple "hints" to the quest for all the people that are getting stuck on 'play with celebi'. I don't know if there is much else that can be done without directly spoon-feeding the player, which I would like to avoid as much as possible. 
Closes #2880


DataCrusade:
Dialogue changes look about as good as they can be without straight up telling people to catch Celebi

Jaaslet (approved):
Looks good. I have done so Oak can be both in Johto and Kanto. It will not confuse players that there are two in different regions... And he just follows you back each time...
Thanks for the PR

PR #2878, opened 2022-08-20
Detective Pikachu Quest

Farboo171:
Adds Detective Pikachu, Detective Raichu, and an associated quest.


Farboo171:
[Detective Pikachu Test Save.txt](https://github.com/pokeclicker/pokeclicker/files/9387777/Detective.Pikachu.Test.Save.txt)
A save for if you want to test things without growing your own enigma berries

src/scripts/towns/TownList.ts
```diff
@@ -323,6 +323,33 @@ const CelebiProfIvy = new NPC ('Prof. Ivy', [
...
+    'In a shady warehouse, you find the informant. He is a Mr. Mime, and he doesn\'t seem willing to divulge the information you need.',
+],
+{ requirement: new MultiRequirement([new QuestLineStepCompletedRequirement('Detective Pikachu', 2), new QuestLineStepCompletedRequirement('Detective Pikachu', 4, GameConstants.AchievementOption.less )]) });
+const Informant2 = new NPC('Informant', [
+    'The Mr. Mime signals to you that this is an illicit drug called R, and that it is frequently used in the underground fighting rings near the Battle Frontier',
+],
+{ requirement: new MultiRequirement([new QuestLineStepCompletedRequirement('Detective Pikachu', 4), new QuestLineStepCompletedRequirement('Detective Pikachu', 6, GameConstants.AchievementOption.less )])});
+
+const MewTwo1 = new NPC('Mewtwo', [
+    'You were wise to seek my out. Howard Clifford has been decieving you. He is the one making the R drug, and....',
```
CorgiOnNeptune, line 336: Couple spelling errors.
'You were wise to seek me out. Howard Clifford has been deceiving you. He is the one making the R drug, and...',

src/scripts/towns/TownList.ts
```diff
@@ -323,6 +323,33 @@ const CelebiProfIvy = new NPC ('Prof. Ivy', [
...
+    'A sphere of energy envelops Mewtwo, and he is dragged away by some sort of helicopter. The helicopter has a logo on it: Clifford Industries!',
```
CorgiOnNeptune, line 337: We've been using italics to kind of denote "non-speech" text with NPCs. <i></i>
Same can be applied to the Temp Battles that aren't speech.

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -1125,6 +1125,76 @@ class QuestLineHelper {
         App.game.quests.questLines().push(ancientGolemsQuestLine);
     }
 
+ public static createDetectivePikachuQuestLine() {
```
CorgiOnNeptune, line 1128: Could you move all the Detective Pikachu quest stuff in between Celebi and Hoenn quests? To stay consistent with the organization of this file. 
Same with the `loadQuestLines()` below.

Jaaslet:
For some reason, my comment did not get added, so i will try again...
It's a new questline, so the crew needs to test it. I can set up a link in a couple of days, if you don't know how to (right now, i have set up an other test).
But we need a sort explanation of what the quest is trying to do. I am guessing the questline follows the movie? What is it balanced for? When are we expecting the player to be able to complete it?

Farboo171:
I don't know how to test it officially,  but ran through it several times during writing the changes including once start to finish with the latest code as of fixing the essential issues.

The quest is a reenactment of the Detective Pikachu movie, with a slightly different ending because I couldn't implement Ryan Reynolds as a pokemon (did Detective Raichu as seen in PoGo instead). The movie is similar the game,  but I used the movie as thr basis.

The quest is not unlocked until early Kalos, because the P2 lab is part of the quest and I wanted to make sure people could complete the quest shortly after unlocking it. Temporary battles are likewise supposed to be able to be completed as long as one can get to early kalos and unlock the quest.

Jaaslet:
(This might take some iterations...)
I have now played through it. It's very confusing, and i have no idea if that is a good or bad thing.
Most people will look up a guide for this... But that might be intended, as it is "detective" pikachu? And it will only be for players, who have already done a lot of farm mutations... So they might be used to the game not being super simple?
But if it's not intended, a lot of the quests needs to tell the player, where they are going.
And one of the lasts quests, it's weird to fight mewtwo in johto, but then need to talk to it in kanto...
And i think the temp battles might be too easy. If you think the level should be mid kalos, then take a gym from there, add the HP up and double it up. That should be the total HP for temp battles.

I will try to set a crew-test up tomorrow

Farboo171:
That commit should say "Added Hints" not hits.

I added hints to some of the more difficult to understand quest steps, so that the next location is at mentioned or clearly hinted to by name by an NPC or the quest description.

I upped the difficulty of the temporary battles so that the final boss requires a little over 3.5 million damage per attack. This puts it at a little more difficult than the Kalos docks gym (2.8 million).

Mewtwo jumping around to talk to is a little weird but it's mostly in keeping with how the movie went (other than the whole region the movie takes place in not being implemented), so I'd rather keep that if possible.

CorgiOnNeptune:
I just pushed a PR to you. I pushed some new images, also ended up moving the quests starting point to Johto Bulletin since it made more sense to have the quest start in the region with the first quest step.

Jaaslet:
We kinda made a new Pikachu in the sevii pr... Which is why this is hard to merge into...
You need to move your Pikachu to id 25.15, before we can resolve the conflicts.
I will hopefully set a test up tomorrow, if the Pikachu have been merged

src/modules/GameConstants.ts
```diff
@@ -1357,6 +1358,12 @@ export const TemporaryBattles = [
     'Spectrier',
     'Dyna Tree Birds',
     'Regigigas',
+    'Aipom Alley',
+    'Mime Interview',
+    'Underground Fighting Ring',
```
Jaaslet, line 1363: These needs to be in the same order as in the temporary battle list. And yes, that will require an update function, but you can just see how it's done above.
But yeah, it will be easier to read, if they are in the correct order

src/scripts/temporaryBattle/TemporaryBattleList.ts
```diff
@@ -75,7 +75,12 @@ TemporaryBattleList['Cue Ball Paxton'] = new TemporaryBattle(
         new TemporaryBattleRequirement('Biker Goon 3'),
     ]
 );
-
+TemporaryBattleList['Mime Interview'] = new TemporaryBattle(
+    'Mime Interview',
+    [new GymPokemon('Mr. Mime', 160225334, 100)],
```
Jaaslet, line 80: Temporary battles are sorted by when we expect the players to defeat them, not where they show up.
So because we expect players to complete them in early kalos, they should all be listed as the first, in kalos

Jaaslet (changes requested):
I will set the test up now. But i have some code changes you can do while the test is going on :)

Jaaslet:
I also just noticed that the two NPCs are missing credits

Jaaslet:
I moved the questboard again... :)
It's expected for the player to find it around Kalos, so it should be in kalos

Farboo171:
Corgi, were all those new sprites yours? I can add them to the credits, if so

Farboo171:
I'm not sure what happened, but I'm not able to get the quest to show up on the Kalos bulletin board anymore. I tried taking Johto out of the src\scripts\GameConstants.d.ts, to match up with \src\modules\GameConstants.ts but that didn't fix it

src/modules/GameConstants.ts
Farboo171: Moved and added associated lines to update.ts

Farboo171:
[New Detective Pikachu Test Save.txt](https://github.com/pokeclicker/pokeclicker/files/9436434/New.Detective.Pikachu.Test.Save.txt)

This new test file has Detective Pikachu ready, to run through the test. The old test file is still valuable to test from the enigma wanderer stage if someone wants to. I did, and it seems to have updated the temporary battle index correctly.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2881, opened 2022-08-20
Add setting to allow Quest Menu to stay open when quest slots

### LINKED ISSUE #2423: [SUGGESTION] Add setting to allow Quest Menu to stay open when quest slots are full
**Current Behavior**

Currently, as soon as you've filled your quest slots, the quest menu will close.

**Improved Behavior**

Allow a setting, like with the hatchery, that allows the user to set when the quest menu closes, choosing between the current setup of closing as soon as all the quest slots are filled, and never.

**Reason for change**

Sometimes, I grab a bunch of quests including one that I know will be done within a few seconds.  Having the option to not have the quest list auto-close on me saves me having to open it up again when that quickly done quest is completed so I can select other quests to do.  It's not a major issue, but it is annoying if I'm doing a lot of quests and the menu closes for a 'hatch 1 egg' quest or 'gain x money / DT' that is finished instantly.

**Detailed Description**

**Screenshots:**

QuantariusRay:
Add a setting to keep the display of the quest modal after selecting a quest. If the quests setting is set to 'never', the modal will remain open, otherwise it will close.

Removed the auto-close functionality in the View that was immediately closing the modal after resolving the quest selection functionality.
Closes #2423


QuantariusRay:
image

Jaaslet:
I have not tested it yet, but why is it a dropdown? If there are only two options, it's just an extra click for (as far as i can see) no real gain?

QuantariusRay:
It could be a checkbox. I can make that change.

Jaaslet (approved):
I have finished it by turning it into a boolean, and made sure it works.
Thanks for the PR

PR #2882, opened 2022-08-22
Adjust Hatchery Lock Slot Image

CorgiOnNeptune:
If you have a locked hatchery slot, the lock awkwardly hangs outside the container.
Adjusted the margin to fix it.

<details><summary>Before</summary>

before</details>

<details><summary>After</summary>

after</details>


RedSparr0w:
I feel like we should probably make the lock a bit smaller too,
I think I may have accidently enlarged it in another PR where I adjusted the size of the Pokérus symbol

CorgiOnNeptune:
Moved the margin to the stylesheet and adjusted the height/width there as well. Looks much better.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2892, opened 2022-08-24
Sevii4567 dungeons now have hoenn size

Jaaslet:
Sevii4567 is located in Kanto, but is Hoenn difficulty. This allows them to be Hoenn sized, while being in Kanto.
This also affects when dungeon loot will be nerfed.
Bonus: Ruby Path is now hidden until Hoenn is unlocked, to not confuse new players


RedSparr0w (approved):
Works great for me, Thanks for the PR! 👍

PR #2895, opened 2022-08-24
NPC Images

CorgiOnNeptune:
Giving the NPCs a little character. Literally!

Added some new images for NPCs and used existing images where appropriate. My goal is to help make NPCs just a little bit more interactive.. People hate reading, but love looking at pictures. Most people have better association with an image, so very small chance this will help with some people actually reading and understanding NPCs texts. Either way, looks good to actually interact with a person/figure instead of just as text wall.

All new images are either credited correctly or ripped from mainline games. And of course tinified to save our precious bandwith and server load.


Jaaslet (approved):
It makes the world feel a lot better. I went through some NPCs, and they look fine. I don't have time to go through them all, but if something is placed wrong, people will tell us, and it's not a big problem.
Thanks for the PR

PR #2897, opened 2022-08-24
Updated Sprites

CorgiOnNeptune:
Updated a bunch of images to standardize some sizing and re-scale some sprites that weren't scaled very well/improperly. What I'm using as a standard is any Gen 3 sprite being used is set to 64x64, while anything gen 4 or up/custom is 80x80.


Jaaslet:
Any reason why we have two standards? It would be easier if they where all the same, but i am guessing there is a reason for them to be different?

CorgiOnNeptune:
All the gym leaders / E4 / Champion are now higher quality 80x80 sprites.

I will mark this as ready, while there are some more trainer sprites I can do, they are less important to this and I will clean some more of those up over time. 

Hoenn is a mix of remade sprites and a handful of B2W2 official sprites. Kanto are all remade custom sprites.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2900, opened 2022-08-25
New EV Math

jk13pclick:
New EV formula post 50 EVs. Makes the line nice and continuous. Slight nerf to EVs Attack Bonus immediately after 50, larger (in magnitude) buff 100EVs. Also uncaps the amount of EVs that can apply to the attack bonus. We can add this back if desired, but I don't think it is especially radical. Desmos graph provided containing a relevant threshold comparison table.

Graph: https://www.desmos.com/calculator/2wsrhpeqrk


Jaaslet (approved):
The numbers looks fine. My attack went down by 13k on my main save... So basicly the same.
It will give a reason to continue catching mons.
Yeah, i think this is safe to release.
Thanks for the PR

PR #2902, opened 2022-08-25
Eternamax Region.none

Qwertypop04:
Eternamax is currently galar native, which makes him appear in battle frontier. This PR makes it none native, to be consistent with the gigantamax pokemon


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2905, opened 2022-08-26
Sevii NPCs

Qwertypop04:
NPCs for Sevii
Also made Ruby disappear immediately instead of after next step in questline


Jaaslet (approved):
Looks good. The image did not have a file extension, so i fixed that.
I am guessing the image is taken from the game, and that is why there is no credit?
But thanks for the PR

PR #2906, opened 2022-08-27
Battle Frontier : the earnings of one complete run can now split.

RegisCoaxans:
Well, I know how this may be controversial but hear me out !

I have made a couple of changes touching the Battle Frontier :
1. Added the ability the leave earlier **with earnings**.
2. Added a new variable `lastEarningsCheckpoint` into data, which stores the last stage whom the player gets his earnings at.
3. Twisted the computation of the BF earnings so the above earnings get counted down from the next earnings.
4. Added a new display in the BattleFrontierInfo component so a message tells the player about that `lastEarningsCheckpoint`.
5. Added a new log entry when the player exits with earnings.
6. Changed the notification and entry contents : it does not mention earnings if there are none.
7. (Minor) the Pokédollars earnings are correctly displayed now.

The pursued objective is not to make BF generate more but to make their generation feel smoother to the players.

Currently, getting earnings from the BF means completing the whole run, from 1 to the end. You sure can quit the BF and go back later but your time spent does not grant you anything unless you finish the run; meaning that you only postpone your income. When you do not constantly do runs, it is not even worth as it feels like a waste of time before you eventually get the earnings that you deserve : you spend more time **waiting** for the run **to end** than actually being able to spend your earnings...

Then, how does it would work now ? Well, there are few circumstances that changed a bit.
Lets assume that you previously exited at stage X before and earned `X ^ 2` Pokédollars :
- Going back in at stage X + 1 and to stage Y, will give you `Y ^ 2 - X ^ 2` Pokédollars once you exit again or lose. Thus, the two runs actually give you the expected total of `Y ^ 2` Pokédollars of one run.
- Going back in at stage 1 to stage Y **greater than X** will also give you `Y ^ 2 - X ^ 2` Pokédollars once you exit again or lose. Same amount, but with the extra first stages not getting you anything more.
- Going back in at stage 1 to stage Z **lesser than X** will give you nothing as `Z ^ 2 - X ^ 2` is lesser than 0 (in fact, you already got those earnings before).

There is only two ways of updating that `lastEarningsCheckpoint` :
- Exiting at a higher stage (X > `lastEarningsCheckpoint`), X is affected to `lastEarningsCheckpoint`.
- Losing at a higher stage (X > `lastEarningsCheckpoint`), 0 is affected to `lastEarningsCheckpoint`; the whole thing is reset and you get earnings from beating early stages again since you fairly lost the run.

Here are the advantages of such a case :
- Any time spent **pushing a run further** is rewarded; less pain to the players.
- Battle Points become a more consistent currency. Not that inconsistency is bad at all, but this kind of inconsistency is (it is like upgrading a building in Clash of Clans but you can barely do anything else beside).
- It makes screwing the run from reducing DPS less useful as you can already and fairly get your income while keeping the ability to pursue the run later.
- It does not change anything to alternative reasons to do runs (leaving early for egg steps or XP will still not grant BP and Pokédollars).
- It actually encourages pushing a run further because each time you do so feels fruitful when you pause it.
- It does not even need to increase the gains, or so my calculations say.

The only real downside that I can think of is that earnings would feels more random if you do not know how it works. And I have added the display for that to be avoided.
A display of the amount of BP and Pokédollars that the runs is currently worth would also be interesting. I am up to make it if required.

This whole explanation could have been shorter but, as I am not only suggesting an update but have made it so it can be tested better, I want the idea behind it to be clear.


Jaaslet:
I am going to close this PR.
Even with the explaining text, it's still confusing. And i think few people will really have any benefit of it

Issue #1358, opened 2021-04-22
[SUGGESTION] Let us choose the maximum number of pokemon my hatchery can hold.


Mega-Rayquaza-8121:
**Current Behavior**
 Currently, the game increases our hatchery size by exponents of 2 after we beat each region. 

**Improved Behavior**
 However, I'd like us to be able to choose the number of pokemon we want our hatchery to hold. For example, there could be a drop down menu that says "Max pokemon in hatchery: <choose>" Or something along those lines. Let us choose between 4 mons, 8 mons, 16 mons, 32 mons and so on. Of course the larger numbers should only be unlocked once we advance to a new region.

**Reason for change**
 Breeding is arguably better in a smaller hatchery instead of a big one. (personal opinion, don't grill me for it). I feel that with a smaller hatchery, it is easier to focus on a set of pokemon and breed them to shiny before moving on to others. It also allows for faster re-breeding of desired mons. Again my brain works in a weird way don't kill me for it

**Detailed Description**
 Image below. Specifically i'd like the number of mons that can be held in the circled part to not be fixed based on advancement, but i want us to be able to choose how many mons it can hold. Like that part currently holds 64 mons (i think) because I am in kalos, but I want to be able to choose to reduce it to 32 or 16 (u get the idea hopefully)

**Screenshots:**

![image](https://user-images.githubusercontent.com/62550729/115718724-fb8e2a80-a3ad-11eb-9285-85dacb2bcce6.png)

Mega-Rayquaza-8121:
hi please comment your thoughts!! i know my request is weird but im sure im not the first one to favour a smaller hatchery over a bigger one~

PR #2912, opened 2022-08-28
Hatchery queue QoL improvements

CypherX:
Adds several QoL improvements to the hatchery queue:

- Display amount of pokemon currently in queue and total amount of queue slots available.
- Add a Clear button to clear the entire queue.
- Setting to limit the queue size to any of the previously obtained regional increments (4, 8, 16, 32, 64, 96) or disable the queue entirely.
    - From: #1358
    - Queue size setting will not be displayed if the queue is not unlocked.
    - Switching to a smaller queue size will remove any extra pokemon from the queue.

image
image


src/scripts/breeding/Breeding.ts
```diff
@@ -456,4 +463,35 @@ class Breeding implements Feature {
         }
     }
 
+    private usableQueueSlots = ko.pureComputed(() => {
```
Jaaslet, line 466: It should be public, now that it's used in the html. Works without, but private might make people think that it's not used elsewhere

Jaaslet (changes requested):
Looks good. I have some small changes:
- I think the clear button needs a confirmation dialog. It's easy to hit, and can ruin a lot of clicks.
- The queue size should be moved to the settings window. It will not be changed by many, and the people who wants to use it, will only change it once, so no need to use UI space in the hatchery

CypherX:
Thanks for the feedback.

- Added confirmation to the Clear queue button.
- Made usableQueueSlots public.
- Moved the queue size setting to the regular settings window under the Other tab. Let me know if there's a better way to handle the setting, the options being dynamic based on player progression make it a bit different from the other settings that have static options.

RedSparr0w:
The current way the setting is implemented, it won't update when the player travels to a new region.
You could bind to update the settings whenever the `queueSlots` value is updated.
Or just allow the player to enter any number instead of having a dropdown (probably add a bind to limit to max amount on change).

CypherX:
Alright, changed the setting to use a textbox instead of dropdown to allow the player to enter any (valid) number. My only reservation is using -1 to denote 'no limit' (also the default value) may be confusing to some players. I've included a tooltip to hopefully help with any confusion. I can't see many people wanting to use this setting anyway, lol.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2915, opened 2022-08-29
Merging some credits

Jaaslet:
Some credits had gotten split up. This should fix it


RedSparr0w (approved):

PR #2919, opened 2022-08-30
Lock placement for dungeon bosses

CorgiOnNeptune:
Before:

After:


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2924, opened 2022-08-31
Fix Unova E4 teams and change E4 to be challenged in any order in Unova, Kalos and Alola

Qwertypop04:
Changed the Unova E4 to have their initial team instead of their rematch team, so as to match every other gym leader and E4 member. Also changed the E4 of Unova, Kalos and Alola to be clearable in any order, to match how they work in the core games


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2927, opened 2022-09-01
Fixed Alola island 3 blocked

Jaaslet:
So... this bug is weird...
When Ultra wormhole was turned from a temp battle, into a gym, we just moved the "completed" count over... So a lot of old players had defeated it more than once...
And for some reason, the default was changed from GameConstants.AchievementOption.more to equal in the latest update for TemporaryBattleRequirement, which broke the ultra wormhole


RedSparr0w (approved):

PR #2929, opened 2022-09-01
Pinkan Dodrio

Qwertypop04:
Makes Pinkan Dodrio the Tanoby Ruins shop mon. Also adds Pinkan Exeggutor to Pinkan Forest and refunds the QP spent on it


Jaaslet (approved):
I have fixed the update function. The old code always returned an array empty, which is true in typescript (weird, i know), while it now returns the length. And 0 is false.
But thanks for the PR!

PR #2930, opened 2022-09-01
Now showing orange league badges

Jaaslet:
Orange league badges does now show up, when you have gained the first badge


RedSparr0w (approved):

PR #2932, opened 2022-09-01
Fixes egg steps of Detective Raichu and Darmanitan (Zen)

Qwertypop04:
They have been increased to the amount their base form has.


RedSparr0w:
I'd rather we put unachievable evolution requirements for their base forms where possible than adjust the egg steps manually _(see Phione)_,
As otherwise this creates a lot of work if the formula for egg steps ever changes.

Qwertypop04:
That makes sense. I've changed them to work like Phione

RedSparr0w (approved):

PR #2933, opened 2022-09-01
Fix Plasma Questline NPCS

CorgiOnNeptune:
I goofed the reqs..


Jaaslet (approved):
I have played through the quest again with your changes, and i could get through it, and stuff looks correct.
Thanks for the PR

PR #2935, opened 2022-09-02
Hatchery stores Pokémon ID instead name in the save

Crobat4:
This PR will make hatchery stores the Pokémon IDs instead the names, as well as update the data of the Pokémon in the hatchery.


src/scripts/breeding/Breeding.ts
```diff
@@ -285,7 +285,8 @@ class Breeding implements Feature {
             this._eggList[index](new Egg());
             this.moveEggs();
             if (this._queueList().length) {
-                const nextEgg = this.createEgg(this._queueList.shift());
+                const pokemon = PokemonHelper.getPokemonById(this._queueList.shift());
+                const nextEgg = this.createEgg(pokemon.name);
```
Jaaslet, line 289: I think createEgg should take the id as a parameter. Yes, you need to change createFossilEgg and createTypedEgg to fetch the pokemon first, but this code happens way more often, so removing a lookup here, should be better

src/scripts/breeding/Egg.ts
```diff
@@ -136,26 +136,26 @@ class Egg implements Saveable {
...
                 setting: NotificationConstants.NotificationSetting.Hatchery.hatched,
             });
         }
         App.game.party.gainPokemonById(pokemonID, shiny);
 
         // Capture base form if not already caught. This helps players get Gen2 Pokemon that are base form of Gen1
-        const baseForm = App.game.breeding.calculateBaseForm(this.pokemon);
-        if (this.pokemon != baseForm && !App.game.party.alreadyCaughtPokemon(PokemonHelper.getPokemonByName(baseForm).id)) {
+        const baseForm = App.game.breeding.calculateBaseForm(PokemonHelper.getPokemonById(this.pokemon).name);
+        if (PokemonHelper.getPokemonById(this.pokemon).name != baseForm && !App.game.party.alreadyCaughtPokemon(PokemonHelper.getPokemonByName(baseForm).id)) {
```
Jaaslet, line 158: This file has a couple of lookups. Can we do a const instead?

src/scripts/Update.ts
```diff
@@ -1129,6 +1129,16 @@ class Update implements Saveable {
             if (saveData.party.caughtPokemon.filter(p => p.id === 103.02).length) {
                 saveData.wallet.currencies[1] += 50000;
             }
+
+            // Replace Pokémon names to IDs
+            saveData.breeding.eggList?.forEach(pokemonName => {
+                const pokemonEgg = PokemonHelper.getPokemonByName(pokemonName);
+                Update.renamePokemonInSaveData(saveData, pokemonName, pokemonEgg.id);
```
Jaaslet, line 1138: It's maybe a bit hack to use this function here. I was confused when i read it.
I am not totally sure how i think it should look, but if people uses `renamePokemonInSaveData` later on, it will break the save. So we either need to rename the function, or remove it. But if we just rename it, the old update functions will look weird...
If you can't find a good solution either, i will try to find something, when i get the PR back

src/scripts/breeding/Egg.ts
```diff
@@ -41,8 +41,8 @@ class Egg implements Saveable {
             return this.totalSteps - this.steps();
         }, this);
 
-        if (this.pokemon) {
-            const dataPokemon: DataPokemon = PokemonHelper.getPokemonByName(this.pokemon);
+        if (this.pokemon != 0) {
```
Jaaslet, line 44: `!= 0` should not be needed. 0 will always be false

Jaaslet (changes requested):
Yay! Refactor!
I have moved the update function to 0.9.15. It works!
I have clicked around in the hatchery and tried different stuff, and nothing broke.
I have some small code comments for you:

src/scripts/Update.ts
Crobat4: From what I understand, `renamePokemonInSaveData` is only used to change the Pokémon name in the hatchery and the queue, so with these changes, it wouldn't be needed anymore (unless you want to change the Pokémon ID).
Would adding `renamePokemonInSaveData` later on actually break the save? Let's say I want to change the Nidoran (F) name using the rename function, the function wouldn't find `Nidoran (F)` in the hatchery because it's saved as `29` and just do nothing.
It works like this, right?

Jaaslet: Yeah, you are right. Should not break anything. But the function will then be sorta useless in later update functions, so we can maybe rename it to `changeHatcheryKey` or something?

Crobat4:
Updated with the changes, from what I tested, it's working fine.
Tell me if something else is needed or if something is broken.

Jaaslet (approved):
I just noticed that the `createEgg` method does still need the eggstep, so yeah we are not saving the lookup i thought. But i still think it looks better, so let's keep it.
But yeah, looks good and still works.
Thanks for the PR

PR #2937, opened 2022-09-02
Sevii 4567 shop and dialogue tweaks

HyruleTeam64:
Changed Prof. Ivy's dialogue to say Dodrio instead of Exeggutor, added some Johto and Hoenn items to shops and changed Pummelo island's shop into a League Shop.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2939, opened 2022-09-03
Show images in shop notifs & sound plays when buying new Pokémon

CorgiOnNeptune:
Added sound to notification for newly purchased Pokémon (that are previously uncaught/shiny). Also added a notification for trades with Berry Masters and added images to notifications for all purchases/trades.

The sound also works for new trade Pokémon (shard Furfrou).


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2940, opened 2022-09-03
Ash\'s and Pink Butterfree

Qwertypop04:
Adds Ash\'s Butterfree and Pink Butterfree as roamers in the Sevii Islands after beating the Orange League Champion


Jaaslet (approved):
Sevii roamers! Yay!
Thanks for the PR

PR #2943, opened 2022-09-03
Make Togepi, Riolu and Toxel baby Pokemon

Qwertypop04:
Lowered Pikachu forms egg cycles to the same as their base form.
Made Togepi, Riolu and Toxel babies due to them being such in canon, and to give them the egg steps babies are supposed to have. Also lowered event Togepi egg steps to match the egg steps their base forms will now have, as is done for Spiky-Ear Pichu


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2945, opened 2022-09-04
Five Island dock can now be selected in the town

HyruleTeam64:
Made the Five Island dock accessible in the town itself like all other docks.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2946, opened 2022-09-04
Princess Diancie Questline

Farboo171:
A quest for unlocking the pokemon Diancie. This quest:

1. Will allow players to unlock Diancie after beating the 4th Elite 4 member, with a lot of effort
2. Will provide Aegislash Blade Forme as a bonus reward
3. Is balanced to be more difficult than the first four E4 members, but weaker than Diantha


Jaaslet:
I just noticed that this quest is not optional. You need it to complete the Kalos dex.
So we might want to lower the grinding needed

Farboo171:
The quest is placed to give players something to work on during the Diantha grind, and is pretty easy to get AFK if they sit on the right routes for a few hours.

The underground might be a bit excessive, though.

Would 250 catch, 25 layers, and 2500 gems be a better set of requirements?

Jaaslet:
We don't want to use optional pokemon for grinding quests. Stuff that takes a long time to grind, should be optional, so you can always skip it with de recommended challenges.
And we kinda want to get rid of that Diantha grind. At least take her a lot closer to the rest of the Elite 4

Farboo171:
Lowered down to 100 catches, 5 layers, and 1000 gems. That's a little more intense than a set of daily quests, but should be easy enough for people to do quickly.

Jaaslet:
I have gotten some feedback from others:
- The other legendary quests includes a step, which is to catch the pokemon. This one should too.
- The movie involves Yveltal. Is there a reason why it's not mentioned in this questline?

Farboo171:
1. I'll add a step to catch Diancie, that makes sense
2. The movie plot for Yveltal and Xerneas was much much different from the game (no Team Flare involved, there's a forest that gets destroyed and reborn), so I didn't want to mix together the existing implementation of those two mons with the movie / this quest.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2948, opened 2022-09-04
Hiding Orange badges until first unlocks

Jaaslet:
So my original fix also made sure that the badges did not show up before the first was unlocked, but i kinda forgot to push that commit... So here it is!


Aegyo (approved):
:+1:

PR #2951, opened 2022-09-04
Unnecessary Sounds

### LINKED ISSUE #2923: [SUGGESTION] Make the "Egg ready to hatch" noise not proc when a helper hatches an egg
**Current Behavior**
Currently the "egg ready to hatch" noise plays every time an egg is ready to hatch. Be it from the hatcher or not. In other words, the noise plays when the hatcher hatches an egg despite the fact they they immediately grab a new egg.

**Improved Behavior**
I would like it if eggs hatched by hatchers made no noise since no user input is required.

**Reason for change**
It would make more sense for said noise to only be used so the user knows a new interaction can take place, in this case, hatch the egg and place a new one.

**Detailed Description**
The fact the hatcher also makes the noise is distracting since the noises are meant to grab the user's attention and, in the case of automatic hatching, there is no interaction to be made.

jk13pclick:
Closes #2923

### DIFF
--- a/src/scripts/breeding/Egg.ts
+++ b/src/scripts/breeding/Egg.ts
@@ -67,7 +67,7 @@ class Egg implements Saveable {
         this.shinyChance = newChance;
     }
 
-    addSteps(amount: number, multiplier: Multiplier) {
+    addSteps(amount: number, multiplier: Multiplier, helper = false) {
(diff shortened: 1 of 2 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2953, opened 2022-09-05
Tooltip Orientation

### LINKED ISSUE #2886: [SUGGESTION] Hide/Move popup windows on Underground buttons
Windows obscure the underground map

**Current Behavior**
This is a minor annoyance, but w
Screen Shot 2022-08-23 at 2 38 56 PM
hen you hover over (and therefore as you click) the black windows on the buttons can obscure some of the spots at the bottom of the underground map

**Improved Behavior**
Maybe make the black popups transparent, or make them appear below the buttons rather than above

**Reason for change**
Would remove impaired view of the underground

**Detailed Description**

**Screenshots:**
Screen Shot 2022-08-23 at 2 38 56 PM

jk13pclick:
Closes #2886


Jaaslet (changes requested):
I don't think moving them right is going to fix the problem, just move the problem.
What i suggest is to set left tooltip on the left buttons, right tooltip on the right buttons and set top/bottom in the middle

jk13pclick:
> I don't think moving them right is going to fix the problem, just move the problem. What i suggest is to set left tooltip on the left buttons, right tooltip on the right buttons and set top/bottom in the middle

The left-right distinction is good. The ``bottom`` placement doesn't work in the modal, probably because it is too close to the bottom? I have left the two center ones as ``right`` as a result.

Jaaslet:
Now all of them are covering something.
It looks like tooltips can't go outside the div it's in, as a default... So we need to figure out how to stop that, before my solution can be used.
I will look at it later, but i am writing this comment, in case i forget. Here is a mockup of the positions i was thinking about using:
image

Jaaslet (approved):
I have moved the placement around a bit, and i think i am happy now.
Thanks for the PR

Issue #2783, opened 2022-07-29
[SUGGESTION] Multiselect filters for hatchery on Region and Type


jaahay:
**Current Behavior**
My helpers can only be configured by:
- Region IN `[ ALL | <regions> ]` AND
- Type IN `[ ALL | <types> ]` AND type IN `[ ALL | <types> ]`

**Improved Behavior**
Region and type support multi-selection. This multi-selection will behave as an OR operator.

**Reason for change**
- smarter helpers
- less clicks to enqueue (ex., I want to enqueue all eggs that are not shiny from either Kanto or Johto)

**Detailed Description**
- Allow multiple selections for region
  - ex., either Kanto or Johto
- Allow multiple selections for type
  - ex., either Grass or Fire

PixLSteam:
Related to/partial duplicate of #2762

Alderi-Tokori:
PR #2955 made this exact behaviour available for regions but not types.

Awec4:
PR #4046 adds this behaviour for types

Issue #2762, opened 2022-07-26
[SUGGESTION] Multiple Region filter for dex and | or Hatchery


lenormandSeb:
**Current Behavior**
Only one Region can be filter

**Improved Behavior**
Enhance filter by region with possibility to filter through multiple region

**Reason for change**
Can be usefull for breeding

Aegyo:
Closing as implemented in #2955

PR #2955, opened 2022-09-05
Hatchery : users can now select several regions on the filter.

RegisCoaxans:
Somehow related to [#2783](https://github.com/pokeclicker/pokeclicker/issues/2783) and [2762](https://github.com/pokeclicker/pokeclicker/issues/2762).

There, I have made the region filter a bitfield so all the regions can fit in and the filter allows for multi-selection.
The value from older versions correctly updates and the default value becomes `1` (as that is set to this in Kanto).

The interface looks like that for now and only displays regions up to the highest one :
image


src/components/breedingRegionModal.html
```diff
@@ -0,0 +1,27 @@
...
+				<h5>Regions</h5>
+				<button type="button" class="close" data-dismiss="modal" aria-label="Close">
+					<span aria-hidden="true">&times;</span>
+				</button>
+			</div>
+			<div class="modal-body p-0">
+				<ul class="list-group" data-bind="foreach : Object.keys(GameConstants.Region).filter(function(r){ return r <= player.highestRegion() && r >= 0; })">
+					<li class="list-group-item">
+						<label class="col-6" data-bind="attr : {for : 'breedingRegion' + $data}, text : GameConstants.camelCaseToString(GameConstants.Region[$data])">Region</label>
+						<input class="col-5"type="checkbox" data-bind="attr : {name : 'breedingRegion' + $data, id : 'breedingRegion' + $data, onchange : 'var bfield = BreedingFilters.region.value() ^ ' + (1 << $data) + ' | (this.checked ? ' + (1 << $data) + ' : 0); BreedingFilters.region.value(bfield); Settings.setSettingByName(\'breedingRegionFilter\', bfield);'}, checked : !!(Settings.getSetting('breedingRegionFilter').observableValue() & (1 << $data))"/>
```
Jaaslet, line 14: Some linebreaks would make this more readable

Jaaslet (changes requested):
Sorry it has taken me so long to comment on this. I have had a hard time figuring out what feedback to give.
It takes a lot more now, to just for example switch region. You need to wait for a modal to open, uncheck the old one and check the new one. It's weird to make it harder to use the UI, for a feature i don't think many players will use. Either we need to come up with something smarter, or it needs to be a setting.
And it's weird that you need to open it, to see what region(s) are selected. It could at least be something like `Johto + 2 more` if you have Johto+kalos+alola selected (for example).

RegisCoaxans:
Well, I can not agree more, it adds more click to the UI...
I am pretty sure that this multiselection will satisfy enough people to make it worth the code, though.
I am all about changing it.

Enabling that menu via a setting is an easy improvement; I can do that.
I had another idea but I did not really like it as is : the select comes back and rightclicking options stacks them, while clicking still sets the region as the selected one.
I know that multi-selects are a thing but they are not native to BootStrap, and they do not save clicks at all. Moreover, rightclicking does not close the select menu, clicking does. The actual issue rather is that it feels hidden as rightclicking in a select is not intuitive.
I like the idea of a setting better but I wanted to share this idea that I had when I made this PR as a hint for a better solution.

Jaaslet:
> I had another idea but I did not really like it as is : the select comes back and rightclicking options stacks them, while clicking still sets the region as the selected one.

I think i like this idea better. But if you don't think you can do it in a good way, the setting solution is fine too

RegisCoaxans:
Back from business ! Sorry for the late change.

Select could not fit in because `<option>` fires no rightclick event and the select text can not be updated as expected...
However, I have managed to make it work without extra interface or setting thanks to Bootstrap's dropdown menu.

It seems to stick out a bit compared to the old select menu but I think that it only looks weird at first.

image

Anyway, it works as wanted :
- clicking sets the selected region as the clicked region.
- rightclicking stacks the clicked region on top of the other selected regions.
- the dropdown toggler tells the highest selected region and if there is any other.

Now, I guess that it still misses a way to inform the players that such a thing exists.

Jaaslet:
The game does not load. I get the following error:
image

RegisCoaxans:
I had this error once.
Did you just load a file whom you already tested the thing with in a previous dev version ? It can not update the filter twice but the released version changed inbitween. This crash happpens when the filter value does not match with a region ID anymore, which should not occur with a proper file.

Anyway, I could add a test for it if you feel like it is required.

Jaaslet:
I most likely did... I have reimported my main save, and it works fine now. I will test it now

Jaaslet (approved):
I have cleaned up the code a bit (we don't use tabs here), made sure it says "all" when all is selected and made sure it does not trigger right click menu, when that setting is off.
When you cover over, it covers the text. I had no idea how to fix it, and i feel like it's a small bug, so i am okay with merging without.
Thanks for the PR

PR #2967, opened 2022-09-08
Ilex Forest re-forested

DataCrusade:
Turned Ilex Forest into an actual forest rather than just being a generic dungeon sprite on the map


Jaaslet (changes requested):
The overlay for the forest needs to be ajusted a bit. It's a bit to the right.
Other than that, it looks good

DataCrusade:
should be fixed now

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2969, opened 2022-09-08
Lazy load pokedex and hatchery elements

Aegyo:
Defers loading pokedex and breeding modal items until you scroll to the bottom, shows a spinning pokeball when there are more things to load.

Fixes these modals taking a minute to open on first load.

`lazyLoad` might be useful in other places, tried to make it generic - anywhere we do a `foreach: observableArray()` can hopefully be simply replaced with `foreach: lazyLoad($element, observableArray)` and not need any other changes to get it working.


Jaaslet:
I am doing to tests to see how much quicker stuff is, and it's hard to compare. Have you done any tests?
I am not sure if filtering is slower now, or it's the same

Aegyo:
it was noticeably faster for me with a large party on first opening the modals, I think I get some actual numbers for you later though.

Filtering shouldn't have any difference in speed, or maybe a very small imperceptible one with the extra knockout layer. I'll see about getting numbers for that too.

Aegyo:
Taken some measurements on my laptop with a few different save files for various situations.
All times are an average of 5 measurements.

| | New Save | Hoenn Save | Alola Save |
| --- | --- | --- | --- |
| Pokedex Initial Modal Open (lazy load) | 176.67ms | 236.67ms | 303.34ms |
| Pokedex Initial Modal Open (current) | 73.34ms | 913.35ms | 2153.30ms |
| | | | |
| Pokedex Subsequent Modal Open (lazy load) | 56.67ms | 66.65ms | 193.38ms |
| Pokedex Subsequent Modal Open (current) | 30.01ms | 96.67ms | 186.67ms |
| | | | |
| Pokedex FIltering (lazy load) | 43.34ms | 130.00ms | 153.38ms |
| Pokedex Filtering (current) | 36.67ms | 263.34ms | 400.01ms |

Time saved in lazy load filtering is likely because I didn't scroll down to increase number of pages loaded - if I did that I expect it would have similar numbers to the current filtering.

The Hatchery is a bit more difficult to time since it opens almost immediately but takes a moment to load content, the method I used for pokedex didn't work well for measuring it. I do also notice a similar significant time save on first open for those saves with a large party however.


Code used for timing modal open
```js
const timeOpen = (modalId) => {
  console.time(modalId);
  $(modalId).modal("show");
  requestIdleCallback(() => console.timeEnd(modalId));
}
```

Code used for timing filtering
```js
const timeFilter = (region) => {
  console.time("filter")
  $('#pokedex-filter-region').val(region)
  PokedexHelper.updateList()
  requestIdleCallback(() => console.timeEnd("filter"))
}
```

Jaaslet (approved):
Looks good. Thanks for the PR

PR #2974, opened 2022-09-10
Possible fix for the disappearing pokemon bug

CypherX:
After some extensive testing I believe the primary cause of the hatchery eating pokemon is this piece of code in Egg.ts:

```
// Deferring this because it wants to access App.game.party, which isn't avaliable yet
setTimeout(() => {
    this.partyPokemon(this.type !== EggType.None ? App.game.party.getPokemon(PokemonHelper.getPokemonById(this.pokemon).id) : null);
}, 0);
```
Yes, I realize this bug initially appeared before this deferment was moved from Breeding.ts to Egg.ts in #2758, but keep reading.

I was able to replicate the issue on an old, low-spec laptop using the gym auto-repeat feature with a gym I could one-shot click and an autoclicker set to 100ms. I could not replicate the issue using the exact same setup on a high-end desktop or mid-tier laptop. All testing was done in firefox.

In some cases, mainly on low-spec hardware, it appears that the deferred statement doesn't get executed before the egg is hatched resulting in the breeding flag not getting set to false. The browser may delay executing this timeout if the event loop is busy with other calls.

After making the changes in this PR I repeated the same test on the low-spec laptop and was no longer able to replicate the bug, pokemon no longer disappeared when hatching.

The timeout was originally needed to defer that statement until after loading when App.game.party was available, it served no purpose once the game was loaded. My solution is to get rid of the timeout entirely and use the checkAndFix() function after loading to set the Egg partyPokemon for the loaded eggs, that way App.game.party is available and the execution order is guaranteed for all eggs.


Jaaslet (approved):
It did not break anything. And we kinda need to release it, to test if it fixed anything. I will merge it.
Thanks for the PR!

PR #2975, opened 2022-09-10
Added update function for new Galar gyms

Jaaslet:
5 new badges got added to Galar. Because Orange League is in the bottom, we need an update function for this. This PR adds that update function.

I tried making a better solution, so we don't need an update function each time we add new badges, but could not find a clean way of moving the Orange Leagues badges up the list...


RedSparr0w (approved):

PR #2976, opened 2022-09-10
Evolve Milcery

Jaaslet:
So in the game, the player needs to give Milcery a sweet, and spin around to evolve it into Alcremie.
I am trying to build a small modal, which replicates this behavior, so the player can all the forms.
I am still trying to figure out the details, so i will update this PR, as i work on it.

But the basic idea is that you have some daily spins. When you use a spin, you "buy" a sweet for berries. Then you select a direction and a duration. You then get an evolution based on all these information


src/scripts/towns/BattleCafe.ts
```diff
@@ -0,0 +1,253 @@
...
+                ];
+            case GameConstants.AlcremieSweet['Ribbon Sweet']:
+                return [
+                    {berry: BerryType.Chople, amount: 20},
+                    {berry: BerryType.Haban, amount: 20},
+                ];
+        }
+    }
+
+    private static buildEvolutions() : {[sweet: string] : {[spin: string]: PokemonItem}} {
```
Aegyo, line 172: This looks like it returns the same data every time, can we just make it a static value instead of a function we call every spin?

Also I think we can avoid the double enum wrapping if you pass sweet/spin in directly instead of as `AlcremieSweet[sweet]` etc

Jaaslet: Yeah, i don't know how i ended up with that. That function had some iteration.
But i have fixed it now

src/scripts/towns/BattleCafe.ts
```diff
@@ -169,85 +169,85 @@ class BattleCafeController {
...
-        evolutions[GameConstants.AlcremieSweet[GameConstants.AlcremieSweet['Ribbon Sweet']]][GameConstants.AlcremieSpins[GameConstants.AlcremieSpins.dayClockwiseAbove5]] = new PokemonItem('Alcremie (Ribbon Caramel)', 0);
-        evolutions[GameConstants.AlcremieSweet[GameConstants.AlcremieSweet['Ribbon Sweet']]][GameConstants.AlcremieSpins[GameConstants.AlcremieSpins.at7Above10]] = new PokemonItem('Alcremie (Ribbon Rainbow)', 0);
+        evolutions[GameConstants.AlcremieSweet['Strawberry Sweet']] = {};
+        evolutions[GameConstants.AlcremieSweet['Strawberry Sweet']][GameConstants.AlcremieSpins.dayClockwiseBelow5] = new PokemonItem('Alcremie (Strawberry Vanilla)', 0);
+        evolutions[GameConstants.AlcremieSweet['Strawberry Sweet']][GameConstants.AlcremieSpins.dayCounterclockwiseBelow5] = new PokemonItem('Alcremie (Strawberry Ruby Cream)', 0);
+        evolutions[GameConstants.AlcremieSweet['Strawberry Sweet']][GameConstants.AlcremieSpins.nightClockwiseBelow5] = new PokemonItem('Alcremie (Strawberry Matcha)', 0);
```
Aegyo, line 177: I was thinking something like this would be cleaner:
 
```ts
    private static evolutions: Record<GameConstants.AlcremieSweet, Record<GameConstants.AlcremieSpins, PokemonItem>> = {
        [GameConstants.AlcremieSweet['Strawberry Sweet']]: {
            [GameConstants.AlcremieSpins.dayClockwiseBelow5]: PokemonItem('Alcremie (Strawberry Vanilla)', 0),
            [GameConstants.AlcremieSpins.dayCounterClockwiseBelow5]: PokemonItem('Alcremie (Strawberry Ruby Cream)', 0),
            [GameConstants.AlcremieSpins.nightClockwiseBelow5]: PokemonItem('Alcremie (Strawberry Matcha)', 0),
            ... 
        },
        [GameConstants.AlcremieSweet['Love Sweet']: {
            ...
        },
        ...
    }
```

Jaaslet: Thanks. I tried to do this, but i had no idea how the type should look... But it works! I have submitted it

Aegyo (approved):
I think this needs a few things before release, but this all seems to work fine :+1: 

It needs a Help tab explaining to use it + when you get more spins, and maybe something to track which Alcremie you have managed to get so far?
Possibly a tooltip on the berries for people who don't know them by picture.
Sweets you can't afford could be greyed out and pop the "not enough berries" notif when trying to click?

All of that can be done separately though, doesn't have to be on this PR

PR #2979, opened 2022-09-11
Only put origin and pathname in discord login redirect uri

Aegyo:
Fixes people having issues linking discord when they have a hash in the url (eg https://www.pokeclicker.com/#)


RedSparr0w:
Unfortunately this won't work with the client as it will just redirect them back to `file://`

RedSparr0w (approved):
Nvm, am big dumb, misread part of it.

PR #2982, opened 2022-09-12
NPC cleanup

Jaaslet:
Tech Enthusiast is no longer needed, as the chests now shows that they make the pokemon stronger, and they should already have noticed that they can see the dungeon, after they have opened some chests.
Big Spender was too spammy... Yes, we get it. You are rich and travel a lot! Now stop following us around!


RedSparr0w (approved):

PR #2984, opened 2022-09-12
Alola and Galar map updates

Qwertypop04:
Seperates Ula'ula and Poni into seperate subregions, adds Poni map, and updates, South Galar, North Galar and Crown Tundra maps


Jaaslet (approved):
The maps looks good. Thanks to the mapmakers and thanks for the SVG!
Approved!

PR #2985, opened 2022-09-12
Hide Snorlax trainer during temp battles

Qwertypop04:


Jaaslet (approved):
I forgot... Thanks...

PR #2990, opened 2022-09-15
Support for Shopmon EVs

jk13pclick:
Multiple Mon purchases supported, code could be made smoother but this will work


Jaaslet (approved):
The code looks correct, and the numbers in my testing looks correct too. At least it's a lot better than the old code.
Thanks for the PR

PR #2991, opened 2022-09-16
Added gem furfrou

Jaaslet:
Adds gem furfrou + some gem master code cleanup.
Still not sure about the numbers, but i don't think we will get closer to the perfect numbers, so it's ready for review.


RedSparr0w (approved):

PR #2993, opened 2022-09-17
Boost new Pokemon egg steps

Qwertypop04:
Boosts the egg steps of Pinkan Arbok, Poliwhirl and Weezing, and Aegislash (Blade) to match their base forms

### DIFF
--- a/src/scripts/pokemons/PokemonList.ts
+++ b/src/scripts/pokemons/PokemonList.ts
@@ -645,7 +645,7 @@ const pokemonList = createPokemonArray(
     {
         'id': 12.02,
         'name': 'Pink Butterfree',
-        'catchRate': 15,
+        'catchRate': 45,
         'type': [PokemonType.Bug, PokemonType.Flying],
         'levelType': LevelType.mediumfast,
         'exp': 178,
@@ -662,7 +662,7 @@ const pokemonList = createPokemonArray(
     {
         'id': 12.03,
         'name': 'Ash\'s Butterfree',
-        'catchRate': 15,
+        'catchRate': 45,
         'type': [PokemonType.Bug, PokemonType.Flying],
         'levelType': LevelType.mediumfast,
         'exp': 178,
@@ -1927,7 +1927,7 @@ const pokemonList = createPokemonArray(
     {
         'id': 45.02,
         'name': 'Pinkan Vileplume',
-        'catchRate': 20,
+        'catchRate': 45,
         'type': [PokemonType.Grass, PokemonType.Poison],
         'levelType': LevelType.mediumslow,
         'exp': 221,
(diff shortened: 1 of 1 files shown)


RedSparr0w:
For the evolutions, can you please just put them as unobtainable evolution methods, so we don't need to manually change the egg steps if it's adjusted down the road.

Qwertypop04:
Okay. Although that'll look really weird in bot coins lol

RedSparr0w:
I think we need to just eventually add a `evolutionStage` argument or similar so the egg steps get calculated off that when it doesn't have actual evolution stages.

Qwertypop04:
Yeah. That would also solve the wierd cases like burmy, where a Pokémon evolves from another form of the same Pokémon

Jaaslet:
@RedSparr0w, it seems like this solution will do so the bot will give a lot of false information. Don't know if you just want to make an exception for the bot or something?
But if the solution is fine, you can just approve it and merge it. I have already tested it, and it looks good!

src/scripts/pokemons/PokemonList.ts
```diff
@@ -657,7 +662,7 @@ const pokemonList = createPokemonArray(
     {
         'id': 12.03,
         'name': 'Ash\'s Butterfree',
-        'catchRate': 45,
+        'catchRate': 15,
```
RedSparr0w, line 665: Was this intentional?

src/scripts/pokemons/PokemonList.ts
```diff
@@ -1911,11 +1927,11 @@ const pokemonList = createPokemonArray(
     {
         'id': 45.02,
         'name': 'Pinkan Vileplume',
-        'catchRate': 45,
+        'catchRate': 20,
```
RedSparr0w, line 1930: Was this intentional?

RedSparr0w (approved):
Looks good, just a couple of places where values changed, that I'm not too sure were intentional 😅 

> Don't know if you just want to make an exception for the bot or something?

Will work something out, for now will probably just ignore the words `Pinkan` and `Valencian`

src/scripts/pokemons/PokemonList.ts
```diff
@@ -640,11 +645,11 @@ const pokemonList = createPokemonArray(
     {
         'id': 12.02,
         'name': 'Pink Butterfree',
-        'catchRate': 45,
+        'catchRate': 15,
```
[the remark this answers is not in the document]
Qwertypop04: No.

src/scripts/pokemons/PokemonList.ts
Qwertypop04: No.

src/scripts/pokemons/PokemonList.ts
Qwertypop04: No. I have undone all 3 of those mistakes

PR #2995, opened 2022-09-17
Lugia and Ho-oh Quests

Farboo171:
Adding quests for Lugia and Ho-oh, revolving around the Kimono Girls. 

The single temporary battle is meant to be balanced to be slightly harder than Lance2.

Ho-oh is dependent on catching the legendary beasts, but not the quest itself for ease of play for people past Johto already.


Osheyking:
I would say for this quest starts once your the champ and after you have the 3beasts kanpy for ho-oh

Farboo171:
Messed up the file list with the code test. Closing to make a clean PR.

PR #2998, opened 2022-09-19
Overworld Sudowoodo

Qwertypop04:
Sudowoodo is now in the overworld instead of route 36, and is required to be defeated and caught to move on to route 37


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3005, opened 2022-09-20
Snorlax Display Names

Qwertypop04:
Gives both Snorlax's the display name of 'Snorlax', instead of 'Snorlax route 12' and 'Snorlax route 16'


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3007, opened 2022-09-20
Changed pinkan values

Jaaslet:
Bumped the harvest amount to make getting the news mons a lot more reasonable.
Bumped the farm points and exp to be not useless.


RedSparr0w (approved):

PR #3011, opened 2022-09-20
Bill turned into Clefairy Temporary Battle

HyruleTeam64:
Adds a Temporary Battle against Clefairy to the Sea Cottage, located at the end of route 25, that rewards a Moon Stone. Beating the battle is required to enter Vermilion City Gym.


Jaaslet:
Why do we want this add this? It is not a fight in the game? It seems random

HyruleTeam64:
It adds a memorable moment from the original game using PokéClicker's mechanics. It also gives players an incentive to visit route 24 and 25 and presents them with evolution items for the first time.

Qwertypop04:
> It adds a memorable moment from the original game using PokéClicker's mechanics.
...no it doesn't. This battle doesn't occur in any game

HyruleTeam64:
> ...no it doesn't. This battle doesn't occur in any game

I thought turning it into a battle would make for a fun distraction in between gyms, but if devs think it isn't appropriate we can just close the PR

Jaaslet:
I will close this, as it really lower the bar for what should be a temp battle (which the players will get tired of).
And we already have a reason for completing route 25: you need to do it, to reach route 5

PR #3015, opened 2022-09-20
Fixup hatchery sounds

RedSparr0w:
Was previously not making sounds unless the hatchery queue was empty, this should resolve that.
Helpers eggs should still not make noise with this change.


Aegyo:
#2923

Make it a setting?

RedSparr0w:
Uhh, will make it a setting in a different PR, not exactly sure how it should be handled

PR #3016, opened 2022-09-20
Remove ShardType enum

Qwertypop04:
Not sure what this effects, but they clearly should be there


Jaaslet (changes requested):
The shard type enum is not used for anything. I think we should just remove it instead

Qwertypop04:
Okay. Removed it.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3018, opened 2022-09-20
Remove Factory Key

Qwertypop04:
It has no purpose, and there are no plans to make use of it. Therefore, the Factory Key key item should be removed.
Not sure if this needs any update function?


Jaaslet (approved):
It looks like there is no need for an update function. Thanks for the PR!

PR #3019, opened 2022-09-20
Singular Berry Quest Text 

CorgiOnNeptune:
Update harvest berries quest text to account for a single berry.


RedSparr0w (approved):

PR #3020, opened 2022-09-21
Pinkan Theme Park typos

CorgiOnNeptune:
Some minor typos and rename to split Themepark, as it should be two words. Also put the Sinnoh questline into proper capitalization.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3021, opened 2022-09-21
change(underground): Reworked item generation for new layers

Arkive86:
I noticed that the RNG for generating items in a new layer is heavily dependant on the random item and position. Currently, max items more closely represents number of attempts to add items. This means that, as the value of max items grows, it becomes more difficult for the game to generate a layer that actually has the maximum number of items on it.

This PR aims to mitigate that by deciding how many items to place before generating items and by allowing the game to retry placing any given item multiple times before giving up.


src/scripts/underground/Mine.ts
```diff
@@ -33,29 +35,37 @@ class Mine {
...
+            }
+            // If item can be added, add it
+            // Else, if we haven't hit the minimum number of items yet, try again with a new random item
+            // Otherwise, just skip to next item
             if (res) {
                 Mine.addReward(x, y, item);
-                added = added + 1;
-                //This should loop until it's added.
+            } else if (i < App.game.underground.getMinItems()) {
+                i--;
```
Arkive86, line 68: Just as a note, this could be changed to just a normal `else` statement if we wanted this to be applied to all items and not just the minimum required ones.

Aegyo: This scares me a little because it looks like we can get stuck in an infinite loop if no item can be added anywhere on the grid.

But we are already doing an infinite loop until `min` and if we remove it, it's possible to get layers with less items than we are promising with the min items upgrade...

Problem for another day I think.

Aegyo (approved):
LGTM :+1:

Some numbers to see how much of an effect it has, using the following code
```js
App.game.underground.upgradeList[1].level = 4;
App.game.underground.upgradeList[2].level = 4;
loops = 10000
totals = {};
for (i=0; i<loops; i++) {
  Mine.loadMine();
  const n = Mine.rewardNumbers.length;
  totals[n] = totals[n] ? totals[n] + 1 : 1
}
console.log(Object.entries(totals).map((([k, v]) => [k, `${100*v/loops}%`])))
```
| # of items | Current | This PR |
| --- | --- | --- |
| 5 | 65.18% | 26.77% | 
| 6 | 26.68% | 27.75% | 
| 7 | 07.65% | 26.44% | 
| 8 | 00.49% | 19.04% | 

Weird that 8 is lower than the others, maybe we have an issue in `Rand.intBetween`...

PR #3023, opened 2022-09-21
Remove nativeRegion from event Pokemon

Qwertypop04:
Removing as it is useless code, since it assigns the native region they would be automatically assigned to. Presumably a leftover from when they had negative IDs


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3031, opened 2022-09-22
Fixed Johto SVG error

DataCrusade:
Fixed a gap that I thought I already fixed, oopsies


Jaaslet:
I think i reverted your fix, before merging the other PR.
It's weird to have the SVG for the town go outside the town. We don't do that anywhere else.
I know they are connected in the games, but i think we should move the town closer, in that case. At least i don't like this solution.

DataCrusade:
I'm not a fan of moving the town closer and adding a tiny bit of route right there would be "wrong". Not sure what the best solution would be here

Qwertypop04:
I'm not sure what the problem would be with moving Azalea and Slowpoke Well closer to Ilex is? It would even match the games more closely.

Jaaslet:
I am closing this PR, as this is not the solution we want. I don't think we let the SVG go outside any other towns, so it's weird to do it here

PR #3036, opened 2022-09-23
Updating Qualot to be Pink instead of Yellow

arduousFrivolity:
Small change to make Qualot's color value pink instead of yellow, as this makes more sense with it being one of the seven pink berries that is dropped in Pinkan Mountain and used to mutate Pinkan Berries. Also changed the wanderers it attracts to match the change in color.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3037, opened 2022-09-23
Allow Enigma hint to be displayed in Berrydex once unlocked

CypherX:
(I dedicate this PR to tater in discord)

Currently if a player unlocks the Enigma berry without receiving all 4 hints from the Kanto berry master the hint in the Berrydex will be incomplete or missing entirely and the berry master will no longer give the hints, making it difficult to mutate again in the future if needed. This change will allow the full hint to be displayed if the berry has been unlocked and mutated previously.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3039, opened 2022-09-23
Fix protein filter dropdowns "resetting"

CypherX:
Fixes the protein Region & Type dropdown filters resetting when closing and re-opening the modal. Additionally these values were already persisted in the settings but were not getting set upon load.


Jaaslet (approved):
The code had two solutions for the problem. I removed one of them, so now it looks good.
Thanks for the PR

CypherX:
The removed code was needed to filter the protein list after a reload. Without it the dropdowns are set correctly but the pokemon list doesn't get filter because the settings are never applied to the filters:
image

Jaaslet:
I see... Yeah, `PartyPokemon.hideFromProteinList()` is not perfectly written. The code you wrote, should not be needed. I will try to fix `PartyPokemon.hideFromProteinList()`

PR #3040, opened 2022-09-23
fix: the plate amount in a player's mine inventory now only counts plate/gem items

Arkive86:
Currently, the "Plate" count when hovering over your underground net worth displays the total number of items that aren't classified as Diamond or Fossil items, which includes more than just Plate items. This just fixes that amount to display the correct number.


Jaaslet (approved):
I changed it so it shows total shards too.
Looks good. Thanks for the PR

PR #3050, opened 2022-09-24
Add move to dungeons for Icefall Cave and Malie Garden

Qwertypop04:
I forgot to add one for Icefall Cave when Sevii was added, and Malie Garden is somewhat hard to tell apart from the rest of Malie City now.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3055, opened 2022-09-25
Checkbox styles

Crobat4:
For the layout changes, since checkboxes can be used anywhere, I think it needs its own PR.

checkbox


RedSparr0w:
Haven't had a chance to test this out yet, but does it take the coloring from the different themes?

RedSparr0w (approved):

PR #3059, opened 2022-09-26
Fixed pinkan pikachu images

Jaaslet:
DraK4y fixed her pikachus (they where not centered), so here they are!


RedSparr0w (approved):

PR #3061, opened 2022-09-26
In-game FAQ update

xslk:
Updates In game FAQ with updated EV information


src/components/FAQModal.html
```diff
@@ -125,7 +125,40 @@ <h5>Frequently Asked Questions</h5>
             Roamer Modifier: <code>50×</code><br/>
             Repeatball Modifier: <code>5×</code></span></p>
             <p><strong>What is the EV Attack Bonus?</strong><br/>
-            <span class="text-muted">The EV bonus is multiplicative, and scales linearly until 1.5x (50EVs). The bonus thereafter has diminishing returns relative to the number of EVs gained, and maxes at 2x (1770EVs).</span></p>
+            <span class="text-muted">The EV bonus is multiplicative, and scales linearly until 1.5x (50EVs). The bonus thereafter has diminishing returns relative to the number of EVs gained. EVs can be gained indefinitely. </span></p>
+            <p><strong>What are the EV damage multipliers?</strong><br/>
+            <span class="text-muted">
+                <style>table, th, td {border: 1px solid black;border-collapse: collapse;}
```
Jaaslet, line 131: This affects all other tables. You either need to use the html attribute `style=""` on the table, or give the table a class and add the css onto that class (in a css file)

src/components/FAQModal.html
```diff
@@ -125,7 +125,40 @@ <h5>Frequently Asked Questions</h5>
...
+                th, td {padding: 3px;}
+                </style>
+                <table style='width: 30%'>
+            <tr>
```
Jaaslet, line 135: Indentation is weird. Table should be moved in and tr should be moved out

Jaaslet (changes requested):
Text looks good, but i have some changes

src/components/FAQModal.html
RedSparr0w: I would rather we don't manually adjust the style too much (especially colors) as otherwise it won't change with whichever theme the user has selected.

PR #3062, opened 2022-09-26
Update Kalos Route 1 encounters

Farboo171:
Kalos Route 1 does not have any encounters in the base game. This changes the encounter list from the somewhat jarring 100% Rattata to a lore-friendly pair of regional mons.

This does not affect balance, as the added mons are in the very next route.


Jaaslet (approved):
Looks good. Thanks for the pr

PR #3063, opened 2022-09-26
Update Diancie Questline requirement

Qwertypop04:
The Diancie questline is supposed to unlock after beating the Kalos Elite 4 (but before the Champion), so its requirement was the last Elite 4. However, recently the Unova, Kalos and Alola Elite 4 were changed to be challengeable in any order as per the games, so this ended up making it just require a random one of the Elite 4 that used to be the last one. This PR makes it so it requires all 4, as originally intended.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3065, opened 2022-09-26
Remove Kalos Route 1 (make it visual only)

kittenchilly:
I saw #3062 and I wondered "why is kalos route 1 even in the game in the first place?"
So I did this. This makes kalos route 1 only visual, there are no encounters. Its a lot like Skyarrow Bridge.


Jaaslet:
We went with the other PR, so i will close this one

PR #3066, opened 2022-09-27
Shiny Flying Pikachu sprite off by one pixel

osherwott:
Shiny for Flying Pikachu was one pixel higher


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3068, opened 2022-09-28
Held Items

jk13pclick:
Initial PR Draft, might add to more dungeons. This is 95% canon locations.


Jaaslet (approved):
I have removed the pokemon drops. It's a bit weird that you can unlock held items, before you can use them... And i don't really have a great solution yet... But we can always readd the code again, from this PR, when we find a solution.
But the dungeon ones looks good. Thanks for the PR

PR #3069, opened 2022-09-28
Cleanup the code for protein filters

Jaaslet:
- Removed the protein filters class. Now we just directly uses the setting
- Made the protein filter method look a lot nicer
- Fixed problem with region+type filters not being applied on game load


src/scripts/party/PartyPokemon.ts
```diff
@@ -215,33 +213,33 @@ class PartyPokemon implements Saveable {
...
-                const { type: types } = pokemonMap[this.name];
-                if (type !== null && !types.includes(type)) {
-                    return true;
-                }
+        }
+        const type: (PokemonType | null) = Settings.getSetting('proteinTypeFilter').observableValue() > -2 ? Settings.getSetting('proteinTypeFilter').observableValue() : null;
+        if (type !== null) {
+            const { type: types } = pokemonMap[this.name];
+            if (type !== null && !types.includes(type)) {
+                return true;
```
Aegyo, line 232: I can't get the suggestion comment to play nice with the braces just after here that would also need changing, but it looks like this code can be simplified to just:

```suggestion
        const type = Settings.getSetting('proteinTypeFilter').observableValue();
        if (type > -2 && !pokemonMap[this.name].type.includes(type)) {
            return true;
```

Aegyo (approved):
:+1:

Commented on another bit of code that could be cleaned up, I don't mind if you do it or just go ahead and merge now

PR #3070, opened 2022-09-28
Add more Pokemon to typed egg pools

Qwertypop04:
Adds more pokemon to typed egg pools to make the egg exclusive Pokemon less common, since in recent regions often only the exclusive Pokemon are in the pools


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3072, opened 2022-09-29
Fix static Kebia Bloom

arduousFrivolity:
Quick PR to replace the static KebiaTreeBloom.png with it's original animated version. No more holding your breath when growing Kebia!


RedSparr0w (approved):

PR #3073, opened 2022-09-29
Pinkan Bloom animation fix

arduousFrivolity:
Blooming Pinkan Trees were vibing at a slower rate than every other plant. Made them vibe harder. 

(Updated delay between frames from 100 to 50 to match all other berries)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3075, opened 2022-09-29
Rebalance late Kalos

Jaaslet:
I think the game needs a full rebalance, and i think Ultimas rebalance project is the way to go. But because that is a bigger project, i have nerfed the biggest problem a bit (late kalos), so we don't need to wait a year to fix the problem. Here is the comparison i used:
image


erzads:
I am not the most qualified to talk about balance but I can say that it _feels_ steep.

Aegyo (approved):

PR #3076, opened 2022-09-29
Fixed galar professor

Jaaslet:
The professor npc only tells you how many pokemon are left, if you have unlocked next region. And it checks next region, by checking if the starting town is unlocked.
Because we have not added gen 9 yet, the galar professor did not want to tell how many mons was left. The solution was either to have a list of all champions or add a town to region final. I think the last one is the best, so i went with that one.


Aegyo (commented):
This will require us to update the Final Region Town requirement with each new region, wouldn't it be better to instead add a condition on the ProfNPC text to see if its next region is final?

```diff
- if (!nextRegionUnlocked) {
+ if (!nextRegionUnlocked && this.region + 1 !== GameConstants.Region.final) {
```

Jaaslet:
The idea is that the text should change when defeating the elite 4. The professor starts out saying something like "come see me when you become the champion", and after you are the champion, the professor tells you how many pokemon are left.
I think with your code, the professor will always just tell how many mons are left?

Aegyo:
True, then can we have a comment around the region enum reminding us to update the final town req?

Jaaslet:
Sure. Done

Aegyo (approved):
:+1:

PR #3077, opened 2022-09-29
Galar Questline NPCs

Qwertypop04:
Adds NPCs to the questline "The Dojo's Armor". Also adds some NPCs to towns to explain how to get certain alternate forms and evolutions, and reorders the questline "The Ancient Golems"


Jaaslet (approved):
Looks good. I have not tested everything, but if there is more problems, we will catch them during the galar test

PR #3079, opened 2022-09-30
Skippable Hoenn Badges

Qwertypop04:
Hoenn in Pokeclicker is based on Emerald, so I have changed the Knuckle Badge to only be required to challenge Norman, and the Feather Badge to only be required to challenge the Elite 4, as in Emerald.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3083, opened 2022-10-01
More Galar NPCs (Hints & Darkest Day Questline)

CorgiOnNeptune:
I had some written that conflicted a bit with Qwerty's PR. The only one I left in that still conflicts a bit is the Hiker in Hammerlocke for the Cramorant. The cook isn't really needed also, but I thought it was funny so I kept it in. If we feel it just doesn't serve a purpose to keep in, that's fine too.

NPC talking about Farfetch'd's leek is commented out for now till that evo is ready so I have a better idea where to place it.


src/scripts/towns/TownList.ts
```diff
@@ -4390,40 +4394,140 @@ const FreezingtonShop = new Shop([
...
-    'Our Marnie\'s Morpeko is so cute, don\'tcha think?',
-    'When it\'s hungry, though, it gets so angry it changes forms',
+    'Our Marnie’s Morpeko is so cute, don’tcha think?',
+    'When it’s hungry, though, it gets so angry it changes forms.',
     'If you want to catch it in that form, you would probably be best trying to attract it with a berry that electric Pokémon like.',
 ], {image: 'assets/images/trainers/Team Yell Grunts.png'});
+
+/* To be added when held item evos are ready.
+// TODO: Add to town near a dungeon/shop that has the Leek available.
+const FarfetchdArtist = new NPC('Artist Duncan', [
```
Jaaslet, line 4515: Leek will be added as a drop item for Galarian Farfetch'd, so you can just finish this NPC

src/scripts/towns/TownList.ts
```diff
@@ -4723,11 +4834,11 @@ const PeonyComplete = new NPC ('Peony', [
 const ProfMagnolia = new ProfNPC('Prof. Magnolia',
     GameConstants.Region.galar,
     'TODO: Add text before Galar is released',
-    'TODO: Add text before Galar is released');
-
+    'TODO: Add text before Galar is released',
```
Jaaslet, line 4837: Any chance you can add these texts? Can't remember what they need to say, but you can look at the other ProfNPC's

src/scripts/temporaryBattle/TemporaryBattleList.ts
```diff
@@ -231,7 +231,7 @@ TemporaryBattleList['Suicune 6'] = new TemporaryBattle(
 TemporaryBattleList['Spiky-eared Pichu'] = new TemporaryBattle(
     'Spiky-eared Pichu',
     [new GymPokemon('Spiky-eared Pichu', 3178500, 20)],
-    '',
+    '<b><i>You caught the Spiky-eared Pichu!</i></b>',
```
Jaaslet, line 234: It's defeated message will now show up (before the window just went away).
So we now need to add an image for the pichu, so the modal has something to show

Jaaslet (changes requested):
Looks good! I don't know if there are too many lore-NPCs. I will hopefully get a feeling, next time i play through Galar. But i think 2-4 per region is what we want. Don't know how many you added.
And also some more comments:

src/scripts/towns/TownList.ts
CorgiOnNeptune: Prof Kukui was also missing this text for Alola. I added them in, with the next region line for Magnolia noting that there is no ETA when the next region is available, with a commented out line to be changed to when it is.

src/scripts/temporaryBattle/TemporaryBattleList.ts
CorgiOnNeptune: I used the directory to not have to add another image in the temporaryBattle folder. If we like this solution I can clean up some of the other potentially unnecessary image files in the temporaryBattle folder to save up a tiny bit of space using the same method in a different PR.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3084, opened 2022-10-01
Add caught indicator to Berry deals

Aegyo:
So people can see which Pinkans they have bought easily

Pinkan Berry deals


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3086, opened 2022-10-01
Battle cafe improvements

Jaaslet:
Jessica had some ideas for improvements to the battle cafe. The PR might expand if i find more stuff, before it's merged.
- Added tooltip to the berry, which shows the name of the berry
- Find a way to disable/show disable sweets, if you can't buy them
- Changed how you see your berry inventory
- Added javascript check to make sure the player can't spin for more than 20 seconds
- Moved spin and berry used to after the spin, so people don't lose it on reload
- Added caught status to sweets
- You could not read the text of selected berries on some themes. I have found an other class. It still looks bad on some themes, but it's readable
- A lot of html cleanup, with help from the pokecrew
- Changed which berries and the cost. Berries is based on color, and a spread of how hard the sweets should be. Cost is based on baileys spreadsheet. For sweets using max gen 3 berries, it should take 4 hours to farm the berries, without boosts. 12 hours for gen 4 and 48 hours for gen 5. Strawberry should be basicly free, so people can complete the dex without using the farm
- Easteregg (see code for more details)


src/scripts/towns/BattleCafe.ts
```diff
@@ -50,21 +50,26 @@ class BattleCafeController {
...
 
     private static unlockAlcremie(clockwise: boolean, spinTime: number, sweet: GameConstants.AlcremieSweet) {
         let spin: GameConstants.AlcremieSpins;
         const curHour = (new Date()).getHours();
+        if (spinTime == 3600) {
+            (new PokemonItem('Milcery (Cheesy)', 0)).gain(1);
+            return;
+        }
         if (curHour == 18 && !clockwise && spinTime > 10) {
             spin = GameConstants.AlcremieSpins.at7Above10;
```
RedSparr0w, line 74: Shouldn't this be `at6Above10?` or `curHour == 19`?

RedSparr0w (approved):
Changes look good, thank you for the PR! 👍 

Left 1 small comment.

src/scripts/towns/BattleCafe.ts
Jaaslet: You are right. Nice catch!

PR #3087, opened 2022-10-01
DT Flute -> EV Flute

jk13pclick:
A change I've been wanting to make. I didn't have the heart to get rid of the Time Flute so DT is the odd one out. We have also removed all of the other currency Flutes, so this makes some sense.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3088, opened 2022-10-01
Fixed achievement bonus

Jaaslet:
The achievement bonus might had been a bit too OP after my changes to achievement handler... Fixed the bug!


Aegyo (approved):
:+1:

PR #3092, opened 2022-10-02
Make farm auras properly computed

Aegyo:
Fixes auras not coming back when freeze mulch runs out, and also just feels like the better way to have this coded.

Definitely needs some more thorough testing, I only did some basic checking that numbers are the same and everything updates when it is supposed to.

~~May want to rateLimit the emitted aura observables.~~ Added the rateLimit.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3093, opened 2022-10-02
Fix nested/stacked modals not appearing on top

### LINKED ISSUE #2483: [BUG] Skip layer popup spawning behind the Underground modal.
**Info**
 - OS: Windows 10
 - Browser: Chrome
 - Version: v0.9.5

**Description:**
After opening the Underground via hotkeys while another modal is open, clicking on the "Skip" button to skip an underground layer will make the confirmation popup spawn behind the modal.

**Steps to Reproduce:**
1. Open the Farm modal
2. Use a hotkey to open the Underground (default: U)
3. Click on the "Skip" button
4. Observe that the confirmation popup spawns behind the modal.

**Expected Behavior:**
The confirmation popup should spawn in front of the modal at all times.

**Additional Information:**
Clicking on the close button will cause the modal to close but not the confirmation popup so you can still skip that way.
Everything works as intended, it only is a position issue.

CypherX:
Using a hotkey to switch to a modal while another is open will cause any nested/stacked modals to appear behind the current modal due to the z-index values being the same. (eg switching from the farm to the underground using the underground hotkey then clicking the Skip button)

This changes the z-index calculation to use the highest modal z-index (if any) rather than calculating it based on number of visible modals.

Reported on discord by BobTheDragon#4638
Closes #2483

### DIFF
--- a/src/scripts/index.ts
+++ b/src/scripts/index.ts
@@ -24,7 +24,7 @@ document.addEventListener('DOMContentLoaded', () => {
 // preventing us from getting into that messy situation.
 // Copied from https://stackoverflow.com/questions/19305821/multiple-modals-overlay#answer-24914782
 $(document).on('show.bs.modal', '.modal', function () {
-    const zIndex = 1040 + (10 * $('.modal:visible').length);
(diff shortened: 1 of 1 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3098, opened 2022-10-02
Fix Darkest Day Questline

Qwertypop04:
adds some missing requirements for the Darkest Day questline


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3099, opened 2022-10-02
Pokerus Spread Bug Fix

jk13pclick:
Should solve issue identified on Discord (ID: 1020794749441036389). Did a bit of testing and it all seems to work well


Jaaslet (approved):
I think it fixed the bug? At least it looks like it. Thanks for the PR

PR #3104, opened 2022-10-03
More Galar Quest NPCs 

Qwertypop04:
Added NPCs to Sword and Shield, and the latter half of The Darkest Day. Also fixed an error with the HP of Sordward & Shielbert


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3105, opened 2022-10-03
Fix Cyrus for players updating v0.9.8/9 -> 0.9.16+

CorgiOnNeptune:
Players updating an old save from either v0.9.8 or v0.9.9 to v0.9.16 with the 'A new world' quest completed, have had to start the quest over because of the rename changing the capitalization. 

What happened was they were receiving the questline again because the `Update.startQuestline` function in v0.9.10 couldn't recognize the 'A New World' questline as the already completed 'A new world'. So it started the questline again before receiving the v0.9.16 update to rename the questline, making it impossible to complete the Cyrus temp battle step since it was already completed before the update.

This resets the Cyrus battle for players who have affected saves that restarted the questline from the update jump, as well as preventing other players with old saves from being forced to restart the quest in the future.

[Discord bug thread w/ relevant save files for testing](https://discord.com/channels/450412847017754644/1024087097126162432)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3106, opened 2022-10-03
Genders

Crobat4:
This PR adds genders to Pokémon following the official types and ratios. Wild/Trainer Pokémon will be either male, female or genderless with a random chance calculated with the gender ratio and type.

Gender properties of forms are the same as the base form unless stated otherwise (ex. Ash-Greninja and Cap Pikachus being male-only, Spiky-eared Pichu being female-only, etc.).

Pokémon with gender differences can be filtered in the Pokédex, there is also a setting to change the sprites to the female one (like the hide shiny one).

You can also select the default sprite for Pokémon with gender differences in the Pokémon statistics menu, said setting will change the sprite in:
* Pokemon list container
* Pokemon statistics menu
* Hatchery menu
* Hatchery queue
* Damage calculator

Total/Pokémon gender statistics were also added.

I also made some other changes due convenience while I was making the code, such as:
* Moved Pokémon name from battle views to a template (gender character is still there just in case but commented)
* Moved Pokémon sprite from battle views to a template
* Moved statistics increments (captured/encountered/defeated/hatched) to another function in PokemonHelper
* Unified PokemonHelper.getImage(), PokedexHelper.getImage() and PokedexHelper.getImageStatistics() to a single function in PokemonHelper as their task was pretty much the same
* Fix: Captured statistics increasing only by 1 when bulk buying a Pokémon

~~Draft because gen 8 gender properties and female sprites are missing.~~

Some pics:
genderpr1
genderpr2
genderpr3


Jaaslet:
There is a lot of stuff, so it might take me a while to get through this PR (thanks for the work!).
I have three questions right now. I will hopefully have time to look more into the PR later today.
- Why do we need the update function? No one on the main branch will have the data in their savefile already, right?
- Can you explain to me what the gender options means? Type means if there exists male, female, both or they don't have genders, right? Difference is if they use different images? And ratio is only set, if there is not a 50/50 spawnrate?
- Can we turn GameConstants.MALE_ONLY and friends into an enum, instead of having 4 different consts? And ratio does not need a const, right? There is no reason to change MALE_25 at any point, so we might as well just add numbers

Crobat4:
> * Why do we need the update function? No one on the main branch will have the data in their savefile already, right?

I saw statistics lines in the update function for older updates so I though they were needed, they can be removed if they aren't.
> * Can you explain to me what the gender options means? Type means if there exists male, female, both or they don't have genders, right? Difference is if they use different images? And ratio is only set, if there is not a 50/50 spawnrate?
* Type can be male-only, female-only, genderless and male/female, defaults to male/female.
* Difference is if they have female difference, for the sprite, defaults to false.
* Ratio can be 12.5%, 25%, 50%, 75%, 87.5%, defaults to 50%. These ratios belongs to the male chance.
> * Can we turn GameConstants.MALE_ONLY and friends into an enum, instead of having 4 different consts? And ratio does not need a const, right? There is no reason to change MALE_25 at any point, so we might as well just add numbers

I'll change the types to enum later, as for the ratios, I personally like them as constants because they are easier to read than the actual numbers, at least for me, if you prefer, they can be removed.

Jaaslet:
> I saw statistics lines in the update function for older updates so I though they were needed, they can be removed if they aren't.

I can't see why they should be needed, but there might be something i miss. I can look at it, when i test the PR.

> as for the ratios, I personally like them as constants because they are easier to read than the actual numbers, at least for me, if you prefer, they can be removed.

You could move the calculation to the generateGender method? So if a pokemon has a 25% chance for being a male, the number will just be 25 or 0.25. And then you can change how you calculate the chance in the method. I think that will look cleaner.

> Difference is if they have female difference, for the sprite, defaults to false.

Can we change it to "visualDifference" or something like that?

But i will most likely don't have time to look further into this PR the next couple of days, but i will hopefully review the rest sunday.

src/components/breedingDisplay.html
```diff
@@ -72,7 +72,7 @@
     <!-- ko if: App.game.breeding.queueList().length -->
     <div id="hatcheryQueue" class="card-footer p-1" data-bind="foreach: App.game.breeding.queueList()">
       <div class="queuePokemon clickable bg-primary" data-bind="click: function(){ App.game.breeding.removeFromQueue($index()) }">
-        <img src="" data-bind="attr: { src: PokemonHelper.getImage(pokemonMap[$data], App.game.party.alreadyCaughtPokemon(pokemonMap[$data].id, true)) }">
+        <img src="" data-bind="attr: { src: PokemonHelper.getImage(pokemonMap[$data].id, App.game.party.alreadyCaughtPokemon(pokemonMap[$data].id, true), App.game.party.getPokemon(pokemonMap[$data].id).defaultFemaleSprite()) }">
```
Jaaslet, line 75: This code is duplicated a couple of times. You might want to add a `PokemonHelper.getPartyImage()`, which does all the lookups. Maybe just add it on PartyPokemon class? I think it fits better there

src/components/pokedex.html
```diff
@@ -87,6 +87,20 @@ <h4 style="margin-bottom: 0">Pokédex</h4>
                                             value="false" onclick="PokedexHelper.updateList()"> Hide alternate forms
                                 </label>
                             </div>
+                            <!--Gender differences-->
+                            <div class="col-12">
+                                <label class="form-check-label">
+                                    <input class="form-check-input" type="checkbox" id="pokedex-filter-gender-diff"
```
Jaaslet, line 93: I can't really see the usecase for this checkbox? Can you explain when you think it's useful?

src/components/pokedex.html
```diff
@@ -87,6 +87,20 @@ <h4 style="margin-bottom: 0">Pokédex</h4>
...
+                                            value="false" onclick="PokedexHelper.updateList()"> Gender differences
+                                </label>
+                            </div>
+                            <!--Show female images-->
+                            <div class="col-12">
+                                <label class="form-check-label">
```
Jaaslet, line 99: We might just want the pokedex to show the selected gender for each pokemon, instead of having a global toggle

src/modules/GameConstants.ts
```diff
@@ -1579,6 +1579,39 @@ export enum GalarSubRegions {
...
 
+// Gender Types
+export enum Genders {
+    Genderless = 0,
+    MaleFemale, // Default
+}
+
+// Gender Ratio
+export const FEMALE_0 = 0;
+export const FEMALE_12_5 = 0.125;
```
Jaaslet, line 1590: The idea behind changing the numbers to something that made sense, was so we no longer needed them as GameConstants :)

src/modules/GameConstants.ts
```diff
@@ -1579,6 +1579,39 @@ export enum GalarSubRegions {
...
+export const FEMALE_25 = 0.25;
+export const FEMALE_50 = 0.5; // Default
+export const FEMALE_75 = 0.75;
+export const FEMALE_87_5 = 0.875;
+export const FEMALE_100 = 1;
+
+// Gender Text
+export const TEXT_GENDERLESS = '';
+export const TEXT_MALE = '<i class="gender-icon-male">♂</i>'; // If Bootstrap Icons: <i class="bi bi-gender-male gender-icon-male"></i>
+export const TEXT_FEMALE = '<i class="gender-icon-female">♀</i>'; // If Bootstrap Icons: <i class="bi bi-gender-female gender-icon-female"></i>
```
Jaaslet, line 1600: It's weird to have HTML here. It should be moved to the template

src/modules/GameConstants.ts
```diff
@@ -1579,6 +1579,39 @@ export enum GalarSubRegions {
...
+
+// Wild/Trainer/BF Pokémon Gender
+export enum BattlePokemonGender {
+    NoGender = 0,
+    GenderMale,
+    GenderFemale,
```
Jaaslet, line 1606: Is there a reason why they are called "GenderMale" and "GenderFemale"? Gender is already in the enum name, so it should be clear

src/scripts/items/PokemonItem.ts
```diff
@@ -35,6 +35,16 @@ class PokemonItem extends CaughtIndicatingItem {
 
         const pokemonID = PokemonHelper.getPokemonByName(pokemonName).id;
         App.game.party.gainPokemonById(pokemonID, shiny, true);
+        // Only for statistics
+        for (let i = 0; i < amt - 1; i++) { // -1 because gainPokemonById will add 1 to statistics
+            const gender = PokemonFactory.generateGenderById(pokemonID);
+            // Add to shiny statistics depending of the number of shinies obtained
```
Jaaslet, line 41: Can we mash this for loop together with the shiny for loop? They are doing basicly the same, so might as well add them together

src/scripts/pokemons/PokemonFactory.ts
```diff
@@ -292,4 +298,56 @@ class PokemonFactory {
...
+        return this.generateGender(pokemon.gender.ratio, pokemon.gender.type);
+    }
+
+    /**
+     * Calculate which gender has the pokemon.
+     * @param chance Base chance, should be from GameConstants under Gender Ratio comment
+     * @param genderType Gender type (Genderless, male only, etc.), should be from GameConstants under Gender Types comment
+     * @returns {number} 0 = Genderless, 1 = male, 2 = female
+     */
+    public static generateGender(chance: number, genderType: number): number {
```
Jaaslet, line 317: Why is this returning a number and not the enum? If it returned the enum, the rest of the code will be easier to read

Jaaslet (changes requested):
I think it looks great. I have a ton of comments (sorry!), but they are all small changes (i hope).

- You can add the gender icon to the battle screen. We are currently moving stuff a bit around, so there is room for a pokerus icon, so there should be room for the gender icon. But can you make it a bit smaller than how it looks on the suggestion? They are a bit big there.
- The update function: What you have done currently is not needed (i have tested it). But we could consider setting the `totalMale...` stats to be equal the old stats? Just say that all mons caught/defeated so far was male? That kinda fits with our old sprite selection

src/components/pokemonStatisticsModal.html
```diff
@@ -73,6 +73,36 @@ <h5 class="modal-title text-light" data-bind="if: App.game.party.getPokemon(poke
...
+                          <code data-bind="html: PokedexHelper.getGenderRatioData(pokemonMap[$data]).male + '%'"></code>
+                          <br />
+                          <code data-bind="html: '♀: '"></code>
+                          <code data-bind="html: PokedexHelper.getGenderRatioData(pokemonMap[$data]).female + '%'"></code>
+                        <!-- /ko -->
+                      </td>
+                    </tr>
+                    <!-- ko if: App.game.party.alreadyCaughtPokemon($data) && pokemonMap[$data].gender.visualDifference -->
+                    <tr>
+                      <td class="text-left align-middle">Default sprite</td>
```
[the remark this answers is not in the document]
Crobat4: Party, breeding, Pokédex, etc. Not battles though.
I changed it to Party Pokémon gender

src/components/pokemonStatisticsModal.html
```diff
@@ -97,6 +127,51 @@ <h5 class="modal-title text-light" data-bind="if: App.game.party.getPokemon(poke
                       <td class="text-left">Hatched</td>
                       <td class="text-left tight"><code data-bind="text: App.game.statistics.pokemonHatched[$data]().toLocaleString('en-US')">-</code></td>
                     </tr>
+
+                    <!-- Gender Statistics -->
+                    <!-- ko if: pokemonMap[$data].gender.type == GameConstants.Genders.MaleFemale && pokemonMap[$data].gender.ratio != GameConstants.FEMALE_0 && pokemonMap[$data].gender.ratio != GameConstants.FEMALE_100 -->
+                    <tr>
+                      <td class="text-left align-middle">Male/Female Encountered</td>
+                      <td class="text-left tight">
+                        <code data-bind="html: '♂: '"></code>
```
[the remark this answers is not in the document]
Crobat4: Merged them with the encountered/defeated/captured/hatched rows.

src/components/breedingDisplay.html
Crobat4: getImage is not just the sprite from party/Pokédex/breeding but also wild/trainer Pokémon, since they did the same task (getting the sprite) I decided to merge them all into a single function.
I made some changes so I can remove the shiny and gender parameters from some calls in the html files.
Since it's also being used for wild/trainer sprites, I didn't move it yet to PartyPokemon, but it can be done if you want.

src/components/pokedex.html
Crobat4: It's just a filter for Pokémon with gender differences, nothing else. I removed it.

src/components/pokedex.html
Crobat4: Removed it.
Now the Pokémon sprite in the Pokédex will change according to its gender option selected.

src/modules/GameConstants.ts
Crobat4: Removed the constants.

src/modules/GameConstants.ts
Crobat4: Moved to a template.

src/modules/GameConstants.ts
Crobat4: It had another name iirc, changed the names to Male and Female respectively.

src/scripts/items/PokemonItem.ts
Crobat4: Done.

src/scripts/pokemons/PokemonFactory.ts
Crobat4: It was a number before, forgot to change it.

Crobat4:
Alright, I made the requested changes.
For the update function, I decided to give total statistics according to the gender type, not just to the male ones so genderless and female can have statistics too (female statistics gets the ones from female-only Pokémon).

Jaaslet (approved):
Looks good. It's a lot of code to look at, but i can't find anymore i think should be changed.
Everything seems to work. We have some Galar testing running atm, so hopefully it will catch anything i missed.
Thanks for the PR

PR #3109, opened 2022-10-04
Move Azalea Town and Slowpoke Well

Qwertypop04:
Moves Azalea Town and Slowpoke Well to the left, lengthens route 33 slightly


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3112, opened 2022-10-04
Update Berry Obtained Stats Index

CorgiOnNeptune:
We missed the berriesObtained statistic with the Pinkan Berry in both .9.15 and .9.16. This will look a little funny since people will have obtained berries in the meanwhile, but stats from pre .9.15 should be corrected.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3115, opened 2022-10-05
fix: farm hands will no longer spend energy trying to plant berries the player doesn't have

Arkive86:
Currently, if a farm hand is set to plant a berry that the player currently has 0 of, then the farm hand will spend energy even though they can't do anything. This fixes that so the farm hand won't spend energy in that scenario.

I also removed a redundant `worked = true;` line, just for fun.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3118, opened 2022-10-05
Fix error in Galar Questline

Qwertypop04:
The step for clearing Slumbering Weald Shrine had a typo in it and asked for a different dungeon


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3119, opened 2022-10-05
Various Galar fixes and changes

Qwertypop04:
Leon is now part of the Darkest Day questline. Marnie is moved slightly, Professor Oak refers to the correct professor upon becoming champion, fixed error with Hop NPC being locked behind the quest step it was needed for, and some small text fixes


RedSparr0w (approved):

PR #3120, opened 2022-10-05
Minor route correction and additional pokemon forms

osherwott:
I noticed a singular Linoone in Galar didn't have the aforementioned adjective before its name :)
Also added Hoopa Unbound, as well as other trivial pokemon forms to the list as comments in case anyone wants to implement them in the future.


Jaaslet:
We... Don't agree that they are "trivial"...
The galar fix is good, but we don't really want a ton of code, that we will never use a big part of. We have enough pikachus. We don't need to add all the pokemon go ones

osherwott:
Reply is noted. I opened a request to just fix Linoone

PR #3125, opened 2022-10-07
Princess Diancie NPC and Text Fix

Farboo171:
Added an NPC in Reflection Cave to let players know to check their bulletin boards when they are 'powerful', to hint at how to start the Diancie quest.

Specified a city for the Examine Doublade quest step so people stop asking how to do that.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3126, opened 2022-10-07
Farm Point Hatchery Helper

Farboo171:
Farm Point Hatchery Helper added.

Used Farm Points of 250,000 as the unlock condition, so it won't show up unless players have FP to spend.

Cost is 25 FP/Hatch. At one hatch every 15 seconds, this helper will use 6000 FP/hour. A full farm of lums being auto-harvested will easily feed this helper, as will several other less intense setups.

This helper starts at 75% for both values, as FP are slightly harder to get than DT (The helper for that starts at 50%/50%), and slightly easier than QP (for which helpers have higher stats)


Jaaslet (changes requested):
It's a bit weird that you need the currency to unlock the helper. For the rest of the currency based helpers, you need to buy them from a shop.
But yeah, the numbers look fine. But it's hard to say

Farboo171:
I was trying to figure a way to base it on "must progress some amount in the farm", and went with FP total. 

It could be changed to unlocking Lum, that's another medium-difficulty farming milestone.

Jaaslet:
We could also just sell it for like 100k farming points or something?

Farboo171:
Good plan. I'll add it to the Hoenn berry master for 100k FP.

Jaaslet:
Might want to update the hatchery helpers achievements on this PR

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3127, opened 2022-10-07
EV Status for Routes and Dungeons

jk13pclick:
WIP. My attempt at displaying the EV progress for Routes/Dungeons. I'm not totally thrilled with where I moved the information, but it seemed like the best spot. Also gets rid of our route text overlap issue

Dungeon trainers and Gym trainers + their team images will still overlap with their Pokémon names. Also the PR does not have support for mimics. All things that can be improved on, but I want feedback first.
Screen Shot 2022-10-10 at 11 32 28 AM
Screen Shot 2022-10-10 at 11 32 43 AM
Screen Shot 2022-10-10 at 11 41 38 AM


RedSparr0w:
I'm not a huge fan of the layout, it's probably just because I'm accustomed to the old one though 😅 
Could you try make the bottom similar to the top (transparent black bar, with info over top, background underneath), will likely have to enlarge the battle area slightly for it.

jk13pclick:
@RedSparr0w updated as requested
Screen Shot 2022-10-11 at 2 40 51 PM
Screen Shot 2022-10-11 at 2 41 01 PM
Screen Shot 2022-10-11 at 2 41 08 PM

src/components/dungeonView.html
```diff
@@ -116,5 +100,61 @@ <h4><strong data-bind="text: DungeonRunner.timeLeftSeconds() + 's'"></strong></h
...
+                  <knockout data-bind="using: App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(DungeonRunner.dungeon.name)]()"</knockout>
+                  <knockout data-bind="text: $data.toLocaleString('en-US')">0</knockout>
+                  &nbsp;Clears
+                </td>
+              </tr>
+              <tr>
+                <td width="50%">
+                    <span style="display: inline;">
+                        Pokémon Attack:
+                        <span data-bind="
```
Jaaslet, line 137: Can we throw this span into a template? It's used a lot, and is quite big

src/scripts/wildBattle/RouteHelper.ts
```diff
@@ -64,6 +63,26 @@ class RouteHelper {
         return true;
     }
 
+    public static getPossibleRoutePokemon(route: number, region: GameConstants.Region, includeHeadbutt = true): PokemonNameType[] {
+        return RouteHelper.getAvailablePokemonList(route, region, includeHeadbutt);
```
Jaaslet, line 67: Why have we added this method? It's not big enough to should be it's own method

Jaaslet (changes requested):
I have fixed the textsize, so it looks better now.
You are still missing attack for temp battles.
I have some small code comments, but other than that, it looks good

Jaaslet:
Towns also needs to be fixed

src/scripts/wildBattle/RouteHelper.ts
jk13pclick: It is superfluous, removed

src/components/dungeonView.html
jk13pclick: Yes, but I can't get the tooltip to work. Anybody that understands HTML a bit more than I feel free to try and fix up pokemonAttackTemplate.html

Jaaslet (approved):
I fixed the tooltip. I think stuff looks good now. At least i can't find a better way of fitting all of the information.
Thanks for the PR

PR #3128, opened 2022-10-07
Regi Trio Quest

Farboo171:
The new Regi Trio Quest, without all the develop merge issues from before.

This quest is meant to be done mid way through Hoenn, when the Sealed Chamber is first unlocked.

There are Braille images, but every image has a translation with an NPC that lives in the Sealed Chamber. It should be easy to complete this quest if players use this Braille Enthusiast NPC.

PR is a draft until crew testing is done.


Jaaslet (approved):
It works really well! Better than i expected. Thanks for the PR

PR #3131, opened 2022-10-08
Fix highest mutation ordering

CypherX:
Mutations are not in the same order as berries with Kasib and Starf being added after Enigma. This can cause Enigma to not appear in the berrydex even after receiving a hint.

This change sorts the mutation list by berry id (descending) when determining the highest mutation rather than simply reversing the list. Figured we might as well future proof it instead of just rearranging the mutations.

Save file where this scenario occurs, provided by raynedance25#8878 in discord:
[Pre-Starf_no_Enigma_Shown.txt](https://github.com/pokeclicker/pokeclicker/files/9738488/Pre-Starf_no_Enigma_Shown.txt)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3132, opened 2022-10-08
Achievement fixes

Jaaslet:
- Questlevel achievement required too much, so they have been lowered.
- Default value for the setting "achievementsCategory" was invalid. This is now fixed.


Aegyo (approved):
👍

PR #3134, opened 2022-10-08
Resized Galar Icons

DraKay:
Added all galar icons that needed to be resize. As it is my first ever PR I hope i didnt break anything


Jaaslet (approved):
Sweet images. Thanks for the PR

PR #3135, opened 2022-10-08
New EV Mechanics

jk13pclick:
WIP. Lots to do here. This way of doing trainer EVs feels janky, could definitely be improved. Needs the other mechanic as well, so big WIP.

What it does: Gain EVs by defeating **some** trainer Pokémon. All Gym Trainer and Dungeon Trainer Pokémon can yield EVs, only those that are included in the filtered Pokémon list will gain EVs. Dungeon trainers receive the dungeon boost to yield - trainer base yield is 1/10 base catching yield. Also included are icons for trainer EVs on the modals.


Jaaslet:
Can you update the description with what it ended up doing?

jk13pclick:
> Can you update the description with what it ended up doing?

Done 👍

Aegyo:
Why are stone evo base pokemon and all evolutions of pokemon in the list also included in the TrainerEVList? If those pokemon weren't picked up by the gym and dungeon trainer checks, when will they be gaining EVs from a trainer battle?

AnnoyingPriannacompy:
too bad You couldn't include all pokemon in the Battle Frontier.
It would make Farming Level in the Battle Frontier even more worth it. Plus it only add pokemon you have access too.

jk13pclick:
See #3954 and subsequent relevant PRs

PR #3136, opened 2022-10-08
Moved sevii dungeons achievements

Jaaslet:
I found a clean way to move the sevii dungeon achievements to the Sevii category!


RedSparr0w (approved):

PR #3138, opened 2022-10-09
New Image for the Pinkan Pokémon Reserve

HyruleTeam64:
Changed Pinkan Pokémon Reserve's image for one of the actual building from the anime instead of the one shared with Pinkan Mountain.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3142, opened 2022-10-10
Lugia and Ho-oh Quests addition

Farboo171:
Adding quests for Lugia and Ho-oh, revolving around the Kimono Girls.

The single temporary battle is meant to be balanced to be slightly harder than Lance2.

Ho-oh is dependent on catching the legendary beasts, but not the quest itself for ease of play for people past Johto already.

A fixed version of #2995 without the extra files added from the test website.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3143, opened 2022-10-10
Small Galar fixes

Qwertypop04:
Hid the dungeon bosses locked behind quests until unlocked, small text fix. Also fixed Reshiram having lower HP than Zekrom


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3147, opened 2022-10-10
Dragon Fang fix and Soul Dew

Qwertypop04:
Fixes Dragon Fang not appearing in the held items menu, and adds a new held item, Soul Dew, currently without an obtain method


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3152, opened 2022-10-11
Bf pokemon reward and baby can be shiny

VodkaFR:
Pokemon obtained via Battle frontier rewards couldn't be shiny, so I added a shiny role there.
Same for baby pokemons. 

Both have their own shiny chance constants which is currently at 1024 (same as shop pokemons)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3153, opened 2022-10-12
Update held item images and text

CorgiOnNeptune:
Held item images are now a consistent 120x120, cleaning up the UI a bit. I also updated some of the text associated with the held items.


Jaaslet (approved):
Looks good. Thanks for fixing my mistakes :)

PR #3156, opened 2022-10-12
Sprite fixes

osherwott:
Moved positions to line up shiny and normal sprites:
- Pinkan Rhydon
- Furfrou Diamond, Debutante, Matron, Dandy, La Reine, Kabuki, and Pharaoh
- Bewear
- Zeraora

Replaced with new:
- Wormadam Sand and Trash given gen 5 sprites to match with Plant cloak
- Reverted Average size Pumpkaboo sprite that was mistakenly replaced with with Small size Pumpkaboo when its forms were added
- Mareanie normal and shiny sprites replaced with new ones that now match
- Silvally is sourced from the same place its shiny was and now matches it (the previous one was from a fangame with a non canon shiny sprite)

Changed sprites:
- Ash's Butterfree, Pinkan Pidgey, and Pinkan Bellsprout sprite touch-ups
- Shiny Santa Snorlax's gift matches with its normal sprite
- Various Vivillon now more faithfully represent their canon colors and designs


DraKay:
Why did you changed ash butterfree right wing adding dark color? That part is the transparent part of the wing that shows under the feet. If you look what you did is add a black spot to the wings that doesn't exist in butterfree. Also you changed the light in right  antena and one pixel in the upper part of the wing without any reason (just to make a visible corner?)

Also, thanks for ask me before editing my sprite (yes, is sarcasm)

osherwott:
There is no need for sarcasm, this is a public repository. Whatever you upload here is subject to be played around with by others. I changed Ash's Butterfree to match its shiny sprite, and only the few pixels necessary to do that.  If you compare the two it looks like the regular sprite is missing part of its foot, for example. If you prefer, I could change the shiny sprite instead.

arduousFrivolity:
This looks like multiple PRs all bundled into one to me. Is it possible to separate them out into maybe a different one for each category (position fixes, sprite replacements, and sprite touchups)? That should make it easier to review and not have the simple fixes get delayed due to discussion over the others.

DraKay:
Something being public doesn't mean you have the right to touch at your own. At least not in art there is something called property rights that exist just from the moment you create something. And I prefer you don't touch my sprites even more if you don't take the time to ask first. 

Being public doesn't mean you can change everything because you like. For that there is a discord channel, if you would take the time to collaborate instead of try to became the project yours you could be told that have no sense to edit kalos and alola sprites since there is already an user working in all of them (almost finished iirc) with a more complete pr and with devs aproval

osherwott:
Good idea @arduousFrivolity

PR #3158, opened 2022-10-12
Muscle Band Held Item

HyruleTeam64:
Intended to be a worse, easier to obtain version of the type Held Items. It grants a 5% Bonus to any Pokémon's attack and can be bought in the Battle Frontier for 1000 Battle Points. I'm not completely sure about the price. With the current one, players will get enough BP to buy one in a single run around Late Sinnoh/Unova.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3159, opened 2022-10-12
Add missing exp values

Qwertypop04:
Some Pokemon have an exp value of null, for some reason. This gives them the correct values


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3163, opened 2022-10-12
New sourced sprites

osherwott:
- Wormadam Sand and Trash cloaks are now also from gen 5, like Plant cloak
- Non-shiny Average size Pumpkaboo brought back after its accidental replacement with Small size sprite when its forms were added
- Mareanie sourced from Smogon and now has a matching normal and shiny sprite
- Silvally sourced from Smogon instead of a fangame and matches its shiny sprite


DraKay:
Can you not do anything with sprites made from smogon(gen 6+)? Someone is doing something with them and it might make something weird happen

PR #3164, opened 2022-10-12
Linking cord

DraKay:
Update Linking cord image to the same style as others. Hope I didn't brake anything I used same fork as the other pr but different brancb


Jaaslet (approved):
Fits way better. Thanks for the PR

PR #3167, opened 2022-10-13
Held Item NPC

CorgiOnNeptune:
Adds an NPC explaining held items to Violet City in Johto. 

Also ended up adding some other NPC images (while unsuccessfully searching for a Primo sprite) and various other text fixes/changes.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3170, opened 2022-10-13
Resistant to Pokerus Notif/Log

CorgiOnNeptune:
Adds a notification and log when a Pokemon becomes resistant to pokerus.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3172, opened 2022-10-13
Dream orbs

Jaaslet:
Inspired by the best game ever, Pokemon Dream Radar... or at least the few parts we can use...
- Each time you are offline, you gain pink orbs!
- Pink orbs can contain stuff (based on the game)
- One of the things is 'Tornadus (Therian)'
- When that is unlocked, you can get green orbs! and so on
Eugene will fill in the content. The plan is for this to contain the 4th mon, when Hisui is implemented


src/components/dreamOrbsModal.html
```diff
@@ -0,0 +1,28 @@
...
+    <div class="modal-content">
+        <div class="modal-header">
+            <h5 class="modal-title">Open Dream Orbs</h5>
+        </div>
+        <div class="modal-body">
+            <select id="dream-orbs-selector" class="custom-select" onchange="App.game.dreamOrbController.selectedOrb(App.game.dreamOrbController.orbs[this.value]); App.game.dreamOrbController.item(undefined)" style="margin-right: 8px">
+                <!-- ko foreach: App.game.dreamOrbController.orbs.filter((o) => !o.requirement ||o.requirement.isCompleted()) -->
+                <option data-bind="attr: { value: $index }, text: $data.color"></option>
+                <!-- /ko -->
+            </select>
```
Aegyo, line 13: A neater way I've found for writing these select lists.
The filter thing also appears in a couple of places, may be worth making it a computed property on dreamOrbController to make sure everywhere agrees.
```suggestion
            <select id="dream-orbs-selector" class="custom-select" onchange="App.game.dreamOrbController.item(undefined)" style="margin-right: 8px"
                    data-bind="options: App.game.dreamOrbController.orbs.filter((o) => !o.requirement ||o.requirement.isCompleted()),
                               optionsText: 'color',
                               value: App.game.dreamOrbController.selectedOrb"
            >
            </select>
```

src/scripts/towns/DreamOrbController.ts
```diff
@@ -0,0 +1,123 @@
...
+        return {
+            orbs: [
+                ...this.orbs.map((o) => {
+                    return {amount: o.amount(), color: o.color};
+                }),
+            ],
+        };
+    }
+    fromJSON(json: Record<string, any>): void {
+        json?.orbs?.forEach((o) => this.orbs.find((o2) => o2.color == o.color).amount(o.amount));
```
Aegyo, line 99: Just in case we remove a color (spelling or name change maybe?)
```suggestion
        json?.orbs?.forEach((o) => this.orbs.find((o2) => o2.color == o.color)?.amount(o.amount));
```

Aegyo (changes requested):
The notification change is the only one I want for sure, I've included a couple of stylistic suggestions you can accept or reject however you like

Jaaslet:
I agreed with all your comments, so i have merge them, resolved conflicts and made sure stuff still works. Should be easy to rereview, i guess?

Aegyo (approved):
:+1:

src/scripts/towns/DreamOrbController.ts
```diff
@@ -0,0 +1,119 @@
...
+        new DreamOrb('Orange', new ObtainedPokemonRequirement(pokemonMap['Thundurus (Therian)']), [
+
+            new DreamOrbLoot({type: ItemType.item, id: 'Staryu'}, 0.133),
+            new DreamOrbLoot({type: ItemType.item, id: 'Ralts'}, 0.133),
+            new DreamOrbLoot({type: ItemType.item, id: 'Rotom'}, 0.134),
+            new DreamOrbLoot({type: ItemType.item, id: 'Landorus (Therian)'}, 0.075),
+            new DreamOrbLoot({type: ItemType.underground, id: 'Fire Stone'}, 0.25),
+            new DreamOrbLoot({type: ItemType.item, id: 'Protein'}, 0.275),
+        ]),
+        //new DreamOrb('Blue', new ObtainedPokemonRequirement(pokemonMap['Enamorus (Therian)']), [new DreamOrbLoot({type: ItemType.item, id: 'Enamorus (Therian)'}, 0.01)]),
```
RegisCoaxans, line 54: I know that this is commented but, following the previous pattern, should the requirement not be Landorus instead of the Enamorus that we obtain in here ?
Also, Enamorus appears twice is this pool, making the sum of weights 1.01.

Farboo171: Good catch, I'll give this a fix

PR #3174, opened 2022-10-14
Shard Price Rebalance

Farboo171:
Updated Shard costs to the following:

20 total shard cost for eggs, previously priced at 100
40 total shard cost for stones, previously priced at 100
50 total shard cost for special stones, which were previously priced at 125

No changes to QP costs


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3179, opened 2022-10-16
Galarian Birds Questline extension

Qwertypop04:
Adds some more stuff to make the Galarian Birds quest feel less short compared to the other galar questlines


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3180, opened 2022-10-16
More female sprites

DraKay:
Adding new female (and male) forms to alt forms as they missed them 

As always I hope I didn't break anything.

For the credits soul, arduous and qwerty made the other version of their owns sprites while I made mine and changed nice dice ones as and the clone one, but as they are very small changes and mostly paste the original sprites parts I don't think it need to be credited


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3181, opened 2022-10-17
Add missing fossil mon requirements

Qwertypop04:
Forgot to have some of the wild fossil mons in galar self require. This should fix that


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3182, opened 2022-10-17
Adamant and Lustrous Orb Held Items

HyruleTeam64:
Adds the Adamant Orb and the Lustrous Orb Held Items. They give a 50% attack bonus to Dialga and Palkia respectively and can be dropped by them.


Jaaslet (changes requested):
We don't really want the boosted droprate for now. We have added Soul Dew with standard droprate, to see if it will be useful or not. And because it does not lock any alt forms, we think that it's fine. At least for now. So there is no reason for these items to have increased droprate.

HyruleTeam64:
Ok, I've removed the boosted droprate.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3183, opened 2022-10-17
Beastballs defaults to ultra ball

VodkaFR:
Goal is to allow catching UltraBeast AND other pokemons with the same settings. 

During the first release of BeastBalls, regular ball setup ended trying to catch Ultrabeasts too, resulting in unwanted use of "scarce" Beastballs. 

Here we turn it around, so Beastballs setting, defaults to using Ultraballs if the encounter isn't an Ultrabeast. 

ie : BeastBall on caught = catches everything you already caught


Jaaslet:
Problem with this solution is people also wanted to be able to hunt ultrabeasts, without catching other mons. We considered this solution, and came to the conclusion that the solution we currently have, is the best until we figure out how we want to do the filters

VodkaFR:
I understand this, I only find it not consistent with how the game works currently. 
You can't filter any pokemon when you are trying to catch something throughout the whole game. 
You can't for instance catch only roamer. 
It feels a bit weird that UltraBeasts get treated differently. 

But since it's a different mechanic this can be the expected way. 



Also I have found I forgot to remove some other PR change in this one causing potential merge issues. What is the best way to check for merge conflict before making a PR ?

Jaaslet:
> Also I have found I forgot to remove some other PR change in this one causing potential merge issues. What is the best way to check for merge conflict before making a PR ?

You can update the branch here on the website, and then look at files changes, also here on the website

Jaaslet:
It's weird, yes. And we want a solution for better filtering, but we just have not found out yet, what it should be.
The solution is not to remove the filtering we have. I will close this PR

PR #3184, opened 2022-10-17
All pokemon shiny roll on obtention

VodkaFR:
Allow pokemon quest rewards to be shiny. 

We could eventually use another const than SHINY_CHANCE_SHOP  to control the rate of those shiny. 

Also i'm not sure but maybe we don't want to allow "Ash pokemeon" to be shiny on the reward since we already know if they are shiny or not in the anime


Jaaslet (changes requested):
The changes looks good. But i have some comments:
- For some reason, your PR contains your changes from earlier PR. Both in the commit history and the files changes. I don't like merging something that looks like that, so you might want to open a new PR, where you use a fresh branch.
- Don't change anything in the docs folder
- Can we add a method for obtaining mons for questrewards? In QuestHelper or something. Would make the code cleaner, now that more stuff in happening on mon questrewards. Let it gain the pokemon, with the correct shiny chance. You can also move the notification in there. Most of them have a generic text, so you can just use that as default, but allow the function to give an other notification

docs/index.html
```diff
@@ -36463,7 +36463,7 @@ <h5 class="modal-title" id="pickStarterModalLabel">Professor Oak</h5>
                            <input class="image-starter" type="image"
                                   data-bind="
                                     attr: { src: 'assets/images/pokemon/' + $data + '.png' },
-                                    click: () => { App.game.party.gainPokemonById($data); player.regionStarters[player.region]($index()); $('#pickStarterModal').modal('hide'); }
+                                    click: () => { App.game.party.gainPokemonById($data, PokemonFactory.generateShiny(GameConstants.SHINY_CHANCE_STARTER)); player.regionStarters[player.region]($index()); $('#pickStarterModal').modal('hide'); }
```
Aegyo, line 36466: It doesn't look like SHINY_CHANCE_STARTER actually gets defined anywhere

Aegyo (changes requested):
Not much has happened on this one for a while, do you need some help doing what Jaaslet suggested earlier?

Jaaslet:
I will close this PR. It seems abandoned, and i think if we want to implement it, we rather want to make a `rewardPokemon` method somewhere, which handles this and notification. Would make it a lot easier to reward new mons, and fix stuff like this in the future

PR #3187, opened 2022-10-18
Zarude Village

Qwertypop04:
Changed Glimwood Tangle pokemart to be called Zarude Village


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3192, opened 2022-10-18
Change Held Item Drop Locations to Avoid Debuffs

arduousFrivolity:
Added, removed, and moved Held Item drop locations to ensure the following:
- **All held items are able to be obtained in at least one non debuffed dungeon at all times (including upon reaching Galar)**.
- All basic type enhancing items (sans Metal Coat and Hard Stone) are available in Johto when Held Items are unlocked.
- No two held items are dropped in the same dungeon.
- No held item is dropped in more than one dungeon in the same region.
- There is at least some baseline logic as to where they drop even if it means departing from the games.

[Basic logic behind where I placed them is in this spreadsheet](https://docs.google.com/spreadsheets/d/1_G6UmKCwTjZ3oIUxDPhPH9FFcYZ2kh3ziLqd6DqdIuA/edit?usp=sharing) for the curious.


Jaaslet (approved):
jk has approved the locations. Thanks for the PR

PR #3194, opened 2022-10-19
Slightly boost chance of undiscovered fossils appearing in the underground

CypherX:
With the addition of shards to the underground loot pool there has been an uptick in players, mainly in Kanto with the initial slowness of the underground, getting stuck unable to find the specific fossil they need to complete the regional dex.

This change slightly increases the odds of fossils that have not been found previously (but are unlocked) to appear, giving them a weight of 1.5 vs a weight of 1.0 for any other item.

Some quick numbers from a save near the end of Kanto needing a Helix fossil and Amber fossil -

Average # of items generated before a needed fossil appears
- Prior to this change: 22
- After this change: 15


Jaaslet (changes requested):
It would look a lot cleaner, and be easier to adjust in the future, if the weight was added as a parameter on the UndergroundItem class.
I don't think we need the weight to go away when the pokemon is caught. But if you still think that, you can turn the weight into a function on the UndergroundItem class.
And let's just keep the fossil weight to standard, for sinnoh and onward. Making it a bit harder later on. People got time there, while breeding for attack

CypherX:
Added weight parameter to the UndergroundItem class, it can either be a number or function.

Changed the increased weight to only apply to pre-Sinnoh fossils as suggested, but I did opt to keep it conditional so once the fossil or pokemon has been acquired it returns to the standard weight. I don't think making these fossils have a permanent higher weight would be a good idea as it would ultimately hurt later-game players when hunting for diamond or other items.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3195, opened 2022-10-20
Improved Evo Stone and Item Drop Notifs

CorgiOnNeptune:
Added item images to 'evo stone used' and 'x enemy dropped item' notifications. And some minor text/style changes in itemModal.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3196, opened 2022-10-20
Pokédex layout

Crobat4:
This PR changes the Pokédex layout a bit so the filters container don't get lost the more you scroll through the Pokémon list anymore.
It also changes the checkboxes with the checkbox sliders.
pkdx-layout-1
pkdx-layout-2
pkdx-layout-3


Jaaslet (changes requested):
> the filters container don't get lost the more you scroll through the Pokémon list

We are not sure what that means... I was expecting the filters to be fixed in place when scrolling, but it does not happen for me. And i think my screen should be big enough that it's not hit by any small-screen fixes.

And for some reason, your code triggers a lot of lazy loads when you open the pokedex. So it ends up loading like half of the pokemon on load, even if the player don't scroll.

But the rest looks cool

Crobat4:
> We are not sure what that means... 

Currently, if you scrolled through the Pokémon list, when you open the filters, they appear at the top of the list, having to scroll all the way to the top to use them, in the PR I changed the filters container to stay in the screen regardless how much you scroll, it's weird it's not working for you since it's working fine in my end. [(Video)](https://user-images.githubusercontent.com/104547700/196953515-dec02846-6906-4945-9122-40e118aa8b81.webm)

The small screen fix commit just add a scroll to the filters containers for small screens (width lower than 768 px) because the container is too long for them and the filters at the bottom were inaccessible.

As showed in the video, I don't see lazy load triggers when open the Pokédex, and it's working with normality when scroll.

If this layout is not very convenient across different screens, what I can try later is to move the filters to another modal.

Jaaslet:
Just tested it again. It works as you described in chrome, but it's super broken in firefox. I have not tested other browsers

Crobat4:
It should be fixed for Firefox now.

Jaaslet (approved):
Looks good. I think it works fine on phones too.
Thanks for the PR

PR #3197, opened 2022-10-20
Fix pokedex breaking on uncaught pokemon

Jaaslet:
After the gender code was added, the pokedex broke when it needed to find the image of a uncaught pokemon. This PR fixes that.


RedSparr0w (approved):

PR #3198, opened 2022-10-20
Fixed trainer images randomly floating

Jaaslet:
With jks changes for the pokerus icons, the background image got shifted, which made the trainers randomly float.
I like how the images look after they have been shifted, so i have just moved the trainers to fit the background.
For some reason, it did not affect dungeons. I will maybe look into that later, but because they look like they did before, i have not done anything to them.

Current patch:
image

Before my change:
image

After my change:
image

### DIFF
--- a/src/components/gymView.html
+++ b/src/components/gymView.html
@@ -37,7 +37,7 @@
           </div>
         </div>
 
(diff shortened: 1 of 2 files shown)


RedSparr0w (approved):

PR #3200, opened 2022-10-20
Fixed BF notifications/log entries not displaying actual earnings.

RegisCoaxans:
Pokédollars announced now match the pokédollars earned up.
I have done the same for BP too, although the amount does not change for now.


Jaaslet (changes requested):
I think it should just ignore the bonus instead. I know the pokedollar reward is not high, but if we one day balance it better, it feels like bad design that you want your buffs to be active at the end of the run to get the bonus. gainMoney() has a parameter to ignore the bonus

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3202, opened 2022-10-20
Turn Hard Stone into a held item + add Nugget

kittenchilly:
Switches Hard Stone from a diamond item into a held item that boosts rock types. You can still get it from the underground. Also adds a new diamond item to replace it, Nugget. It's the exact same as the old hard stone except with a new sprite and name. This is so we don't have one less diamond item.


Jaaslet:
My plan was for hard stone to still be sold as a diamond item. I want to test out stuff with dual purpose, and hard stone seems like the perfect candidate

kittenchilly:
Ah right. I just thought we could get this in without having to overhaul the item code like you said. Unless it actually is easier to do.

Jaaslet:
I rather want to wait for the correct solution, instead of picking an other solution. I will close this PR

PR #3203, opened 2022-10-20
Add held items to all pokemon that canonically hold them

kittenchilly:
I thought it was weird that held items weren't held by wild pokemon, so I did that. Another way to get held items besides dungeons.

### DIFF
--- a/src/scripts/pokemons/PokemonList.ts
+++ b/src/scripts/pokemons/PokemonList.ts
@@ -703,6 +703,7 @@ const pokemonList = createPokemonArray(
         'gender': {
             'visualDifference': true,
         },
+        'heldItem': {type: ItemType.item, id: 'Silver_Powder'},
     },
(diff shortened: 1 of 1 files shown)


Jaaslet:
jk has this solution ready for the patch, but i ended up reverting it.
The problem is that we don't want players to get held items before johto. They already get a ton of new stuff to think about in kanto, so we want to spread it out.
We did not find a solution in time for the patch (and we still have not), and this PR does not have a solution for that either

jk13pclick:
I think we should leave most Held Items in Dungeons for now. We can revisit this idea when Held Items is a bit more fleshed out/stable as a mechanic, but for now having held items be mostly Dungeon exclusive loot is good, if only to make Dungeons more interesting. Going to close this, but I could see this implemented in the future - we have to let Held Items exist for a bit first

PR #3206, opened 2022-10-21
Fix Tunnel to the Top

Qwertypop04:
Fixes Tunnel to the Top incorrectly having the same encounters as Lakeside Cave


RedSparr0w (approved):

PR #3207, opened 2022-10-21
Adding a caching feature - work in progress

erzads:
Hi everyone. First of all, I am new to NodeJS/Typescript (I know a little Angular but that's it).

I am trying to implement a Caching Feature but I am struggling to understand how to properly import modules and classes.
Could someone please take a look at this PR and give me some insights?

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -83,6 +83,7 @@ const srcs = {
         'node_modules/knockout/build/output/knockout-latest.js',
         'node_modules/bootstrap-notify/bootstrap-notify.min.js',
         'node_modules/sortablejs/Sortable.min.js',
+        'node_modules/node-cache/index.js',
         'src/libs/*.js',
     ],
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "pokeclicker",
-      "version": "0.9.17",
+      "version": "0.9.19",
       "license": "ISC",
       "dependencies": {
         "@types/bootstrap": "^4.3.1",
@@ -19,6 +19,7 @@
         "intro.js": "^2.9.3",
         "jquery": "^3.5.1",
         "knockout": "^3.5.1",
+        "node-cache": "^5.1.2",
         "popper.js": "^1.16.0",
         "sortablejs": "^1.10.2"
       },
@@ -11140,6 +11141,25 @@
       "resolved": "https://registry.npmjs.org/nice-try/-/nice-try-1.0.5.tgz",
       "integrity": "sha512-1nh45deeb5olNY7eX82BkPO7SSxR5SSYJiPTrTdFUVYwAl8CKMA5N9PjTYkHiRjisVcxcQ1HXdLhx2qxxJzLNQ==",
       "dev": true
+    },
+    "node_modules/node-cache": {
+      "version": "5.1.2",
+      "resolved": "https://registry.npmjs.org/node-cache/-/node-cache-5.1.2.tgz",
+      "integrity": "sha512-t1QzWwnk4sjLWaQAS8CHgOJ+RAfmHpxFWmc36IWTiWHQfs0w5JDMBS1b1ZxQteo0vVVuWJvIUKHDkkeK7vIGCg==",
+      "dependencies": {
+        "clone": "2.x"
+      },
+      "engines": {
+        "node": ">= 8.0.0"
+      }
+    },
+    "node_modules/node-cache/node_modules/clone": {
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/clone/-/clone-2.1.2.tgz",
+      "integrity": "sha512-3Pe/CF1Nn94hyhIYpjtiLhdCoEoz0DqQ+988E9gmeEdQZlojxnOb74wctFyuwWQHzqyf9X7C7MG8juUpqBJT8w==",
+      "engines": {
+        "node": ">=0.8"
+      }
     },
     "node_modules/node-libs-browser": {
       "version": "2.2.1",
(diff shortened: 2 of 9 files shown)


Aegyo:
What's the motivation for caching calculatePokemonAttack? It only takes a few ms to run on my machine.
Also I don't think it is a good candidate for caching - the result also depends on the attack of every pokemon you have caught, not just the arguments passed into it.

<hr>

> properly import modules and classes

Only possible in the modules folder, exactly as you have done it already.
In the scripts folder everything is floating around in global scope, we want to move all of this into proper modules eventually but it is a bit of tangled mess right now.

Also, you shouldn't need to load node-cache in index.html, typescript will bundle it along with your Caching.ts since you have imported it there.

Aegyo:
Clicking shouldn't be calling calculatePokemonAttack directly, I think it's more likely you are one hitting pokemon and the tooltip is updating for the new pokemon types. The only other time I see it being called is the auto attack once per second.

Usually for this kind of thing we just put a rateLimit on the computed observable, for a tooltip I think we would have to explicitly create the computed on Battle or something like Weather does, so it can be extended with rateLimit and then used in pokemonAttackTemplate

erzads:
You got me there. I did notice these knockout observables and it took me a while to figure that rate limit stuff. But changing things regarding the tool tip is a while new level.

Also, another big performance issue I noticed happens on some update logic regarding the achievements.

I'd you got some time we could to some pair programing. I can show you the debug console running from my phone.

Also, I am gonna link you to a post I made on discord regarding this issue.

Aegyo:
Could you open an issue for the achievement thing?
That one looks to me like you have it sorted by progress, so pretty much any action is triggering a recalc. Could do with a rateLimit somewhere down that path too I think.

erzads:
@Aegyo unbelievable! I changed the sorting to default and it got noticeably faster!
Is there a place in the code where if it's not sorted by default it triggers a calculation? I will look into creating an issue for it.
I have one created about modal windows already but that seems too particular to mobile so I don't believe anyone will be interested in that...

Aegyo:
There is no place in the code that says "go recalculate this", it is just the behavior of computed observables.

Essentially, a computed observable will keep track of every observable value that is used to calculate its value, and subscribes to all of them (and triggers a recalc when one of those things updates).
The achievement modal creates a computed observable for the page of achievements it wants to see, which needs the sorted list, and if you are sorting by progress it inevitably has to go through every achievement and look at their progress (which are all observable values).
So that computed page of achievements is now subscribed to the progress of every achievement - if any of them gain some progress it will recalculate the whole thing. If you are unlucky it might trigger twice from the same action, if there are multiple achievements tracking the same thing (there are :sweat_smile:)

Making [the sorted list calc](https://github.com/pokeclicker/pokeclicker/blob/develop/src/scripts/achievements/AchievementHandler.ts#L46) a rateLimited computed observable should avoid the aggressive recalculations.

erzads:
@Aegyo I am gonna take a closer look at that.
Btw, I have been thinking about creating a configuration that would allow players to increase the default rateLimit by a percentage. Perhaps a value from 1 to 3 that could be multiplied on some of the rateLimits around the code, meaning the game would take a little longer to update some values for the user but could mean some players on slower computers/mobile could have some better responsiveness.
What do you think?

Aegyo:
I'm not sure a rate limit multiplier would have much effect, if these calculations are taking anywhere in the 500ms region most of our ratelimits are at, the game is likely completely unplayable on that device anyway. The bottleneck making the game lag would probably be somewhere else in that case.

erzads:
@Aegyo  it's very weird indeed. This is my device https://www.gsmarena.com/motorola_edge_20-11034.php

The top 3 bottlenecks  in order are:
1- Whenever any modal (even the Oak Items modal) is opened or closed there is a period of around 3-5 seconds of unresponsiveness (can't scroll the screen, interact with anything)
2- The breeding modal loads quickly the first time, because of the pagination but if closed and reopened the pagination doesn't "reset" so everything gets loaded at once and there is a big hit on performance)
3- Quickly tapping to attack is very laggy when the achievements list is filtered/sorted.

Aegyo:
1. Might be resolved with a bootstrap update, otherwise might have to look into a game setting to disable the fade animation or see if we are doing anything weird on all modal open/closes
2. I can add a resetTrigger to lazyLoad
3. Suggestion above should help with this, I haven't tested it to check if there are other achievement things happening too though.

erzads:
> 1. Might be resolved with a bootstrap update, otherwise might have to look into a game setting to disable the fade animation or see if we are doing anything weird on all modal open/closes
> 2. I can add a resetTrigger to lazyLoad
> 3. Suggestion above should help with this, I haven't tested it to check if there are other achievement things happening too though.

I did try removing the "fade" behavior from a modal but the issue remained. I can take a look at updating bootstrap tough, just for the sake of keeping it updated.

erzads:
@Aegyo FYI I am trying to update the dependencies but I am facing some issues regarding tooltips.
Apparently, bootstrap 5 uses popper/core, node popperjs, which I already updated too. Looks like an issue of order/conflict of dependencies.
`Uncaught TypeError: $(...).tooltip is not a function
    at SaveSelector.loadSaves (SaveSelector.ts:24:38)
    at HTMLDocument.<anonymous> (script.min.js?v=0.10.0:28894:18)`

`script.min.js?v=0.10.0:26362 Uncaught (in promise) TypeError: Unable to process binding "template: function(){return { name:'otherItemTemplate',data:'Protein'} }"
Message: Unable to process binding "tooltip: function(){return {
        title:ItemList[$data] ?`<u>${ItemList[$data].displayName}</u><br/>${ItemList[$data].description || ''}`:'',trigger:'hover',placement:'top',html:true,boundary:document.getElementById('showItemsModal'),} }"
Message: Cannot read properties of undefined (reading 'config')
    at update (script.min.js?v=0.10.0:26362:29)
    at a.$.l (knockout-latest.js:79:379)
    at Function.yd (knockout-latest.js:55:359)
    at Function.zd (knockout-latest.js:55:45)
    at Function.ha (knockout-latest.js:54:384)
    at a.o.a.$ (knockout-latest.js:52:147)
    at knockout-latest.js:79:366
    at Object.D (knockout-latest.js:11:52)
    at p (knockout-latest.js:78:408)
    at k (knockout-latest.js:76:170)`

`  "dependencies": {
    "@popperjs/core": "^2.11.6",
    "@types/bootstrap": "^5.2.5",
    "@types/bootstrap-notify": "^3.1.36",
    "@types/intro.js": "^2.4.7",
    "@types/knockout": "^3.4.72",
    "@typescript-eslint/parser": "^3.6.1",
    "bootstrap": "^5.2.2",
    "eslint": "^7.4.0",
    "intro.js": "^2.9.3",
    "jquery": "^3.6.1",
    "knockout": "^3.5.1",
    "sortablejs": "^1.10.2"
  }`

I am gonna keep looking at this.

PR #3210, opened 2022-10-22
Dungeons handle debuff escape attribute on loots

RegisCoaxans:
Loot tables now allow for items with undebuffable rate.
Following Jessica's idea, the runner tries to pick an undebuffable item from the undebuffed loot table, then, if it fails, tries and only picks from debuffable ones.
However, I have kept the loot tables as is; I can add the attribute to desired items myself if required.


Aegyo:
Can we pull the debuff check out of getRandomLootTier and calculate that when opening the chest, so both getRandomLoot and getRandomLootTier are passed the same debuff flag? Just feels a little disjointed at the moment.

Jaaslet:
We want it on the zen form for now, so you can just add it there

RegisCoaxans:
> Can we pull the debuff check out of getRandomLootTier and calculate that when opening the chest, so both getRandomLoot and getRandomLootTier are passed the same debuff flag? Just feels a little disjointed at the moment.

I have somehow done it. Although, `debuffed` parameter in `dungeon.getRandomLoot` was only meant to filter out undebuffable loots on the second draw. I tweaked the variable name and changed the [condition](https://github.com/RegisCoaxans/pokeclicker/blob/814b41572e78a6849e7bf18a0ab34d1e12af5520/src/scripts/dungeons/DungeonRunner.ts#L124); it should feel more both readable and consistent now.

Jaaslet (approved):
I think it works. I have changed debuffable to ignoreDebuff in the code.
Thanks for the PR

PR #3211, opened 2022-10-22
Add polyfill requestIdleCallback

RedSparr0w:
Required for Safari browsers to load the Hatchery.

Compatibility:
https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback#browser_compatibility
Refer Discord thread here:
https://discord.com/channels/450412847017754644/1031490985257095248/1031681769193353257

### DIFF
--- a/src/modules/index.ts
+++ b/src/modules/index.ts
@@ -1,3 +1,6 @@
+// Get our polyfills loaded first
+import './polyfill';
+
 import './temporaryWindowInjection';
 
 // Ensure that the Knockout Extenders are injected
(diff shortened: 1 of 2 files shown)


Aegyo (approved):
Testing on [Surf](https://surf.suckless.org/) since I don't have easy access to safari, but it shouldn't make too much difference.

It looks like scrolling down to the second page doesn't load things quick enough and we just end up loading all pages, but that is almost what we had before the lazy loading anyway and it beats being completely broken :+1:
Probably something we can do in the lazy loader to fix that, but I don't think it is needed for this PR.

PR #3215, opened 2022-10-22
Fix some Hoenn background scenes

CorgiOnNeptune:
Changes the background for Jagged Pass from Mansion to Cave. 
Adds Magma Hideout to Fire.
Adds Aqua Hideout to PowerPlant.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3218, opened 2022-10-22
Fix new game not starting

Jaaslet:
Fixed the bug in the following thread:
https://discord.com/channels/450412847017754644/1033293034152001586
If you start a new save, but still have cookies from an old, you get this error. So i added a null check


RegisCoaxans:
Your check seems to make the text of the button always « All » or « None ».
Right here, only three regions selected out of four but « All » is displayed instead of « Sinnoh & more ».
image

The null check is already a thing but is nested in the « All » one.
I think that flipping them instead could suffice for what you want to do.

src/components/breedingModal.html
```diff
@@ -58,7 +58,7 @@ <h5 class="modal-title" id="breedingModalLabel">Day Care</h5>
                                 <label>Regions</label><br/>
                                 <div class="dropdown show">
                                     <button type="button" class="text-left custom-select col-12 btn" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" data-bind="text :
-                                        BreedingFilters.region.value() == 2 ** (player.highestRegion() + 1) - 1 ?
+                                        BreedingFilters.region.value() || BreedingFilters.region.value() == 2 ** (player.highestRegion() + 1) - 1 ?
```
Aegyo, line 61: As mentioned by Regis, this ternary always returns 'All' for any truthy region filter value now.
Maybe something like
```suggestion
                                        BreedingFilters.region.value() == undefined || BreedingFilters.region.value() == 2 ** (player.highestRegion() + 1) - 1 ?
```

Aegyo (changes requested):

src/components/breedingModal.html
RegisCoaxans: Actually, `undefined` is not the issue.
People stuck in the loading screen somehow kept the old value `-2` from before the update; but on a fresh new save...
It breaks because that can not be properly handled as a bitfield and results in inconsistent behaviors.

src/components/breedingModal.html
```diff
@@ -58,7 +58,7 @@ <h5 class="modal-title" id="breedingModalLabel">Day Care</h5>
...
                                             'All' :
                                             (BreedingFilters.region.value() ?
```
RegisCoaxans, line 63: Changing this null check to `BreedingFilters.region.value() > 0` will stop `-2` from being truthy as well.

Aegyo (approved):

PR #3229, opened 2022-10-23
Makeing Vivillion Forms availiable after finishing The Great Vivillon Hunt!

Nydaleclya:
Vivillion Forms available after completing "The Great Vivillon Hunt!" so people may farm EV's on them if they wish to.


Jaaslet:
I will close this PR.
It's just weird for them to stick around. And it will make it a lot harder to get 500 clears in that dungeon.
We have solutions planned for EVs

PR #3232, opened 2022-10-23
WIP - trying to limit the attack calculations

erzads:
@Aegyo Would you please take a look at this? I don't know if this is close to what you imagined. I am missing something because it is still calling the calculate attack method every time I one shot a route by clicking.


Aegyo:
Pretty close, but it references enemyPokemon in the tooltip binding still so skips right past your rateLimit.

have your observable return the full text used in the tooltip (<number> against <pokemon>)

Also that name is quite long, I would just call it pokemonAttackTooltip

erzads:
@Aegyo how about it now? I am not so sure it worked as intended because I have a console.log at Party.calculatePokemonAttack and it logs everytime a one hit a pokemon ( I am testing on route 1)

Screenshot from 2022-10-24 12-20-57

erzads:
I think I got it. I believe the new function I created that returns a KnockoutComputed must use observables in it.

Aegyo (approved):
:+1:

PR #3233, opened 2022-10-23
Boss are completely excluded from enemy counter in dungeon

RegisCoaxans:
Bosses are excluded from the enemy counter at first but still increment it on defeat.
It leads to the dungeon showing [one too many defeated encounter](https://cdn.discordapp.com/attachments/1031364212032815144/1031364212607426650/Screenshot_2022-10-16_16.31.47.png) when the player tries and catches the boss Pokémon.
This PR solves that.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3235, opened 2022-10-23
Fix Flowering Celebi Gender

CorgiOnNeptune:
Fixed Flowering Celebi gender and future-proofed the commented Megas gender ratios.


Jaaslet (approved):
I considered if we needed an update function. But if we are turning it genderless, it will just show total caught... Which is the correct number... So it's not needed
Looks good. Thanks for the PR

PR #3237, opened 2022-10-24
Fix: Remove BF milestones from save if corresponding Pokémon is missing

PixLSteam:
This PR makes the 0.10.1 update function remove cleared milestones from the save data if the corresponding Pokémon is not in the party.
I also removed the `addPokemonToSaveData` helper function as data structure changes might cause older update functions to break if they rely on this. I removed the Deoxys forms from the 0.6.0 update function as they relied on this helper and the 0.10.1 update should take care of them anyway.

Link to post with affected save: https://discord.com/channels/450412847017754644/1034075029366906950


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3242, opened 2022-10-24
Give the player more battle cafe spins

### LINKED ISSUE #3221: [SUGGESTION] Alcremie EVs grind
**Current Behavior**
Currently Battle Cafe has a daily limit of 3 spins a day. It was discussed on discord that it would take months to have alcremie resistant with EV challenge off and years to fully make all 63 alcremie resistant to pokerus with the EV challenge on. 

It was also brought up that in the Sw/Sh games, you were only 1 spin a day.

**Improved Behavior**
Change the Battle Cafe to allow less of a grind for EVs after getting all alcremie forms.

**Reason for change**
Doing the change would make the grind to fully resist all alcremie forms less of a needed grind for those who want to fully resist all pokemon. #3173 is being worked on, but this would allow another way to gain EVs while still using the mini game created.

**Detailed Description**
Two options were presented.
1. Create a new achievement for getting all the alcremie forms. After achievement is triggered, the cap on spins per day can be removed or increased to a much higher number like 10.
2.  A quest line afterwards that would remove the limit or increase daily spin limit.

**Screenshots:**
N/A

Jaaslet:
Closes #3221
I saw some math on how long it would take to fully resistant all the cake mons, so i decided to give the player a bit more spins.
Each time they get all mons from a berry, they gain an additional daily spin (triggers the next day).


RedSparr0w (approved):

PR #3244, opened 2022-10-25
Redid Braille Images

arduousFrivolity:
Redid the Braille Images for the Regi Quests based on feedback as to their readability. 

I hope I never have to make these exact 5 images ever again.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3245, opened 2022-10-25
Wider profile layout and shiny/gender options

Crobat4:
Just a small change to the profile layout, plus adding the shiny and gender checkboxes.
_Note: Female was changed to Gender_
profile-3
profile-4


Jaaslet (changes requested):
Looks good. I just have one comment:
The gender box is kinda weird, when they don't have a gender difference. I am considering if it should just be hidden, when the mon has no gender difference. I am not sure, so if you think this is the best solution, i am fine with it.
I think it's fine how the shiny checkbox works. Because it's something you can do something about, so players would like to know the option is there

Crobat4:
Gender checkbox is hidden now if the Pokémon has no gender differences

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3248, opened 2022-10-25
Apply a license

FlorentPoinsaut:
Hello,

Thank you for this great game!
Please, could you apply an open-source software license to your code? I suggest you add the LICENSE GPLv3.

Help by GitHub: https://help.github.com/en/articles/adding-a-license-to-a-repository

Thank you in advance.

Regards,


Jaaslet:
I am not the one who should decide of we should have a license or not. But here is my take:
It feels like a bad idea to add a license for a project, which contains a lot of stuff we don't own. We have a lot of official Pokemon art, and a lot of images we are only allowed to use, if we give credits. I am not sure how the rules work, but i don't think we can give permission to use this

FlorentPoinsaut:
Thanks for your answer @Jaaslet . It's a big problem, it's true.
What I can propose is that you can apply this license only to the code.
Would it be better?

Aegyo:
Applying the license to just the code is fine, but we would need permission first from everyone who has contributed code (maybe just code we are still using?)

I'm getting this information from https://opensource.guide/legal/#what-if-i-want-to-change-the-license-of-my-project

FlorentPoinsaut:
That right!

About one hundred contributors according to this command:
```sh
git log --format="%an" | sort -u | wc -l
```
> 107

I'll be looking for a success story about a license change...

PS: If you want to retrieve their emails, you can use this:

```sh
git log --format="%an <%ae>
```

RedSparr0w:
Going to close this and move it to an issue - #3296

PR #3255, opened 2022-10-27
Berrydex : display « Total Life Span » instead of « Total Growth Time »

RegisCoaxans:
*Total Growth Time* is already presented along with stages anyway.
Therefore, this PR introduces the wither time in a nice manner, useful information that the player could not read in game.
image

### DIFF
--- a/src/components/berryDexModal.html
+++ b/src/components/berryDexModal.html
@@ -27,8 +27,8 @@
                   </thead>
                   <tbody>
                     <tr>
-                      <td class="text-left">Total Growth Time</td>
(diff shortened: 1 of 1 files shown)


erzads:
I think that makes it confusing. It's not clear that the growth stages are not additive.
I would maybe add the information you want, instead of replacing the current information.

RegisCoaxans:
Players learn from using the farm that the stage durations are not additive. Thus the information looks redundant.
this is why I replaced the information instead of puting both. Moreover, I think that having both could confuse people into thinking that the lifespan does not include the growth time.

Aegyo:
I wouldn't rely on players learning anything, safer to just include both lines imo.

RegisCoaxans:
There go these players !

PR #3266, opened 2022-10-28
Prevented BF from spoiling in pokédex.

RegisCoaxans:
Battle Frontier still generates enemies based on the highest region visited : its diversity is not compromised.
Although, their statistic does not change upon defeat. Thus they do not appear in the pokédex if they have not been counted outside of the BF, nor do the unseen pokémons up to those.


src/scripts/pokemons/BattlePokemon.ts
```diff
@@ -1,5 +1,10 @@
 /// <reference path="../../declarations/GameHelper.d.ts" />
 
+interface battleDefeatOptions {
```
erzads, line 3: Shouldn't this interface's name start with a capital letter?

erzads (approved):

Jaaslet:
I am going to close this. I don't mind the BF spoiling base forms. They are not really a secret, as they just follow the main games.
But even if we wanted to fix this, not adding to the statistics is not the correct solution. You have still defeated them, even if they are in the BF

PR #3268, opened 2022-10-28
Use underground item found setting for treasure scanner bonus items

### LINKED ISSUE #3264: [SUGGESTION] Disable notifications for bonus Treasure Scanner items
**Current Behavior**
Disabling "Item found when mining" notifications will suppress the notification for the initial item, but still display notifications for any bonus items found via Treasure Scanner. Currently, the only way to suppress the Treasure Scanner notifications is to disable all notifications.

**Improved Behavior**
Two alternative improvements:
- Disabling "Item found while mining" notifications also suppresses the Treasure Scanner notifications
- A separate option to disable the Treasure Scanner notifications

**Reason for change**
Users should have a way to suppress the Treasure Scanner notifications without having to disable all notifications.

Aegyo:
Fixes #3264


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3269, opened 2022-10-29
Update daycare modal Pokemon styling

### LINKED ISSUE #3261: [SUGGESTION] Have names wrap when overflowed in Day Care
Long Pokemon names can extend super wide and end up either off-screen or underneath other names, at least within the Day Care.

**Current Behavior**
This image is from the Day Care. As you can see, super long names get pushed off to the sides.
image

**Improved Behavior**
Have the names wrap on overflow. I've created a rough mock-up of what it could potentially look like:
image

**Reason for change**
It would be nice to know exactly what Pokemon I'm selecting, instead of having to guess based on the parts of the name that are currently visible.

**Detailed Description**
Pokemon with long names, such as the Alcremie forms, can have their names go off-screen or otherwise become difficult to read because they're wider than the Day Care list is capable of showing. By wrapping the names around to the next line, you could mitigate this issue, allowing players to more easily read Pokemon names.

CorgiOnNeptune:
Allows Pokemon name text to wrap in the daycare modals list. 
<details><summary>The name overflows were starting to get a little crazy</summary>

</details>

<details><summary>w/ text wrapping</summary>
 
</details>

<details><summary>Elements Re-Organized</summary>
   
</details>

Fixes #3261


src/styles/breeding.less
```diff
@@ -68,6 +67,27 @@
     opacity: .5;
     pointer-events: none;
   }
+  .pokemonName {
```
Jaaslet, line 70: These class names are a bit generic. We have pokemon names all over the place, but this only affects one of the places

Jaaslet (changes requested):
I like the design. I just have one code comment:

src/styles/breeding.less
CorgiOnNeptune: Made the class names more specific to their actual use-case.

PR #3272, opened 2022-10-29
Fix undefined pokerus image in dungeonView

CorgiOnNeptune:
Wild Pokemon uninfected with Pokerus were giving a broken image link in dungeons.

And a couple random '_x_-type' typos...


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3273, opened 2022-10-29
More pokemon

xslk:
Changed Floette Red in Poni meadow to blue, few sources show it should be blue here and not read.
Added additional pokemon to routes with obtained first requirement, this allows EV farming for more pokemon and not have them locked.
Added another farm wanderer.


src/scripts/wildBattle/RouteData.ts
```diff
@@ -474,6 +474,8 @@ Routes.add(new RegionRoute(
     'Pinkan Forest', GameConstants.Region.kanto, 41,
     new RoutePokemon({
         land: ['Pinkan Caterpie', 'Pinkan Weedle', 'Pinkan Pidgey', 'Pinkan Vileplume', 'Pinkan Paras', 'Pinkan Venonat', 'Pinkan Mankey', 'Pinkan Exeggutor'],
+        special:
+    [new SpecialRoutePokemon(['Pinkan Pidgeotto'], new ObtainedPokemonRequirement(pokemonMap['Pinkan Pidgeotto']))],
```
Jaaslet, line 478: It does not show up in the forest, in the anime, which is why we did not add it to the routes.
For the rest of the mons, i will ask someone who have played more of the games to look it through

xslk: This one was the only one I took the liberty of adding that is outside of normally obtainable areas.

Jaaslet (changes requested):
Notes from Qwerty:
- The Vespiquen, Swalot and Yanmega in Unova are all from Sinnoh or before, they don't need the requirements. Same goes for the Dragonair in Kalos, the Shelgon, Ambipom, Pyroar and Dragonair in Alola.
- The Toucannons and Salandit in Alola aren't egg exclusive, I don't think they need the requirement (Salandit is dungeon or egg only currently, so maybe it does, but I personally don't think so.) 
- The Galarian Darumaka on Snowslide Slope does not need a requirement, and is in fact already on the route. Not sure why sLk did that, but it should be removed, as unnecessary
- The Clobbopus on Alola route 9 should definitely not be there. It's a Galar Pokémon. It should be on Galar route 9.

Ultima1990:
Floette in Meadow is red because blue is supposed to be wanderer exclusive.

Qwertypop04:
> Floette in Meadow is red because blue is supposed to be wanderer exclusive.

Flabébé Blue is still wanderer exclusive 🤷‍♂️

Ultima1990:
> Flabébé Blue is still wanderer exclusive 🤷‍♂️

So?

xslk:
> Notes from Qwerty:
> 
> * The Vespiquen, Swalot and Yanmega in Unova are all from Sinnoh or before, they don't need the requirements. Same goes for the Dragonair in Kalos, the Shelgon, Ambipom, Pyroar and Dragonair in Alola.

I had the complete pokedex challenge off when adding the catch first requirements for these. I can fix it if I shouldn't consider it.

> * The Toucannons and Salandit in Alola aren't egg exclusive, I don't think they need the requirement (Salandit is dungeon or egg only currently, so maybe it does, but I personally don't think so.)

Toucannon might not need the obtain first, but Salandit was dungeon only so I added it that way.

> * The Galarian Darumaka on Snowslide Slope does not need a requirement, and is in fact already on the route. Not sure why sLk did that, but it should be removed, as unnecessary

Too many words in those weather list, I think missed it on accident. 

> * The Clobbopus on Alola route 9 should definitely not be there. It's a Galar Pokémon. It should be on Galar route 9.

Yup it should be Galar. My mistake!

xslk (commented):
Updated files with the said changes, also spoke with Qwerty and added additional pokemon in new upload.

Jaaslet (changes requested):
Some of the special route mons have weird indentation/linebreaks. Can you change it so it looks more consistent?

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3275, opened 2022-10-29
Set subregion for all towns

Daniferrito:
Currently, all town and dungeon towns (except a few Sevii islands) have no subregion information (which default to subregion 0)

This PR sets the subregion as a mandatory field for all towns, and adds the appropiate subregion for all of them.

While this does not solve any issue in this project currently, it makes sure that all future towns have the correct information, and makes easier implementing things related to subregions, like the sevii archievements.


Jaaslet (approved):
Looks good. Thanks for doing all the boring work.
I have done some code cleanup, but not much.

PR #3279, opened 2022-10-29
Unova seasons

Nydaleclya:
Deerling und Sawsbuck forms are now available dependent on the current month as in the BW games.

Making SpecialEvent code better to read.


Jaaslet:
Og god... We are not that mean... I will close this before people complain too much

DataCrusade:
I don't think it's a bad idea exactly but it's something that we would need to let people use the #2200 Event Calendar to be able to change the season

PR #3280, opened 2022-10-30
Giratina Questline

HyruleTeam64:
Adds a questline for Giratina, focused around finding an entrance to the Distortion World. It is intended to be completed in late Sinnoh.
Also adds the Griseous Orb Held Item. It can be dropped by Giratina Altered and gives it a 50% attack bonus. Giratina Altered can evolve into Giratina Origin by reaching level 70 with the Griseous Orb equipped.


Farboo171:
A few comments from the devs, relayed here:

- The overall structure of the quest looks good!

- The method for obtaining Giratina-A and O at the end is somewhat inconsistent with the general approach to Legendary quests we're shooting for. Rewarding an already-existing alt form is the main inconsistency

- It would be more in line with other quests if you kept Giratina-A as the boss for Distortion World, replaced the reward stage with a Griseous Orb reward, and made Giratina-O a held item evolution of Giratina-A

- Rather than the TemporaryBattleRequirement('Giratina') quest start requirement, completion of the "A New World" questline is probably more appropriate.

Do these seem like changes you could make?

HyruleTeam64:
> It would be more in line with other quests if you kept Giratina-A as the boss for Distortion World, replaced the reward stage with a Griseous Orb reward, and made Giratina-O a held item evolution of Giratina-A

I'm not sure if I have understood this correctly, should I keep the Giratina temporary battle after A New World and have it give the Orb instead of Giratina-A or remove the battle and give the Griseous Orb during the Giratina quest?

> Rather than the TemporaryBattleRequirement('Giratina') quest start requirement, completion of the "A New World" questline is probably more appropriate

I used that requirement because, if players who have beaten Sinnoh completed the quest without beating Giratina, there would be two Distortion Worlds in the map. If I remove the temporary battle requirement, I don't think the A New World one would be required, since you need to have completed the quest in order to beat Cynthia.

I'll change the PR into a draft in order to make all the necessary changes.

Farboo171:
Keeping the Giratina-A temporary battle is ok. Now that I read the quest steps more closely, I'd say move the Orb to be a reward for beating Zero.  You can make the last steps:
1. Beat Zero, which gives you an Orb
2. Catch Giratina-A, who is the boss of Distortion World
3. Evolve Giratina-O using the Orb

As Ultima brought up, the Giratina temporary battle should be part of a quest line. I think adding it to the beginning of this quest would work fine if you don't want to muck around in already finished quests.

Qwertypop04:
The Giratina O temp battle really needs to be part of "A New World" or to not exist at all. It would be necessary for progression, where this quest would presumably be a quest board questline, and it would be somewhat confusing to have it not be part of A New World when it's required to go to Sunnyshore

HyruleTeam64:
> The Giratina O temp battle really needs to be part of "A New World" or to not exist at all. It would be necessary for progression, where this quest would presumably be a quest board questline, and it would be somewhat confusing to have it not be part of A New World when it's required to go to Sunnyshore

I initially added it outside of A New World so players who have completed A New World but haven't caught Giratina-A could still get it. But if we are doing Giratina-O as a held item evolution, I don't think the Giratina temp battle should exist, since players could still catch Giratina-A as a boss in the Distortion World.

Qwertypop04:
> I initially added it outside of A New World so players who have completed A New World but haven't caught Giratina-A could still get it. But if we are doing Giratina-O as a held item evolution, I don't think the Giratina temp battle should exist, since players could still catch Giratina-A as a boss in the Distortion World.
That makes sense. At this point it has no essential purpose, other than flavour and canon, so including it or not works

HyruleTeam64:
> You can make the last steps:

Beat Zero, which gives you an Orb
Catch Giratina-A, who is the boss of Distortion World
Evolve Giratina-O using the Orb

I think that would be the best way to do it. Giratina-O should probably be a Questline restricted evolution in order to prevent players who already have Giratina from wasting the Orb before catching the Pokemon for the quest.

Qwertypop04:
> I think that would be the best way to do it. Giratina-O should probably be a Questline restricted evolution in order to prevent players who already have Giratina from wasting the Orb before catching the Pokemon for the quest.
>
Given that the goal of the player during the questline, isn't to obtain Giratina (Origin), but to return to the Distortion World, I don't think that obtaining Giratina-O actually needs to be something required to complete the quest. Maybe it would make sense to have the quest finish upon catching Giratina-A? Then we wouldn't have to give the Griseous Orb for free (could have it be dungeon loot or a drop from Giratina-A?).

HyruleTeam64:
I think I've implemented all of the suggested changes. I'll mark the PR as ready for review again. Let me know if there is anything else that needs to be changed.

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -766,6 +766,65 @@ class QuestLineHelper {
...
+        const obtain10PurpleShards = new CustomQuest(10, undefined, 'Obtain 10 Purple Shards.', () => player.mineInventory().find(item => item.name == 'Purple Shard').amount());
+        giratinaQuestLine.addQuest(obtain10PurpleShards);
+
+        const talktoAzelf = new TalkToNPCQuest(ValorAzelf, 'Ask Azelf about the Distorsion World in Lake Valor.');
+        giratinaQuestLine.addQuest(talktoAzelf);
+
+        const obtain10OchreShards = new CustomQuest(10, undefined, 'Obtain 10 Ochre Shards.', () => player.mineInventory().find(item => item.name == 'Ochre Shard').amount());
+        giratinaQuestLine.addQuest(obtain10OchreShards);
+
+        const talktoUxie = new TalkToNPCQuest(AcuityUxie, 'Ask Uxie about the Distorsion World in Lake Acuity.');
```
Aegyo, line 794: Distorsion -> Distortion
```suggestion
        const talktoMesprit = new TalkToNPCQuest(VerityMesprit, 'Ask Mesprit about the Distortion World in Lake Verity.');
        giratinaQuestLine.addQuest(talktoMesprit);

        const obtain10PurpleShards = new CustomQuest(10, undefined, 'Obtain 10 Purple Shards.', () => player.mineInventory().find(item => item.name == 'Purple Shard').amount());
        giratinaQuestLine.addQuest(obtain10PurpleShards);

        const talktoAzelf = new TalkToNPCQuest(ValorAzelf, 'Ask Azelf about the Distortion World in Lake Valor.');
        giratinaQuestLine.addQuest(talktoAzelf);

        const obtain10OchreShards = new CustomQuest(10, undefined, 'Obtain 10 Ochre Shards.', () => player.mineInventory().find(item => item.name == 'Ochre Shard').amount());
        giratinaQuestLine.addQuest(obtain10OchreShards);

        const talktoUxie = new TalkToNPCQuest(AcuityUxie, 'Ask Uxie about the Distortion World in Lake Acuity.');
```

src/scripts/towns/TownList.ts
```diff
@@ -2597,13 +2633,58 @@ const SunyshoreRibbonerJulia = new NPC('Ribboner Julia', [
...
+    'You know what? You are fired, I no longer need you. Now, give me the key to the Distortion World or I will take it by force!',
+], {
+    image: 'assets/images/temporaryBattle/Zero.png',
+    requirement: new MultiRequirement([new QuestLineStepCompletedRequirement('Zero\'s Ambition', 11), new QuestLineStepCompletedRequirement('Zero\'s Ambition', 12, GameConstants.AchievementOption.less)]),
+});
+
+const SendoffSpringZero2 = new NPC('Zero', [
+    'Seems like you have managed to catch the Pokémon that rules the Distortion World.',
+    'But this doesn\'t look like the Renegade Pokémon from legend, it seems like some sort of altered form.',
+    'I once read that, deep within the Distortion World, there is an object capable of carrying the realm\'s power. Maybe, if you train Giratina while it has such object equipped, it can turn into it\'s true form.',
```
Aegyo, line 2682:
```suggestion
    'I once read that, deep within the Distortion World, there is an object capable of carrying the realm\'s power. Maybe, if you train Giratina while it has such object equipped, it can turn into its true form.',
```

Aegyo (dismissed):
Mostly just looked over the text for any issues, haven't run the code.

Did you check if adding that bulletin board in the middle causes any problems for old saves?

HyruleTeam64:
> Did you check if adding that bulletin board in the middle causes any problems for old saves?

It didn't cause any problems when I tested it, but I may have missed something.

Aegyo:
You've got a conflict here because of items moving into modules

HyruleTeam64:
I've fixed the conflict, thanks for letting me know.

src/scripts/towns/TownList.ts
```diff
@@ -2623,13 +2659,58 @@ const SunyshoreRibbonerJulia = new NPC('Ribboner Julia', [
 const FightAreaAceTrainer = new NPC('Ace Trainer Quinn', [
     'Something amazing happened on top of Mt. Coronet. We could see it all the way from here. I\'m sure everyone in the entire region saw it.',
     'What? You were there? What happened? What was that purple thing?',
-    'The Distortion World? Hold on, I think I\'ve read about that in an old book in the Canalave City library. But according to that book the only entrance to the Distortion World is hidden deep within Sendoff Spring. This is quite the discovery my friend.',
+    'The Distortion World? Hold on, I think I\'ve heard that name before, there was a guy around here named Zero looking for a way to enter that place.',
 ], {image: 'assets/images/trainers/Ace Trainer (male).png'});
 
+const FightAreaZero1 = new NPC('Zero', [
+    'You\'re the Champion, right? I need your help on a small errand.',
+    'My old friend Newton used to study the Distortion World, he was working on opening a gate to it. I\'ve heard that you were able to enter the Distortion World from a portal at the top of Mt. Coronet, so I was wondering if you could help me open a new portal to fufill my friend\'s wish.',
```
CorgiOnNeptune, line 2667:
```suggestion
    'My old friend Newton used to study the Distortion World, he was working on opening a gate to it. I\'ve heard that you were able to enter the Distortion World from a portal at the top of Mt. Coronet, so I was wondering if you could help me open a new portal to fulfill my friend\'s wish.',
```

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -766,6 +766,65 @@ class QuestLineHelper {
...
+            ], 'Charge the key to the Distortion World.');
+        giratinaQuestLine.addQuest(chargeDistortionKey);
+
+        const talktoZero3 = new TalkToNPCQuest(SendoffSpringZero1, 'Talk to Zero in Sendoff Spring.');
+        giratinaQuestLine.addQuest(talktoZero3);
+
+        const clearZero = new CustomQuest (1, 0, 'Defeat Zero to protect the key.', () => App.game.statistics.temporaryBattleDefeated[GameConstants.getTemporaryBattlesIndex('Zero')]());
+        giratinaQuestLine.addQuest(clearZero);
+
+        const catchGiratina = new CaptureSpecificPokemonQuest('Giratina (Altered)', 'Catch the ruler of the Distortion World', 1, false, 0, undefined);
```
CorgiOnNeptune, line 819:
```suggestion
        const catchGiratina = new CaptureSpecificPokemonQuest('Giratina (Altered)', 'Catch the ruler of the Distortion World.', 1, false, 0, undefined);
```

CorgiOnNeptune:
Commented for two small typos.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3285, opened 2022-10-31
Fix a few typos in NPC interactions

equinox:
Noticed a missing apostrophe in an NPC's dialogue. Then while trying to find that in the code, I found a few more that needed fixing up.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3287, opened 2022-11-01
Removes overlap on Alola Route 17

Qwertypop04:
Removes an overlap in the SVG of Alola Route 17


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3288, opened 2022-11-01
Skip confirmation when quitting quests that have no progress

### LINKED ISSUE #3027: [SUGGESTION] Quest 'Quit' option to skip confirm if no progress
**Current Behavior**
Currently, when you press the 'quit' option on a quest, it will ask for confirmation and state that you will lose all progress if you quit. This happens regardless of the actual quest progress.

**Improved Behavior**
A possible QoL feature would be to skip confirmation if the quest doesn't have any progress yet.

**Reason for change**
This way, it's more convenient to change your mind on which quests you're doing.

**Detailed Description**
The quest-quit confirmation dialog will not appear if the user presses quit while at 0 progress; instead, it quits the quest right away.

CypherX:
When quitting a quest that has no progress the confirmation prompt will be skipped.

Closes #3027

### DIFF
--- a/src/components/questDisplay.html
+++ b/src/components/questDisplay.html
@@ -31,7 +31,7 @@
                     </td>
                     <td width="30%" class="p-0">
                         <!-- ko ifnot: $data.isCompleted() -->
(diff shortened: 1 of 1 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3289, opened 2022-11-01
Revamping the collapsibility of the QuestDisplay

RegisCoaxans:
Clicking on the QuestDisplay header now only collapses a daily quest block.
Questline blocks individually collapse when we click on their own respective header.

With more and more questlines, this change helps keep the UI reasonably well-sized and readable while starting all the quests that we want at once.

qUI


Aegyo:
While we're here, doesn't this list feel a bit cramped to anyone else? How does it look with a small margin between each questDisplayBlock?

Aegyo (approved):

RegisCoaxans:
There is already a small 1px margin between each block. Do you think that it should be wider ? How much ?
I can not really say as it still looks readable to me. It might also come from the fact that I use a theme which strongly compacts things.

Here are some examples; 1, 3, 5px between blocks :
qUI

Aegyo:
I like the 5px one but I don't really mind if it is changed or not, I'll leave it for Jaaslet to decide

Jaaslet:
All 3 looks good to me... Sorry i am not more helpful :p

RedSparr0w:
I think I like 1px, with 5px margin at the bottom of the progress bars 🤷

RegisCoaxans:
I agree with that last idea, and I think that it looks good. Do I push this change ?
qUI

RedSparr0w:
Looks good to me! 👍

RegisCoaxans:
There we go. Thank you for the advices.

src/styles/quest.less
```diff
@@ -37,15 +37,22 @@
...
-  }
+	
+	table {
+		margin-bottom : 4px;
+	}
 }
+
+.questDisplayBlock:last-child table {
+	margin-bottom : 0px;
+}
```
RedSparr0w, line 58: I believe this should also work with less, without having to do it separately: 
```suggestion
    }
    &:last-child table {
        margin-bottom : 0px;
    }
}
```

RegisCoaxans:
It does work.

RedSparr0w:

PR #3292, opened 2022-11-02
Use proper ko bindings for pokedex filters

Aegyo:
Should fix the Rare Held Item toggle defaulting on that some people are experiencing, and removes the need for manually asking the pokedex list to update, querying the dom for filter values, and manually adding html for some of the selections (populateFilters).

I wrapped the highestDex calc in a computed we peek at to avoid subscribing to all of the defeat/capture/seen statistics, so we aren't running this filter on every hatch/defeat, with the tradeoff of the highestDex number only updating when the modal opens or you change a filter setting.


src/scripts/pokedex/PokedexHelper.ts
```diff
@@ -3,6 +3,16 @@ import TypeColor = GameConstants.TypeColor;
 class PokedexHelper {
     public static toggleStatisticShiny = ko.observable(true);
     public static hideShinyImages = ko.observable(false);
+    public static filters = {
```
Jaaslet, line 6: Is it possible to just bind to the setting, instead of having this object?
Then the filters should be saved too, right?

Jaaslet (changes requested):
Looks good. And i think it will solve the problem.
I found one bug: If you set type=normal, it's the same as type=all.
And then i also have a code comment:

src/scripts/pokedex/PokedexHelper.ts
Aegyo: :+1:

Aegyo:
Update with requested changes, Normal type filter seems to be working fine, and filters are remembered through Settings.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3299, opened 2022-11-03
Fix Hiker Lenny's pokemon

GLoarer:
Hiker Lenny in the Rock Tunnel Dungeon has 3 geodudes and 1 onix in Pokeclicker. As I was translating the Wiki page and looking for his french name, I saw on Pokepedia, Bulbapedia etc... That he in fact has 3 geodudes and 1 machop (in all versions or the pokemon games). So I changed Onix to Machop.


Jaaslet (approved):
More machops! <3
Thanks for the PR

PR #3305, opened 2022-11-05
[module]: Move some Item classes into modules

Aegyo:
Everything seems to be working, but I could have missed something.
The only thing I am unsure about is this part of Item.ts constructor I had to remove:
```ts
        if (!ItemList[this.saveName]) {
            ItemList[this.saveName] = this;
        }
```
I needed to remove it to avoid a cyclical dependency, and I couldn't see why this was ever needed.


Jaaslet (approved):
The code looks good. I have tested some different items stuff, and it all works.
For the code you removed, i can see a usecase for it. All the Alcremies are not added to the ItemList (because i was lazy), which will be a problem when we need to find all shopmons for jk's EV changes... But i should just fix it instead, so we don't need that code... I will do it on his PR, when we get there.

Thanks for the PR

PR #3309, opened 2022-11-05
BreedingEfficiency+ Setting

jk13pclick:
Working?


src/scripts/breeding/BreedingController.ts
```diff
@@ -206,7 +206,7 @@ class BreedingController {
             case 'baseAttack': return `Base Attack: ${pokemon.baseAttack.toLocaleString('en-US')}`;
             case 'eggSteps': return `Egg Steps: ${App.game.breeding.getSteps(pokemonData.eggCycles).toLocaleString('en-US')}`;
             case 'timesHatched': return `Hatches: ${App.game.statistics.pokemonHatched[pokemonData.id]().toLocaleString('en-US')}`;
-            case 'breedingEfficiency': return `Efficiency: ${((pokemon.baseAttack * (GameConstants.BREEDING_ATTACK_BONUS / 100) + pokemon.proteinsUsed()) * BreedingController.calculateRegionalMultiplier(pokemon) / pokemonMap[pokemon.name].eggCycles).toLocaleString('en-US', { maximumSignificantDigits: 2 })}`;
+            case 'breedingEfficiency': return `Efficiency: ${((pokemon.baseAttack * (GameConstants.BREEDING_ATTACK_BONUS / 100) + pokemon.proteinsUsed()) * BreedingController.calculateRegionalMultiplier(pokemon) / pokemonMap[pokemon.name].eggCycles * (Settings.getSetting('breedingEfficiency+').observableValue() ? pokemon.calculateEVAttackBonus() : 1)).toLocaleString('en-US', { maximumSignificantDigits: 2 })}`;
```
Jaaslet, line 209: Can we get some linebreaks here? To make it easier to read

src/components/settingsModal.html
```diff
@@ -36,6 +36,7 @@ <h5 class="modal-title">Settings</h5>
                                 <tr data-bind="template: { name: 'MultipleChoiceSettingTemplate', data: Settings.getSetting('eggAnimation')}"></tr>
                                 <tr data-bind="template: { name: 'MultipleChoiceSettingTemplate', data: Settings.getSetting('hideHatchery')}"></tr>
                                 <tr data-bind="template: { name: 'MultipleChoiceSettingTemplate', data: Settings.getSetting('farmDisplay')}"></tr>
+                                <tr data-bind="template: { name: 'BooleanSettingTemplate', data: Settings.getSetting('breedingEfficiency+')}"></tr>
```
Jaaslet, line 39: I think we rather want this in the hatchery window, as it's something players might often want to turn on and off while looking there

Jaaslet (changes requested):
The description is not very describing :)
I have two comments for you:

src/components/settingsModal.html
jk13pclick: I considered this, and decided against it for two reasons.
1. I didn't want to bother with the HTML
2. The changes also affect the Pokémon List

I could see adding an additional option in the hatchery, or just moving it there entirely. I'd have to see how easy/difficult the html is

jk13pclick: Added the setting, it looks bad because I'm bad at HTML. Feel free to fix or reject

Jaaslet (approved):
I have removed the setting from the settings window, done some code cleanup and gave it a better name.
Thanks for the PR

Issue #3253, opened 2022-10-26
[SUGGESTION] Alternate forms to not trigger base forms in pokedex


osherwott:
**Info**
 - OS: All
 - Browser All
 - Version All

**Description:**
Catching a pokemon with an alternate form triggers an "Unseen" box for the base form, upsetting the harmony of the pokedex

**Steps to Reproduce:**
1. Catch a pokemon's alternate form without catching the base form
2. Go to pokedex
3. Scroll down to the pokemon
4. See the error

**Expected Behavior:**
Alternate forms show up instead of uncaught base forms

**Screenshots:**
![alcremie](https://user-images.githubusercontent.com/114451054/197996632-13e8ba1f-d66a-4b80-9bf1-36d2c4859b18.PNG)

**Additional Information:**
I have tried looking at the code to see what causes this. Could changing line 10 in PokedexHelper.ts to include "pokemon not seen" being along "no caught alternate forms" help?

Resolving this issue would greatly benefit pokemon of all alternate forms, not just Alcremie.

Qwertypop04:
This could maybe warrant consideration as a suggestion, but it is not a bug.

osherwott:
I wrote it as a bug because it also happened with Unown and is presumably not intended. I can change the title to [Suggestion] if you'd like, however.

PixLSteam:
I can see how it could confuse people when their Pokédex shows uncaught Pokémon even though they can move on to the next region, but in the end, the base form is... the base form. I'm not sure if this should be the default or if we should add a setting. Although not everything needs a setting...

osherwott:
I wasn't suggesting a setting, but another check in the code. I don't believe adding another line that specifies to "turn off" the grey box when an alternate form is caught would mess anything up. I'd test it myself and submit a PR but I'm not that experienced with coding...

osherwott:
Something like this, for example. The code isn't exactly correct, but hopefully the idea is. Something to check whether you've caught an alternate form instead of the base one, and replace the unseen or uncaught base with the form.
![pokedexhelper](https://user-images.githubusercontent.com/114451054/198028002-689cc854-401f-4015-8613-069b4bcddf04.PNG)

PR #3311, opened 2022-11-05
Caught Alternate forms replace Base Uncaught forms

osherwott:
This should work. Please give feedback if there are any issues.

Currently only works if a pokemon has less than 99 forms.


Jaaslet (changes requested):
I am not sure what this PR is trying to solve? At least i can't see any changes.
And instead of looping over the IDs, you can just look at the pokemonList what mons there are (filter by floor(id))

osherwott:
Apologies! I have a bad habit of rushing without properly explaining.
This PR attempts to solve #3253. The loop checks if there are any caught alternate forms of that pokemon species id. From what I can see, filtering by floor(id) only works if it is based on the id of the alternate form, which is already checked for in the earlier if statement. I am trying to make it so that, given a base pokemon's id, that same base form can be checked for and returned false. If you have an idea on how to use floor(id) to make this check please feel free to elaborate!

Below are examples of the code in action:
uncaught
rubycream
ruby and vanilla

Jaaslet:
Closes #3253 
(Just linking stuff while i remember. Will look at it tomorrow)

Jaaslet (approved):
I cleaned up the code.
Looks good. Thanks for the PR

PR #3312, opened 2022-11-06
Add Kecleon Temporary Battles

Farboo171:
Added a Kecleon blocking the road after Fortree City, as it occurs in Emerald.

Battle difficulty is a required 100k attack against normal type, less difficult than the gym itself (111k attack against a mix of types), as some of the badge requirements were moved around.

Added two bonus hidden Kecleon battles for no reason other than flavor, slightly more difficult than the blocking temporary battle.


src/scripts/towns/TownList.ts
```diff
@@ -1955,6 +1955,18 @@ const FortreeRanger = new NPC('Pokémon Ranger Catherine', [
...
 
+const Steven1 = new NPC('Steven', [
+    'I have been investigating the behavior of Kecleon, the Color Swap Pokémon.',
+    'There are a few hidden near the city. If you can find them all, I will give you a wonderful prize!'],
+{requirement: new TemporaryBattleRequirement('Kecleon3', 1, GameConstants.AchievementOption.less),
+});
+
+const Steven2 = new NPC('Steven', [
+    'Behold! A picture of a camouflaged Kecleon! That is your reward.',
+], {image: 'assets/images/npcs/other/KecleonDrawing.png',
```
Jaaslet, line 1966: We can't expect players to have read the first NPC, so this might need some more context.
Other than that, i have done some code cleanup. The rest looks good

Farboo171: Changed text to give more context if the player misses Steven1

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3313, opened 2022-11-06
Updated text in Legendary Beasts quest to tell people Kanto is East of Johto

Farboo171:
This gets asked all the time and is not explicitly written out. Adding these two words may save some headaches.


RegisCoaxans:
I do not like the spoonfeed. I think that the quest was made great and the player should somehow figure out how to find Suicune.
What do you thing about not mentioning Kanto but still mentioning regions ?
I am not good enough at English and the best I can come up with now is « East beyond Johto ». My idea remains to tell people that Suicune is not in Johto but can be found East anyway without just telling them to look for it in Kanto.

Farboo171:
> I do not like the spoonfeed. I think that the quest was made great and the player should somehow figure out how to find Suicune. What do you thing about not mentioning Kanto but still mentioning regions ? I am not good enough at English and the best I can come up with now is « East beyond Johto ». My idea remains to tell people that Suicune is not in Johto but can be found East anyway without just telling them to look for it in Kanto.

Another option that has been discussed is adding an NPC somewhere that says Kanto is East of Johto, but not stating it explicitly in the quest text. Would that match with what you're thinking?

RegisCoaxans:
I remember reading an issue about that. While it looks good, I am afraid most people do not read the NPCs unless their quest tells them to do so or they can not go around those. Therefore I think that it would also work, if the NPC is « obvious » enough.

arduousFrivolity:
> Another option that has been discussed is adding an NPC somewhere that says Kanto is East of Johto, but not stating it explicitly in the quest text. Would that match with what you're thinking?

The idea I had was a hiker NPC in Blackthorne who says something along the lines of "I hiked west over the mountains in Kanto for days, and to my surprise I found another region here!" or something a bit more eloquent than that. Still a little spoonfeedy, but at least it fits into the world (there is a mountain between the two regions) and requires at least one to two neurons to fire to figure out (if Johto is west of Kanto, what is east of Johto..?).

Ultima1990:
The _actual_ best idea, which of course came from me no my ego's not too big what are you talking about, is to have an NPC in Goldenrod doing some exposition on the Magnet Rail and saying it goes east to Kanto. Two birds, one stone. You do some worldbuilding about the magnet rail and also the player still has to connect the dots.

Farboo171:
I have caved to pressure and made a lore-friendly NPC instead of changing the quest text

Jaaslet (approved):
Let's hope it helps a bit. Thanks for the PR

PR #3314, opened 2022-11-06
Restore wrongfully deleted NPCs

Farigh:
Those NPCs were removed with the previous PR, it does not seem right.


Jaaslet (approved):
Great catch! Thanks! (i looked through it again, and did not find anything else...)

PR #3315, opened 2022-11-06
Made Temporary Battle Names more consistent

Nydaleclya:
Updated Temporary Battle Names with numbers to be consistent with each other.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3318, opened 2022-11-06
Add held item modal search functionality

CypherX:
Adds search functionality to the held item modal pokemon list (essentially the same as the protein modal search). Also added the ascending/descending tooltip to the sort direction button to keep it consistent with the button elsewhere in the game.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3321, opened 2022-11-07
Type order

Qwertypop04:
Some Pokemon had their types in the wrong order. There may be more but these are all the ones I was able to find


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3330, opened 2022-11-07
Removes the certain Unown A from Ruins of Alph

Qwertypop04:
Unown A was added to be always be in Ruins of Alph if it was uncaught, due to people complaining the Pokedex still had an uncaught Pokemon. This is no longer necessary.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #2737, opened 2022-07-23
Feature/add nickname

lenormandSeb:
Issue : https://github.com/pokeclicker/pokeclicker/issues/2707

PR #3348, opened 2022-11-10
Can now set pokemon party nickname

### LINKED ISSUE #2707: [SUGGESTION] Optional Nicknames for Caught Pokémon
In short, the ability to set custom nicknames on any caught Pokémon for purely cosmetic reasons.

**Current Behavior**
Currently, the game has a system for categorizing Pokémon with a category. While it would be possible to create a custom category to use as a nicknaming device, this would not only clutter up the category list but would also not be displayed anywhere outside of the hatchery list as nothing but a colored dot. Additionally, only one tag category be applied to a Pokémon at any given time, which would mean using a category would overrule the usage of any other categories that would be used for actual filtering functions.

**Improved Behavior**
What the better way of going about this would be is as follows. Being able to simple set a proper nickname for a Pokémon, entirely separate from categories or any other currently existing methods. The nickname would be able to be set from the Pokédex itself, and would display in all instances (Excluding Quests, NPC dialogue, and other things referencing the Pokémon as a species rather than the one caught by the player) the Pokémon's name would normally be shown. There could be a setting that would change the way the nickname is displayed; either entirely in place of the Pokémon's name, or alongside it.

**Reason for change**
The ability to nickname your Pokémon has been a staple feature of the series since day one. It helps the player/user feel more connected to both themselves and their Pokémon. Pokéclicker does not currently have a feature that does this exact thing, and I feel it would be a fun addition as a QoL change.

**Detailed Description**
By selecting the Start Menu, Pokédex, and then any Pokémon you are brought to the detailed information on that Pokémon. At the top of the detailed information list, the Pokémon's name is displayed alongside it's Pokédex number, caught/shiny status, and Pokérus status. With the proposed update, a text box would be displayed underneath the Pokémon's name. This box can be filled in with a nickname that has a maximum of twelve characters (The same max character length as in the games). An empty box would simply not register a nickname to the Pokémon, and a box with characters in it would display those characters as a nickname. Nicknames can be changed, added, or removed at any time and for any captured Pokémon. The actual detailed list of the Pokémon's information would be elongated slightly, and would require either a scrollbar or to be re-centered on the screen vertically to avoid said scrollbar.

**Screenshots:**
- Displayed: Potential layout for proposed feature.
image

- Displayed: Potential display of nickname next to actual Pokémon name
image

Jaaslet:
Nicknames! nincknames! come and set your nicknames!

You can change it in the pokedex window (like with gender and shiny), and it should only show up in the areas that is related to your own pokemon.
I have piggybacked a lot off the translation project.

Pokedex window looks weird, i know. But i did not really know how to make it better, and players won't use it very often

Closes #2707


Aegyo:
Something that translations also didn't do, but I think name sorting should work off the name we are displaying, currently it uses the english name. [link to code](https://github.com/pokeclicker/pokeclicker/blob/2c1ba76c47b0b4a02af475b27697ac3c368a9aaf/src/modules/settings/SortOptions.ts#L39)

Aegyo (approved):
Approved other than the comment above, you can either fix that here or we can handle it separately

Aegyo:
Oh actually looks like someone had done some work on this same thing in #2737 / #2707 - the name input there is above the table so you have more control over the width, seems better than what you have now. Steal that bit?

Aegyo (changes requested):
Requesting above change(s)

Crobat4:
If you put a special character in the nickname, the game won't save.
image

Jaaslet:
> Oh actually looks like someone had done some work on this same thing in #2737 / #2707 - the name input there is above the table so you have more control over the width, seems better than what you have now. Steal that bit?

I think Corgi already moved the name up there like a couple of weeks ago?

Jaaslet:
I have also fixed the encoding problem, the sorting problem and made the display look nicer

src/components/pokemonStatisticsModal.html
```diff
@@ -118,6 +119,15 @@ <h5 class="modal-title text-light" data-bind="if: App.game.party.getPokemon(poke
                       </td>
                     </tr>
                     <!-- /ko -->
+                    <tr>
+                        <td colspan="2" class="text-left align-middle">Party Nickname</td>
```
Aegyo, line 123: I think it looks a little better with this text centered
```suggestion
                        <td colspan="2" class="text-middle align-middle">Party Nickname</td>
```

Aegyo (changes requested):
Small change requested, everything else looks fixed to me :+1:

Aegyo (approved):

PR #3349, opened 2022-11-11
Cleanup Battle Names (Bruno2, Hop2, Circhester2, etc.)

### LINKED ISSUE #934: [BUG] Labeling/Spelling/General wording issues

### LINKED ISSUE #2920: [BUG] Alola Trials are named Gyms
**Info**
 All versions

**Description:**
Right now, all trials in Alola displays gyms in the label of the button to start the fight. It is specifically specified that there are no gyms in Alola (see screenshot)

**Steps to Reproduce:**
Go to Alola
Check any "gyms"

**Expected Behavior:**
The trial challenge button should display Trial and not Gym

**Screenshots:**
image

CorgiOnNeptune:
Should clean up all cases of the display values for gym leader/town names in the achievements and quests. 

Also changed the Alola 'gyms' to the would-be canon trial names.

Closes #934, closes #2920


src/modules/requirements/TemporaryBattleRequirement.ts
```diff
@@ -11,6 +11,6 @@ export default class TemporaryBattleRequirement extends Requirement {
     }
 
     public hint(): string {
-        return `Requires beating ${this.battleName}.`;
+        return `Requires beating ${this.battleName.replace(/\s(?=[\d])\d/g, '')}.`;
```
Jaaslet, line 14: Temp battles have a display name. I know this is in modules and temp battles are not, so it's not as easy. But this seems like a temp solution

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -391,13 +391,21 @@ class AchievementHandler {
...
          * REGIONAL
          */
-        const addGymAchievements = (gyms: string[], category: GameConstants.Region | GameConstants.ExtraAchievementCategories) => {
+        const addGymAchievements = (gyms: string[], category: GameConstants.Region | GameConstants.ExtraAchievementCategories, subregion?: string) => {
             gyms.forEach(gym => {
-                const gymTitle: string = gym.includes('Elite') || gym.includes('Champion') ? gym : `${gym} Gym`;
+                const displayName = GymList[gym]?.displayName;
+
+                // Name of person's title if elite/champion, else the gym's town name
+                const gymTitle: string = gym.includes('Elite') || gym.includes('Champion') || gym.includes('Gym') ? gym.replace(/\d/g, '') : `${gym.replace(/\d/g, '')} Gym`;
```
Jaaslet, line 399: This solution is most likely breaking if the player tries to track for example a Bruno achievement. We will have 2 achievements with the same name, and then we don't know which one to track, when we open a save

src/scripts/quests/questTypes/DefeatGymQuest.ts
```diff
@@ -39,10 +39,15 @@ class DefeatGymQuest extends Quest implements QuestInterface {
...
+        const trial = GymList[this.gymTown]?.displayName?.includes('Trial');
         const desc = [];
-        desc.push(`Defeat ${this.gymTown}`);
-        if (!this.gymTown.includes('Elite') && !this.gymTown.includes('Champion')) {
-            desc.push('gym');
+
+        if (this.gymTown.includes('Elite') || this.gymTown.includes('Champion') || this.gymTown.includes('Gym')) {
+            desc.push(`Defeat ${townName}`);
+        } else {
+            desc.push(`Defeat ${leaderName}'s ${trial ? 'trial' : 'gym'} at ${townName}`);
```
Jaaslet, line 50: Can we also use displayName here? Or add an other displayName for this situation?

src/scripts/gym/Gym.ts
```diff
@@ -70,13 +70,16 @@ class Gym extends TownContent {
         {
             quest = true,
             achievement = true,
-        }: gymFlags = {}
+        }: gymFlags = {},
+        public displayName?: string
```
Jaaslet, line 74: Can we use the displayName for all the `...2` gym leaders? Then we can remove all the `.replace(/\d/g, '')` code we have

Jaaslet (changes requested):
It looks good. But still needs some stuff:

src/modules/requirements/TemporaryBattleRequirement.ts
CorgiOnNeptune: It's definitely a temp solution haha, I have no idea how to reference the displayName from scripts here...

Jaaslet: We will leave this like this for now...

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3350, opened 2022-11-11
Trainers' School NPC

CorgiOnNeptune:
Adds a NPC in Alola's Trainer School, making note of the new floor system in dungeons at that point of the game.

Also added images to the client island NPCs.


Jaaslet (approved):
Looks good. Thanks for the pr

PR #3351, opened 2022-11-11
Removes repeated Pokemon from routes

Qwertypop04:


Jaaslet (approved):
There where a lot :o
Thanks for the pr

PR #3353, opened 2022-11-11
Fix DreamOrbs going negative

Nydaleclya:
At the moment you can open DreamOrbs while having 0 or less.


Jaaslet (approved):
I was just about to look at this. Thanks for the PR!

PR #3356, opened 2022-11-11
Move player away from Distortion World

Nydaleclya:
Move player away from Distortion World during A New World Questline as it should no longer be accsesible till clearing Sendoff Spring once.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3357, opened 2022-11-11
Translate setting display names

Aegyo:
Translations for setting display names added in:
- pokeclicker/pokeclicker-translations#22

I've left all of the strings we currently have where they are, and they are now used as a backup value in case a translation can't be found.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3364, opened 2022-11-12
Add lock icon for currently impossible evos in stone item modal

Aegyo:
Adds a lock icon and tooltip for any stone evolution caught indicators which can't actually be obtained right now.
I think this only affects eevee currently, but could be expanded in the future by removing the isSatisfied check in PokemonHelper.getPokemonsWithEvolution, but that would require some extra handling to prevent selecting a mon / using a stone when no evolutions are possible.

The tooltip will display all unsatisfied requirement hints 
stone-evo-lock


Jaaslet (changes requested):
Looks good. Just one comment:
On the region restricted ones, it just says the index of the region instead of the region itself

Aegyo:
Which pokemon has that restriction?

Jaaslet:
Exeggcute has an Alola only evolution. I think pikachu does too

src/modules/requirements/InRegionRequirement.ts
```diff
@@ -11,8 +11,12 @@ export default class InRegionRequirement extends Requirement {
...
-        return `You must be in one of the following regions: ${
-            this.regions.join(', ')
-        }`;
+        return this.regions.length > 1
+            ? `You must be in one of the following regions: ${
+                this.regions.map((r) => camelCaseToString(Region[r])).join(', ')
+            }`
+            : `You must be in the ${
+                camelCaseToString(Region[this.regions[0]])
+            } region`;
```
Nydaleclya, line 20: Shouldn't this alone work even if you have just one Region?

``` suggestion
        return `You must be in one of the following regions: ${
                this.regions.map((r) => camelCaseToString(Region[r])).join(', ')
            }`
```

Aegyo: Yes, but I think saying "one of the following regions" and only listing one thing looks a bit odd

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3366, opened 2022-11-12
Ultra Necrozma

Qwertypop04:
Makes Ultra Necrozma available. Also adds to the Colress NPC in Mount Lanakila, so he references how to get Ultra Necrozma, as well as the alt forms of Solgaleo and Lunala.


src/modules/pokemons/PokemonList.ts
```diff
@@ -21048,7 +21051,7 @@ export const pokemonList = createPokemonArray(
     },
     {
         'id': 800.03,
-        'name': 'Necrozma (Ultra)',
+        'name': 'Ultra Necrozma',
```
Aegyo, line 21054: Pokemon names are used as translation keys, so this will also need an update in the [translation repo](https://github.com/pokeclicker/pokeclicker-translations)

Qwertypop04: Yep, I noticed that, already working on it.

Aegyo (approved):

PR #3367, opened 2022-11-12
Mega stone items

Jaaslet:
Adds the code for mega stones as items. This code is used if you want to give out mega stones either as dungeon loot or sell them in stores.
Adds Abomasnow to Frozen Caverns as example.
Also does so dungeon loot can't show up when they are not available


src/modules/items/MegaStoneItem.ts
```diff
@@ -0,0 +1,30 @@
+import { Currency } from '../GameConstants';
+import Item from './Item';
+import { ShopOptions } from './types';
+
+export default class MegaStoneItem extends Item {
+    constructor(private pokemonId: number, megaStoneName: string, basePrice: number, currency: Currency = Currency.questPoint, options?: ShopOptions, displayName?: string) {
```
Aegyo, line 6: I think we should pass in a PokemonNameType instead of the id, easier to check things are correct.

Jaaslet: I started with that solution, but finding the image got a bit harder. Then i needed to look up the id somehow...
But if you think it's a better solution, i can try again

Aegyo: What do you need the id for? We have `party.getPokemonByName` and `party.alreadyCaughtPokemonByName`

Aegyo: oh I see now, the image

Aegyo: Do you get errors with `pokemonMap[name].id`?

Jaaslet: Yeah, that works. Thanks! I have submitted that change now

Aegyo (commented):
Tested this out and your code looks good, but dungeon loot has weird handling for giving out items which means it never actually goes through your gain function. I'll make a separate PR for that dungeon loot bug.

Aegyo (changes requested):
Never mind, after looking it that loot gain code more, it needs a bigger rethink than I want to do right now. Can you add a `ItemList[input] instanceof MegaStoneItem` exception before [this line](https://github.com/pokeclicker/pokeclicker/blob/a9e9dbe4fb557ef691c7a3d024407c87185c1cb5/src/scripts/dungeons/DungeonRunner.ts#L175) please, and call `ItemList[input].gain(amount)` inside.
<hr>

Ignore that hardcoded default of xAttack nothing - gets there anyway because the line I linked catches everything (instanceof has a higher operator precedence than ||, so that line is checking... if those classes exist, or the item is an evostone...

It looks like we aren't calling the item's own gain method because a couple of them assume it was bought, and that increments some stats?? But we have a buy method on item! why aren't the purchase stat increments happening there?? 
/rant

Jaaslet:
You now gain the item. I did the boring solution. But yeah, i will maybe solve it in an other PR, so it goes through the gain function

Aegyo (approved):

PR #3369, opened 2022-11-13
Move some small pokemon classes to modules

Aegyo:
Moves into modules:
- DataPokemon
- EnemyPokemonInterface
- RoamingPokemon
- RoamingPokemonList


Jaaslet (approved):
Looks like stuff still works. Thanks for the PR!

PR #3340, opened 2022-11-09
Few alternate forms order and rotating ID function

osherwott:
Switched order of forms in PokemonList.ts and PokemonNameTypes.ts to accurately reflect official and game listings:
* Pikachu (World Cap) between (Partner) and (Flying)
* Vivillon (Icy Snow) as 666.00, (Meadow) as 666.06, and (Poké Ball) as 666.18
* Flabébé, Floette, and Florges (Orange) between (Yellow) and (Blue)
* Minior in *ROGBIV* order: (Red), (Orange), (Yellow), (Green), (Blue), (Indigo), (Violet)
* Alcremie by cream: (Vanilla), (Ruby Cream), (Matcha), (Mint), (Lemon), (Salted), (Ruby Swirl), (Caramel), (Rainbow), and within each cream subdivided by sweet: (Strawberry), (Berry), (Love), (Star), (Clover), (Flower), (Ribbon); ID subdivisions to two decimal places changed accordingly

Re-named files of images (shiny and non-shiny) to match new IDs:
* Pikachu 25.08 - 25.12
* Vivillon 666 - 666.18
* Flabébé 669.02 and 669.03
* Floette 670.02 and 670.03
* Florges 671.02 and 671.03
* Minior 774.01 - 774.07
* Alcremie 869.01 - 869.86 (tenths and hundredths numbers changed accordingly)

Re-added Pumpkaboo (Average) non-shiny sprite that had been omitted for some time after being replaced with Pumpkaboo (Small)

Adjusted code in PokedexHelper.ts to account for Vivillon (Meadow)'s new ID

Resolves #2020

PR #3371, opened 2022-11-13
Pokemon List cleanup and Mega data

Qwertypop04:
Moves all Pokemon IDs to n.0x where some were previously n.x, and moves around the ordering of some alternate forms
Moves all Mega forms to be normal alt forms, adds their Mega Evolutions, and Mega Stones. This does not make any more Megas available, just makes it easier to make more available later.


src/scripts/Update.ts
```diff
@@ -1420,6 +1420,207 @@ class Update implements Saveable {
...
+                [877.1, 877.01],
+                [888.1, 888.01],
+                [889.1, 889.01],
+                [892.1, 892.01],
+                [893.1, 893.01],
+                [898.1, 898.01],
+                [898.2, 898.02],
+            ];
+
+            consistentIDs.forEach(([oldID, newID]) => {
```
Aegyo, line 1591: This isn't going to work for the Alcremie ones, you are overwiting data of a different Alcremie, so it won't be there when you try to move that id later. Osherwott is working on a method of rotating ids around in #3340, I would wait for that before these Alcremie changes.

eg .53 overwrites into .54, and then later you try to move .54 into .51, but .54 now contains stats for the old .53

osherwott:
Would it be possible for you to move Pikachu (World Cap) to after (Partner Cap), since that way it would more accurately reflect the main games' pokedex entries and pokeclicker's acquisition order? Everything else looks stellar! We'll try to get that update function out as soon as possible!

Qwertypop04:
> Would it be possible for you to move Pikachu (World Cap) to after (Partner Cap), since that way it would more accurately reflect the main games' pokedex entries and pokeclicker's acquisition order? Everything else looks stellar! We'll try to get that update function out as soon as possible!

No. The way it is is intentional.

Aegyo:
What's the reasoning?

Qwertypop04:
> What's the reasoning?

It is essentially the "Galar Cap". So it is grouped with the rest of the region caps. Partner Cap is a seperate thing.

src/scripts/Update.ts
```diff
@@ -1513,6 +1518,253 @@ class Update implements Saveable {
...
+                }
+                // Update our statistics
+                saveData.statistics.pokemonEncountered[newID] = saveData.statistics.pokemonEncountered[oldID] || 0;
+                saveData.statistics.pokemonDefeated[newID] = saveData.statistics.pokemonDefeated[oldID] || 0;
+                saveData.statistics.pokemonCaptured[newID] = saveData.statistics.pokemonCaptured[oldID] || 0;
+                saveData.statistics.pokemonHatched[newID] = saveData.statistics.pokemonHatched[oldID] || 0;
+                saveData.statistics.shinyPokemonEncountered[newID] = saveData.statistics.shinyPokemonEncountered[oldID] || 0;
+                saveData.statistics.shinyPokemonDefeated[newID] = saveData.statistics.shinyPokemonDefeated[oldID] || 0;
+                saveData.statistics.shinyPokemonCaptured[newID] = saveData.statistics.shinyPokemonCaptured[oldID] || 0;
+                saveData.statistics.shinyPokemonHatched[newID] = saveData.statistics.shinyPokemonHatched[oldID] || 0;
```
Aegyo, line 1757: Needs the male/female variants of all of these too

Qwertypop04: Ah, of course. I believe that should be all of them?

Jaaslet (approved):
I have fixed the problems i found. There are no more missingNOs, and i tested a couple of mons to see if the stats are transferred correct, and i found no problems.
Thanks for the PR

PR #3373, opened 2022-11-13
Rainbow Rocket credits

CorgiOnNeptune:
Also changed Kukui since Meltan is available, as well as all of Alola's dex.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3374, opened 2022-11-13
Move most PokemonHelper functions to modules

Aegyo:
Leaves the bot helpers and one function that can't be moved yet in scripts, the rest are now in modules.

One of the big blockers preventing other things moving to modules! When I get PokemonFactory across too, a bunch of things should be possible to start moving.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3376, opened 2022-11-13
Fix translations for Notifications settings

Aegyo:
Turns out NotificationSetting doesn't extend Setting, so it needs its own displayName handling


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3378, opened 2022-11-14
Mega Audino

Farboo171:
This PR allows players to obtain Mega Audino after they get all three Therian Formes from Dream Orbs. The Audinate can be obtained pre-Kalos, but can't be used until players can actually buy mega stones.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3383, opened 2022-11-14
Meltan and Melmetal regions

Crobat4:
I noticed Meltan line still have the region set as none despite being available on next update, meaning that they can't appear in the Pokédex. This PR fix that.

### DIFF
--- a/src/modules/pokemons/PokemonList.ts
+++ b/src/modules/pokemons/PokemonList.ts
@@ -21251,7 +21251,6 @@ export const pokemonList = createPokemonArray(
     {
         'id': 808,
         'name': 'Meltan',
-        'nativeRegion': Region.none,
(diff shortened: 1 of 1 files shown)


Jaaslet:
The reason why they are none, is because they are not Alola Pokemon. If we set the region to Alola, they will be required for the Alola dex. Which is not something we want.
I will close this PR

Aegyo:
I think we just want to add a `&& !alreadyCaught` check to [this check](https://github.com/pokeclicker/pokeclicker/blob/develop/src/scripts/pokedex/PokedexHelper.ts#L60-L60) in PokedexHelper.

Should let Meltan/Melmetal show up in the pokedex after being caught, and doesn't make it required for progression. We won't get the silhouette when you have galar mons, but I think we are fine with that.

PR #3388, opened 2022-11-15
Fix Meltan not showing in hatchery

Aegyo:
Should make any None region pokemon appear under the All and None region filter in hatchery (You can get to None by right clicking all selected options)

Also simplifies some of this bitfield code to actually use bitwise operations
- `1 << n` shift 1 to the left by n places, fill in 0s behind it, used to select a region bit
- `n ^ b` toggle all bits in `n` which are set in `b`, used to toggle a region bit on right click
- `~n` inverse, used to get the inverse of 0 (nothing), ie all bits set (All)


Jaaslet (changes requested):
It looks good, but the text shown in the box does not really work that well anymore.
If you press all and then right-click any region (removing them) it just shows none.
And if you leftclick Galar and then rightclick the rest, it will show "Galar and more" even if all is selected

Aegyo:
Will be able to fix this in a few hours, I guess `~0` wasn't such a good idea after all...

Jaaslet (changes requested):
Meltan no longer shows up under none. But it works fine with All, and the text in the select now works again

Aegyo:
95% sure I got everything this time :crossed_fingers:

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3391, opened 2022-11-16
Translate hotkey settings which aren't really a Setting

Aegyo:
These ones aren't actually a Setting, so I missed them in my initial scraping of setting keys.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3393, opened 2022-11-16
Default starter selection as Grass in StarterRequirement

Aegyo:
We have a bunch of rival temp battles where all of the pokemon have a starter requirement, but no handling for none starter.
Having None starter selected means these temp battles have no pokemon and the game crashes because we expect at least 1.

<hr>

[Save file](https://discord.com/channels/450412847017754644/942364514400227358/1042240204322373682) for testing.
Fight Hugh 1 over at route 19 Unova, game crashes immediately on develop.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3394, opened 2022-11-16
Update client notice text

CypherX:
As discussed in discord, update the client notice text on the save selector to remove mention of the dreaded _server_ word. Also reduced the font size slightly to keep it on one line for most resolutions.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3396, opened 2022-11-16
Fix creating new RoamingPokemon in special events

Aegyo:
I may or may not have broken the current event on v0.10.2


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3399, opened 2022-11-16
dream orb notifications shows up after animation

Jaaslet:
The notification showed up when the animation started, so i moved code around.


Aegyo (changes requested):
It is possible to start opening on one orb type, and then switch to another (that you possibly have <= 0 of) and get an item from that one instead.

Should probably leave the incrementObservable and item rolling outside of the timeout.

Aegyo:
or if that is a problem, keeping a note of the orb used and reference that instead if this.selectedOrb()

Jaaslet:
I did the 2nd solution

Aegyo (approved):

PR #3408, opened 2022-11-18
Fixup layering for Magikarp Jump map

RedSparr0w:
This PR just fixes some of the layering issues the Magikarp Jump map had.


Jaaslet (approved):
It works! Thanks for the PR

PR #3412, opened 2022-11-18
Prevent route achievements generating for NullRequirement SubRegions

Aegyo:
Getting route achievements for Magikarp Jump on dev right now.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3413, opened 2022-11-19
Grisious Orb should drop from Altered, not Origin

Qwertypop04:


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3414, opened 2022-11-19
Fix giratina evo and orb

Aegyo:
These slipped onto the wrong form after a merge


Aegyo:
Qwerty was faster than me :cry:

PR #3419, opened 2022-11-19
Fix Legendary Beast quest bug

### LINKED ISSUE #3417: [Bug]: Legendary Beasts quest bugged
### Confirmations

### What happened?

upload(2)
I just updated and download my save at the save time and this happened.
What should I do?

I expected a normal save file without any problem happen.

### Steps to Reproduce

1. Wait for an update
2. Update and download a save file at the same time (12 hours save if nessessary)

Windows 7

Opera

### Relevant log output

_No response_

### Additional Information

_No response_

Aegyo:
Checks if you are on the derek step with an `initial` which is an array of values, and moves you onto the catch step instead.
Should work for people updating from pre 0.10.2 and the people with the NaN bug now.

Fixes #3417

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pokeclicker",
-  "version": "0.10.2",
+  "version": "0.10.3",
   "description": "PokéClicker repository",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3423, opened 2022-11-19
Only show Silver 3 fight after clearing the burned tower quest step

Aegyo:
If you had cleared Burned Tower before the RIvals update, you could beat silver 3 before that quest step and then get stuck at that step, because the fight disappears after clearing.

Changed Silver 3 temp battle to require the previous quest step, and fix softlocked saves by setting their initial to 0 (which should make it auto complete)

stuck save file example:
https://discord.com/channels/450412847017754644/1043508747689005066/1043609852980777030


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3424, opened 2022-11-19
Fix double escaping translated text

Aegyo:
Fixes translations passing in values with apostrophes, like "Let's Go Eevee" logbook entries, and having that apostrophe displayed as the raw html escaped version.

Knockout text bindings I think are doing a second round of escaping, so we don't need to be doing it when translating too.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3425, opened 2022-11-19
Update handling of negative quest items after loading

CypherX:
Updates the quest line section of  `checkAndFix()` to _actually_ fix negative quest progression, specifically underground items. Previously it would "fix" the display but a refresh before completing the quest would undo progress because the updated `initial` wasn't persisted to the save file. It will also handle `MultipleQuestsQuest` quests now, previously it wouldn't update them at all.

Of course this isn't a full solution but a temporary fix until a proper solution can be implemented.

### DIFF
--- a/src/scripts/Game.ts
+++ b/src/scripts/Game.ts
@@ -232,8 +232,17 @@ class Game {
         // Check that none of our quest are less than their initial value
         App.game.quests.questLines().filter(q => q.state() == 1).forEach(questLine => {
             const quest = questLine.curQuestObject();
-            if (quest.initial() > quest.focus()) {
-                quest.initial(quest.focus());
+            if (quest instanceof MultipleQuestsQuest) {
+                quest.quests.forEach((q) => {
+                    if (q.initial() > q.focus()) {
+                        q.initial(q.focus());
+                    }
+                });
+            } else {
+                if (quest.initial() > quest.focus()) {
+                    quest.initial(quest.focus());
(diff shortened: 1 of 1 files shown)


src/scripts/Game.ts
```diff
@@ -232,8 +232,17 @@ class Game {
...
             const quest = questLine.curQuestObject();
-            if (quest.initial() > quest.focus()) {
-                quest.initial(quest.focus());
+            if (quest instanceof MultipleQuestsQuest) {
+                quest.quests.forEach((q) => {
+                    if (q.initial() > q.focus()) {
+                        q.initial(q.focus());
+                    }
+                });
+            } else {
```
Aegyo, line 241: I think you can add a curQuestInitial fix for the multiquests with this:
```suggestion
            if (quest instanceof MultipleQuestsQuest) {
                quest.quests.forEach((q) => {
                    if (q.initial() > q.focus()) {
                        q.initial(q.focus());
                    }
                });
                questLine.curQuestInitial(quest.quests.map((q) => q.initial()));
            } else {
```

CypherX: I didn't add that originally because `curQuestInitial()` is only ever used in the QuestLine `toJSON()` function for quests that are not `MultipleQuestsQuest`. I can add it if you feel it should be there though.

Aegyo: No you are right, should be fine without it :+1:

Aegyo (approved):

PR #3427, opened 2022-11-19
Fix shard quest progress not updating if you had never had that shard

Aegyo:
If you never had a shard before and start a quest for it, the progress function crashes and doesn't update.


RedSparr0w (approved):

PR #3428, opened 2022-11-20
Fix Silvally (Psychic) images

Aegyo:
id 773.10, but images need to be 773.1


RedSparr0w (approved):
Nice catch, thanks for the PR! 👍

PR #3432, opened 2022-11-20
Fix quest check and fix bug

Aegyo:
If you load the game, and the code has changed such that a questline you were on is now completed (either through bugfixes or potentially also through quest requirement changes), the check and fix function crashes because the questline has not had time to update its state, and the dummy quest object we return doesn't have initial or focus.

Now skipping over questlines if the current step is greater than however many quests there are supposed to be.

Save file with this issue:
https://discord.com/channels/450412847017754644/1043878685532508310/1043879271921369169


RedSparr0w (approved):

PR #3450, opened 2022-11-22
Magikarp jump Gym leaders now have the correct fishies

Jaaslet:
Changed the magikarps to the correct ones for the gymleaders, and made so they reward their own fish


src/scripts/gym/GymList.ts
```diff
@@ -1589,21 +1589,41 @@ GymList['Friend League'] = new Gym(
...
-    [new RouteKillRequirement(10, GameConstants.Region.alola, 32)]
+    [new RouteKillRequirement(10, GameConstants.Region.alola, 32)],
+    () => {
+        if (!App.game.party.alreadyCaughtPokemonByName('Magikarp Skelly')) {
+            Notifier.notify({
+                message: 'You were awarded a Magikarp Skelly!',
+                type: NotificationConstants.NotificationOption.success,
+                setting: NotificationConstants.NotificationSetting.General.new_catch,
+            });
+            App.game.party.gainPokemonById(pokemonMap['Magikarp Skelly'].id);
```
Aegyo, line 1604: Probably want to pass `suppressNotification = true` for these, getting two notifications which basically say the same thing right now.
Also this doesn't look like it will roll for shiny, do we want it to? I would probably have gainPokemonById handle the generation like it does for gender, if we explicitly pass `shiny = null` or something. Or just leave it always be not shiny.

Aegyo (changes requested):

src/scripts/gym/GymList.ts
Jaaslet: Fixed

Aegyo (approved):

PR #3451, opened 2022-11-22
Magikarp jump temp battles

Jaaslet:
- Added magikarp jump temp battles
- Gave them the correct fishies
- The give the fishies, if no gym has it and no temp battle before it
- Added the correct requirements for them
- Added them to towns

Still missing balance and defeat text


src/scripts/Update.ts
```diff
@@ -1559,6 +1559,29 @@ class Update implements Saveable {
                 }
             }
         },
+        '0.10.5': ({ playerData, saveData }) => {
+            saveData.statistics.temporaryBattleDefeated = Update.moveIndex(saveData.statistics.temporaryBattleDefeated, 159);
```
Aegyo, line 1563: We do these massive moveIndex walls so often, maybe we want a `Update.moveIndexRange(159, 179)`

RedSparr0w: Could only really be used for new indexes, so probably something like `insertNewIndexRange`

RedSparr0w (approved):
Looks good to me 👍 

Just a couple questions, don't we have badges implemented for the sub region?
Shouldn't these be gyms?
Or are the badges coming from somewhere else?

Jaaslet:
> Just a couple questions, don't we have badges implemented for the sub region? Shouldn't these be gyms? Or are the badges coming from somewhere else?

I have added 10 gyms in an other pr. Made these as temp battles because 30+ gyms felt like too many

PR #3452, opened 2022-11-22
Added magikarp jump achievement category

Jaaslet:
- Magikarp jump subregion gets it's own achievement category
- It's too small for 50%, so it gets 25%. To prevent a weird total, we will give Hisui 75%
- Routes currently does not show up, because of the null requirement
- There are currently no achievement for catching them all. I don't know if we want one, and how many we want. Will figure that out, after i have added all the forms


Aegyo (commented):
My only issue with this is that the achievements are going to be shown before they are possible to complete, unless the region is done before next release.

Can we get a way for us to force an achievement category to be hidden with this?

RedSparr0w:
We could create a separate branch for the Magikarp jump stuff, and have PRs made to that branch, so we know everything in the branch has already been tested/approved when we merge it?

Aegyo:
We could, but it would need develop merging into it frequently and might get messy with moveIndex numbers having to be updated if things changed in develop. Unsure what other potential problems there might be.

If we can make sure the wip region stuff just doesn't affect the game until we enable it, I think things will be easier.

Jaaslet:
Agree with Aegyo.
I have also hidden the achievements for now.
I tried to use a questlinestartedrequirement for it, but the code is loaded before quests are added... And i don't really want to fix that in this PR... But in that attempt, i ended up making the introduction quest (without Mayor Karps real text)

Aegyo (approved):
Not sure if you intended to make the subregion available with this PR, so I'll leave it for you to merge

Jaaslet:
The idea is that you need to complete the quest to gain access to it. And the quest is locked behind a `DevelopmentRequirement`, so it's not available in the live version. At least that is what i attempted

PR #3457, opened 2022-11-23
Update function to fix Darkest Day temp battles

CorgiOnNeptune:
Implements a checkAndFix for The Darkest Day questline. Players who didn't revert their save after the accidental Galar release get blocked by the messed up temp battle index. This will check and fix any instance of it as well as prevent the initial in that instance from being -1.

[A bugged file on quest step 3](https://cdn.discordapp.com/attachments/1033710705788072006/1043847975803306064/v0.10.3_PokeClickerSave_2022-11-20_11_16_24.txt)


RedSparr0w:
This PR seems to have changes from a previous PR remaining.

Does this need to be in the `checkAndFix` function as those are run every time the game loads, I think it should be resolvable in an update function?

CorgiOnNeptune:
> 

Fixed the other PR issue.. Since it's already merged haha.

I could move most of it to an update. Only concern is somebody can do Hop 7 and Hop 6 out of order if they already have the Circhester badges. Then they'd be stuck again if they weren't paying attention and did those in the wrong order.

Aegyo:
Add a hop6 requirement to hop7 showing?

CorgiOnNeptune:
> 

Oh ya I guess that'd work lol

CorgiOnNeptune:
Ok, moved to update function and added Hop 6 as a req to see Hop 7 so they can't be completed in the wrong order. 
And the initial quest values were fixed with Jess' PR, so I didn't need that anymore.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3458, opened 2022-11-23
Kick player off Client Island if not on client

CorgiOnNeptune:
Kicks player back to One Island if they have abandoned the client, but last saved on Client Island.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3459, opened 2022-11-23
Magikarp Jump SVG and Map update

Qwertypop04:
Removes the unused leagues from the map, and makes the SVG more accurate to the map


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3461, opened 2022-11-23
Fix quest checkAndFix, again

Aegyo:
I broke it, almost immediately after multiquest fixing was added...


RedSparr0w (approved):

Do we need anything in the update functions?
Or will the quest affected by this just remain in the previous state?

Aegyo:
Multi quests that needed fixing would just not have been fixed by this function when they should, no additional breakages or anything we need to handle in an update function.

It's the negative shard thing we don't have a proper fix for yet that is mainly needing this, lots of people using their plates/shards for the new giratina questline.

PR #3462, opened 2022-11-23
Use correct LogContentKey for some shiny logs

Aegyo:
Some non-hatch shiny logs were using the hatched text.


RedSparr0w (approved):
Nice catch, thanks for the PR! 👍

PR #3467, opened 2022-11-24
Fix nicknames and translations for held item modal

Jaaslet:
The modal still used name instead of display name


Aegyo:
Search still uses name here too, otherwise looks good :+1:

Jaaslet:
fixed

Aegyo (approved):

PR #3474, opened 2022-11-25
Fix up the Pokémon ID shuffle

RedSparr0w:
The hatchery wasn't being updated, so any mons that the old ID doesn't exist now, were becoming `MissingNo.`.
Also removed some code that wasn't needed.
All the rotating ID Pokémon should also update in the hatchery now, instead of becoming a different mon.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3481, opened 2022-11-26
Multiple Safari Zone Framework

jk13pclick:
Rough framework for Multiple Zone support. More or less copies what we do for the Deals. Lots of room for improvement with this PR.


src/scripts/safari/Safari.ts
```diff
@@ -374,8 +373,8 @@ class Safari {
         return GameConstants.SAFARI_BASE_POKEBALL_COUNT;
     }
 
-    static completed(shiny = false) {
-        return SafariPokemon.list.reduce((all,poke) => {
+    static completed(shiny = false, region) {
```
Jaaslet, line 376: `region` is not used here

Jaaslet (changes requested):
The game crashes when it's in a region without a safari zone. The code expects `SafariPokemonList.list[region]` to always exist

Jaaslet (changes requested):
So, currently the "leave" button allows you to continue your safari run later. And with this, you can start a safari run, make an overworld mon spawn, leave an encounter it in an other region. Which is weird. But I am not sure what we should do about it... It would make sense that leave really leaves, but it also seems mean to take it away from the player now.
Also the sinnoh safari should be removed before merge, as it's currently only there for testing

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3484, opened 2022-11-27
Added battle against Red

Jaaslet:
You can now fight Red at mt. Silver. He gives you a Light ball each time you beat him.
Also buffed some of the pokemon specific held items a bit, because they where a bit behind the legendary specific held items


Aegyo (approved):
You've got a conflict here, looks good to merge otherwise

PR #3488, opened 2022-11-28
Chansey shop image

DraKay:
Change the shop image of a common Chansey for a charitative chansey image to make it more representative of the buyable item


RedSparr0w (approved):
Looks great, thanks for the PR! 👍

PR #3493, opened 2022-11-29
Keeping attack achievements unlocked after reload

RegisCoaxans:
I have provided a generic solution to ephemeral achievements. Only used on attack achievements for now.

I had to change `unlocked` from boolean to observable\<boolean\> or the computed `isCompleted` kept overwriting it anyway. I changed that attribute calls anywhere I could find them but tell me if I missed some.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3495, opened 2022-11-30
[FEATURE?] Dummy evolutions

### LINKED ISSUE #3494: [SUGGESTION]: Dummy Evolutions
### Current Behavior

Currently, dummy evolutions like Phione -> Manaphy or Darumaka -> Darmanitan (Zen) are shown by the bot and might confuse users.

### Improved Behavior

Dummy evolutions should be implemented as a separate evolution type, which is ignored by the bot.

### Reason for change

It confuses users.

### Detailed Description

I'll miss Phione's evolution level 😢 
But I feel like this is the proper way to implement them.

PixLSteam:
This PR implements dummy evolutions. These evolutions have a new evolution trigger: `EvoTrigger.NONE`.
They are skipped by `PokemonHelper.getPokemonPrevolution` and should therefore not show up in the bot's data.
However, this would also remove them from the bot quiz. If this is not wanted, I can look for an alternative solution.

This closes #3494.

### DIFF
--- a/src/modules/pokemons/PokemonList.ts
+++ b/src/modules/pokemons/PokemonList.ts
@@ -23,7 +23,7 @@ import {
     Region, StoneType, Genders, MaxIDPerRegion,
 } from '../GameConstants';
 import BagItem from '../interfaces/BagItem';
-import { EvoData, LevelEvolution, StoneEvolution } from './evolutions/Base';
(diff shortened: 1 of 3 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3496, opened 2022-11-30
[FEATURE] Keep EXP/level in real evolution challenge

### LINKED ISSUE #3439: [SUGGESTION]: Keep the evolved Pokémon level during Real Evolution challenge
### Current Behavior

With Real Evolution challenge on, whenever a Pokémon can evolve (for example : Bulbasaur at lvl 16), it will change to its evolution but his level will be reset to 1.

### Improved Behavior

Not much to change, but just leave the Pokémon to its actual level (Ivysaur to a level 16 instead of 1 in our exemple)

### Reason for change

Feel closer to the actual behaviour of real evolution.

### Detailed Description

Keep the level at which it is when evolving, even if it is a few level above the requirement. In later regions, it is fairly easy to gain 10 levels per encounter.

PixLSteam:
Transfers the EXP to the evolved Pokémon. According to [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Experience#Relation_to_level), experience types are consistent within evolutionary families.
This closes #3439.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3497, opened 2022-11-30
Reduce CPU/GPU usage when window not visible

RedSparr0w:
This PR hides everything when the page isn't visible, which reduces GPU and CPU usage significantly.


Jaaslet (approved):
It reduces the CPU usage by a lot, and takes like 1 sec to load again, on my old laptop in powersaving mode. Yeah, i think this is a really good addition.
Thanks for the PR!

PR #3499, opened 2022-12-01
Fix Lure Ball causing crash when used in a dungeon

CypherX:
Using a Lure Ball in a dungeon causes a crash. The catch bonus logic looks at the pokemon on the players current route to determine if the bonus should be applied, in dungeons the players route is zero resulting in `Routes.getRoute(player.region,player.route())` returning `undefined`.

I separated the logic for routes and dungeons and because dungeons have no concept of "land" or "water" pokemon like routes, the bonus will be applied if the enemy pokemon is a water type. I considered having no bonus while in dungeons but that didn't feel... right?

Discord thread: https://discord.com/channels/450412847017754644/1047702638831079455


src/scripts/pokeballs/Pokeballs.ts
```diff
@@ -74,12 +74,20 @@ class Pokeballs implements Feature {
             }, 1250, 'Increased catch rate on water routes', new RouteKillRequirement(10, GameConstants.Region.hoenn, 101)),
 
             new Pokeball(GameConstants.Pokeball.Lureball, () => {
-                const numLandPokemon = Routes.getRoute(player.region,player.route()).pokemon.land.length > 0;
-                const isWaterPokemon = Routes.getRoute(player.region,player.route()).pokemon.water.includes(Battle.enemyPokemon().name);
+                if (App.game.gameState == GameConstants.GameState.fighting && player.route()) {
+                    const numLandPokemon = Routes.getRoute(player.region,player.route()).pokemon.land.length > 0;
```
Jaaslet, line 78: `numLandPokemon` is not totally correct. Something like `LandPokemonExists` would be more correct. (I know it was there before, you might as well fix the code while you are working with it)

src/scripts/pokeballs/Pokeballs.ts
```diff
@@ -74,12 +74,20 @@ class Pokeballs implements Feature {
...
-                if (numLandPokemon == true && isWaterPokemon == true) {
-                    return 15;
+                    // If route has Land Pokémon and the current pokémon is a Water Pokémon
+                    if (numLandPokemon == true && isWaterPokemon == true) {
+                        return 15;
+                    }
+                }
+                if (App.game.gameState == GameConstants.GameState.dungeon) {
+                    // If player in a dungeon and the enemy pokémon is a water type
+                    if (Battle.enemyPokemon().type1 == PokemonType.Water || Battle.enemyPokemon().type2 == PokemonType.Water) {
```
Jaaslet, line 88: This dungeon code does not fit the description of the lure ball. It claims it helps on fished mons, but this is just water mons. Not sure if i am caught up in a weird detail, or we either need to remove this or change the description a bit

src/scripts/pokeballs/Pokeballs.ts
```diff
@@ -74,12 +74,20 @@ class Pokeballs implements Feature {
...
-                const isWaterPokemon = Routes.getRoute(player.region,player.route()).pokemon.water.includes(Battle.enemyPokemon().name);
+                if (App.game.gameState == GameConstants.GameState.fighting && player.route()) {
+                    const numLandPokemon = Routes.getRoute(player.region,player.route()).pokemon.land.length > 0;
+                    const isWaterPokemon = Routes.getRoute(player.region,player.route()).pokemon.water.includes(Battle.enemyPokemon().name);
 
-                // If route has Land Pokémon and the current pokémon is a Water Pokémon
-                if (numLandPokemon == true && isWaterPokemon == true) {
-                    return 15;
+                    // If route has Land Pokémon and the current pokémon is a Water Pokémon
+                    if (numLandPokemon == true && isWaterPokemon == true) {
```
Jaaslet, line 82: `== true` is unnecessary

Jaaslet (changes requested):
It fixes the crash! but i have some comments:

src/scripts/pokeballs/Pokeballs.ts
CypherX: changed to `hasLandPokemon`

src/scripts/pokeballs/Pokeballs.ts
CypherX: Decided to scrap the dungeon portion after giving it some more thought. It doesn't fit canon and was a cheap way around dungeons not having "fishing" encounters like routes.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3503, opened 2022-12-02
Change egg default setting to egg step instead of percentage 

DraKay:
Right now egg is percentage by default. This setting give very few information and is confusing for new players (seen 2 confusions with it today) so if accepted this change set egg step as default

Was it what you mean pixl? XD


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3504, opened 2022-12-02
Removed max for held items

Jaaslet:
You can now use more than one of each held items and use them on more than 6 pokemon.
A step towards making held items worth using!


RedSparr0w (approved):
👍

PR #3507, opened 2022-12-03
Fix egg slots not filling from queue

### LINKED ISSUE #2540: [BUG] hatchery queue not filling in empty slots still
**Info**
 - OS: [e.g. Windows 10]
 - Browser chrome
 - Version 0.9.7

**Description:**
the hatchery queue not filling in occasionally is still an issue. the change logs state it was fixed in 0.9.6 but I had it happen 3 times within an hour. once with an egg helper then it happened twice in the bottom two slots without an egg helper. i have a full queue but it wouldnt fill in.

CypherX:
Steps to reproduce here: https://github.com/pokeclicker/pokeclicker/issues/2540#issuecomment-1312334946

Egg slots would sometimes get 'stuck' and not pull the next egg from the queue, this would often require emptying and re-populating the queue to fix.

Closes #2540


src/scripts/breeding/Breeding.ts
```diff
@@ -306,19 +311,25 @@ class Breeding implements Feature {
...
+        const nextEgg = this.createEgg(this._queueList.shift());
+        this.gainEgg(nextEgg);
+        if (!this._queueList().length) {
+            Notifier.notify({
+                message: 'Hatchery queue is empty.',
+                type: NotificationConstants.NotificationOption.success,
+                timeout: 1e4,
+                sound: NotificationConstants.NotificationSound.Hatchery.empty_queue,
+                setting: NotificationConstants.NotificationSetting.Hatchery.empty_queue,
+            });
```
RedSparr0w, line 332: Shouldn't this be moved to the check above with the `return;`?

CypherX: That notification is displayed when the queue becomes empty following an egg being moved from the queue to an egg slot. If it were moved to the check above it would display every time there was an open egg slot with no queue and eggs gained progress or hatched.

CypherX:
Changing this to a draft for now - there's something else I want to look into.

CypherX:
Finally had time to get back to this. I wasn't happy with my initial changes and noticed some oddities while doing additional testing.

Firstly, sometimes if a helper slot is empty due to filters it was possible for an egg to move from a non-helper slot to a helper slot when `moveEggs()` was called. I noticed this happening more frequently after my initial changes so I changed the `moveEggs` function to ignore helper slots (if this was intended let me know) so it doesn't happen at all anymore.

Secondly, I adjusted how empty egg slots (whether they were empty to begin with or as a result of an egg hatching) are filled from the queue in `progressEggs`. This guarantees any empty non-helper slots will be filled which fixes the original issue of slots getting "stuck" and not filling from queue.

Jaaslet (approved):
I think it looks good? It fixes the case described in the issue, and it looks like the hatchery works as before.
Thanks for the PR

PR #3508, opened 2022-12-03
Fixup loading older save files

RedSparr0w:
This PR fixes up loading older save files


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3511, opened 2022-12-03
Allow mobile users to remove Vitamins

RedSparr0w:
Adds a button to toggle adding/removing vitamins.
Mostly for Mobile users, as this doesn't toggle what right click does.
image
image


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3512, opened 2022-12-03
Allow search on nickname or translated name (Hatchery & Vitamin)

RedSparr0w:
Allows searching on nickname and translated name for Hatchery and Vitamin menus


Jaaslet (approved):
We should consider doing the same for the hatchery. I think it's only looking for display name

PR #3513, opened 2022-12-03
Magikarp jump: Added route patterns and roamers

Jaaslet:
This adds pattern to routes and some roamers. Here is my logic:
- I took the list of pattern, and removed the mons already obtained from a temp battle or gym
- Then i took all patterns you unlock from a rod and split them between all routes leading up to the next rod unlock
- Then if there where too many, i either added them as roamers or left them out
- The last 4 unlocked after the last league are two roamers and 2 for later
- There should be 3 pattern not obtainable yet. I will add them as a mix of shopmons and questline rewards in a later PR


RedSparr0w (approved):

PR #3514, opened 2022-12-03
Small magikarp jump fixes

Jaaslet:
- Missing roamer group, so it crashed when you selected a route
- Two temp battles has the same name
- Elite 4 league did not show up


RedSparr0w (commented):

RedSparr0w (approved):

PR #3515, opened 2022-12-03
Add a development requirement

RedSparr0w:
The requirement will only be true if running in a dev environment or if set to true in the config.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3516, opened 2022-12-03
Dungeon Megas

Farboo171:
Adding three dungeon megas in Kalos:

Garchomp: Victory Road
Heracross: Santalune Forest
Kangaskahn: Glittering Cave

These are the canon locations for the megas stones in the base game. method of adding was the same as Mega Abomasnow.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3521, opened 2022-12-04
Update Vitamin menu

RedSparr0w:
Allows you to swap between different vitamins while the menu is open:
image


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3523, opened 2022-12-04
Added basic requirements to magikarp jump subregion

Jaaslet:
Added the natural progression. You still can't start the area until #3452 is merged.
Will add damage gates later as requirements, so that is why it's called "basic requirements"


Aegyo (approved):
Are we leaving the NullRequirement on Friend League Bridge for now?

Jaaslet:
The null requirement has been removed in #3452

PR #3524, opened 2022-12-04
Update function to auto-start Suicune quest for people in the middle of the Legendary Beasts quest

Farboo171:
This update function will ensure that if people have gone too far in the Legendary Beasts to trigger the auto-start of the Suicune quest without having started said quest, the update will start it for them.

Tested on every step of hte Legendary Beasts questline to ensure it starts at the right time, and tested on saves that have completed both to make sure it doesn't auto start the quest for a second time.

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -1874,6 +1874,22 @@ class Update implements Saveable {
                     saveData.statistics.temporaryBattleDefeated[GameConstants.getTemporaryBattlesIndex('The Darkest Day')] = 0;
                 }
             }
+            // Suicune Quest autostart for players too far in Legendary Beasts quest
+            const johtoBeastsQuestLine = saveData.quests.questLines.find((q) => q.name == 'The Legendary Beasts');
+            const johtoSuicuneQuestLine = saveData.quests.questLines.find((q) => q.name == 'Eusine\'s Chase');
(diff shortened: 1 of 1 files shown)


src/scripts/Update.ts
```diff
@@ -1874,6 +1874,22 @@ class Update implements Saveable {
                     saveData.statistics.temporaryBattleDefeated[GameConstants.getTemporaryBattlesIndex('The Darkest Day')] = 0;
                 }
             }
+            // Suicune Quest autostart for players too far in Legendary Beasts quest
+            const johtoBeastsQuestLine = saveData.quests.questLines.find((q) => q.name == 'The Legendary Beasts');
+            const johtoSuicuneQuestLine = saveData.quests.questLines.find((q) => q.name == 'Eusine\'s Chase');
+            if (johtoBeastsQuestLine?.state == 4 || (johtoBeastsQuestLine?.state == 1 && johtoBeastsQuestLine?.quest >= 4)) {
```
Jaaslet, line 1880: State will never be 4? It will always be 0, 1 or 2

Farboo171: Changed the function back to state 2, tested on quest steps again to confirm behavior is still working as desired.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3525, opened 2022-12-04
Fix shiny Galarian Slowking sprite filename

Crobat4:
ID changed from `199.1` to `199.01`, but the filename was still `199.1`.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3531, opened 2022-12-06
Make alcremy npc text less confusing 

DraKay:
Welcome to a new episode of silly DraKay PRs, today we have changed 3 words!

After a talk in discord it seem that alcremy extra spins aren't very obvious in the text of the npc and Lucho and me ended with this slightly different text that should be less confusing


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3532, opened 2022-12-06
Only magikarps attack in magikarp jump

Jaaslet:
When you are in the magikarp jump subregion, your magikarps are the only onces that count for base attack and click attack


Aegyo:
Do we want to override the region debuff too?

Jaaslet:
I like the idea that you will get like 13% damage boost by going to Galar before you do this region. But it might make more thematic sense to not have it here? Since it's the home of Magikarps?
But no matter what, i think we should add it in an other PR (so i will add it to the task list)

Aegyo (approved):

PR #3537, opened 2022-12-06
Magikarp jump hp balance

Jaaslet:
Added hp to routes, temp battles and gym battles based on this excel sheet i made, where "damage gate" is now much attack i want the players to have for all their magikarps combined
image


Aegyo:
The temp battles before each gym are harder than the actual gyms, I'm guessing the requirements weren't there when you added these numbers?

Jaaslet:
You have double the amount of time for temp battles, so it should be fine, right?

Aegyo:
That makes sense, didn't notice the time difference

Aegyo (approved):

PR #3544, opened 2022-12-10
Mimic EV Boost + Mimics in Encounter List

jk13pclick:
Uses Dungeon boss modifier for Mimics. Also adds caught mimics to the Encounter List for dungeons.


Jaaslet (approved):
Looks good. I did some changes:
- It now shows a chest on the list (like the boss icon)
- includeMimics->getCaughtMimics. So it's easier to see what it does
- You just created a new requirement, without seeing if it was completed. So i changed it to `App.game.party.alreadyCaughtPokemonByName()`, so now the filtering works.

Thanks for the PR

PR #3547, opened 2022-12-10
Mega Lopunny

Farboo171:
Adds Mega Lopunny as a Battle Frontier reward.

In ORAS, Lopunnite is just found in Mauville City. Since Mega Lopunny is Fighting type, and no other megas are in the BF yet, this seemed a good place to add it.


Jaaslet:
The bf milestone needs an obtained pokemon requirement for Lopunny. Else it will crash the game (i maybe need to look into if the rest of the mega stones have the same problem).
And it feels weird that a insect from the same region is harder to get than a mega stone

Farboo171:
Added catch requirement

I was unsure of if Mega Lopunny should be before or after the Vivillon, and landed on before. Especially after the Kalos rebalance, BF 666 is very close to the end of the region.

Qwertypop04:
Maybe go with 700? It's not too much later than 666, so I doubt it would be especially difficult to get in Kalos. Besides, imo it's fine if it's not particularly easy to get in Kalos, as long as it is possible

Farboo171:
Changed to level 690, because ice stones are on 700.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3549, opened 2022-12-10
Mega stones no longer show up after obtained in dungeons

Jaaslet:
Before the dungeon loot code only checked if items where available, and not if they where sold out. That was fine for mega stones, until the mega stone item code got fixed in #3516


RedSparr0w (approved):

Issue #3482, opened 2022-11-27
[SUGGESTION]: Add option to edit default save download name


BaileyP2SR:
### Current Behavior

Currently when you download a save it automatically formats it "[version#] PokeClickerSave_YYYY-MM-DD HH_MM_SS.txt" (I think?).

### Improved Behavior

Adding a box or setting below the download save button which allows you to edit the default name when you go to download. You could set the syntax for date and time still so users can move that around into whatever makes sense for them.

### Reason for change

The biggest reason is for multiple files, making it even easier to automatically keep track of which file is which. For example you can default your main file to be PokeClickerMain and an All Challenges files to have PokeClickerAC. Obviously you can do this now, and I'm sure I'm not the only one who does, but it is a nice QOL to make it super easy to manage your own saves.

### Detailed Description

Just a text field below the download save button allowing you to set this up, maybe with info similar to the discord integration showing what different syntax you can use.

RegisCoaxans:
There are three locations where a download can occur :
- [Save.download](https://github.com/pokeclicker/pokeclicker/blob/develop/src/scripts/Save.ts#L52) on the save modal
- [update.automaticallyDownloadBackup](https://github.com/pokeclicker/pokeclicker/blob/develop/src/scripts/Update.ts#L1951) before updating a save
- [SaveSelector.download](https://github.com/pokeclicker/pokeclicker/blob/develop/src/modules/SaveSelector.ts#L86) on the main menu

There is no naming convention between them.
The first one contains the date, the two first ones use no reference to the actual save but the SaveSelector includes the save key into the file name.

Should the three of them use the same format now ? Or does this suggestion only have to do with downloads from the save modal ?

RegisCoaxans:
I was planning on doing it but go ahead.
There is no deadline. Just do not claim the issue then quit without a word, lol.

Funk3:
If you are already planning on doing it please go ahead! I would be committing very little time to it, and would rather you do it if you have an idea of how to!

PR #3550, opened 2022-12-10
Added the ability to predefine the name of the save download

RegisCoaxans:
Closes [#3482](https://github.com/pokeclicker/pokeclicker/issues/3482).

- The filename is used in each of the three contexts listed in that issue so it is all standardised.
- The default name matches the one previously used in `Save.ts`.
- The function in `Update.ts` still adds ` Backup` at the end for convenience.
- In case of an empty name, the default one is used instead.


Jaaslet (changes requested):
Looks good. Just one change request:
I think the name should be in settings. Not a lot of players want to change it, and almost none wants to change it often. So it should be hidden in the settings modal

Jaaslet (approved):
Looks good. Thanks for the PR

Issue #2987, opened 2022-09-13
[SUGGESTION] Pokédex, Evolution Items and Profile layout changes


Crobat4:
Just suggesting some layout changes, I can provide PR later if you want them.
**Pokédex**:
![pokedex-1](https://user-images.githubusercontent.com/104547700/189969134-6c7993af-16e3-4fd9-b790-8ecf2401a826.PNG)
![pokedex-2](https://user-images.githubusercontent.com/104547700/189969140-9994483c-ce7c-4524-a1cd-dfb34117baf5.PNG)
**Evolution Items**:
![evo-items-1](https://user-images.githubusercontent.com/104547700/189970099-6398961a-9f63-44a3-976f-433398e684ab.PNG)
**Profile**:
![profile-1](https://user-images.githubusercontent.com/104547700/189970181-b559dc74-c555-4969-ad13-cc59fec6890f.PNG)
![profile-2](https://user-images.githubusercontent.com/104547700/189970195-cd80068b-3e45-4dd2-a8f0-051222460b9f.PNG)

Jaaslet:
I will talk about the images one at a time:
- I don't like that the filters in on the side. It might make the window too wide for smaller screens. And our placement of filters will not be consistent throughout the game.
- I like that message. I would not mind if that was added to the game. I am not sure the last sentence is correct english. Hopefully someone better at english than me, will confirm or deny.
- I am not sure about the new evolution window. It will make it easier for people with smaller screens, but i think we can find other solutions for that. But i don't like that it's not easy to see how many of each evolution item you have. I like being able to see all the items at once. But that might just be me.
- The profile window will be fine to change, i think. I think both this and the new are fine. The shiny maybe looks a bit weird alone. It's the smallest of them all... Maybe move it to the same row as the text color?
- And i like the shiny selection. We can add that one.

CypherX:
"Missing a Pokémon but don't see it here? Be sure to completely explore the region!"

RedSparr0w:
>
> - The profile window will be fine to change, i think. I think both this and the new are fine. The shiny maybe looks a bit weird alone. It's the smallest of them all... Maybe move it to the same row as the text color?
> - And i like the shiny selection. We can add that one.

I think maybe have the `Name` and `Text color` on the same line, and then the `Pokémon` and `Shiny` on the same line, as those are more closely related settings.

Also just noticed this, but we should change the text from `Pokemon` → `Pokémon`

[...]

PR #3551, opened 2022-12-11
Pokédex and Hatchery layout

Crobat4:
Here are the Pokédex and Hatchery layout also requested in https://github.com/pokeclicker/pokeclicker/issues/2987#issuecomment-1329394319.
* Removed the header and footer to increase the amount of visible Pokémon on a single screen.
* Since the sprites of hatchery helpers appear in every tab, I added the `Fire All` button in every tab as well.
* The tooltip from the `?` button was moved to the help tab.
* I also fixed some text color issue in the evolution items modal.

If you want, I can also add to the Pokédex the icon of the rare item they drop and a scrolling text effect for long names in both modals (the effect in the video triggers at 18+ characters regardless the screen size).
[basculin.webm](https://user-images.githubusercontent.com/104547700/206888754-6d3a61ae-24ef-4d96-ad6e-f994e6f1a9c3.webm)

image
image
image
image


Crobat4:
Conflicts solved, I also made some changes:
* Added filter for Pokémon that can mega evolve in the current game version and Mega Pokémon in both modals (Gigantamax too but it's commented).
* Added Category filter and category dots in each Pokémon entry in the Pokédex.
* LazyLoad `pageSize` increased from 40 to 60 in both modals due some scrolling issues.

image
image
image

Jaaslet (approved):
Looks good.
I did some small code cleanup, and i removed the commented gigantiamax code, as i don't think it will make sense with our current idea for the implementation. But it will be easy to add again, if we find out it makes sense. I have also done some small code cleanup.

I have two comments. But they are nice to have, and i really want this merged, so you can solve them in an other PR:
- On mobile view, the pokemon list should not be collapsed as default. It looks very empty when you open it, and it might be hard to see that they pressed anything.
- The name search field should be bigger. Some forms have long names, so the current searchbar is not big enough.

Thanks for the PR

PR #3552, opened 2022-12-11
Generation 9 stat changes

Qwertypop04:
Some existing Pokemon had their base stats altered as of gen 9. This updates these Pokemon to their current correct values


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3557, opened 2022-12-11
Profile modal : shiny icon after each shiny Pokémon in the select menu.

RegisCoaxans:
When looking for a nice Pokémon to display on the trainer card, we might want to show a shiny one.
This PR adds an indicator on our shiny Pokémons in that menu.
image


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3563, opened 2022-12-12
Refactor code to other visibilitychange listener, only hide game element

RedSparr0w:
Instead of hiding everything, this will only hide the main game element,
So it should be less jarring when tabbing back into the game.
image


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3564, opened 2022-12-12
Update FAQ for click attack

BaileyP2SR:
FAQ in game still only showed 50% click attack increase for shinies, updated that and added about Pokerus resistance there as well


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3566, opened 2022-12-12
Make amaze mulch less confusing

DraKay:
Before start I swear I'm not obsessed with the farm! Is just that.... That.... OK, maybe I'm obsessed with the farm, but is perfect because then I change this things...

In today incredible and complex edit we come to change Maze mulch description.. You know that description that make people think amaze mulch isn't going to shuck... (but we know it shucks... 

_Credits_

Problem courtesy of Lucho
Text courtesy of Pixl
File location courtesy of TinyCoyote
Edits courtesy of DraKay


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3568, opened 2022-12-12
Magikarp jump: Now showing temp battles until unlocked

Jaaslet:
Crew felt like you just completed the same temp battle again and again. So now they shows up as disabled, until unlocked.
And bonus fixed some requirements that does not have an impact, but was wrong


RedSparr0w (approved):
👍

PR #3571, opened 2022-12-12
Magikarp jump temp battles now shows trainers

Jaaslet:
Used the 3 different gymleaders we have, and added them in as random pattern as my simple human brain can do


RedSparr0w (approved):
Looks good 👍
Thanks for the PR!

PR #3573, opened 2022-12-13
Town Caught Status

### LINKED ISSUE #2964: [SUGGESTION] shop mons in key
Add shop mons to the uncaught/caught key checker

please even as a toggleable option.

I can never remember where all the shop mon are, feel like I waste 5 minutes a time scouring the universe for missing shopmons

jk13pclick:
Closes #2964. Code is kind of sloppy, could probably be improved on for cleanliness/efficiency. Also does not deal with Pokerus, as Egg Items do not currently have a PKRS indicator.


Jaaslet (changes requested):
A much cleaner solution would be to implement that logic on the `Shop.ts` class. Add the following method and add your logic to it:
`public areaStatus() {
  return areaStatus.completed;
}`
Then everything should just work fine. And when you move the logic, remember that it should be based on the lowest value, and not the last item

jk13pclick:
Did the requested code cleanup, let me know if it looks good!

Jaaslet (approved):
I felt like it was a bit too aggressive to show mapcolors for evo stones and eggs, as they are all over the place (I apparently have no shinies at all). So i changed it to just shopmons.
But other than that, it looks good. Thanks for the PR

PR #3574, opened 2022-12-13
Fire All + Tooltip

### LINKED ISSUE #3491: [SUGGESTION]: Set the Fire All Button from Hatchery on all tabs
### Current Behavior

A nice addition to the game was the inclusion of a Fire All Button to stop all working Hatchery Helpers. However, I find it annoying that it's only on the Helpers tab.

### Improved Behavior

Make that button appear on all tabs (Pokémon, Eggs, Fossils, Help even) so that we don't have to move to the Helpers Tab to remove them.

### Reason for change

Slight gain of time and click, improved qol

### Detailed Description

Add the Fire All button to all Hatchery tabs. In other words, make it permanent at the bottom of the Hatchery modal as long as a Helper is working.

Maybe the same thing could be done to the Farm Helpers, I'll do a separate issue for those.

jk13pclick:
Closes #3491


Jaaslet (changes requested):
If we want to show the button all the tabs, we should remove the code i used for it. So remove all the `onclick` from the nav-items.

jk13pclick:
I don't think we need the button on Eggs, Fossils, or Help, which is why I kept the onclick stuff. Easily changed if we want that, I just didn't think it was necessary.

Jaaslet (approved):
> I don't think we need the button on Eggs, Fossils, or Help, which is why I kept the onclick stuff.

I don't understand why we need it on the pokemon page, so i just thought we wanted it on all of them.
But yeah, if we only want it on those two pages, this looks good. Thanks for the PR

PR #3577, opened 2022-12-13
Increase mega evo requirement

CorgiOnNeptune:
Devs are in conclusion that the attack threshold required to Mega Evolve Pokemon with the Key Stone was too low, resulting in a requirement that many later game players wouldn't even notice.

The previous requirement was to increase the Pokemon's attack to `100x` its base. The requirement is now `500x`.


Jaaslet (approved):
Looks evil. Thanks for the PR

PR #3585, opened 2022-12-14
Dungeon Token Quest Rebalance

jk13pclick:
Experimental new algorithm. Basic premise is that is uses the maximum available token cost for balancing. Token cost is multiplied by a random ``intBetween(1, player.highestRegion()+2)`` to get the ``amount``. The proportion of ``amount``/``Max possible amount`` is multiplied against a constant (``200``) for the reward value.

Notes: Currently the maximum QP obtainable from this quest type is ``(player.highestRegion() + 2) * 200``. So in Galar this would be 1800QP. In Kanto this max value would be 400QP.


jk13pclick:
I pushed a GainMoneyQuest rebalance that models this one. Can be reverted.

src/scripts/quests/questTypes/GainMoneyQuest.ts
```diff
@@ -8,13 +8,17 @@ class GainMoneyQuest extends Quest implements QuestInterface {
     }
 
     public static generateData(): any[] {
-        const amount = SeededRand.intBetween(20000, 60000);
-        const reward = this.calcReward(amount);
-        return [amount, reward];
+        const availableGyms = GameConstants.RegionGyms[player.highestRegion()].filter(x => App.game.badgeCase.hasBadge(GymList[x].badgeReward));
+        const maxGym = GymList[availableGyms[availableGyms.length - 1]];
```
Aegyo, line 12: It's possible the player doesn't have any badges from their highest region yet, which makes this crash

src/scripts/quests/questTypes/GainTokensQuest.ts
```diff
@@ -8,13 +8,17 @@ class GainTokensQuest extends Quest implements QuestInterface {
     }
 
     public static generateData(): any[] {
-        const amount = SeededRand.intBetween(1000, 8000);
-        const reward = this.calcReward(amount);
-        return [amount, reward];
+        const regionDungeons = GameConstants.RegionDungeons[player.highestRegion()].filter(dungeon => TownList[dungeon].isUnlocked());
+        const maxDungeon = dungeonList[regionDungeons[regionDungeons.length - 1]];
```
Aegyo, line 12: Similar issue with the maxGym calc - it's possible for no dungeons to be unlocked

Jaaslet:
image

Jaaslet:
This has been dead for a while, and other people have started on the quest rebalance. So i will close this one

PR #3588, opened 2022-12-14
Added enviroments to magikarp jump

Jaaslet:
They are all water! (big surprise).
There are no town images, because i have no idea what to put there... But i am not sure if we need them. Will at least remove it from the todo list, because i don't want it to hold back the release


RedSparr0w (approved):
Great change, adds a lot of variety to the region 🎉 
Thanks for the PR! 👍

PR #3589, opened 2022-12-14
wip(Magikarp Jump): Gyms now rewards money

Jaaslet:
Based in on the elite 4 of Alola. Should be less in the start and about the same in the end. I have adjusted for each gym only having one mon


RedSparr0w (approved):
Looks good,
Thanks for the PR 💸

PR #3591, opened 2022-12-15
Add Pokerus check for Shopmons

jk13pclick:
Adds a Pokerus Resistant check for Shopmons


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3594, opened 2022-12-15
New Magikarp Jump HP balance

Jaaslet:
Turned out i got some numbers wrong, and my balance was way to easy. So here is the new balance:
image


RedSparr0w (approved):
Seems more balanced ⚖️ 
Thanks for the PR! 👍

PR #3595, opened 2022-12-15
Fill in magikarp jump routes

Jaaslet:
The routes was a bit empty. I have done the following:
- Made sure there where 3 magikarps base forms on each route. I want players to be tired of seeing base forms (like in the game)
- Filled in some forms you could obtain earlier. There are at least two on each route. Makes it a bit more fun. (i should maybe have added 3? Not sure...) **Edit: I added more forms to the later routes**
- One route had too many new forms, so i added a bunch of old ones there


RedSparr0w (approved):

PR #3602, opened 2022-12-16
Making Hop and Leon team requirements on Galar starter

Razikd29:
Hop and Leon teams were pointing on Kanto and Alola starters to determine their teams. Now it's about the Galar starter.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3607, opened 2022-12-16
Normalize Alcremie Day/Night Cycle

Razorflamekun:
This pull request is meant to normalize the Day/Night cycle in the Battle Cafe to make it consistent with the rest of the game.


Jaaslet:
Night should start at 18, right?

Razorflamekun:
Yeah, did I set it to something else?

src/scripts/towns/BattleCafe.ts
```diff
@@ -77,9 +77,9 @@ class BattleCafeController {
             (new PokemonItem('Milcery (Cheesy)', 0)).gain(1);
             return;
         }
-        if (curHour == 19 && !clockwise && spinTime > 10) {
+        if (curHour == 17 && !clockwise && spinTime > 10) {
             spin = GameConstants.AlcremieSpins.at7Above10;
-        } else if (curHour >= 5 && curHour < 19) { // Is day
+        } else if (curHour >= 6 && curHour < 17) { // Is day
```
Jaaslet, line 82: It's day if current hour is less than 17

Razorflamekun: Yes, that should be the case, as dusk is it's own time period, and then night starts at 18, so if this is set to 17, then it should make it day from 6-17, Dusk from 17-18, and Night from 18-19, right?

Razorflamekun: Did I miss something in my changes?  I'm new to coding.

Jaaslet: `<` is less than. So `<17` means as long as the current hour is less than 17. So when the clock turns 17, it's no longer true

Qwertypop04: > Yes, that should be the case, as dusk is it's own time period, and then night starts at 18, so if this is set to 17, then it should make it day from 6-17, Dusk from 17-18, and Night from 18-19, right?

Dusk is the last hour of day. So from 17-18, it should be day, and also dusk

Razorflamekun: I've fixed the         } else if (curHour >= 6 && curHour < 17) { // Is day to be         } else if (curHour >= 6 && curHour < 18) { // Is day

now. This should fix the issues, correct?

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3613, opened 2022-12-18
Shiny Noctowl

Qwertypop04:
Makes Ash's Noctowl shiny. Because we can do that now


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3616, opened 2022-12-18
Adding Safari Zone Statistics

Farboo171:
Adding Safari Zone statistics for future plans and achievements.

No megas in this PR, top be implemented later


RedSparr0w:
Are there any other statistics worth adding at the same time?
(shiny) Pokémon Encountered
(shiny) Pokémon Caught
Steps Taken
Times Entered

Farboo171:
Added steps taken (verified that trying to take steps when you can't, like into a tree, does not increment steps)
Added times entered, only counts when you pay the entry fee.

Did you mean shinys encountered in general, or in the safari zone?

RedSparr0w (approved):

RedSparr0w:
> Did you mean shinys encountered in general, or in the safari zone?

Shinies encountered while in the Safari Zone.
Going to merge this PR as is though, so feel free to do it in another PR if you'd like.
May also be worth adding a statistic for overworld shinies that have appeared, even if not encountered.

PR #3621, opened 2022-12-19
Organized Berry and Color Wanderer code

osherwott:
- Berries are more organized in Farming.ts
- Color Wanderers moved to Berry.ts
- Color Wanderers tweaked based on XY spawns: Ledyba, Illumise, Spewpa, Burmy (Plant), Combee, and Volbeat
- Added Flabébé (Red) as a Color Wanderer to match with (Blue) and (Yellow) being available as wanderers


osherwott:
This first farm-centric PR for additional berry characteristics and organized code is ready to review. A PR regarding the modal UI will come at a later date.

Aegyo:
I think this should be 4 different PRs:
- Reformatting Farming.ts berry data
- Color wanderers
- Extra berry data
- Moving hint text

Also your unit setting shouldn't have two separate SI options since the unit is the same, I don't think a different number format should count as a separate unit. It would also be better for that data to be just a single number, and have a function to convert it into other unit options.

Some ideas for what/how that data could be used for might also be helpful for getting that part merged, I'm guessing something to do with pokeblocks?

osherwott:
Sounds good. I'll split the PR accordingly.

Regarding Units:
- I agree that the number format (comma vs decimal) can be its own setting, but since it would be better to have such a setting affect all numbers I opted to implement what could work for this PR. The scope of a more universal setting is out of my current expertise, sadly.
- The data itself can be a single number that can be converted, but some berries share the same length in one unit while differing in another. This is a minor issue though and if it is achievable, I would rather have a single value as well.

Ideas for having these new elements be functional is a work in progress and linked in the suggestion above. I would very much like these to be more than descriptive! The only block in doing so is my current coding capabilities :).

Aegyo:
> some berries share the same length in one unit while differing in another

Could that be a precision thing, or are the games just giving inconsistent numbers?

osherwott:
> Could that be a precision thing, or are the games just giving inconsistent numbers?

Yes, it's precision. I just compared values and inches are rounded to the first decimal place after being converted. Of course, this also means that the values to be used here would need to be in centimeters, if we are only choosing one number to represent size and converting from that.

osherwott:
I have removed everything new except the new color wanderer code, I hope that's okay. This should be easier to review now that's it's only reformatted code!

Aegyo (changes requested):
Berry data reformat looks fine, I've compared `App.game.farming.berryData` before and after and it just some minor wander additions and re-ordering that is different.

Just this small change with the colorWander config format, and I'll approve and merge :+1:

osherwott:
Works perfectly. Ready to merge.

Aegyo (approved):

PR #3622, opened 2022-12-19
Add confirmation when restarting Battle Frontier run

CypherX:
Adds a confirmation dialog to BF when clicking Start while having a checkpoint. Prevents accidentally losing your checkpoint.


Jaaslet (approved):
I updated the text a bit. Looks good. Thanks for the PR

PR #3627, opened 2022-12-21
Update publish script

RedSparr0w:
Now the publishing will only push through the `/docs` folder (and `./package.json`).
I have also removed `/docs` from the develop branch, which will now hold everything else except that.
This will reduce the amount that needs to be downloaded for the client, and reduce what needs to be downloaded when working on the development of the repo as there was a lot of duplication of files.

NOTE:
Once merged, we will need to manually delete all the files we don't want from the `master` branch, and it should continue working as expected from that point onwards.

Check out the [`live-test` branch](https://github.com/pokeclicker/pokeclicker/tree/live-test) to see what `master` branch would look like.


Jaaslet (approved):
Looks good. Stuff still works in development. I guess the best test would be to push it, and see if the dev site is still working?
Thanks for the PR

PR #3628, opened 2022-12-22
Goodbye lonely pixels - - A Xmas tale of an underground items rework

DraKay:
Welcome ladies and geltemen to this winter tale... Once upon a time there was a mysterious place under the earth, local called it the underground. That place was full of some items people digged for get magic eggs and poddles with weird hairstyles. But that place was so lonely that always was trying to make people dig harder to have people in it more time. But that tricks made people avoid to dig its magic shards using an arcane magic called chains or just trying to find them in chest people just forgot in the wild.

And here we are... Some days before Xmas, our poor underground lonely, waiting the chisels and the bombs, crying where nobody could see it, yelling where nobody could hear it... And it is the perfect moment for a ghost visit, as you all know this tale you know that we start with the past Xmas ghost:
-hi I'm the past underground ghost- said a little gastly- any past is full of regrets and mistakes, yours are that little shards coded with wrong sizes, is time to put 4x3 shard when they are 4x3...

The underground looked inside him and recolocated a few things on it, it purple and crimson shard changed their digging shapes, the fossilized drake recovered a full size being digged the same...

-hi I'm the present underground ghost-said a funny haunter appearing for the back of the underground- no present is perfect, no present is enough... Your present need to get rid of tiles with only few pixels, nobody can see them, you are hurting people with sight problems...

The underground though about it... And with a lot of effort took its shards and fossil parts and make them Biggers in some points being sure they were visible...

-hi I'm the future underground ghost-said a spooky big gengar in front of the underground - the future is our goal, the future need to be work. Work in clean, work in image style, that's your future...

The underground smiled and clean a bit extra pixels and joined the styles of some Evo items....

*What I made: I made all shards visible, purple and crimson shard change the digging layout as now they recover their original shape. I cleaned some pixels and redraw a few Evo items that had a different style*


RedSparr0w (approved):
Changes look good to me, I'm not entirely sure the best way for me to test it though 😅 
Thanks for the PR! 👍

PR #3629, opened 2022-12-22
HGSS Red Sprite

Qwertypop04:
Has Red use the HGSS sprite instead of the non canon sprite currently used


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3630, opened 2022-12-22
Suicune does not longer show up early

Jaaslet:
If you had defeated the johto champion, but had not completed the suicune quest, one of the kanto suicunes would show up on the map. This fixes that


RedSparr0w (approved):
Thanks for the PR! 👍
Changes look good.

PR #3632, opened 2022-12-23
Update hatchery region filter when moving to new region

CypherX:
When a player travels to a new region the hatchery region filter will now update to include the new region if all prior regions are selected. If a player has configured the filter to not have all regions selected, we will not add the new region. Previously the new region would not be selected sometimes leading to confusion when pokemon from the new region did not appear in the hatchery.

Test save that is ready to travel to Galar (provided by Eat Pant): 
[Normal_Pre-Galar.txt](https://github.com/pokeclicker/pokeclicker/files/10292161/Normal_Pre-Galar.txt)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3633, opened 2022-12-23
Dr Splash quest

Farboo171:
A quest to fill out the Magikarp Jump region.

One TODO from Jaas to change the Tackle Machine quest step from water gems to total mons defeated.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3634, opened 2022-12-23
Add a display name to Colress 3 (in code)

CooldudeOmega:
did more than edit a typo and actually added a line of code
fixes typo where Colress 3 name showed as... Colress 3, instead of Colress


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3635, opened 2022-12-23
Vitamins in Dream Orbs

Farboo171:
Proteins were in dream orbs as an abstraction of Rare Candy.

Now that there's more vitamins, might as well put them in the dream orbs too. split protein odds 3 ways to not affect Therian obtain balance.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3636, opened 2022-12-24
Remove Level from sorting options in the Hatchery

EatPant2nd:
Removes the "Level" option in both manual hatchery and Hatchery Helpers. Also, makes Hatchery Helper sort use the same sorting options as manual side for consistency. No more Egg Cycles, yay!


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3638, opened 2022-12-24
Update Santa Snorlax roaming locations

CypherX:
Adds Santa Snorlax to all roaming pools in every region rather than just the first roaming pool. Currently in regions with multiple roaming pools (Kanto & Galar) it was only being added to the first pool, meaning it would be roaming the Kanto mainland but not Sevii, and in Galar it would only being roaming in Galar South. The only place it's excluded from is Galar North as it has no roamers.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3639, opened 2022-12-24
Now hiding shady salesman on the map

Jaaslet:
Mapcolors gave the easteregg away. Either because you where missing a mon from the shop or you changed your completed mapcolor. This just removed mapcolors from the two shady salesman town.
And bonus fix of magikarp jumps salesman requirement


RedSparr0w (approved):
Changes look good,
Thanks for the PR! 👍

PR #3640, opened 2022-12-24
Fixed magikarps eyes location

Jaaslet:


RedSparr0w (approved):
Haven't actually tested this one, but assume it's good if you went through the effort of making the change 😅 
Thanks for the PR 👍

PR #3642, opened 2022-12-24
Comment out unused Mega Evolutions

Qwertypop04:
Comments out Mega Evolutions where the Mega Stone is not yet available, as the completeness of the Key Stone evo item depends on them


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #3643, opened 2022-12-24
[SUGGESTION]: Change Heatran back to canon egg cycles


Farboo171:
### Current Behavior

Heatran has egg cycles of 80, or 3200 steps. This is an outlier from canon, where it has an egg cycle value of 10, or 400 steps

### Improved Behavior

Make Heatran's egg cycles be 10, or 400 steps

### Reason for change

Match the value with canon.

### Detailed Description

#1102 changed the Heatran egg cycles value from 80 to 10 because Heatran was considered to be too powerful. Returning its egg cycles to the canon value would put Heatran at Breeding Efficiency of 4.775, which would make it the highest BE in the game. I think that is now OK for the following reasons:

1. Things should be canon unless there is a good reason to make them not canon
2. The next two strongest pokemon by BE are Gyrados (4.63) and Manaphy (4.6). They are very close to Heatran in BE, so 4.77 is not a huge outlier
3. A legendary pokemon being the strongest pokemon in the game by some measure makes sense
4. From a balance standpoint, Fire and Steel are not very well represented in High BE mons. The current strongest Fire type is Heatmore (BE 1.63) and the strongest steel is Totem Togedemaru (BE 2.18).  Water and fairy dominate the top breeders, and this would help balance this out

RedSparr0w:
With the Vitamin rework,
Once you can apply 40 Vitamins, this would be each of their Efficiency:
![image](https://user-images.githubusercontent.com/7288322/209454251-75900a41-3326-4d83-b68b-1ec2f7c10515.png)

Farboo171:
Added via #3644

PR #3644, opened 2022-12-25
Change Heatran back to canon egg cycles

Farboo171:
Per #3643


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3646, opened 2022-12-27
fix Ultra Megalopolis requirement

Qwertypop04:
locks Ultra Megalopolis temp battle behind Necrozma temp battle


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3653, opened 2022-12-29
Update inaccessible dock notification text when in Kanto

CypherX:
Removes the `Progress further to return to previous regions!` text from the dock notification when the player is in Kanto.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3654, opened 2022-12-29
Clear Safari battle state when modal closed

CypherX:
Unlocking the Holo Caster key item (110 unique pokemon) by capturing a pokemon in the Safari zone forces the Safari modal to close to display the new key item modal. This causes the Safari battle state to remain as `true` preventing any interaction (movement, leaving) upon re-entering the Safari zone requiring the player to reload.

Discord bug thread: https://discord.com/channels/450412847017754644/1057809832628977765


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3655, opened 2022-12-29
Multiquest descriptions for Logbook

CorgiOnNeptune:
Added descriptions to some MultipleQuestsQuest(s) since they were displaying blank in the logbook otherwise. Used super generic descriptions for the Meltan quests. Fixed some other typos along the way.


RedSparr0w (approved):

PR #3656, opened 2022-12-29
Include amount of each color Dream Orb obtained in notification

CypherX:
Updated the Dream Orb notification to list the amount of each color obtained while offline.

image


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3657, opened 2022-12-29
Repel Aura added to game and Micle Berries

Farboo171:
This adds a Repel aura to Micle berries that allows for people to stop wanderers from coming onto their farms.

The aura calculation is additive instead of multiplicative, such that a full elimination of wanderers is possible with 4 Micle berries on the field. This is explained to an extent in the Help tab.

Micle is selected because it does not currently have a use, and it is somewhat difficult to get.

There is no real reason for this PR other than to help out monotype challenge people, but it turns this berry from useless to niche useful,


src/scripts/farming/Farming.ts
```diff
@@ -1027,6 +1031,12 @@ class Farming implements Feature {
             .reduce((acc, p) => acc * (p.emittingAura.value() ?? 1), 1);
     }
 
+    addPlotAuras(auraType: AuraType): number {
+        return this.plotList
+            .filter(p => p.emittingAura.type() === auraType)
+            .reduce((acc, p) => acc + (p.emittingAura.value() ?? 1), 0);
```
PixLSteam, line 1037:
```suggestion
            .reduce((acc, p) => acc + (p.emittingAura.value() ?? 0), 0);
```
This should probably default to 0.

PixLSteam (changes requested):
One minor thing, the rest looks good to me!

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3658, opened 2022-12-29
Thick Club and Quick Powder are now obtainable

Jaaslet:
Thick club drops from Alolan Marowak and Quick Powder drops in pokemon village.
Alolan Marowak only spawns in dungeons, so it kinda follows my "trying to keep held items in dungeons"-plan. Also we can't give it to the kanto one, as players would then be able to unlock held items in kanto (they already need to learn too much in kanto).

And yeah, the items are hard to unlock, but they are not really used for anything other than just a small boost for some mons. So only useful if you really love them, or you have done everything else in the game (and both situations you want to spend time on getting them)


RedSparr0w (approved):

Is it worth adjusting the base price to `Infinity` or something just to kind of indicate in the code it's not a store item?

Jaaslet:
> Is it worth adjusting the base price to `Infinity` or something just to kind of indicate in the code it's not a store item?

Yeah, maybe. But then we need to fix it a lot of places

PR #3659, opened 2022-12-30
Exclude Dream Orb colors from notification when none found

CypherX:
Noticed when testing other things that the Dream Orb notification update in #3656 was displaying all unlocked colors even if none of that color were found.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3660, opened 2022-12-30
Release the Magi-kraken!

Jaaslet:
I am happy with magikarp jump, so this PR removes the DevelopmentRequirements, so it will be live in the live version


RedSparr0w (approved):
👍 🎉

PR #3662, opened 2022-12-30
Calcium and Carbos are available later

Jaaslet:
To make it less complicated for new players, the new vitamins are unlocked later.
The solution is not great, but it's what i have time for...

Also bumped the prices a bit, so they are not too cheap for lategame players


RedSparr0w (approved):

PR #3665, opened 2022-12-30
Add display name to Jump Champs

CorgiOnNeptune:
Add display name instead of having Jump Champ 1/2/3.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3666, opened 2022-12-31
Fix eggs not gaining steps after update

CypherX:
Eggs that were ready to hatch prior to updating and had been given proteins will not gain egg steps after updating due to the egg  `notified` property being set to `true`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pokeclicker",
-  "version": "0.10.5",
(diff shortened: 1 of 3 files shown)


CypherX:
Discord bug thread: https://discord.com/channels/450412847017754644/1058525527838826496

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3668, opened 2022-12-31
Mega Mewtwo and Gengar fix

Qwertypop04:
Fixes Mega Gengar having its types flipped, and both Mega Mewtwos having the same ID


RedSparr0w (approved):

PR #3670, opened 2022-12-31
Fix broken quest rewards due to ID changes

Farboo171:
Fixes Spike Eared Pichu, Blade Aegislash, and Magerna's ID #'s in quest rewards


RedSparr0w (approved):

PR #3673, opened 2022-12-31
Fix several more broken rewards due to ID changes

CypherX:
Fixed:

- Urshifu (Rapid Strike) (was giving MissingNo.)
- Pinkan Pikachu (was giving Let's Go Pikachu)
- Pikachu (World Cap) (was giving Pikachu (Partner Cap))

Also switched other alt-form unlocks to be name-based rather than id-based.

We will most likely need an update function or something in `checkAndFix()` to unlock the proper Pokemon for players that may have completed these things while broken.

### DIFF
--- a/src/scripts/dungeons/Dungeon.ts
+++ b/src/scripts/dungeons/Dungeon.ts
@@ -8976,7 +8976,7 @@ dungeonList['Tower of Waters'] = new Dungeon('Tower of Waters',
     ],
     2000000, 36,
     () => {
-        App.game.party.gainPokemonById(pokemonMap['Urshifu (Rapid Strike)'].id);
+        App.game.party.gainPokemonByName('Urshifu (Rapid Strike)');
         Notifier.notify({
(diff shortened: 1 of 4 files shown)


RedSparr0w:
We're probably better off making a new method `gainPokemonByName` which takes the `PokemonNameType` so we will hopefully be able to avoid this in the future.
```js
    gainPokemonByName(name: PokemonNameType, shiny = false, suppressNotification = false, gender = -1) {
        const pokemon = pokemonMap[name];
        this.gainPokemonById(pokemon.id, shiny, suppressNotification, gender);
    }
```
This way it will validate it's a valid Pokémon name before reaching the main files.

CypherX:
I'll update this PR to make use of `getPokemonByName` when #3674 gets merged.

CypherX:
Updated to use newly added `getPokemonByName()` when awarding alternate forms.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3674, opened 2022-12-31
Add `gainPokemonByName` method

RedSparr0w:
This PR adds a `gainPokemonByName()` method to `Party`, this way we can instead use the Pokémon name rather than ID so we don't end up with more MissingNo.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3679, opened 2022-12-31
Solving Grey shard mistake

DraKay:
Sorry, sorry, sorry and sorry... It was my mistake, Grey shard is bigger than I write is. This single 1 change should fix my mistake, is surprising is the first I do or maybe isn't...


RedSparr0w (approved):

PR #3683, opened 2022-12-31
Update function to give player missing reward Pokemon

CypherX:
Some players may be missing one-time reward pokemon as a result of the ID changes in the previous update:

- Spiky-eared Pichu
- Aegislash (Blade)
- Magearna (Original Color)
- Pinkan Pikachu
- Pikachu (Partner Cap)
- Urshifu (Rapid Strike)

This will give the player the missing pokemon if they have the relevant criteria (questline, temp battle, dungeon) completed. Not sure if any other statistics need to be incremented. Also didn't include anything to remove Missingno as I'm not sure if we just want to remove it for everyone or only those affected by this issue.

Included some re-usable helper methods in Update that can be used in the future if needed.

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -1901,9 +1901,9 @@ class Update implements Saveable {
             Update.giveMissingQuestLineProgressRewardPokemon(saveData, 'Unfinished Business', 8, 172.01);
             Update.giveMissingQuestLineProgressRewardPokemon(saveData, 'Princess Diancie', 6, 681.01);
             Update.giveMissingQuestLineProgressRewardPokemon(saveData, 'A Mystery Gift', 1, 801.01);
-            Update.giveMissingTempBattleRewardPokemon(saveData, 'Ash Ketchum Pinkan', 25.14);
-            Update.giveMissingTempBattleRewardPokemon(saveData, 'Ash Ketchum Alola', 25.08);
-            if (saveData.statistics.dungeonsCleared[GameConstants.getDungeonIndex('Tower of Waters')] > 0) {
+            Update.giveMissingTempBattleRewardPokemon(saveData, 123, 25.14); // Ash Ketchum Pinkan
+            Update.giveMissingTempBattleRewardPokemon(saveData, 151, 25.08); // Ash Ketchum Alola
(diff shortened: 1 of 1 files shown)


src/scripts/Update.ts
```diff
@@ -1897,6 +1897,15 @@ class Update implements Saveable {
         },
 
         '0.10.6': ({ playerData, saveData }) => {
+            // Give the player any missing questline or temporary battle rewards
+            Update.giveMissingQuestLineProgressRewardPokemon(saveData, 'Unfinished Business', 8, 172.01);
+            Update.giveMissingQuestLineProgressRewardPokemon(saveData, 'Princess Diancie', 6, 681.01);
+            Update.giveMissingQuestLineProgressRewardPokemon(saveData, 'A Mystery Gift', 1, 801.01);
+            Update.giveMissingTempBattleRewardPokemon(saveData, 'Ash Ketchum Pinkan', 25.14);
+            Update.giveMissingTempBattleRewardPokemon(saveData, 'Ash Ketchum Alola', 25.08);
+            if (saveData.statistics.dungeonsCleared[GameConstants.getDungeonIndex('Tower of Waters')] > 0) {
```
RedSparr0w, line 1906: Unfortunately we need to use magic numbers in here, as the index could change in the future.

src/scripts/Update.ts
```diff
@@ -2300,6 +2309,27 @@ class Update implements Saveable {
...
 
+    static giveMissingQuestLineProgressRewardPokemon(saveData, questLineName: string, questStep: number, pokemonId: number) {
+        const quest = saveData.quests.questLines.find((q) => q.name == questLineName);
+        if (quest?.state == 2 || (quest?.state == 1 && quest?.quest >= questStep)) {
+            Update.giveMissingPokemon(saveData, pokemonId);
+        }
+    }
+
+    static giveMissingTempBattleRewardPokemon(saveData, tempBattleName: string, pokemonId: number) {
+        const tempBattleIndex = GameConstants.getTemporaryBattlesIndex(tempBattleName);
```
RedSparr0w, line 2320: Same thing here, not sure what the best solution would be though

CypherX:
Changed to use the indices as of the current update.

RedSparr0w (approved):

PR #3685, opened 2023-01-01
Update function to adjust tempBattle stats in saveData

CorgiOnNeptune:
Created a function for this as well as these temp battle issues are becoming a more common occurrence, so it's nice to have a quick and easy fix to toss in an update rather than editing individual affected files.

Not sure what broke the stats for the battles I included, could be script related, could not be. But, doesn't hurt to fix.

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -1908,10 +1908,9 @@ class Update implements Saveable {
             }
 
             // Set 'Team Plasma Grunt 1' to 0 if quest step not completed
-            Update.fixTempBattleState(saveData, 'Team Plasma Grunt 1', 0, 'Quest for the DNA Splicers', 0);
-
-            // Set Cyrus as complete if you've completed 'A New World'
-            Update.fixTempBattleState(saveData, 'Galactic Boss Cyrus', 1, 'A New World', 3);
+            Update.fixTempBattleState(saveData, 64, 0, 'Quest for the DNA Splicers', 0);
+            // Set Cyrus as complete if 'A New World' completed
+            Update.fixTempBattleState(saveData, 57, 1, 'A New World', 3);
         },
     };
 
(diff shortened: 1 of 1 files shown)


src/scripts/Update.ts
```diff
@@ -2309,6 +2315,25 @@ class Update implements Saveable {
         }
     }
 
+    // Use setBattleState as 0 or 1 to manipulate battles to what status they should be based on related questline progress.
+    static fixTempBattleState = (saveData, battleName: string, setBattleState: number, questLineName: string, questStep: number) => {
+        const ql = saveData.quests.questLines.find((q) => q.name === questLineName);
```
RedSparr0w, line 2320: May be worth just chucking a return in here if no quest found?

src/scripts/Update.ts
```diff
@@ -2309,6 +2315,25 @@ class Update implements Saveable {
...
+
+        if (setBattleState === 1) {
+            // set to complete if related questline/step is completed
+            if (ql?.state === 2 || ql?.quest > questStep) {
+                saveData.statistics.temporaryBattleDefeated[GameConstants.getTemporaryBattlesIndex(battleName)] = 1;
```
RedSparr0w, line 2325: Unfortunately we can't make use of the `getTemporaryBattlesIndex` as the indexes can change from update to update, so we need to use magic numbers (current index as of this update).

src/scripts/Update.ts
```diff
@@ -2309,6 +2315,25 @@ class Update implements Saveable {
...
+            }
+        }
+
+        if (setBattleState === 0) {
+            // set to not complete if related questline/step isn't complete
+            if (ql?.state < 2 && ql?.quest <= questStep) {
+                saveData.statistics.temporaryBattleDefeated[GameConstants.getTemporaryBattlesIndex(battleName)] = 0;
```
RedSparr0w, line 2332: Same as above with magic numbers

CorgiOnNeptune: Changed battleNames to the index and added a return if `ql` is undefined

RedSparr0w (approved):

PR #3686, opened 2023-01-01
Implement pluralizeString() function

CorgiOnNeptune:
Adds global function to check for plural cases on strings. Currently utilized in shops, deals, and dungeon loot notifs. 

Definitely doesn't cover every case it might run in to, but for now it covers common cases for `Berry -> Berries`, `Mulch -> Mulches`, `Carbos -> Carbos`. More cases can always be implemented later.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3688, opened 2023-01-01
Luxury ball currencies

### LINKED ISSUE #1408: [SUGGESTION] What the luxury ball can do.
**Current Behavior**

Currently, the luxury ball doesn't do anything, but I believe that can change.
**Improved Behavior**

I think the luxury ball can make the Pokemon you catch have a chance of being shiny.
**Reason for change**

 The luxury ball doesn't have any features currently, and I think this could make it interesting.
**Detailed Description**

When you catch a pokemon, there will be a very low chance for that pokemon to be shiny. Maybe around 1/2048 chance. The odds could work like how evolution stones work, along with safari zone encounters.
**Screenshots:**

Gameonlp:
Added a new enum for currency conversions (based on master balls)
Added extra effect for luxury ball
closes #1408


Farboo171:
I'd suggest commenting out the whole Contest Token portion for all files. We'll want to turn it on later, but it would be strange and confusing to reward contest tokens right now.

Gameonlp:
Contest tokens are commented out, the conversion rates feel like they have to be tweaked, but i wanted to use the one currency equivalent we had for all currencies as base

Farboo171:
Looking at balance, the values are a bit high. 5 diamond (or 5/6 diamond after current currency number and odds are included) for ~ 30 free gen 1, ~ 30 cheap gen 2 and ~10 easy gen 3 berries would make this a brokenly effective source for that currency. Similarly, it blows early BF gains out of the water.

It would be better to lower the amounts by a factor of 5. People will still cheese it for diamonds but it won't be as unbalanced

Farboo171:
The DT looks ok (on a per ball basis, only enough for ~one Johto dungeon), but all the others should be lowered.

Also you can make the contest tokens value equal to the QP value (still commented out), we're getting some numbers for that sorted out.

Gameonlp:
Values are currently scaled on the DT you would usually get on the route. If you would get 15000 DT for a catch you should with the current numbers get 2x of the stated values.

Farboo171:
> Values are currently scaled on the base DT you would usually get on the route. If you would get 15000 DT for a catch you should with the current numbers get 2x of the stated values.

Oh. That's still probably way too much then. That means that by Galar with reasonable achievement % and token collector players can be getting 10+ diamonds on a catch.

Will the code still work if you raise the DT amount? Do the other currencies round up?

Gameonlp:
> Will the code still work if you raise the DT amount? Do the other currencies round up?

The code will work, all gains are rounded up so, that existing code does not error

Farboo171:
Ok, I think that upping the DT to 15,000 will work with these numbers then. It lets players get 2 diamonds per 6 catches if they're not trying too hard by Kalos, and caps out at ~6 diamonds per 6 catches in Galar if they're trying really hard.

Thanks for your patience on the balance checking!

Gameonlp:
Really no problem with that. If anything is up with the code/balancing after merging you can also just @ me.

src/modules/GameConstants.ts
```diff
@@ -360,6 +360,16 @@ export enum Currency {
...
 }
 
+export enum CurrencyRate {
+    money = 200_000.0,
+    questPoint = 60.0,
+    dungeonToken = 15_000.0,
+    diamond = 1.0,
+    farmPoint = 60.0,
+    battlePoint = 10.0,
+//    contestToken = 60.0,
```
RedSparr0w, line 370: Could you update these to just use standard numbers, just for code consistency:
```suggestion
    money = 200000,
    questPoint = 60,
    dungeonToken = 15000,
    diamond = 1,
    farmPoint = 60,
    battlePoint = 10,
//    contestToken = 60,
```
_(same thing with the `.d.ts` file too please)_

Gameonlp: Changed it, should i keep the switch numerical or do you rather want the string version?

RedSparr0w: Should probably use the enums rather than magic numbers

Aegyo: If you make this a Record<Currency, number> you won't need the switch case at all - can just lookup the value directly by currency

src/modules/GameConstants.ts
```diff
@@ -360,6 +360,16 @@ export enum Currency {
     contestToken,
 }
 
+export enum CurrencyRate {
```
Jaaslet, line 363: Your names are a bit generic. If someone just saw `CurrencyRate` in GameConstants, they would not be able to guess what it's used for

Gameonlp: CurrencyExchangeRate maybe?

Jaaslet: The values are only used for luxury balls, right? Or are you planning to use the values for something else?

Gameonlp: I am not planning on using them for anything else

Jaaslet: Then include Luxury ball in the name, so other coders knows that it's what the numbers are balanced for

src/modules/GameConstants.ts
Gameonlp: Thanks for the advice combined both

src/modules/GameConstants.ts
Gameonlp: Done, should be more clear now

src/modules/GameConstants.ts
```diff
@@ -360,6 +360,16 @@ export enum Currency {
...
 }
 
+export const LuxuryBallCurrencyRate: Record<Currency, number> = {
+    [Currency.money]: 200000,
+    [Currency.questPoint]: 60,
+    [Currency.dungeonToken]: 15000,
+    [Currency.diamond]: 1,
+    [Currency.farmPoint]: 60,
+    [Currency.battlePoint]: 10,
+    [Currency.contestToken]: 60,
```
Aegyo, line 370: The previous balance suggestion didn't take into account that achievement % is only applying to money and dt, could you bump up the numbers on the rest to bring it more in line with what was intended?
```suggestion
export const LuxuryBallCurrencyRate: Record<Currency, number> = {
    [Currency.money]: 200000,
    [Currency.questPoint]: 600,
    [Currency.dungeonToken]: 15000,
    [Currency.diamond]: 10,
    [Currency.farmPoint]: 600,
    [Currency.battlePoint]: 100,
    [Currency.contestToken]: 600,
```

Gameonlp: Done. Tell me if they feel fine like that.

Aegyo (approved):

Issue #3667, opened 2022-12-31
[SUGGESTION]: Show Breeding Efficiency in Vitamins menu


PixLSteam:
### Current Behavior

Shows `Attack Bonus` and `Egg Steps`.

### Improved Behavior

Show `Breeding Efficiency` as well.

### Reason for change

This would allow players to see the changes to the Pokémon's BE immediately.

### Detailed Description

Add an additional column to the Vitamin Pokémon table, showing the Pokémon's Breeding Efficiency.

PR #3696, opened 2023-01-01
Added Breeding Efficiency in Vitamin Modal and fixed it

RegisCoaxans:
The Vitamin Modal now has a dedicated column for BE. Closes [#3667](https://github.com/pokeclicker/pokeclicker/issues/3667).

BE was broken : prior to 0.10.5, the Attack Bonus was not rounded down when computing that.
From 0.10.4 to 0.10.5, with no vitamin, over 600 Pokémons got their BE reduced, by up to a fifth, the change on the sorted list affected over 800 Pokémons.

The Attack Bonus in Vitamin Modal now shows raw Attack Bonus instead of rounded down one : I think that it helps see the exact difference when adding calciums.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3698, opened 2023-01-02
Egg and Hatchery Changes to support Monotype runs and other challenges

Farboo171:
This PR serves to make a few small changes that will leave game balance for normal playstyles unchanged, but provide a huge help for players working on special challenge runs

1. Move access to the Hatchery to being locked behind Route 3, not Route 5. This change will allow players to use mystery eggs and fossils as soon as they can obtain them. Most importantly, it will allow anyone doing a restricted run to begin to hatch mons for attack power after Pewter City. This change is a slight deviation from canon in terms of when the Pokemon Daycare can be accessed, but I think this small change is worth it.

2. Grass and water eggs have their pools slightly changed to remove redundant mons that allow for better type distribution from eggs. Paras replaces Bellsprout to allow access to Bug types in eggs, and does not affect balance (Bellsprout is still accessible between the 1st and 2nd gym). Slowpoke replaces Psyduck, adding Psychic type to the egg pool. This does not affect balance because both mons become accessible as soon as the Super Rod is unlocked.

3. Added three pokemon to the Mystery Egg pool: Jigglypuff to add Normal and Fairy, Gastly to add Ghost, and Geodude to add Ground and Rock. This will allow all types to be represented in Mystery eggs, making all monotype runs except Dark theoretically doable (Maybe not Fire if it can't beat Brock, but that's a different issue)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3702, opened 2023-01-02
[FIX] Egg.stepsRequired when Pokémon is not in party

PixLSteam:
`Egg.stepsRequired` was left uninitialized when loading eggs from JSON and the Pokémon is not in your party.
This should fix that and allow those eggs to gain steps again.

[Discord bug thread with affected save file](https://discord.com/channels/450412847017754644/1059561403331129405)


RedSparr0w (approved):

PR #3703, opened 2023-01-03
Toggle for showing non-boosted berry ripe times

### LINKED ISSUE #3336: [SUGGESTION] Option to show non-boosted berry ripe time
**Current Behavior**
Right now, you can set the options to see the time until the berry is ripe, or to the next stage of growth.

**Improved Behavior**
I am suggesting there be an option, somewhere in the settings menu, that if turned on, would show the time until the berry is ripe/next stage (based on current settings), but below would also show the time it would take without any growth speed boost.

**Reason for change**
My change would allow for more efficient farming without hassle, if you're trying to farm liechi berries, you have the wacan berries to increase the growth speed but want to switch to passho berries when the berries would need 6 hours left to grow, without boost. This change would allow you to see when you need to change it without needing to do the math. 

**Detailed Description**
I am suggesting it as an option to turn on/off, as if it was just added and was on, it would probably confuse new players. just making it a setting would allow those who don't care as much for efficiency to just not have to deal with it.

osherwott:
Only difference from jk's request is "altered from XX:XX" instead of "boosted from XX:XX" due to some growth auras slowing down growth times

Closes #3336


src/scripts/farming/Plot.ts
```diff
@@ -95,6 +97,28 @@ class Plot implements Saveable {
...
+                return '';
+            }
+            let timeLeft = 0;
+            if (this.age < this.berryData.growthTime[3]) {
+                timeLeft = Math.ceil(this.berryData.growthTime[3] - this.age);
+            } else {
+                timeLeft = Math.ceil(this.berryData.growthTime[4] - this.age);
+            }
+            return GameConstants.formatTime(timeLeft / 1);
+        });
```
Aegyo, line 120: Can you refactor these with the regular ones above so we don't have duplicated code?
For instance, you could have a `calcFormattedTimeLeft = (includeGrowthMultiplier: boolean) => {` which can be called in the regular and base computeds with true/false depending on if they want to use the growth multiplier or not. Just saves us having to modify two places in the future if anything changes.

osherwott: >`calcFormattedTimeLeft = (includeGrowthMultiplier: boolean) => {`

Would that be instead of my added calculation on line 100 or the original on line 86? And how would I go about formatting a boolean setting? I'm having trouble visualizing how I could include the data and accompanying text in the same function and result line, if that's what you're describing.

Aegyo: There are currently these two functions in the code with your changes
```ts
        this.formattedTimeLeft = ko.pureComputed(() => {
                return '';
            }
            let timeLeft = 0;
            } else {
            }
            const growthMultiplier = App.game.farming.getGrowthMultiplier() * this.getGrowthMultiplier();
            return GameConstants.formatTime(timeLeft / growthMultiplier);
        });

        this.formattedBaseTimeLeft = ko.pureComputed(() => {
                return '';
            }
            let timeLeft = 0;
            } else {
            }
            return GameConstants.formatTime(timeLeft / 1);
        });
```

They are identical except for the one line defining a growthMultiplier, and the last line where one uses and the other just hardcodes a value of 1.

Instead, you should do something like this
```ts
        this.calcFormattedTimeLeft = (includeGrowthMultiplier: boolean) => {
                return '';
            }
            let timeLeft = 0;
            } else {
            }
            const growthMultiplier = includeGrowthMultiplier
                ? App.game.farming.getGrowthMultiplier() * this.getGrowthMultiplier()
                : 1;
            return GameConstants.formatTime(timeLeft / growthMultiplier);
        });

        this.formattedTimeLeft = ko.pureComputed(() => {
            return this.calcFormattedTimeLeft(true);
        });

        this.formattedBaseTimeLeft = ko.pureComputed(() => {
            return this.calcFormattedTimeLeft(false);
        });
```

and something similar for the StageTimeLeft functions.

osherwott:
Condensing the rest of the code was harder than I thought. This much will suffice.
Ready to merge!

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3706, opened 2023-01-04
Arrows for BerryDex Modal

osherwott:
Assistance is welcome!
To do:
- make unobtained berries skippable
- skip the "None" berry while going backwards
(able to with
``click: () => (App.game.statistics.selectedBerryID() + GameHelper.enumLength(BerryType) - 1) % GameHelper.enumLength(BerryType) - 1)``
as that skips to two berries over, so that might give you some ideas)

And once this is done maybe we can add arrows to pokemon too!

### DIFF
--- a/src/components/berryDexModal.html
+++ b/src/components/berryDexModal.html
@@ -8,6 +8,10 @@
                     <span aria-hidden="true">&times;</span>
                 </button>
             </div>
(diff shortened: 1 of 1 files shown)


Jaaslet:
Can we get a description of what arrows are, and what they are trying to solve?

PR #3710, opened 2023-01-05
Use gainPokemonByName instead of ID in our rewards functions

RedSparr0w:
Switch to using `gainPokemonByName` instead of `gainPokemonByID` in our reward functions, as less likely to encounter bugs.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3712, opened 2023-01-05
Enhanced Vitamin Management

CypherX:
Adds a way to easily manage vitamins for all pokemon from a single place. There is a new button labeled "All Vitamins" below the vitamin list in the inventory that will open this new modal. Clicking each individual vitamin in the inventory will still open the current modal.

Clicking/tapping the Plus (+) and Minus (-) buttons in the grid will add or remove vitamins for the pokemon in that row. Green signifies that vitamins can be added or removed and red signifies that vitamins cannot be added or removed (no vitamins remaining or at max vitamins)

image
image

The `Remove All xx` options all display a confirmation dialog.

### DIFF
--- a/src/components/itemModal.html
+++ b/src/components/itemModal.html
@@ -73,7 +73,7 @@
                                     </div>
                                 </div>
                             </div>
-                            
+
                             <!-- Amount buttons -->
                             <div class="col-12 pt-3">
                                 <div class="row justify-content-center">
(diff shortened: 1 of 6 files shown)


RedSparr0w:
Made a small change, just added the `table-responsive` class around the tables, so they can scroll horizontally on mobile, without the whole modal body also scrolling.

But the buttons on the All Vitamins modal are still overflowing:
image

Probably also a good idea to split up the filters to maybe 2 per row on smaller screens.

CypherX:
Added `toLocaleString()` back to the breeding efficiency value:
image

CypherX:
Everything should nicely stack now on smaller screens.
image

CypherX:
Also tweaked the plus/minus display so instead of red or green the unusable button(s) will simply be invisible. If people prefer the other way we can change it back.
image

RedSparr0w:
I personally like it with the `-/+` always visible, but fine with either option.
Could maybe try `text-muted` instead?

CypherX:
Good call - looks good to me.

image

RedSparr0w (approved):

PR #3715, opened 2023-01-05
Changed some functions to pure computed, Added more lazy loads

RedSparr0w:
Attempting to reduce some lag.

Changed a few of the `PartyPokemon` params to be `pureComputed`.
Added `LazyLoad` to a few of the list of Pokémon.
Added `reset` option to `LazyLoad`, this should be an observable, whenever this value changes, the list will reset to only 1 page displayed.
Removed the "Include EV" toggle in breeding efficiency as it should always be enabled due to always directly affecting the efficiency.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3721, opened 2023-01-06
PokemonStatiticsType -> PokemonStatisticsType

LegitSi:
Whoops, you missed an S. 11 times, as a matter of fact. Someone probably noticed this, but figured it was too low on the priority list to care. I decided to go ahead and fix it now.

And if I've just ruined an ongoing joke amongst the developers, whoops. All good things come to an end though, right?


Jaaslet (approved):
Looks good. Thanks for the pr

PR #3728, opened 2023-01-07
Added Aerodactylite to the underground

Jaaslet:
An other mega! Yay!
It's rare, and you can only find one!


Jaaslet:
Added two more mega stones to the underground, because we have like 48 we need to get through, and this is a good enough idea for more than one

RedSparr0w (approved):

PR #3729, opened 2023-01-07
Cycling keyboard shortcut, dungeon edition + fixes on Ruby Path

### LINKED ISSUE #3705: [SUGGESTION]: Make Alola make more user friendly
### Current Behavior

Currently if you do not have quest or some kind of map color sitting on a dungeon, they are almost impossible to find. The maps look great but it takes forever to find some dungeons.

### Improved Behavior

Do something to make dungeons more obvious, rather an outline, glow, etc to make it easier to locate alola dungeons. Adding a specific hotkey to navigate to the next or previous dungeon like routes could be another solution.

### Reason for change

Makes navigating and locating dungeons easier on the player.

### Detailed Description

Locating dungeons in Alola sucks. If you have caught all pokemon, completed all achievements, EV'd all dungeons, locating dungeons in alola is a huge pain due to how well they blend into the map. Adding something the make dungeons more obvious or a way to navigate to the next dungeon via hotkeys would make the region more user friendly.

RegisCoaxans:
Closes #3705, maybe.

Added the ability to cycle through the dungeons of a region.

Fixed Ruby Path being disposed wrong on the map.


Jaaslet (changes requested):
The routes only cycles through unlocked routes, which makes it a bit smoother. Other than that, it looks fine

RegisCoaxans:
It does not seem to even in dev version. A notification just pops and tells the requirement for the route that we try to access.
Should I change both to ignore locked routes and dungeons ?

Jaaslet:
Thoughts i had made it like that...
Yeah, please fix both :)

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3737, opened 2023-01-08
Twisted repel aura

### LINKED ISSUE #3714: [Bug]: Repel aura displaying with 0%
### Confirmations

### What happened?

The external auras is showing repel effect even when having no micles, repel aura

When there is no micles on the farm, it remains hidden

### Steps to Reproduce

Go to farm and see the external auras on the bottom left corner to see it being displayed

windows 10

Chrome

### Relevant log output

_No response_

### Additional Information

_No response_

RegisCoaxans:
Closes #3714.

I changed additive aura's base value from 0 to 1 as they are supposed to increase values of already existing mechanics.


Aegyo:
Is it not confusing to have separate Attract and Repel aura numbers?
Will people understand what an `External Auras: Attract 2x, Repel 0.4x` actually means for their wander rates?

Farboo171:
Different aura types were needed because as a multiplicative aura, Attract can never zero out wanderers. That's why it was made an additive, separate aura

Aegyo:
Right, but you could still hide that from the user by saying
```ts
this.externalAuras[AuraType.Attract] = this.multiplyAuras(AuraType.Attract) * this.addAuras(AuraType.Repel)
```

RegisCoaxans:
Since we are at it; there still are some things odd : nothing tells the users that this aura is additive, in fact, it even gets displayed as multiplicative on the plot tooltip...

osherwott:
A simple solution would be to make a single mature Micle's aura multiply the wander rate by 0. I don't see anyone who deliberately plants one wanting to limit the wander rate any less than that (if someone wants to repel all pokemon they will plant the full amount of 4 Micles at the same time anyway), and having the current growth stage auras as additives seems to cause more trouble than its worth.

Farboo171:
OK, so I think I have a solution: Percentages
If we display the Repel amounts as percentages instead of '#.#x', that could clear the confusion and is a bit easier to see as additive.

Each berry would say "Repel: 30%"
The tool top could display Repel only for values over 0%
If the tooltip says something like "Repel: 120%" it is pretty easy for a player to understand that's more than 100%, so things are totally repelled.

(Right now the PR is giving some confusing data about the berry emitting a negative aura but the total effect being a positive number. I get it but it might be a bit hard to understand)

RegisCoaxans:
Percentages will not work because of the difference between multiplicative and additive auras not being made : while Micle displays 30%, Passho, for instance, displays 130%, which seem more confusing than values around 1, especially if the players have to find out what is multiplicative and what is not.
I will stand by Jessica's side and say that a short term solution is to hide repel behind attract in the external aura tooltip (and revert the negative thing, of course).
A long term solution would be a proper implementation for us to handle both additive and multiplicative auras view-wise and model-wise if we are going to add more additive auras.

Farboo171:
I meant only applying percentages to the Repel aura for displays, not all the auras.

RegisCoaxans:
Oh, well, I misunderstood, my bad. I guess that it is doable, then. Especially if all additive auras are handled that way now and after.

Aegyo:
Having the berries say +30% would make the additive nature even clearer

RegisCoaxans:
Bip boup. I finally had time to work on that.
- Previous changes are undone.
- Micle displays its aura as a percentage and its bonus is not written as a multiplicative aura.
- The external auras tooltip changes accordingly.
- Micle aura being nerfed by Lum is fixed as well.

Aegyo:
I'm not sure about the `×`s being moved to before the numbers, it doesn't seem like a natural way to write multipliers to me

RegisCoaxans:
I did it because of consistency with the `+` being added. Plus, the `×` at the end looked weird to me now that it is the actual sign, lol.

RegisCoaxans:
image
image
image

Aegyo:
`×` before the number looks fine :+1: 

However it was mentioned on discord that 30.00% is technically two extra digits of precision compared to ×1.30 - can we have the percentages displayed with no decimal places instead?

Aegyo:
Oh wait I misread how people voted, the decimal places are fine ignore my last comment

src/scripts/farming/Farming.ts
```diff
@@ -1610,9 +1610,9 @@ class Farming implements Feature {
     }
 
     addPlotAuras(auraType: AuraType): number {
-        return this.plotList
+        return Math.max(0, this.plotList
```
Aegyo, line 1613: Why this `Math.max`?

RegisCoaxans: It remains from my original idea and followed Eugene's usage of the function. It is useless as of now.

Aegyo: I don't think we should have it - it isn't needed for Micle to work and would prevent negative auras being summed if we ever add one.

Aegyo (approved):

PR #3740, opened 2023-01-08
Added optional underground items list module

Malayena:
Added an optional module below Pokémon list to have a direct access to underground items list


Jaaslet:
Can you explain why you think this is useful? Are underground items important enough to be on the main screen? And why are they more important than all the other items?

Malayena:
> Can you explain why you think this is useful? Are underground items important enough to be on the main screen? And why are they more important than all the other items?

The base idea was to monitor my shard progress as they can be farmed in dungeons. I could have stopped, moved to a shop or openned the underground to the right tab and scrolled but this was breaking my momentum with the dungeons.
As I was sharing my idea to have this shard list, people rightfully suggested to include every kind of underground items.
Why not other items ? Because they are already listed in the place they are used and don't get farmed as easy or need to be farmed a lot before being used.
Berries barely drop from pokemons so it has little value to list them directly on the main screen.
Battle items are already there.
Hold and evolution items are also barely ever dropped and only need little monitoring in my opinion.

I understand that the module isn't useful for everyone or at all times and this is why it comes with a setting to be hidden.

Jaaslet:
I am going to close this PR. I have some ideas for a more general item modal, that i have wanted to do for a long time. I have found a way to implement a better solution for your problem, in that modal.
Sorry for waiting this long. I did not know what to do with this PR

PR #3741, opened 2023-01-08
Making docks more visible on the maps

RegisCoaxans:
Following a convo on Discord, I made docks and the S. S. Anne flash.
That happens when clicking on the map image (oceans, mountains, ...), clicking on it again or clicking on any dock stops it all.


Jaaslet:
Can you explain what problem this PR is trying to solve?

RegisCoaxans:
Docks sometimes are kind of hard to see on the maps. The flash makes them obvious to locate.

Jaaslet:
Are players clicking randomly on the maps, when they can't find the dock? It does not feel super intuitive that clicking randomly on the map should highlight the docks

RegisCoaxans:
They might. They may rather eventually click on the map by accident or on purpose and find out by themselves through playing.

This does not look that intuitive, but a new button did not seem to fit the container. I could see a keyboard shortcut works well, altough I am not sure about how necessary using a key for that is.

Now that I think about it, the map tooltip could inform the player as well.

What is your opinion ?

Jaaslet:
Could we solve the problem by adding a big ship to the docks like with kanto?

DraKay:
Not sure if all maps have space for big ship...

RegisCoaxans:
Well, the blinking had the advantage of being easily spotted.

Some places can not stand a horizontal ship. Unova can not have anything as the docks are right at the edge of the map, there is not so much water to look at, though.

Jaaslet:
I don't have a good solution, but i really don't like this one. It will trigger a lot, just by players missclicking stuff. I think a lot of players will find it annoying

RegisCoaxans:
Do you like better to hide it behind an accessibility setting ? Default would be `false`, of course.

Jaaslet:
Sure. I am hoping to find a better solution, but hiding it behind a setting will be fine, until i do

osherwott:
Is there a reason this is preferred over adding it to the shortcut list with the Farm, Mart, and Underground?

osherwott:
> Overlapping map functionnalities sucks.

I see. I've always thought it would make it easier for people on mobile or smaller screens, especially on more crowded maps. Personally I think the location of a big button is easier to memorize than all the docks across all the regions and sub-regions. More overlapping choices never hurt anyone, but your PR definitely seems helpful! Hopefully the dock is easier to spot now.

DraKay:
Dock shortcut have been rejected before, that's why martius is using another solution

RegisCoaxans:
Upon further investigations, I like better the idea for a ship, but not everywhere.
Some places might need a vertical one, though.
Maps which, I think, would need boat: Sevii 4567, Hoenn maybe, Akala Island (vertical), Ula'Ula Island (vertical), Poni Island (vertical) and Magikarp Jump (vertical as well).

Leafsw0rd:
Would it not be possible to just permanently highlight docks as a different colour, or no colour at all? 
Right now they inherit the colour of their attached city, which helps them blend in with the rest of the background. But when not highlighted, they have a fairly distinct cream/light wood colour and barcodey pattern to them.

RegisCoaxans:
Closing as this implementation sucks. The new idea is to add ships next to barely distinguishable docks and someone will take care of it eventually.

PR #3742, opened 2023-01-08
Now hiding that Feebas is a magikarp

Jaaslet:
PokemonItem can now use displayName, and using it for the feebas/magikarp


src/scripts/items/PokemonItem.ts
```diff
@@ -66,7 +66,7 @@ class PokemonItem extends CaughtIndicatingItem {
     }
 
     get displayName(): string {
-        return this._translatedName();
+        return this._displayName ?? this._translatedName();
```
RedSparr0w, line 69: We should probably still be showing the `_translatedName` here, and default that to using `_displayName ?? pokemon`

Jaaslet: fixed

RedSparr0w (approved):

PR #3743, opened 2023-01-08
Only apply time traveler flag if it's a new day

RedSparr0w:
Hopefully solve the issue of some people getting the flag applied incorrectly.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3744, opened 2023-01-08
Update FAQ to explicitly say that EV stands for Effort Value

Farboo171:
Right now, the game doesn't say that EVs are effort values. This will fix that.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3746, opened 2023-01-09
Added a checkbox to hide pokémon already holding an item in held item… (Closes #3719)

### LINKED ISSUE #3719: [SUGGESTION]: Add a toggle to, when equipping held items, to hide pokemon equipped with held items
### Current Behavior

Game does not have the ability to hide pokemon that have equipped held items in the list when equipping held items.

### Improved Behavior

Add a toggle similar to the toggle in the vitamins modal.

### Reason for change

Makes it easier to equip held items. Less scrolling past pokemon already equipped with held items, especially if people plan to equip every pokemon with items. With over 1k pokemon, this would surely become a chore.

### Detailed Description
The toggle would be either be between the search bar and the table or between drop down menu and search bar.

Vodovik:
… modal

looks like this
image


src/scripts/party/PartyPokemon.ts
```diff
@@ -359,6 +359,10 @@ class PartyPokemon implements Saveable {
         if (!new RegExp(Settings.getSetting('heldItemSearchFilter').observableValue() , 'i').test(this.displayName)) {
             return true;
         }
+        if (Settings.getSetting('heldItemHideHoldingPokemon').observableValue() && typeof(this.heldItem()) !== 'undefined') {
```
CypherX, line 362: `&& typeof(this.heldItem()) !== 'undefined'` can just be `&& this.heldItem()`

Vodovik: ah thanks, i wasn't sure how undefined would work so my old JS ways recalled this existing x)

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3750, opened 2023-01-09
Changed evolution item of Spooky Togetic from Shiny Stone to Dusk Stone.

LuchoDreamer:
Pant gave the idea that Spooky Togetic should evolve with a Dusk stone instead of a Shiny stone, since it kind of makes sense because Spooky Togetic is part ghost type, being its counterpart part fairy type.

Also there are many more pokemon that evolve from Shiny stone than Dusk stone, this would balance them more a bit.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3752, opened 2023-01-09
Exclude mega stones from daily deals

Jaaslet:
No trading for mega stones!


RedSparr0w (approved):

PR #3754, opened 2023-01-09
Add NPC for Feebas Salesman

Farboo171:
Adds an NPC that points in the direction of the Feebas NPC, so folks can find it without asking the discord or wiki if they read NPCs.


Jaaslet (approved):
Yeah, this is maybe fair... Thanks for the PR

PR #3756, opened 2023-01-10
Underground, mega stones, daily deals

RedSparr0w:
#3752 unintentionally changed some of the underground daily deal trades,
This PR resolves that, and still excludes Mega Stones from underground daily deal trades,


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #3578, opened 2022-12-13
[SUGGESTION]: Quest Pokemon's silhouettes should show after an NPC mentions them


InQuognito:
### Current Behavior

Silhouettes for certain pokemon (such as legendaries) are hidden until they are encountered.

### Improved Behavior

Silhouettes should now show in the pokedex list after an NPC mentions the Pokemon for the first time. Take, for instance, the questline with Celebi or Lugia. The girls say something like "We call upon Lugia" so the player now knows of that Pokemon's existence, even if they don't know what it looks like. They at least know it needs a slot in the pokedex at that point. Same with Celebi, when Oak is like "I used to play with that Celebi", etc.

### Reason for change

Seems more realistic and immersive, and also helps new players with their checklist of which Pokemon remain to be caught.

### Detailed Description

![image](https://user-images.githubusercontent.com/44120299/207418928-5723aa14-310e-4c98-a7f6-e3eaf15f0baf.png)
Note how, even though my questlines are almost complete, Celebi and Lugia remain hidden in the pokedex.

CooldudeOmega:
I think I had suggestion way back when but with the roamers, makes sense.

CooldudeOmega:
#1878
or https://github.com/pokeclicker/pokeclicker/issues/1878
im not sure which one works better

PR #3758, opened 2023-01-10
Added seen count (if never encountered) on roamer npc dialog

### LINKED ISSUE #1787: [SUGGESTION] New Event - Godzilla/Kaiju Themed
Godzilla/Kaiju Event

**Current Behavior**
Kaiju Event doesn't exist

**Improved Behavior**
Adding Kaiju Event

**Reason for change**
Just thought the idea would be a fun event

**Detailed Description**
Reskinned Pokémon:
Hydreigon with a gold palette for King Ghidorah
Tyranitar with a green/black palette for Godzilla
Larvitar could also be a Minalla/Baby Godzilla
Crawdaunt with little change in design for Ebirah
Muk with a black/red palette for Hedorah
Moltres or Fearow with a dark red palette for Rodan
Rhyhorn as Anguirus
Mothim or Butterfree as  Mothra
My thoughts for which should be quests/wandering are maybe if they can fly canonically they are wandering and if they can't they get a quest

Time Limit: Maybe around the anniversary of when the OG godzilla released with around 2 weeks of time

### LINKED ISSUE #1878: [SUGGESTION]: Make the roamer NPCs add any roamers shown to your Pokédex
**Current Behavior**
Roamer NPC shows you the Pokémon sprites but the Pokémon shown do not get added to your Pokédex.

**Improved Behavior**
Make the roamer NPCs add any roamers shown to your Pokédex if they aren't already there.

**Reason for change**
It would be more convenient to have the Pokémon added to your Pokédex as it will show you more Pokémon in the dex, though to be honest, this isn't particularly useful.
And it's just cool.
**Detailed Description**
The roamer NPC would now show you the route with increased roamer odds but would also add the roamers shown to your Pokédex.

**Screenshots:**

before
after

Vodovik:
This is made so roamers are seen in the pokédex after talking to the npc, might not be wanted as it tweaks statistics.

Note that suggestion #3578 also wants that and more, but I haven't seen a simple way to do the same with the quest npc dialog.

Closes #1878


Jaaslet:
This solution will lead to too many questions in discord. If we want to do this, we need to have a "seen" boolean saved somewhere, and use that one.

Vodovik:
So you're telling me to add a new statistic? cause i'm willing, but if you don't find it worth it, which i understand, then we can close the PR. or are you telling me to do something else ?^^

Jaaslet:
Add a new one. But do so it's only saved, if the pokemon is never encountered. To make the save file snaller

Vodovik:
Tried doing what i understood, thanks for taking the time! I hope i didn't declare any useless property ^^'

src/modules/pokemons/PokemonHelper.ts
```diff
@@ -114,6 +114,7 @@ export function incrementPokemonStatistics(pokemonId: number, statistic: Pokemon
     const pokemonStatistics = {
         Captured: App.game.statistics.pokemonCaptured[pokemonId],
         Defeated: App.game.statistics.pokemonDefeated[pokemonId],
+        Seen: App.game.statistics.pokemonSeen[pokemonId],
```
Aegyo, line 117: Without any gender/shiny/total variants, I don't think it makes sense to have this in this function. Just increment it manually from the RoamerNPC instead.

CooldudeOmega:
My Issue is being added after almost one year? That’s pretty cool

Aegyo (approved):

PR #3764, opened 2023-01-12
Time Flute now increases catching speed too

RegisCoaxans:
I felt like the Time Flute was kind of weak compared to the other flutes, so I improved it.

It is sometimes used to pass a wall, if you are not willing to breed, and that is all... few minutes each day, lol.
The other flutes have long term interesting and competitive effects. Even the black flute can be used for much longer periods of time...

On the other hand, there is nothing affecting the catching speed, somehow with good reasons.

So, now, it also reduces the catch time by up to ~18% with maximal achievement bonus. That is not much so it does not break anything, but still is something and hopefully will make players think about it more often after they are done with badges.


Jaaslet:
We already think time flute is strong enough, and don't need to do more stuff.
I will close this PR

PR #3765, opened 2023-01-12
Add LocationDex

Dylan-Brotherston:
There are a lot of locations in Pokéclicker: Towns, Dungeons, Gyms, Shops, Etc.

Apart from clicking on every location the best way of finding a specific location is to check the wiki.

This PR adds a way to search for locations within the game itself.

Only locations that the player can access are shown (to avoid spoilers and clutter).
Locations can be filtered by name, type (Towns, Dungeons, Gyms, Shops), and/or region.

Screenshot 2023-01-12 134230 (2)

Only unique shops are shown (ie no Poké Mart).


Jaaslet:
I will close this PR, as it's the wiki's job

PR #3766, opened 2023-01-12
Fish shop tweaks

Farboo171:
Moves the fish shop to a more reasonable location and changes the fish hints to point to it better, based on crew feedback


Jaaslet (changes requested):
I think it should either have an environment or a background, now that it's out in the water

Farboo171:
> I think it should either have an environment or a background, now that it's out in the water

background is now a lake from the Hoenn region, from the anime.. Different episode, but close

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3768, opened 2023-01-13
Do androids dream of electric sheep?

Qwertypop04:


src/scripts/towns/TownList.ts
```diff
@@ -4102,7 +4102,7 @@ const CoumarineBirdwatcher = new NPC('Birdwatcher', [
 const CoumarineElectricTrainer = new NPC('Electric Trainer', [
     'My Ampharos has gotten a boost since a friendly soul gave me an Ampharosite!',
     'Luckily my Ampharos was already strong enough to use it!',
-    'If you find that fella, tell him the code <b>ELECTRIC-GIRAFFE</b> and he might help you too!',
+    'If you find that fella, tell him the code <b>DO-MAGEARNA-DREAM-OF-MAREEP?</b> and he might help you too!',
```
Vodovik, line 4105: why the sudden all caps kebab when old codes don't use '-' ?
image

Qwertypop04: Dunno. That's how it already was

Jaaslet: It would be easy if the codes where predictable

Vodovik: fair enough, i did blind try that exact code without the '-' back in 2022 xD

Jaaslet: (serious answer) It would maybe be more consistent if we did it the same way, but i think the joke is easier to read with the `-`, so i added them. But yeah, we can always change them later, if we think it's too weird

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3770, opened 2023-01-13
Refactored TimeRequirement to handle timed evolutions with DayCycles …

Vodovik:
…instead of hardcoded hours


src/modules/pokemons/PokemonList.ts
```diff
@@ -21192,10 +21193,10 @@ export const pokemonList = createPokemonArray(
...
         'catchRate': 190,
         'evolutions': [
-            TimeRestrictedLevelEvolution(6, 17, 'Rockruff', 'Lycanroc (Midday)', 25),
-            TimeRestrictedLevelEvolution(17, 18, 'Rockruff', 'Lycanroc (Dusk)', 25),
-            TimeRestrictedLevelEvolution(18, 5, 'Rockruff', 'Lycanroc (Midnight)', 25),
-            TimeRestrictedLevelEvolution(5, 6, 'Rockruff', 'Lycanroc (Dusk)', 25),
+            TimeRestrictedLevelEvolution(DayCyclePart.Day, true, 'Rockruff', 'Lycanroc (Midday)', 25),
+            TimeRestrictedLevelEvolution(DayCyclePart.Night, true, 'Rockruff', 'Lycanroc (Midnight)', 25),
+            TimeRestrictedLevelEvolution(DayCyclePart.Dusk, true, 'Rockruff', 'Lycanroc (Dusk)', 25),
+            TimeRestrictedLevelEvolution(DayCyclePart.Dawn, true, 'Rockruff', 'Lycanroc (Dusk)', 25),
```
Vodovik, line 21199: this changes the hours from 5-6 to 6-5. Might not be wanted, dawn isn't used anywhere else so we could move it to 5 AM (instead of 6). this way dawn is part of the night and Dusk is part of the day. Which would make this (to get current, replace dawn but : 6 -7)

Dawn : 5 - 6
Day:  6 - 18
Dusk: 17 - 18
Night: 18 - 5

Aegyo:
You have a conflict here

src/modules/requirements/TimeRequirement.ts
```diff
@@ -1,33 +1,26 @@
...
-            // If the start time is after the end time, only 1 needs to be true
-            : currentHour >= startHour || currentHour < endHour;
-
-        return Number(satisfied);
+        return Number(this.dayCycleParts.includes(DayCycle.currentDayCyclePart()));
     }
 
     public hint(): string {
-        return `Your local time must be between ${this.startHour}:00 and ${this.endHour}:00`;
+        return `Your local part of the day must be ${this.dayCycleParts.map((dayCyclePart) => DayCyclePart[dayCyclePart]).join(' or ')}`;
```
Aegyo, line 24: This feels less informative, unless there is somewhere telling people exactly what time the dayCycleParts start/end?
If there isn't, can we do that somewhere, or change this hint to give the actual times (either in addition or instead of the day part name)

Vodovik: There is a display of the part of the day now, but that doesn't allow knowing when the part starts. Maybe should I add an NPC? that describes their day : "i wake up at dawn, at 6 AM every day" and so on ?

src/modules/GameConstants.ts
```diff
@@ -1818,3 +1818,7 @@ export enum ExtraAchievementCategories {
     sevii,
     magikarpJump,
 }
+export enum DayCycleStartHours {
+    dawn = 6,
+    dusk = 17,
+}
```
Aegyo, line 1824: Make this a `Record<DayCyclePart, number>` and fill it in for all DayCycleParts, so we aren't relying on dusk and dawn being 1 hour long.

src/modules/dayCycle/DayCycle.ts
```diff
@@ -2,18 +2,19 @@
...
         let dayCyclePart: DayCyclePart;
 
-        if (curHour === 17) {
+        if (curHour === DayCycleStartHours.dusk) {
             dayCyclePart = DayCyclePart.Dusk;
-        } else if (curHour === 6) {
+        } else if (curHour === DayCycleStartHours.dawn) {
             dayCyclePart = DayCyclePart.Dawn;
-        } else if (curHour > 6 && curHour < 17) {
+        } else if (curHour > DayCycleStartHours.dawn && curHour < DayCycleStartHours.dusk) {
```
Aegyo, line 17: When you have filled in all DayCycleStartHours, this if/else can be replaced with finding the smallest DayCycleStartHour which is greater than the current hour, or the last DayCyclePart if none was found.

src/modules/dayCycle/DayCycle.ts
```diff
@@ -35,12 +36,12 @@ export default class DayCycle {
     });
 
     public static dayCycleMoments: { [dayCycle in DayCyclePart]?: DayCycleMoment } = {
```
Aegyo, line 38: Is there a reason a Record wasn't used here?
```suggestion
    public static dayCycleMoments: Record<DayCyclePart, DayCycleMoment> = {
```

src/modules/dayCycle/DayCycleMoment.ts
```diff
@@ -9,7 +9,21 @@ export default class DayCycleMoment {
 
     get tooltip(): string {
         const tooltip = [];
-        tooltip.push(this.description);
+        const dayCycleSchedule: Array<string> = [
+            'Dawn : 06:00 - 07:00',
+            'Day: 07:00 - 17:00',
+            'Dusk: 17:00 - 18:00',
+            'Night: 18:00 - 06:00',
+        ];
```
Aegyo, line 17: Should not be hardcoded, use DayCycleStartHours to generate it

src/modules/dayCycle/DayCycleMoment.ts
```diff
@@ -9,7 +9,21 @@ export default class DayCycleMoment {
...
+
+        dayCycleSchedule.forEach((schedule, dayCyclePart) => {
+            if (dayCyclePart === this.type) {
+                tooltip[dayCyclePart] = (`<b>${schedule}</b>`);
```
Aegyo, line 21: The bold doesn't stand out very well to me (maybe just because of dark theme?), could we get an icon or something in front of the current part as well?

src/modules/pokemons/evolutions/Restrictions.ts
```diff
@@ -89,21 +90,20 @@ export const weatherRestrict = <T extends EvoFn>(evo: T) => (
 );
 
 export const timeRestrict = <T extends EvoFn>(evo: T) => (
-    startHour: number,
-    endHour: number,
+    dayCycleParts: DayCyclePart[],
```
Aegyo, line 93: This should be renamed to `dayCyclePartRestrict`

src/modules/requirements/TimeRequirement.ts
```diff
@@ -1,33 +1,26 @@
 import { AchievementOption } from '../GameConstants';
 import GameHelper from '../GameHelper';
 import Requirement from './Requirement';
+import DayCyclePart from '../dayCycle/DayCyclePart';
+import DayCycle from '../dayCycle/DayCycle';
 
 export default class TimeRequirement extends Requirement {
     private updateTrigger = ko.observable(0);
-    constructor(public startHour: number, public endHour: number, option = AchievementOption.more) {
+    constructor(public dayCycleParts: DayCyclePart[], option = AchievementOption.more) {
```
Aegyo, line 9: Rename to `DayCyclePartRequirement`

src/modules/requirements/TimeRequirement.ts
```diff
@@ -1,33 +1,26 @@
...
         super(1, option);
         setInterval(
             () => GameHelper.incrementObservable(this.updateTrigger),
             60 * 1000,
         );
```
Aegyo, line 14: There shouldn't be any need for this setInterval or the updateTrigger property if the `DayCycle.currentCyclePart` observable is updating properly

src/components/FAQModal.html
```diff
@@ -111,6 +111,41 @@ <h5>Frequently Asked Questions</h5>
...
+                    <td><center><strong>Icon</strong></center></td>
+                    <td><center><strong>Start</strong></center></td>
+                    <td><center><strong>End</strong></center></td>
+                </tr>
+                <tr>
+                    <td><center><span class="text-muted text-center">Dawn</span></center></td>
+                    <td><center><span class="text-muted text-center"><img style="background: #25b6a0; width:30px;" src="assets/images/dayCycle/Dawn.png"/></span></center></td>
+                    <td><center><span class="text-muted text-center">06:00</span></center></td>
+                    <td><center><span class="text-muted text-center">07:00</span></center></td>
+                </tr>
```
Aegyo, line 128: All of these `<tr>`s should be generated from DayCycleStartHours

Aegyo (changes requested):
A few things I think need changing, mostly to make adding new DayCycleParts in the future easier

src/modules/dayCycle/DayCycle.ts
```diff
@@ -2,24 +2,14 @@
...
-        } else if (curHour === 6) {
-            dayCyclePart = DayCyclePart.Dawn;
-        } else if (curHour > 6 && curHour < 17) {
-            dayCyclePart = DayCyclePart.Day;
-        } else {
-            dayCyclePart = DayCyclePart.Night;
-        }
-
-        return dayCyclePart;
+        return Number(Object.entries(DayCycleStartHours).reverse().find(([, startHour]) => startHour <= currentHour)?.[0] ?? DayCyclePart.Night);
```
Aegyo, line 12: Just in case Night isn't always the last one,
```suggestion
        return Number(Object.entries(DayCycleStartHours).reverse().find(([, startHour]) => startHour <= currentHour)?.[0] ?? Object.keys(DayCycleStartHours).slice(-1));
```

src/modules/dayCycle/DayCycleMoment.ts
```diff
@@ -1,15 +1,28 @@
...
+
+            if (DayCyclePart[dayCyclePart] === DayCyclePart[this.part]) {
+                tooltip.push(`<span class="text-success"><b>${DayCyclePart[dayCyclePart]}: ${GameHelper.twoDigitNumber(startHour)}:00 - ${GameHelper.twoDigitNumber(endHour)}:00</b></span>`);
+            } else {
+                tooltip.push(`${DayCyclePart[dayCyclePart]}: ${GameHelper.twoDigitNumber(startHour)}:00 - ${GameHelper.twoDigitNumber(endHour)}:00`);
+            }
+        });
+
         return tooltip.join('<br>');
     }
```
Aegyo, line 27: This can be condensed a bit to
```ts
    get tooltip(): string {
        const dayCycleStartHours = Object.entries(DayCycleStartHours);
        return dayCycleStartHours.map(([dayCyclePart, startHour], index) => {
            const [, endHour] = dayCycleStartHours[index + 1] || dayCycleStartHours[0];
            const content = `${DayCyclePart[dayCyclePart]}: ${GameHelper.twoDigitNumber(startHour)}:00 - ${GameHelper.twoDigitNumber(endHour)}:00`;

            return DayCyclePart[dayCyclePart] === DayCyclePart[this.part]
                ? `<span class="text-success"><b>${content}</b></span>`
                : content;
        }).join('<br>');
    }
```

src/modules/requirements/TimeRequirement.ts
Aegyo: This was talking about the class name, sorry I wasn't clear. It doesn't take time inputs anymore, we are restricting by DayCycleParts, so the class name should reflect that

src/modules/pokemons/evolutions/Restrictions.ts
Aegyo: This was talking about the `const timeRestrict`, since it is restricted by DayCycleParts now and the name should match

Vodovik: that's what i thought at first,  but then i thought the line was meaningful ^^ not easy to place that in a review : )

src/modules/requirements/TimeRequirement.ts
Vodovik: that's what i thought at first, but then thought the line was meaningful, like for timeRestrict :^^ not easy to place that in a review : )

Vodovik:
@Aegyo should be good to go : )

src/modules/requirements/DayCyclePartRequirement.ts
```diff
@@ -0,0 +1,21 @@
...
+    private updateTrigger = ko.observable(0);
+    constructor(public dayCycleParts: DayCyclePart[], option = AchievementOption.more) {
+        super(1, option);
+    }
+
+    public getProgress(): number {
+        this.updateTrigger();
+
+        return Number(this.dayCycleParts.includes(DayCycle.currentDayCyclePart()));
+    }
```
Aegyo, line 16: I checked everything twice looking for one last thing to complain about, and I got a bite:

`updateTrigger` isn't needed, you got the setInterval but these slipped through
```suggestion
    constructor(public dayCycleParts: DayCyclePart[], option = AchievementOption.more) {
        super(1, option);
    }

    public getProgress(): number {
        return Number(this.dayCycleParts.includes(DayCycle.currentDayCyclePart()));
    }
```

Vodovik: oops right, let's try that "commit suggestion" button

Vodovik:
thanks for the vigilance and suggestions : ) @Aegyo

Aegyo (approved):
:tada:

PR #3776, opened 2023-01-14
Fixed computed not being refreshed.

Vodovik:
@Jaaslet this should be part of the next releases a it fixes a bug : )

### DIFF
--- a/src/modules/dayCycle/DayCycle.ts
+++ b/src/modules/dayCycle/DayCycle.ts
@@ -2,10 +2,11 @@
 import { Computed } from 'knockout';
 import DayCycleMoment from './DayCycleMoment';
 import DayCyclePart from './DayCyclePart';
+import GameHelper from '../GameHelper';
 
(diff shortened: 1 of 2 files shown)


Jaaslet:
It would make it easier to review if you explain what bug

Malayena:
> It would make it easier to review if you explain what bug

The hour was calculated once on start-up and never refreshed due to new Date() not being observable. The game needed to be restarted to show the correct new moment of the day.
Using GameHelper.currentTime(), which is observable, fixes that.

Vodovik:
sorry, i created the bug on discord. The new time of day icon isn't refreshed when you enter a new time part, like day => dusk, dusk => night, .. now it should be doing what malayena just explained : ) (and it is in my tests, the refresh only happens once every minute, so if you time travel to test it, you'll need to wait 1 min at least : ))

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3777, opened 2023-01-14
Fix Vitamin Modal UI issues

CypherX:
Fixes the flashing scrollbar that appears when the pokemon list is not fully loaded due to the spinning pokeball loading image. Also fixes the entire list being loaded immediately in Firefox due to the responsive table layout.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3780, opened 2023-01-14
Gardevoirite Temp Battle

Leafsw0rd:
Per discussion in Mega Stones chat, this branch gives Gardevoirite via a Temp Battle with Diantha at Lumiose City. 
This rematch uses her Battle Chateau rematch team, with levels set as their highest possible.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3781, opened 2023-01-14
Luxuryball effect now shows in pokeball selector

DraKay:
Hi again!
Is me! The maker of short pr changing a single word or sentence. This time the visible change are.... 🥁🥁 5 WORDS! Almost a record of length...
And why change that 5 words? Because now luxury ball makes something... 🙀 so selector should say more than "a luxury ball" something to indicate what it makes... You know, to don't think "oh luxury... But I'm f2p..." 

Also as bonus I edited the comment in code saying TODO for adding an effect as it have already an effect. 

And this PR is finished! (and again I took more time write than make the change) if you want more small little changes keep tuned with your favorite person (or a person that you like... Just a little...) !


Jaaslet:
The new effect is not merged yet, and the PR already contains an update for the luxury ball text.
So i will close this as duplicate

PR #3783, opened 2023-01-14
Fixup broken image

Farigh:
The image was not found in-game due to it's extension being upper-cased.

This resulted in: 
image

Instead of:
image

(at least in Firefox)


Farigh:
Duplicate of #3772

PR #3788, opened 2023-01-14
Changes the message prompt when you click the daycare before being unlocked to Route 3

LuchoDreamer:
Day Care actually unlocks after Route 3 now, not after Brock's Gym (or Route 5). This small PR fixes the message


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3790, opened 2023-01-14
Fix Maelstrom appearing after quest

Farboo171:
Fixes Maelstrom appearing after the Weather Trio quest is completed on the overworld map


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3791, opened 2023-01-15
Give object statistics a proper type

Aegyo:
Because we can, and `any` is like not using typescript at all.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3793, opened 2023-01-15
Show all stone evolutions up to the current region

Aegyo:
We already show locked evos if there are other valid ones for the pokemon, this just brings that feature to all the rest.


Jaaslet (changes requested):
I am unsure what this change is doing. The only change i can see, is that it shows the Galar evolutions, when i am only in Kalos. And i don't think that is correct (from the PR title and because i don't think it's a good idea).
image

Qwertypop04:
> I am unsure what this change is doing. The only change i can see, is that it shows the Galar evolutions, when i am only in Kalos. And i don't think that is correct (from the PR title and because i don't think it's a good idea).
> image
> 

I would say that it's better than players seeing that they have an incomplete Thunder Stone evolution when they are in Alola prior to obtaining Vikavolt and the game offering 0 explanation why. At least this way they know the requirement

Aegyo:
Forgot to filter out RegionRestricted ones, I'll add that.

The PR mostly makes time based stone evolutions visible without having to be in the valid time period, so players can see all the possible stone evolutions and why they may not currently be available.

Aegyo:
Updated to filter out options only available in a future region.
The caught + pokerus indicators on stones could also use this filtering (problem already exists on develop), but I didn't want to bloat this PR so I'll add it separately after this has been merged.

## Screenshots taken on a Sinnoh save

Thunder stone with no Magneton option:
Screenshot_2023-01-20_23-47-20


Soothe Bell showing Riolu outside of the usual time:
Screenshot_2023-01-20_23-41-35

RedSparr0w (approved):
Changes look good to me, thanks for the PR! 👍

PR #3795, opened 2023-01-15
Move RouteData to modules

Aegyo:
Other than the obvious moving of the file, I've added an index file in `modules/routes` which imports RouteData to perform the setup side effect, and re-exports Routes (and others) so that they can import a version of Routes which definitely has the RouteData set up.

Also had to add something the gulpfile declaration fiddling to stop it butchering the re-export lines and making typescript cry.

Originally had the Routes import in RouteKillRequirement use the index file, but that led to a circular dependency.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3797, opened 2023-01-15
Give Totem Marowak Thick Club

Qwertypop04:
As it is an alt form of Alolan Marowak


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3799, opened 2023-01-16
Add punctuation to Skyarrow Bridge notification

### LINKED ISSUE #3316: [SUGGESTION] Missing punctuation on Skyarrow Bridge
In the Unova regional map, there is an area you can click on called: "Skyarrow Bridge". You cannot go there, the only thing that pops up is: "What a lovely view" without punctuation. Obviously, the sentence needs a period or some other form of punctuation at the end. All other notifications that pop up on your screen (as far as I know) have punctuation at the end. For example, "You hatched an Onix!" 

I know this is little, but I felt like saying something about it anyway. 
I would just change this myself, but I don't know how to do that. 

Here's a small image of the pop-up that shows up when you click on Skyarrow Bridge: 
What a lovely view
 As you can see, no punctuation.

Aegyo:
Closes #3316

### DIFF
--- a/src/components/UnovaSVG.html
+++ b/src/components/UnovaSVG.html
@@ -852,7 +852,7 @@
         , x: 62, y: 40
         , databind: `attr: { class: MapHelper.accessToRoute(4, GameConstants.Region.unova) ? 'unlockedRoute' : 'lockedRoute' },
             click: () => {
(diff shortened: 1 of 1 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3802, opened 2023-01-16
Update fossilized birds

DraKay:
Humans, aliens and fairy tale criatures... We have a request!

Black zero asked for a more visible fossilized bird, and following our Xmas tale pr (#3628) we changes the image...

Enjoy this slightly bigger fossil but same dig


DraKay:
Remove last line as Red ask

Aegyo (approved):

PR #3803, opened 2023-01-17
Pokéball selector option for contagious pokémon

FredisonP:
This adds an option to the pokéball selector for pokémon infected by (but not resistant to) pokérus. The priority for already caught pokémon should now be Shiny > Contagious > Other, and the option should only be visible to the player once the pokérus mechanic has been unlocked.


FredisonP:
Whoops! Messed that one up. Disregard!

Issue #1712, opened 2021-12-21
Add more Pokeball type selection options


Knirunmund:
**Current Behavior**
You can only choose pokeball type to use under 4 conditions: Caught, Shiny (new), New, Shiny (old)

**Improved Behavior**
Add more categories for choosing pokeball type to use.

**Reason for change**
When trying to quickly do dailies it would be convenient/efficient to be able to only use pokeballs on caught pokemon that meet the requirement for a quest.  Having to throw a ball at all targets is both slow and expensive if you want to use higher grade pokeballs if only 1 of 8 possible pokemon meet the requirement for the quest.

**Detailed Description**
There are a couple possible solutions.  You could add an additional category that checks the current pokemon against all current quests, if it meets the requirements of atleast one it uses the specified ball.

Alternatively, or additionally, you could add an additional category that lets you manually set conditions for which pokemon to throw a ball at.  Another potentially useful condition outside of type that could be useful would be catch chance, to only throw a ball if the chance to catch is atleast x%.  It would allow more efficient dungeon token farming and completion of the dungeon token quest.

RedSparr0w:
Modified title to: `Add more Pokeball type selection options`

The whole Pokéball selection window would probably need to be modified quite a bit to accompany this though.

Maybe even if it was just an "advanced" selection mode, as for most people this amount of customisation wouldn't be needed.

Aegyo:
Requested filter options in other issues:
- Pokemon Type
- Egg steps (why?)
- Group (no idea what was meant by this, maybe hatchery category)
- EV status
- Roamer
- Dungeon Boss
- Specific Pokemon

Farboo171:
#4217  and #3837 have resolved most of this. Aegyo can close this suggestion when they have implemented all they intend to.

PR #3805, opened 2023-01-17
Pokéball selector option for contagious pokémon

FredisonP:
This adds an option to the pokéball selector for pokémon infected by (but not resistant to) pokérus. The priority for already caught pokémon should now be Shiny > Contagious > Other, and the option should only be visible to the player once the pokérus mechanic has been unlocked.


Aegyo (approved):
Any more catch filters will have to wait for #1712, but we are going to allow this one through as an exception as it seems the most useful option of all requested additional filters.

PR #3816, opened 2023-01-19
Update to damage spread graph in the damage calculator

Alderi-Tokori:
Uses a different type spread calculation for the graph if a defender type is selected. This way, the graph shows more accurately which damage type is inflicted to the defender.


Aegyo (commented):
I would prefer if TypeHelper had a method to return the best attacking type given attacker and defender types, have this new method used in the existing TypeHelper.getAttackModifier code, and also called in DamageCalculator to see which type was used.

This current code is liable to be forgotten and not be updated in case of type damage changes.

Alderi-Tokori:
> I would prefer if TypeHelper had a method to return the best attacking type given attacker and defender types, have this new method used in the existing TypeHelper.getAttackModifier code, and also called in DamageCalculator to see which type was used.

I can add a method to TypeHelper, but I don't see how I would go around calling it in TypeHelper.getAttackModifier. It would be the other way around: this new method would call TypeHelper.getAttackModifier with each of the attacking types and would return the best type.

Alderi-Tokori:
Also, if I call the "getBestType" method, it would make the "both types are equal" case trickier I think.

Alderi-Tokori:
What I can try to do, I think, is add a method to TypeHelper (something like compareAttackingTypes). It will call TypeHelper.getAttackModifier with each of the attacking types and return:

- -1 if the first type is better
- 0 if both types are equal
- 1 if the second type is better

Then, in DamageCalculator.getDamageByTypes, I will call TypeHelper.getAttackModifier once with both types (to get the normal damage calculation) and I will call TypeHelper.compareAttackingTypes to determine if I should affect all the damage to one type or the other, or affect half of it to each type.

PR #3817, opened 2023-01-19
fix: bug in underground item placement

Alderi-Tokori:
Currently in Mine.canAddReward, if y is 10 and reward.space.length is 2 (i.e. we're on the second to last row and the item is two spaces high), then the item is placeable, but current comparison is '>= 12' so the method returns false. Using strict comparison fixes this issue.

Additionally, Mine.getRandomCoord also never returns the last row/column, adding one to the call fixes that. This way, if an item is two spaces high, Mine.getRandomCoord(12 - 2 + 1) will return a number between 0 and 10.

### DIFF
--- a/src/scripts/underground/Mine.ts
+++ b/src/scripts/underground/Mine.ts
@@ -86,14 +86,14 @@ class Mine {
     }
 
     private static getRandomCoord(max: number, size: number): number {
-        return Rand.floor(max - size);
+        return Rand.floor(max - size + 1);
     }
 
     private static canAddReward(x: number, y: number, reward: UndergroundItem): boolean {
         if (Mine.alreadyHasRewardId(reward.id)) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


Aegyo (approved):
:+1: 
Changes look fine to me, also verified by running a few thousand mines
```js
async function heatmap(runs) {
  const map = Mine.rewardGrid.map(row => row.map(_ => 0));
  for (i = 0; i < runs; i++) {
    Mine.loadMine();
    Mine.rewardGrid.forEach((row, i) => row.forEach((reward, j) => reward && map[i][j]++));

    if (i % 100 === 0) {
      console.info(`Reached ${i}`);
      await new Promise(r => setTimeout(r,0))
    }
  }
  return map;
}
```
```js
heatmap(5000).then(console.log)
```
`develop` has zeroes on the end edges, and your changes fix that.

Issue #3444, opened 2022-11-22
[Bug]: Pokemon not removed from hatchery on Real Evolution challenge when evolved


BaileyP2SR:
### Confirmations

- [X] I have searched open and closed issues for any existing report of my issue

### What happened?

I placed Eevee in the hatchery, evolved it using a thunderstone, and then 1400 steps later hatched a new Eevee without needing to buy it again.

### What did you expect to happen?

Pokemon should be removed from the hatchery or not be recaptured upon hatching if it was removed due to evolving.

### Steps to Reproduce

1. Place Eevee in the hatchery
2. Evolve Eevee into Jolteon using a thunder stone
3. Hatch the egg and Eevee is now back in your party

### What operating system are you using?

Windows 10

### What browsers are you seeing the problem on?

Chrome

### Additional Information

This is very funny and was done on accident lmao

Aegyo:
Fixed by #3821

PR #3821, opened 2023-01-20
Fix: Real Evolution challenge exploit by evolving a pokémon in the hatchery

Alderi-Tokori:
Fixes issue #3444. If the real evolution challenge is active, you can't use a stone on Pokémons in the hatchery.

### DIFF
--- a/src/modules/items/ItemHandler.ts
+++ b/src/modules/items/ItemHandler.ts
@@ -52,9 +52,9 @@ export default class ItemHandler {
 
         const partyPokemon = App.game.party.getPokemonByName(this.pokemonSelected());
         if (partyPokemon.breeding && App.game.challenges.list.realEvolutions.active()) {
-            // If the real evolution challenge is active, we prevent using stones on pokemons in the hatchery to prevent exploits
+            // If the real evolution challenge is active, we prevent using stones on Pokémon in the hatchery to prevent exploits
             return Notifier.notify({
-                message: 'You can\'t use a stone on a Pokémon if it\'s in the hatchery...',
+                message: 'You can\'t use an evolution item on a Pokémon if it\'s in the hatchery...',
                 type: NotificationConstants.NotificationOption.danger,
             });
         }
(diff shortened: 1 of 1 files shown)


src/modules/items/ItemHandler.ts
```diff
@@ -49,6 +49,16 @@ export default class ItemHandler {
                 type: NotificationConstants.NotificationOption.danger,
             });
         }
+
+        const partyPokemon = App.game.party.getPokemonByName(this.pokemonSelected());
+        if (partyPokemon.breeding && App.game.challenges.list.realEvolutions.active()) {
+            // If the real evolution challenge is active, we prevent using stones on pokemons in the hatchery to prevent exploits
```
PixLSteam, line 55:
```suggestion
            // If the real evolution challenge is active, we prevent using stones on Pokémon in the hatchery to prevent exploits
```

src/modules/items/ItemHandler.ts
```diff
@@ -49,6 +49,16 @@ export default class ItemHandler {
...
+            return Notifier.notify({
+                message: 'You can\'t use a stone on a Pokémon if it\'s in the hatchery...',
```
PixLSteam, line 57:
```suggestion
                message: 'You can\'t use an evolution item on a Pokémon if it\'s in the hatchery...',
```
I believe we refer to all evolution items as stones internally but there are other kinds of evolution items and this message could be confusing.

PixLSteam (changes requested):
Just some small text/comment changes. Looks good otherwise, but I haven't actually tested it.

Razikd29:
Maybe it's a bit too late to ask about that topic, but shouldn't that be a generic behaviour ? Working in a similar fashion as for vitamins that you can't give if the Pokémon is in the hatchery ?

PixLSteam:
Vitamins mess with the Pokémon's step count. Evolving the Pokémon does not actually have any effects on the prevolution outside of this challenge, so it should be fine.

Razikd29:
It was more about making the item behaviour consistent through all scenarios. But I guess it is fine, as you say

Alderi-Tokori:
> Just some small text/comment changes. Looks good otherwise, but I haven't actually tested it.

Committed the suggested changes.

Aegyo (approved):

PR #3830, opened 2023-01-21
Fix evolution stone status indicators counting future region methods

### LINKED ISSUE #3101: [BUG] Thunderstone claiming uncaught pokemon
**Description:**

In Alola, after getting all thunder stone evos and catching a Charjabug and before getting Vikavolt, the thunderstone in shops shows the empty pokeball as if there is another thunderstone evo to get.

**Expected Behavior:**

The thunderstone should ignore the galar only vikavolt stone evo if we haven't reached galar yet

**Screenshots:**

**Additional Information:**

items/EvolutionStone.ts needs an extra filter in the caughtStatus calc to ignore RegionStoneEvos which don't include a region we have access to

Aegyo:
Fixes #3101


RedSparr0w (approved):
Nice work, refactor looks good, thanks for the PR! 👍

PR #3834, opened 2023-01-21
Made player inventory column go hidden from shops if it's unwanted (Closes #3671)

### LINKED ISSUE #3671: [SUGGESTION]: Remove the "inv" column from shopmon shops.
### Current Behavior

Furfrou shard trader and galar fossil trader both have an "inv" column that tells us we have 0 of them in our inventory, which stays at 0 even if the mon is caught. Furfrou gem trader does not have that column.

### Improved Behavior

Do what Furfrou gem trader does.

### Reason for change

useless info displayed

### Detailed Description

remove the useless info from these two shops, maybe others..

Vodovik:
Didn't find a way to infer whether or not the column should be hidden or not. Maybe if $parent.item.itemType is an instance of PokemonItem, but I couldn't make that work.. (Closes #3671)

### DIFF
--- a/src/components/gemMasterModal.html
+++ b/src/components/gemMasterModal.html
@@ -30,7 +30,7 @@
                                 <!-- ko foreach: gems -->
                                 <tr>
                                     <td data-bind="text: App.game.gems.gemWallet[$data.gemType]().toLocaleString('en-US')"></td>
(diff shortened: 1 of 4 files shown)


Aegyo:
You need to put the `Closes #3671` in the body of your PR message, it doesn't work in the title. You can edit it in.

Aegyo (approved):

PR #3838, opened 2023-01-22
Add quest line rewards to pokemon locations function

CypherX:
Follow up to #3709. Adds quest line rewards as an obtainable method for the bot and wiki. I believe the only case missing now are event pokemon.


Aegyo:
> I believe the only case missing now are event pokemon.
Looks like this PR won't pick up suprise togepi, but can't do much about that without changing some code.

CypherX:
> Looks like this PR won't pick up suprise togepi, but can't do much about that without changing some code.

Yeah, was going to submit a separate PR in a few to fix that one case. (probably should have mentioned that)

Aegyo (approved):

PR #3846, opened 2023-01-23
Added decay aura. Turned haban and yache into decay aura. And mad…

### LINKED ISSUE #3626: [SUGGESTION]: Remove growth boost from affecting death timers
### Current Behavior

When using spray duck, boost mulch or wacans, a fully grown berry still gets a "growth" boost to die faster. In the berry dex we can clearly see that growth time is the time it takes to reach the berry stage. So no growing occurs after that, hence the inconsistency, in my opinion. Moreover, there alraedy is a death boost, so why the double dipping?

### Improved Behavior

I would like for growth boosts to not act as death boosts.

Another solution would be to remove the growth time information (which is already available in the growth stages) from the berry dex and keep only the lifespan that is then renamed in growth time, and this way berry stage is still considered as growth and thus boosted by growth boosts.

### Reason for change

It improves consistency of game mechanics and would make it so sprayduck, wacan and boost mulch don't become a liability once their use has passed.

### Detailed Description

```
this.formattedTimeLeft = ko.pureComputed(() => {
                return '';
            }
            let timeLeft = 0;
            } else {
            }
            const growthMultiplier = App.game.farming.getGrowthMultiplier() * this.getGrowthMultiplier();
            return GameConstants.formatTime(timeLeft / growthMultiplier);
        });
```

this part of the code shouldn't use growth multiplier (if it does what i think it does) but only death aura (which is handled in the this.getGrowthMultiplier as an outlined case.

```
 // Handle Death Aura
        if (this.stage() == PlotStage.Berry && this.berry != BerryType.Kasib) {
            multiplier *= this.auraDeath();
        }
```

Vodovik:
…e it so growth aura only affects stage < berry stage (Closes #3626)


src/scripts/farming/Plot.ts
```diff
@@ -421,11 +429,16 @@ class Plot implements Saveable {
...
+            multiplier *= this.auraGrowth();
         // Handle Death Aura
-        if (this.stage() == PlotStage.Berry && this.berry != BerryType.Kasib) {
-            multiplier *= this.auraDeath();
+        } else {
+            if (this.berry !== BerryType.Kasib) {
+                multiplier *= this.auraDeath();
+                multiplier *= this.auraDecay();
+            } else {
+                multiplier *= this.auraDecay();
```
Aegyo, line 440: Move this out of your (berry is Kasib) check, then you won't have to say it twice.

Vodovik: 🤦 haha well i guess it's kind of obvious now that my nose isn't so close. also just noticed "death & decay".. it was the "ultimate" lich spell from warcraft III

src/scripts/farming/Plot.ts
```diff
@@ -143,6 +147,10 @@ class Plot implements Saveable {
                 auraStr.push(`Death: ×${this.auraDeath().toFixed(2)}`);
             }
 
+            if (this.auraDecay() !== 1) {
+                auraStr.push(`Decay: ${this.auraDecay().toFixed(2)}x`);
```
Aegyo, line 151: `x` was moved in front of the numbers after you wrote this
```suggestion
                auraStr.push(`Decay: ×${this.auraDecay().toFixed(2)}`);
```

Aegyo (changes requested):

Aegyo (approved):

PR #3847, opened 2023-01-24
Add redeemable code REFUND-VITAMINS

### LINKED ISSUE #3845: [SUGGESTION]: Add an option to sell vitamins.
### Current Behavior

Presently, there is no way to sell any unused vitamins. No matter how many vitamins a user has bought throughout their PokéClicker "career," they will be stuck with them until the end of time.

### Improved Behavior

Users should be allowed to sell back unused vitamins for a substantially discounted price. Even if the rate of return is drastically low — something in the realm of 5% or 10% of the purchase price — users could still recuperate some of the cost of the vitamins they are no longer using.

### Reason for Change

There is merit to using Protein in earlier regions when the vitamin cap per-Pokémon is very low. Accordingly, it makes sense that the average user would buy plenty of Protein in the early game and max out their Pokémon with it to get the most out of their hatchery steps, especially considering how affordable it is in comparison to the monetary rewards for trainer battles in the earlier regions. As users advance through the regions and the vitamin cap per-Pokémon increases, however, Calcium and Carbos become much more relevant. Thus, it makes sense that the average user would then gradually retract their Protein toward the late game, favoring a healthy balance of Calcium and Carbos. This results in a phenomenon in which users possess an overabundance of Protein that are then no longer used in the late game.

Allowing the user to sell their unused vitamins — even for a _deep_ discount — would prevent them from feeling like their money was wasted in the early game — or, less importantly, from being displeased by the aesthetic dissonance of the enhanced vitamins screen showing several thousand unused vitamins.

### Detailed Description

The following table outlines the number of Protein a user would need for maximum efficiency on every Pokémon for each region, accounting for debuffs. This assumes that the user is playing 100% optimally and shifting their vitamins around as soon as they reach a new region _and_ that the user possesses all Pokémon in the active region as well as all of the forms _(which isn't entirely realistic, but it's the best that could be done for the sake of discussion)_.

image

I can't speak for everyone, but I know that I personally would not have been playing 100% optimally if I'd been running through the game for my very first time post-vitamins update. I also know that I'm not even playing 100% optimally at this very moment. Despite only being 4 Pokémon away from completing the Galar Pokédex, I am still missing a few hundred various forms of Pokémon in the game _(and I imagine the same is true for most people who are this late in the game)_. I also know that most people, like myself, hatch by descending breeding efficiency, which means they are more likely to start pumping vitamins into the most efficient Pokémon — which, unsurprisingly, requires more Calcium and Carbos than Protein due to the higher base attack values of those Pokémon.

By the time a user gets to the late game, the fact that Protein increases the step requirement of a Pokémon means that most people won't put a single Protein on a Pokémon unless they have a Carbos to balance it out for fear of losing efficiency. So the chances are incredibly good that until a user can afford an abundance of Carbos — which the table above demonstrates a substantial need for — they will force their Protein to sit in a pile off to the side until they can counteract the step requirement increase and instead focus on breeding only the Pokémon that they can funnel Carbos into.

### Another Perspective

In addition to the scenario above, there's another factor to consider. Prior to one of the latest updates that introduced Calcium and Carbos, the only available vitamin for raising the step attack bonus for a Pokémon was Protein. As such, it made a lot of sense to buy as much Protein as possible and max out all of your Pokémon with it — even in the late game. Then, the update dropped and vitamins got a rework, which meant that maxing Protein was no longer the most viable option for all Pokémon and a lot of users retracted a lot _(if not all)_ of their Protein to replace them with a healthy balance of Carbos and/or Calcium. This left most pre-vitamins update users with thousands of Protein just sitting there, never to be attached to a Pokémon again. The proposed solution helps alleviate that dilemma.

Finally, there is the case of fat-fingering or mis-clicking. If the user accidentally buys the wrong kind of vitamin in bulk _(or in the wrong quantity)_, there is no way for them to "undo" their mistake and they must simply live with it, mourning their lost money. The proposed solution would help mitigate some of those woes as well.

Aegyo:
Fixes #3845 

Refunds all vitamins in the player inventory, for whatever their currency and base price is.

NPC advertising the code in Nacrene City, south east of Unova - location chosen because Unova is the first region where new players may want to take out a bunch of vitamins and have them sitting around, and the city is near the day care for that region.

### DIFF
--- a/src/modules/codes/RedeemableCode.ts
+++ b/src/modules/codes/RedeemableCode.ts
@@ -3,10 +3,10 @@ import NotificationConstants from '../notifications/NotificationConstants';
 import Requirement from '../requirements/Requirement';
 
 export default class RedeemableCode {
-    constructor(public name: string, public hash: number, public isRedeemed: boolean, private rewardFunction: () => void, private requirement: Requirement = undefined) {
+    constructor(public name: string, public hash: number, public isRedeemed: boolean, private rewardFunction: () => Promise<boolean | undefined>, private requirement: Requirement = undefined) {
     }
 
-    redeem() {
+    async redeem() {
         if (this.isRedeemed) {
             Notifier.notify({
                 message: 'You have already redeemed this code.',
@@ -24,7 +24,7 @@ export default class RedeemableCode {
         }
 
         // If nothing returned, assume it was redeemed fine
-        if (this.rewardFunction() === undefined) {
+        if (await this.rewardFunction() ?? true) {
             this.isRedeemed = true;
         }
     }
--- a/src/modules/codes/RedeemableCodes.ts
+++ b/src/modules/codes/RedeemableCodes.ts
@@ -13,6 +13,7 @@ import ObtainedPokemonRequirement from '../requirements/ObtainedPokemonRequireme
 import RedeemableCode from './RedeemableCode';
 import GameHelper from '../GameHelper';
 import Amount from '../wallet/Amount';
+import Item from '../items/Item';
 
 export default class RedeemableCodes implements Saveable {
     defaults: Record<string, any>;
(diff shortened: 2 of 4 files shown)


RedSparr0w (approved):
Works great for me, thanks for the PR! 👍 

Are we just refunding the base price? not the inflated price?

RedSparr0w (approved):

PR #3849, opened 2023-01-24
Updates README of Tiled with a TIPS section

LuchoDreamer:
I added a TIPS section in the README of Tiled with a single tip for now, saying that the coordinates can be found in the lower left of the screen while hovering, this way other people like me don't count manually...

Since it is its own section now, maybe other people can leave their tips too if they have some.


DraKay:
Can you add that you can use terrain tab to add easily mountains and path without need to search the specific corners times?

LuchoDreamer:
Amazing tip, I had no clue!! I added it, and also changed a couple words to the first letter to uppercase.

Aegyo (approved):

PR #3852, opened 2023-01-26
Added milestone box collapsing

### LINKED ISSUE #3734: [SUGGESTION]: Allow Battle Frontier Milestone Box to be Minimized
### Current Behavior

The town map allows you to click on the top to minimize it, but the battle frontier milestone tracker that fills the same window does not

### Improved Behavior

The battle frontier can also be minimized, and potentially also the dungeon map, though that’s less important

### Reason for change

I at least find the town map way more useful in any given moment than the battle frontier tracker, though I don’t know about others. When I’m doing the frontier, most of the time I’m catching up and nowhere close to the next milestone, or in between milestones with a large gap between them. During those times, I’d prefer to be able to minimize the tracker, either to watch the dynamic background or just for less screen clutter so I can focus better on the hatchery or something. I don’t have much of a reason to make the dungeon map collapsible besides just for consistency, since if my suggestion is added it would be the only one that couldn’t.

### Detailed Description

The battle frontier milestone tracker window can minimize like almost every other window, and potentially the dungeon map would do so too for consistency’s sake.

Vodovik:
you can collapse it now (Closes #3734)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3856, opened 2023-01-26
Hoenn Map Files fix

Farboo171:
Updates some of the map files associated with the new Hoenn map assets. Excluded in error from previous PRs.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3858, opened 2023-01-27
Can now merge $playername$ into npc texts

Jaaslet:
This adds the code to be able to use $playername$ in NPC texts.
Simply use it in the text, and it will be replaced by the player name. Like `Hello $playername$!`.
The code can be used elsewhere, by adding `TextMerger.mergeText()` anywhere else in the code.
If you want to add more texts that can be merge, add them to TextMerger.ts:buildMergeValues()


Razorflamekun (commented):
Seems fine to me!  Thanks for the PR!  (I didn't approve this because I don't have enough coding experience to approve it, but I did have enough to interpret what is going on here.)

Aegyo (approved):

PR #3863, opened 2023-01-28
Catch or hatch patch

Leafsw0rd:
Adjusts all quest-text where Hatching is a valid strategy to mention it specifically. This patch at this point does not adjust any existing Capture or Hatch quests function, it just includes the hatch notifier if it's even slightly possible. Even cases where hatching is only theoretically possible with a lucky roll of the Shiny Code.

This also adjusts two pieces of text in the Manaphy Quest Line - it changes "Canalave Town" to it's correct name of "Canalave City" and changes a use of "Left" to "fled from" for clarity.


Qwertypop04:
It isn't possible under any circumstance to hatch Zacian or Zamazenta. Those being hatchable is unintentional in the first place, but, because nobody had them before the quest was made, they can't be hatched for the quest

src/scripts/quests/questTypes/CaptureSpecificPokemonQuest.ts
```diff
@@ -22,9 +22,9 @@ class CaptureSpecificPokemonQuest extends Quest implements QuestInterface {
             return this.customDescription;
         }
         if (this.amount === 1) {
-            return `Capture ${this.pokemon.name}.`;
+            return `Capture or hatch ${this.pokemon.name}.`;
         }
-        return `Capture ${this.pokemon.name} ${this.amount} times.`;
+        return `Capture or hatch ${this.pokemon.name} ${this.amount} times.`;
```
Aegyo, line 27: `or hatch` is dependent on the includeBreeding parameter, you'll need to make the class remember it (in a new property), and reference here with
```ts
        return `Capture ${this.includeBreeding ? 'or hatch ' : ''}${this.pokemon.name} ${this.amount} times.`;
```

Aegyo (dismissed):
It shouldn't be a problem but you have something weird going on with the commits listed here, did you make a branch off of one your own instead of the main repo develop one?

src/scripts/quests/questTypes/CaptureSpecificPokemonQuest.ts
Leafsw0rd: Incorporated in now.

CypherX:
The addition of the _or hatch_ text may lead to confusion for new players that couldn't possibly have caught the pokemon yet. It may be worth making it dynamic and only include that text if the player has already caught the pokemon in question.

Jaaslet:
> The addition of the _or hatch_ text (for the legendary quests) may lead to confusion for new players that couldn't possibly have caught the pokemon yet. It may be worth making it dynamic and only include that text if the player has already caught the pokemon in question.

I agree. And i don't think we need to make it dynamic, as most old players should had completed it by now. But yeah, that should be an other PR

Jaaslet (approved):
Looks good. Thanks for the PR

Jaaslet (approved):
Missedclicked... This should be an approval

PR #3882, opened 2023-01-29
Use next tier of dungeon flash if holding light ball

### LINKED ISSUE #3380: [TASK] Light Ball Held Item
As per discussion surrounding #1512, this idea has been approved for development. 

Idea: We would like the Light Ball to act as a Pokémon-Specific held item. When the Light Ball is held, an extra "level" of Flash is granted to the player in Dungeons. See #3379 for further details on Flash Levels.

Aegyo:
- If any pokemon is holding a light ball, the next best tier of flash will be used
- No effect above 400 clears as there is no next tier of flash
- Multiple Pikachu holing light ball does not have any extra effect, only one is counted.
- Flash tier chosen when entering dungeon, does not recalc if you equip a light ball while already inside

Closes #3380 

---

### Todo
- [ ] Use different item - Luminous Moss
- [ ] New item will be a ConsumableHeldItem - limited uses before needing to be replaced
- [ ] Make it only work after the first dungeon clear
- [ ] Make the new item also give an attack debuff to the pokemon

### DIFF
--- a/src/scripts/dungeons/DungeonRunner.ts
+++ b/src/scripts/dungeons/DungeonRunner.ts
@@ -318,13 +318,18 @@ class DungeonRunner {
         const clears = App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(dungeonName)]();
 
         const config = [
+            { flash: undefined, clearsNeeded: 0 },
             { flash: DungeonFlash.tiers[0], clearsNeeded: 100 },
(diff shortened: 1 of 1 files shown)


Farboo171:
Closed per discussion on Discord. New dungeon flash makes this no longer needed, and it may be revisited later with other held items, if more are added.

PR #3888, opened 2023-01-30
Makes Sudowoodo show up earlier, plus a couple requirements changes

LuchoDreamer:
In the games you can see Sudowoodo as soon as you have access to Violet Town, but it's a blockade until you beat Whitney. Right now in Pokeclicker you can only see Sudowoodo after you beat route 36. This route is also being blocked by Whitney, which doesn't make sense, because you can access it through route 35 and it has grass, but the Sudowoodo still blocks it until you beat Whitney.

So I've made the following changes:

1. Sudowoodo shows up after you beat Route 31, but you cannot fight it until you beat Whitney.
2. You only need to beat Whitney to unlock Sudowoodo's battle, instead of beating Whitney plus Route 36. This renders Route 35 optional (as some of the routes in Kanto are). In the games, after beating Whitney, you could backtrack from Goldenrod to Violet to do Sudowoodo from there, thus you don't need to step on route 35 ever if you don't want.
3. Route 36 now has two possible unlock requirements, either doing route 35, or Sudowoodo battle.
4. To access route 37 now you need to beat both Sudowoodo and route 36. If you come form Goldenrod side you would do route 36 then Sudowoodo, if you come from Violet side you would Sudowoodo then route 36, which makes sense with how the encounter are laid out in route 36.

This whole PR is not an important/necessary change, but I think showing Sudowoodo from a bit earlier in the game is a nice bit, showing you and reminding you of the blockade.

### DIFF
--- a/src/scripts/temporaryBattle/TemporaryBattleList.ts
+++ b/src/scripts/temporaryBattle/TemporaryBattleList.ts
@@ -306,9 +306,7 @@ TemporaryBattleList.Sudowoodo = new TemporaryBattle(
     'Sudowoodo',
     [new GymPokemon('Sudowoodo', 540000, 20)],
     undefined,
-    [
-        new GymBadgeRequirement(BadgeEnums.Plain),
-    ],
+    [new GymBadgeRequirement(BadgeEnums.Plain)],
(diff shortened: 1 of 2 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3889, opened 2023-01-30
Implemented log book settings to allow disabling saving logs books pe…

### LINKED ISSUE #3725: [SUGGESTION]: Allow players to choose log categories they want to have registered in the Log Book
### Current Behavior

At the moment, filtering logs only shows what you want to see, that's working as intended and it's fine for players that catch everything as soon as possible.
However, the logs that are hidden are still taken into account in the log "memory". If I hide New encounters and go onto a road with uncatched Pokémon, I won't see any logs for them but my other logs (Shiny, items, achievements...) won't stay long (roughly 30 minutes top).

### Improved Behavior

Add a setting where you can tick on/off the type of logs you want to be registered in the log "memory".

### Reason for change

With saves with challenges being a thing, some players might want to check how their game went while away. Not having the proper feedback from the log book is irritating.

### Detailed Description

Add a setting or something similar that allows the registration of specific logs to be stopped or resumed in the log book.

Vodovik:
…r type (CLOSES #3725)
image

Not the sexiest of UIs, if anyone has a better idea, i'm all ears : )


Jaaslet (changes requested):
(have not tested yet) Can we move the settings to the settings window?

Vodovik:
@Jaaslet, @Aegyo requested that these settings be put in the logbook modal, so i put it there! : )

src/components/logBookModal.html
```diff
@@ -40,3 +60,17 @@ <h5 class="modal-title">Log Book</h5>
...
+            setting name
+        </label>
+    </td>
+    <td class="p-2">
+        <input class="clickable" type="checkbox"
+               data-bind="checked: $data.observableValue(), attr: {name, id: 'checkbox-' + $data.name }"
+               onchange="Settings.setSettingByName(this.name, this.checked)"
+        />
+    </td>
+</script>
```
Aegyo, line 76: This isn't needed, the templates defined in the setting modal file should be available to use over here too

src/components/logBookModal.html
```diff
@@ -13,25 +13,45 @@ <h5 class="modal-title">Log Book</h5>
...
+                            </tbody>
+                        </table>
+                    </div>
+
+                    <div id="log-book-settings" class="tab-pane p-3">
+                        <h5><b>Disable or enable types of logs</b></h5>
+                        <table class="table table-striped table-hover m-0">
+                            <tbody>
+                            <!-- ko foreach: Object.keys(LogBookTypes) -->
+                                <tr data-bind="template: { name: 'BooleanSettingTemplate', data: Settings.getSetting(`logBook.${$data}`)}"></tr>
```
Aegyo, line 49: BooleanSettingSliderTemplate looks nicer I think

src/modules/logbook/LogBook.ts
```diff
@@ -17,9 +18,11 @@ export default class LogBook implements Feature {
     public filteredLogs: PureComputed<LogBookLog[]> = ko.pureComputed(() => this.logs().filter((log) => this.filters[log.type.label]?.()));
 
     newLog(type: LogBookType, content: LogContent) {
-        const length = this.logs.unshift(new LogBookLog(type, content));
-        if (length > 1000) {
-            this.logs.pop();
+        if (this.canLog(`logBook.${type.label}`)) {
```
Aegyo, line 21: I don't like how you have to know how the settings names are set up to call `canLog` - I would prefer if `canLog` had the whole LogBookType as the parameter and then that function can do the dirty work of calculating the setting name we need.

Aegyo:
> (have not tested yet) Can we move the settings to the settings window?

I was thinking this didn't have a good location in the settings window, and seemed too small to warrant a new tab. Also just felt like it made more sense being on the modal the settings affect? Not sure, maybe it could be in both modals.

Aegyo (approved):

Jaaslet (approved):

Issue #3850, opened 2023-01-25
[SUGGESTION]: Add Defender type to Hatchery settings


Alderi-Tokori:
### Current Behavior

You can alter breeding efficiency calculation by setting current region to take regional buffer into account.

### Improved Behavior

Add defender type 1 and type 2 to the hatchery settings ("defender type" being the target type like in the damage calculator), so that pokemon who are more or less effective against a given type combination have that effectiveness taken into account into their efficiency.

### Reason for change

This way, people can breed to aim for a particular type combination, this can help people who are blocked by a certain gym or pokemon wall to more efficiently breed toward passing that wall.

### Detailed Description

With no defender types selected (None-None), this would give the same behaviour as the current one. When a defender type 1 (and eventually a type 2) is selected, then the efficiency calculations takes this target type into account: each pokémon will have its breeding efficiency multiplied by its highest type multiplier against the target type.

Alderi-Tokori:
![image](https://user-images.githubusercontent.com/26024673/214692792-fb2c87a1-999b-47c1-9fef-9035b706bbc4.png)
Here, for example, the normal defender type is selected, so combat type pokémons are more efficient, but "naturally efficient pokémons (here, Rotom for the Sinnoh region), can still remain high in the list.

PR #3890, opened 2023-01-31
Add defender type to hatchery settings for breeding efficiency calculation

Alderi-Tokori:
Implements suggestion #3850


CooldudeOmega:
I think defender should be changed to defending but from the slightly more than usual I can gauge from the code this looks good

Alderi-Tokori:
I just realized I put all my code in the "wrong" module; since it's intended to be in the hatchery settings, it shouldn't be in the BreedingFilters, but directly in the BreedingController (in the same fashion as the regionalAttackDebuff), I'll change this so it's coherent with the rest.

I'll also change "defender" to "defending".

Alderi-Tokori:
It's fixed, now the defending type settings are handled in the BreedingController.

Jaaslet:
Sorry for waiting this long to respond. It was not easy to figure out what to do with this PR.
We are going to close it. Very few players will want to use it, and it adds more stuff to an already confusing UI

PR #3892, opened 2023-01-31
Singularize eggs in Hatch egg quest when you only need to hatch one

CooldudeOmega:
I sort of did a copypaste from berries with an ok amount of modification so if it doesn't work please help fix it since I'm sure it isn't right the first time


CooldudeOmega:
check failed, let me go fix that I guess

CooldudeOmega:
No way, it actually looks like it works, I'm just gonna wait for a review then.

CooldudeOmega:
Does the plural of egg need to be defined in the code? I saw a pr about berry pluralization having berry plural defined in dungeon something idk

Aegyo:
That can be done in translation files eventually

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3895, opened 2023-02-01
Used DayCycle instead of hardcoded hours for duskball

Vodovik:
Small refacto to use day cycle for dusk ball.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3897, opened 2023-02-01
Used the berry_discovered sound on berry mutation.

### LINKED ISSUE #2871: [SUGGESTION] Use a different sound for berry mutations
**Current Behavior**
The noise emitted when a berry mutation occurs is identical to the noise that is emitted when a berry is ready for harvest.

**Improved Behavior**
An entirely separate sound should be created for the purpose of denoting berry mutations and the new sound should be emitted in such an instance.

**Reason for change**
 When micro-managing berry plots over an extended period of time, the use of the same noise for both events can cause brief moments of panic for berries with very, *very* short windows of ripeness prior to death when in reality a mutation has simply occurred. Using a different sound would help to avoid conflating a mutation with a berry that is ripe and ready for harvest.

**Detailed Description**
Similar to how the hatchery has two distinct noises — one to indicate when an egg is ready to hatch and one to indicate when all slots in the queue are empty — the farm should possess two distinct noises — one to indicate when a berry is ready for harvest and one to indicate when a berry mutation has occurred.

**Screenshots:**
Not applicable.

Vodovik:
This is a solution, but might not be the best. I like the "new" sound being unique, but then it's useless. If anyone has a better sound to replace that with for mutation, I'm open to suggestions : )

(Closes #2871)


Jaaslet:
I think this solution is worse than the current solution. Players will get annoyed, when they are waiting for a new mutation and keep getting the berry discovered sound, each time they get the wrong mutation.
I think we need someone who can create/find new sounds.
I will close this PR

PR #3900, opened 2023-02-02
Add unit test for module

RedSparr0w:
Only done a couple of modules so far (Credits, GameConstants), not sure if this is the layout we want to go for, or if we wanna add the test to their own folders.

- [X] Also need to solve the typescript update errors 😩


Aegyo:
> Also need to solve the typescript update errors

Was this done already or is this PR more of a draft state right now?
Got a couple of conflicts also

RedSparr0w:
> Was this done already or is this PR more of a draft state right now?
> Got a couple of conflicts also

Yeah, forgot to update the main comment, also resolved the conflicts 👍
Should be all good now 🤞

src/modules/GameHelper.ts
```diff
@@ -2,6 +2,8 @@ import {
     Observable as KnockoutObservable,
     Computed as KnockoutComputed,
 } from 'knockout';
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+import * as ko from 'knockout';
```
Aegyo, line 6: I'm pretty sure this causes any observables created with this ko to not actually trigger any UI updates or computeds made with the global ko - easily tested with the questlist refresh tooltip which uses GameHelper.formatterLetterTimeUntilTomorrow

Last time I had this issue I think I decided it was because this ko is a different one to the one we use when applying bindings? That one coming from the knockout script tag included in the html. I did have an idea though which would save us from having ko as a global in modules - we could remove that script tag and instead temporaryWindowInject whatever `ko` modules is using?

---

Just tried that out for a few minutes and it looks like it isn't as simple as I thought...

Seems like the best way to solve this for now would be to add ko as a global in the jest.config.js and remove this import I am commenting on
```
  globals: {
    ko: require('knockout'),
  },
```

src/modules/utilities/utilities.test.ts
```diff
@@ -0,0 +1,74 @@
...
+jest.spyOn(global.Math, 'random').mockReturnValue(0.123456789);
+
+import SeededRand from './SeededRand';
+import { AchievementType } from '../GameConstants';
+import Rand from './Rand';
+
+describe('Test SeededRand & Rand', () => {
+    it('Test rand', () => {
+        expect(Rand.next()).toBe(0.123456789);
+    });
```
Aegyo, line 13: Not sure that this is a useful test, it is just returning the value you are mocking on Math.random

Aegyo (changes requested):

.github/workflows/build.yml
```diff
@@ -18,4 +18,7 @@ jobs:
       with:
         node-version: 16.x
     - run: npm ci
-    - run: npm run test
+    - run: gulp scripts
+    - run: npm run eslint
+    - run: npm run stylelint
+    - run: npm run jest
```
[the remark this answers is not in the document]
RedSparr0w: Just easier to see in the workflow where something failed, I'm fine with going back to run test though.

src/modules/modules.test.ts
```diff
@@ -0,0 +1,180 @@
...
+describe('Test Credits', () => {
+    it('should have the correct properties', () => {
+        [...SpriteCredits, ...CodeCredits].forEach((credit) => {
+            expect(credit).toHaveProperty('name');
+            expect(credit).toHaveProperty('link');
+            expect(credit).toHaveProperty('image');
+            expect(credit).toHaveProperty('resources');
+        });
+    });
+});
```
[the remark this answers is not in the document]
RedSparr0w: IIRC without it, it showed the lines as having no coverage

src/modules/utilities/utilities.test.ts
RedSparr0w: It's only for coverage, not sure if there's a better way to test this without just keeping Math.random as a random value and saying it should be between `1` and `0`?

src/modules/modules.test.ts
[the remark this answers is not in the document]
Aegyo: Fair enough, a red line in my console every time I run tests would get annoying after a while :sweat_smile:

src/modules/utilities/utilities.test.ts
Aegyo: I'm fine leaving it as is just for the coverage

src/modules/GameHelper.ts
RedSparr0w: I cannot figure out how to get this to work, the globals route just fails to run the tests for me with:
`ko.observable is not a function`

Aegyo: can you push up a separate branch for me to see what you are doing differently? It was working for me when I tried it

src/modules/utilities/utilities.test.ts
RedSparr0w: Actually, if I just test a `Rand.intBetween` or something like that, it should automatically cover this function too.

Aegyo: Yeah that sounds like a good idea :+1:

src/modules/GameHelper.ts
RedSparr0w: Added here: https://github.com/pokeclicker/pokeclicker/tree/sparr0w/testing-jest-ko
Just removed the import line, added the globals.

Aegyo: Keep the `import * as ko from 'knockout'` in the modules.test.ts, just remove the one from GameHelper.
The one in test files should be fine since it doesn't affect the actual game code

Aegyo: Or it might be possible to get the jest global one in your tests with `globalThis.ko.observable()`

RedSparr0w: Gives me the same problem if it remains in `modules.test.ts` or not

RedSparr0w: Gonna make a PR with that branch, see if it fails with the GitHub tests

RedSparr0w: Passes fine on there too 🤔 

Gonna try another clean install and see if that fixed it on my end.

RedSparr0w: Tried clean install, and on another computer which hadn't run any of the tests yet, same error, might be Windows related.

RedSparr0w: Got it to work with `maxWorkers: '1',` not entirely sure why though.
Only thing I could find was this: https://github.com/facebook/jest/issues/11956#issuecomment-1458178986

Aegyo (approved):

PR #3905, opened 2023-02-03
Add hotkeys and relevant settings to swap between Oak Items Loadouts

### LINKED ISSUE #3899: [SUGGESTION]: Hotkeys to switch currently equipped Oak Item Loadouts
### Current Behavior

While similar to https://github.com/pokeclicker/pokeclicker/issues/950, I would like to switch between my loadouts using a hotkey.

Currently, to switch between my loadouts, I have to click the specific loadout every time.

### Improved Behavior

I would like to switch the currently equipped Oak Item Loadout using a hotkey (e.g. `shift-#` where # is a digit from 1-5 to represent the Oak Item Loadout).

### Reason for change

When I am cycling through tasks (e.g. farming, mining, hatching while grinding on a route), I switch my Oak loadout to something to get the exp points for the related item / task, and then switch back to my "grinding" loadout once I'm finished. This requires several clicks between each task. This process happens often enough that I'd like for a hotkey way to quickly swap my current oak item loadout.

### Detailed Description

I don't know exactly what the hotkey combo could be, but a combination press with 1-5 digit seems like a reasonably idea to me.

TheRealJoeFriel:
Changes to address #3899 - Added hotkeys and relevant settings to swap between Oak Items Loadouts.

Closes #3899


src/modules/settings/index.ts
```diff
@@ -303,6 +303,12 @@ Settings.add(new HotkeySetting('hotkey.oakItems', 'Oak Items', 'O'));
 Settings.add(new HotkeySetting('hotkey.underground', 'Underground', 'U'));
 Settings.add(new HotkeySetting('hotkey.pokeballSelection', 'Poké Ball Selection', 'P', { suffix: ' + Number' }));
 
+Settings.add(new HotkeySetting('hotkey.oakItems.loadout.1', 'Activate Oak Items Loadout 1', '!'));
+Settings.add(new HotkeySetting('hotkey.oakItems.loadout.2', 'Activate Oak Items Loadout 2', '@'));
```
Jaaslet, line 307: Not everyone has the same keyboard layout, so this won't work

src/scripts/GameController.ts
```diff
@@ -410,6 +410,41 @@ class GameController {
                         return e.preventDefault();
                     }
                     break;
+                case Settings.getSetting('hotkey.oakItems.loadout.1').value:
+                    // Activate to Oak Item Loadout 1
+                    if (oakItems.canAccess() && !$oakItemsModal.data('disable-toggle')) {
+                        App.game.oakItemLoadouts.activateLoadout(0);
+                        return e.preventDefault();
+                    }
```
Jaaslet, line 418: This can be solved cleaner. See how buttons inside towns works

Jaaslet (changes requested):
I don't like using the shiftkey for something that will not be used that often. Can it be like O+numkey? Can't remember if that is taken.

TheRealJoeFriel:
Hey @Jaaslet, in regards to not using Shift as a modifier key for this, I think that either we will need dedicated hotkeys for each loadout, or always have a modifier key used in conjunction with this. For instance we can't do O+<Number> because if O is the hotkey for something like Oak Item modal (by default) it will repeatedly open and close that modal while you hold down O. Your point about !, @, #, $, % not being valid is a good point as keyboard layouts can change, but how about ctrl or alt? The reason why I can't map it to Shift + <Number> regardless of keyboard mapping is because it sees Shift + 1 as !. But if I did something like control or alt + <Number> they it would not change the input key value like Shift does.

src/scripts/GameController.ts
TheRealJoeFriel: I assume you mean how we handle hotkey presses in towns on lines 343-372. I'm not sure I follow how that could be cleaner. If you are referring to the if condition, I was using the same condition used for opening the oak items modal. If you are referring to the handling of number key presses to select TownContent/NPCs, that would require tying the hotkeys to the number keys. With the current changes it is set up so you could set them to anything. Would you prefer if it was always set to 1-5 (with a modifier) and it wasn't configurable?

RedSparr0w:
Couldn't we just make it work similar to the pokeball selector?
`O` to open the oak items, `1-X` to choose a loadout to load

TheRealJoeFriel:
> Couldn't we just make it work similar to the pokeball selector? `O` to open the oak items, `1-X` to choose a loadout to load

Hi @RedSparr0w, I like your idea but it looks like there is already an implementation for activating/deactivating individual oak items using the number keys when the oak items modal is open on lines 208-220 of GameController.ts. However it only seems to work for 1-9, 0 does not appear to be working. We have more than 9 items in the game, so maybe it would make sense to revise this to swap loadouts instead since not all items are able to be activated/deactivated using the current hotkeys when the modal is open? 

What are your thoughts?

TheRealJoeFriel:
Added changes to change Oak Items loadouts when the Oak Item modal is open. This takes away the ability to activate/deactivate individual oak items using the number keys like mentioned in my previous comment.

Jaaslet (changes requested):
I don't like the idea of taking away a shortcut, even though it did not cover everything

PR #3908, opened 2023-02-03
Never Melt Ice from Mythic to Legendary

EatPant2nd:
All Held Items are Legendary drops. For some reason, Never Melt Ice in Shoal Cave (Hoenn) was Mythic. Now it's Legendary. Consistency!


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #3631, opened 2022-12-22
[SUGGESTION]: Opportunity not to participate in events.


xqmaareekpx:
### Current Behavior

I am new in this game and I guess everyone participates.

### Improved Behavior

Opportunity not to participate in events.

### Reason for change

Not everyone wants custom Pokemon. In a situation where everyone participates, we are somewhat forced into it.

### Detailed Description

Maybe a new challenge, or the ability to remove unwanted pokemon?

Qwertypop04:
Just don't catch them then

CooldudeOmega:
Don't catch them? This isn't something bad, you probably want more Pokémon, no reason not to get them, but if you don't want them just don't get them.

xqmaareekpx:
It depends how You play. I might as well not play for a week. That's not what I'm talking about.

DraKay:
Set new as not catch in ball selector easy peasy... This game have only 4 mandatory pokemons pidgey line and eternatus, anything else can be skipped. If you want to play only with vanilla pokemons there is no problem, just don't catch... If you can solve the problem without add more code why make a complex solution no needed

xqmaareekpx:
I understand, but I would have to turn off all pokeballs, not just new (these pokemons have a shiny version?). This equates to not playing for a week.

CooldudeOmega:
The more rational question is why do you dislike the eventmons so much, why do you not want to play for a week to avoid 3 Pokémon?

EatPant2nd:
I mean, events usually don't happen in all regions so you can move to a different region.

EatPant2nd:
No, I won't call it that. Also, yeah I can see that it would be a bit annoying to remove ball from "new" and "new shiny" for a week. Could be worse though.

CooldudeOmega:
I'm confused, you want only official pokemon, in the pokemon fangame, I doubt that there'll be an opportunity to skip events but going through the trouble of programming this for easy to miss pokemon for a few people that don't like them seems, unproductive.

DraKay:
Just avoid boosted route, not catching a single roamer is easy and if you are unlucky you save scum and done...

xqmaareekpx:
[CooldudeOmega]
Maybe You're right. But firstly, it's a suggestion, and that's what they're for, and secondly, maybe more than a few, You do not know that.

Maybe the release option would be better?

[DraKay]
I know how to avoid catching pokemon by playing my way for a week. That's not my point at all.

:)

[...]

PR #3917, opened 2023-02-05
Add the ability to toggle special events in the settings

TheRealJoeFriel:
Addresses the suggestion raised in #3631. The new "Disable Special Event" setting is set to false by default so that people will be able to participate in events by default when starting a new game. Instead of just checking the setting when an event starts/ends I figured an observable would be the best way to go about this so that the events would end instantly if the events are disabled. 

Tested using custom special events that lasted the entire year. Was able to see special pokemon added to dungeons when the event was ongoing and after toggling the setting the pokemon were removed from the dungeon.


src/modules/specialEvents/SpecialEvent.ts
```diff
@@ -23,6 +24,17 @@ export default class SpecialEvent {
     // TODO: only notify once initially until event about to start/end
     notified: SpecialEventNotifiedStatus;
 
+    disabledSpecialEvents = Settings.getSetting('disableSpecialEvents').observableValue.subscribe((newValue) => {
```
Jaaslet, line 27: You are currently setting the value to the result of a void function. Events are not working because of this

Jaaslet:
We are currently in the process of rewriting some of the event code. After that, we will be able to make a cleaner solution (that works). I will close this PR

PR #3919, opened 2023-02-05
Route 117 Requirement change

LuchoDreamer:
Changed route requirement from Route 110 to May 3.

You are able to access route 117 and Verdanturf after clearing route 110 without being able to even step on Mauville because that's blocked by May 3 and that shouldn't be correct.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3922, opened 2023-02-07
Fixed height to berry seed list

osherwott:
thank you to all those based people out there that made this possible


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3930, opened 2023-02-10
Adds "in Johto" to Unfinished Business quest step

EatPant2nd:
Some people taking the Unfinished Business Questline get a bit confused about the Azalea Town quest step, wondering where it is and why they can't find it in Kanto. This adds that it's "in Johto" to make it clearer that they require further progression to advance said questline.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3933, opened 2023-02-10
Update Contagious Pokemon Catch Logic

morcousw:
If a pokeball is not selected for contagious pokemon, but there is one selected for caught pokemon, then use that on contagious pokemon as well.


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #2079, opened 2022-04-26
[SUGGESTION] Make uncaught bosses have a better chance to appear over caught bosses


TrainerRobo:
**Current Behavior**
Bosses all have equal weight and have the same likelihood to appear.

**Improved Behavior**
Bosses that are uncaught will have a better chance to appear, making it more likely to find an uncaught boss.

**Reason for change**
Attempting to catch a new boss in a dungeon with more bosses can prove frustrating, especially for players who cannot leave their game running overnight to farm dungeon tokens. Grinding dungeon tokens out is quite frankly annoying and adds nothing but annoyance to the game. This is especially true when the last dungeon boss you need is a legendary Pokemon with a low catch rate. You can wind up going 50 to 100 times easily in some dungeons without a master ball. Even if you have a master ball, in a dungeon like P2 Laboratory in Unova or Galactic Eterna Building where there are many boss options, it can still take quite a hefty sum of dungeon tokens for the point in the game you're at.

**Detailed Description**
Add a weight feature to dungeon bosses. If you have already caught a boss, it has a weight of one. If you haven't caught the boss, it has a weight of two, possibly three. This would still require grinding, but it would help a lot in both speeding up the process of grinding tokens, and speeding up the process of clearing a dungeon.

Jaaslet:
I will close this. See linked PR for details

PR #3935, opened 2023-02-11
Increase Uncaught Boss Pokemon Weights

TheRealJoeFriel:
Could be an unwanted change but I did not see any comments on #2079 and it was never closed so figured I would start a discussion around it. 

I had a similar frustration when it took over 50 clears to catch Lunatone in Hoenn's Meteor Falls. I know thats just how RNG can work against you in some cases. Figured as this would only apply to dungeon bosses, and the effect of it seems to just nudge the chance of an uncaught pokemon being the dungeon boss slightly in the players favor. When all boss pokemon are caught or uncaught the bonus is nullified.

Ran some quick numbers for comparison:
Base weight + (uncaught bonus / total bosses) = weight

2 Bosses:
1 + 1/2 = 1.50 / 3.00 = 0.500
1 + 1/2 = 1.50 / 3.00 = 0.500

1 + 0/2 = 1.00 / 2.50 = 0.400
1 + 1/2 = 1.50 / 2.50 = 0.600

1 + 0/2 = 1.00 / 2.00 = 0.500
1 + 0/2 = 1.00 / 2.00 = 0.500

3 Bosses:
1 + 1/3 = 1.33 / 4.00 = 0.333
1 + 1/3 = 1.33 / 4.00 = 0.333
1 + 1/3 = 1.33 / 4.00 = 0.333

1 + 0/3 = 1.00 / 3.66 = 0.273
1 + 1/3 = 1.33 / 3.66 = 0.363
1 + 1/3 = 1.33 / 3.66 = 0.363

1 + 0/3 = 1.00 / 3.33 = 0.300
1 + 0/3 = 1.00 / 3.33 = 0.300
1 + 1/3 = 1.33 / 3.33 = 0.399

1 + 0/3 = 1.00 / 3.00 = 0.333
1 + 0/3 = 1.00 / 3.00 = 0.333
1 + 0/3 = 1.00 / 3.00 = 0.333

4 Bosses: 
1 + 1/4 = 1.25 / 5.00 = .250
1 + 1/4 = 1.25 / 5.00 = .250
1 + 1/4 = 1.25 / 5.00 = .250
1 + 1/4 = 1.25 / 5.00 = .250

1 + 0/4 = 1.00 / 4.75 = 0.210
1 + 1/4 = 1.25 / 4.75 = 0.263
1 + 1/4 = 1.25 / 4.75 = 0.263
1 + 1/4 = 1.25 / 4.75 = 0.263

1 + 0/4 = 1.00 / 4.50 = 0.222
1 + 0/4 = 1.00 / 4.50 = 0.222
1 + 1/4 = 1.25 / 4.50 = 0.277
1 + 1/4 = 1.25 / 4.50 = 0.277

1 + 0/4 = 1.00 / 4.25 = 0.235
1 + 0/4 = 1.00 / 4.25 = 0.235
1 + 0/4 = 1.00 / 4.25 = 0.235
1 + 1/4 = 1.25 / 4.25 = 0.294

1 + 0/4 = 1.00 / 4.00 = .250
1 + 0/4 = 1.00 / 4.00 = .250
1 + 0/4 = 1.00 / 4.00 = .250
1 + 0/4 = 1.00 / 4.00 = .250

### DIFF
--- a/src/scripts/dungeons/Dungeon.ts
+++ b/src/scripts/dungeons/Dungeon.ts
@@ -148,12 +148,16 @@ class Dungeon {
     }
 
     /**
-     * Retreives the weights for all the possible bosses giving a slight bump to bosses which have not been caught.
+     * Retreives the weights for all the possible bosses.
      */
     get bossWeightList(): number[] {
         return this.availableBosses().map((boss) => {
-            const isCaught = App.game.party.alreadyCaughtPokemonByName(pokemonMap[boss.name].name) ? 0 : 1;
-            return (boss.options?.weight ?? 1) + (isCaught / this.availableBosses().length);
+            if (App.game.challenges.list.requireCompletePokedex.active()) {
+                const isCaught = App.game.party.alreadyCaughtPokemonByName(pokemonMap[boss.name].name) ? 0 : 1;
+                return (boss.options?.weight ?? 1) + (isCaught / this.availableBosses().length);
+            }
+
+            return boss.options?.weight ?? 1;
         });
     }
 
--- a/src/scripts/dungeons/Dungeon.ts
+++ b/src/scripts/dungeons/Dungeon.ts
@@ -148,11 +148,12 @@ class Dungeon {
     }
 
     /**
-     * Retreives the weights for all the possible bosses
+     * Retreives the weights for all the possible bosses giving a slight bump to bosses which have not been caught.
      */
     get bossWeightList(): number[] {
         return this.availableBosses().map((boss) => {
-            return boss.options?.weight ?? 1;
+            const isCaught = App.game.party.alreadyCaughtPokemonByName(pokemonMap[boss.name].name) ? 0 : 1;
+            return (boss.options?.weight ?? 1) + (isCaught / this.availableBosses().length);
         });
     }
 


Farboo171:
This may have a minor benefit for the first few dungeon runs to some files (AC mostly) but will have a significant negative impact on limited runs if implemented.

Not a deal breaker, but worth considering

TheRealJoeFriel:
@Farboo171 good point, I was only thinking about the way it affected "vanilla" playthroughs. 

Like I said in the beginning, this might be an unwanted change to begin with but I hadn't seen any discussion on the topic. I can always close this as "will not merge"/"unneeded" but would also ask that someone makes a comment on the original issue saying something similar.

TheRealJoeFriel:
> Can this code only be active if the dex challenge is on? Then it won't apply for weird challenges but will help common runs...

Just to confirm you are referring to the recommended challenge for having to fill the pokedex before moving onto the next region? I think I could make that work, I'll give it a shot.

DraKay:
Yes I was referring to that

Jaaslet:
I will close this PR.
People have over the years added weight to the different dungeon bosses, and spend time thinking about what it should be.
Merging this will kinda negate that work, and replace it with some sorta random values.
If there is a specific case that is too hard/mean, we can talk about changing the balance of that one case, but just nerfing all of them is... Not a good way to solve it...

Issue #3865, opened 2023-01-28
[SUGGESTION]: Display Attack Required to Mega Evolve Pokemon when hovering over them in the Evolution Item Menu


Razorflamekun:
### Current Behavior

The game currently does not display any additional information when hovering over a Pokemon that is displayed when using a Key Stone.

### Improved Behavior

When hovering over a Pokemon after clicking on the Key Stone to use it, have it display the attack required in the tooltips.

### Reason for change

There aren't too many places to go to find out how much attack a Pokemon requires in order to Mega Evolve, and by adding the required attack to the tooltips displayed when hovering over a Pokemon after clicking to use a Key Stone, it adds an additional place to find out what attack is required to mega evolve a Pokemon.

### Detailed Description

Steps:

1. Click on Menu.
2. Click on Items
3. Click on Evolution Items
4. Scroll down and click on Key Stone.
5. Hover over any Pokemon in the list to have the tooltip for it displayed.

This is the box where I'd like the required attack to Mega Evolve a Pokemon to be displayed.  Thanks for reading!

DrunkenGamer:
I would agree with this suggestion - the padlock tooltip for when a mega can not be evolved could be a little more detailed - i found it difficult to work out which I could not evolve as i had not got the related mega stone and which were because i had not got them to a high enough AP 

Had i not looked at the pinned comments in the discord i would not have been able to work it out using just in-game information easily

PR #3939, opened 2023-02-14
Explicit hint for Mega Evo in the Evo Item menu.

RegisCoaxans:
Closes [#3865](https://github.com/pokeclicker/pokeclicker/issues/3865).

The lock tooltip now either says the Pokémon has not got its Mega Stone yet, or that it needs more attack points.
image
image


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3940, opened 2023-02-14
Nerf Blue 3

Qwertypop04:
Significantly lowers Blue 3's health, as it is much higher than was intended.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3942, opened 2023-02-15
[FIX] Pokérus without Kanto starter

### LINKED ISSUE #3331: [BUG] Possible to not get pokerus
**Description:**

Real evolution challenge makes it easily possible to not have the starter pokemon when being given pokerus, so no pokemon will get it.

**Steps to Reproduce:**

1. Enable Real Evolution challenge and disable Dex Requirement
2. Never buy eggs for your starter type, so you never have the base form again
3. Reach the part of the game where you are given pokerus
4. See no pokemon is given pokerus

**Expected Behavior:**

Something else should be given pokerus if starter isn't there

**Additional Information:**

Should probably give it to a random pokemon you have instead in this case, filters will show user which pokemon has it.

PixLSteam:
Real Evo or the starter skip would allow you to get to Sinnoh without a Kanto starter. This PR should infect the first Pokémon in your party if the selected Kanto starter is not in your party.

I also moved the initial infection from onClose to onUnlock.

Closes #3331


CypherX:
Closes #3331

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3948, opened 2023-02-17
Adds Rotom Discord and Surfing Pikachu as bosses into their respective dungeons

LuchoDreamer:
This PR adds Rotom (Discord) and Surfing Pikachu as catchable Pokémon only after you already obtained them, as a way to EV them or any other purpose.

Rotom (Discord) added in Team Galactic Eterna Building with the same requirements as the other Rotoms plus you need to have already obtained Discord (Rotom) for it to appear. It's completely hidden until you already own it.

Surfing Pikachu added in Pikachu Valley with the other Pikachus with the requirement that you need to own it first. Same as Rotom (Discord), Surfing Pikachu is completely hidden until you already own it.

I thought it was a nice idea to have both Pokémon available in the game so they can be EV'd, and there is precedent with the inclusion of the Vivillon coming back during the Lunar Year Event just so they can be EV'd.


Jaaslet:
I will close this PR.
We have planned to add a safari some day, which allows for all mons to be obtained. Adding some of them to random locations, would just be weird... It's something we need to keep track of in the code, and it's something people need to know, when they are EVing all their mons

PR #3953, opened 2023-02-20
Added Thunder Stone Evolution to Detective Pikachu after Quest completion

EatPant2nd:
Allows Detective Pikachu to evolve into Detective Raichu using a Thunder Stone only after completing the relevant Questline.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3954, opened 2023-02-20
Added friendly safari

Jaaslet:
The most friendly of all friendly safaris are here! The main goal of this is a catch-all for being able to EV all the mons that we don't want to spend time on finding EV methods for (which is basicly all of the non-ev-able mons left).

The zone is not finished, but i would claim the PR is still mergable (after i have gotten the new map), because it's better than nothing, and i don't know when i will have time to finish the rest. If we merge it, it's easier for other people to fix the missing stuff/test it. And even if we release an update before that, it should be fine.

List of missing stuff (might add some of it, before the PR is merged):

- [x] Add it to the map (needed before merge)
- [x] Only like 3-10 should show up each day
  - [ ] Make sure it's updated correctly at midnight
- [ ] Normal evolutions does not currently show up
- [ ] Event pokemon does not show up. But this should be fix with the rewrite i want to do for event calendar (one day...)
- [ ] Add random filler mons
- [ ] Maybe add a NPC to explain stuff
- [ ] Add the stuff we have talked about, to make the safari zone more fun to do


Jaaslet:
The safari will not be accessible until we have improved the Safari, so players do not complain about it too much.
But it's still ready to merge, so other people can help improve it

RedSparr0w (approved):
Looks good, thanks for the PR.
Worth noting that the list of encounters would change any time you obtain a new mon too.

PR #3956, opened 2023-02-22
Improved roaming groups

PixLSteam:
Since the new wiki only shows the region of a roaming Pokémon instead of the relevant subregion(s), I decided to improve the way we keep track of roaming groups.
This PR adds a new class for roaming groups, allowing us to track their names and the included subregions.
I have also updated `PokemonHelper.getPokemonRoamingRegions` to return the roaming group as well. 
I will also be opening a PR to the wiki repo to display this information on the individual Pokémon pages.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3958, opened 2023-02-24
Trainer ID

CypherX:
In the pokemon games the trainer id is a six digit randomly generated number between 000000 and 999999 assigned to the player. This is more or less implemented the same way.

Changes:

1. Implement Trainer Id
2. Display Trainer Id on the trainer card / profile
3. Change Enigma berry to use Trainer Id rather than linked Discord Id
4. Update function to derive Trainer Id from linked Discord Id (thanks PixL for the math) to preserve berries required for Enigma


src/components/saveSelector.html
```diff
@@ -38,6 +38,7 @@
     <div class="trainer-card-container mb-3 col-lg-4 col-md-6 col-sm-12 xol-xs-12">
         <div class="trainer-card clickable card font-weight-bold">
             <div class="card-body">
+                <div class="position-absolute small trainer-id" style="top: 0; right: 5px;"></div>
```
Jaaslet, line 41: I think we want something in front of the number. Maybe just a #

CypherX: Updated

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3959, opened 2023-02-24
Mega Diancie

Farboo171:
Adds an additional step to the end of the Diancie quest that provides the mega stone. This fight corresponds to the final fight in the movie the quest is based on.

EDIT: The fight is added such that players who have already completed the quest will still see the temp battle on their map and can fight it to obtain the megastone. Similar to the Red Gyrados implementation.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3963, opened 2023-02-24
Prevent transfering incompatible held items in real evo challenge

### LINKED ISSUE #3961: [Bug]: [Real Evo] Evolution keeps the held item even if normally it couldn't equip it
### Confirmations

### What happened?

If you equip an item to a pre-evo and then evolve it, the game doesn't check if the evolved Pokémon could have equipped this item normally or not and leaves it equipped. It also gives it the correct bonus. So, for example, if you equip an Eevee with a Silk Scarf (Normal type needed to equip) and then evolve it to Vaporeon then Vaporeon will keep the item and even get the bonus.
normal-vaporeon

Evolutions probably shouldn't be able to keep items they normally wouldn't be able to equip since you get issues like getting a Raichu with a Light Ball which would be a higher bonus than normally possible.

### Steps to Reproduce

1. Equip an item to a Pokémon that will evolve to a different typing or into a Pokémon that couldn't normally use that item.
2. Evolve the Pokémon.
3. Check that it has the item equipped and that it receives the bonus.

Windows 11

Chrome

### Relevant log output

_No response_

### Additional Information

_No response_

CypherX:
Fixes #3961


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3973, opened 2023-02-27
Add Kalos Friend Safari to overworld map

Farboo171:
Adds the Kalos Friend Safari to the overworld map, with credits to Soul


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3975, opened 2023-03-01
Legendary Beasts quest accessibility change

Shinesprite41:
To fit more in-line with the main games, I swapped Ecruteak so that the Burned Tower (and the Legendary Beasts questline) is active before Morty, and the gym opens once you beat Silver in the Burned Tower. 

I was attempting to get the Legendary Beasts quest to begin on a successful route 37 clear but for whatever reason it stopped working when I removed the quest from the bulletin, despite not changing anything else. Instead, it activates on the first successful clear of Sudowoodo, which is roughly the same location anyways.


LuchoDreamer:
I was wondering though, is this how we wanna do it? This PR would force an optional quest on us. Which is kind of contradictory. Now you need the start of the quest for progression, but you don't need to complete anymore of it for anything else.

This quest, even if the completion is not needed for progression (without the dex challenge active), forces you to catch three specific pokemon to finish. People with different and specific challenges would never be able to finish it, which is okay if it's from bulletin board, but I think it would be the only forced quest that makes you catch pokemon..

I think it's okay that the quest is unlocked after Sudowoodo instead of after Morty (though honestly I would look how to make it be able to be unlocked after Route 37 which I'm sure is possible). But I would keep it as a bulletin board quest and not make this quest a requirement for Morty.

Shinesprite41:
Lucho's suggestion makes sense so I updated this PR to that. And because I was able to get it working with the Route 37 clear req as a bulletin, it is unlocked at route 37 now instead of the sudowoodo.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3977, opened 2023-03-02
Kalos Route 20 and 21 now unlock at the same time as Snowbelle City

LuchoDreamer:
When rivals were introduced the requirements of Snowbelle City changed but not the ones for route 20 and 21. So as soon as you beat Route 19, both 20 and 21 are available which are past Snowbelle City.

This PR fixes that and now everything unlocks at the same time after the fight with Trevor in Couriway Town.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3980, opened 2023-03-04
 Fixes small details in Crown Tundra map

LuchoDreamer:
Same as the other PR, it fixes small details. In this case it's just corners missing.


LuchoDreamer:
This actually has the old .tmx, gonna close the PR and I might try to replicate the map in a new .tmx at a later date, since it's not in the project anymore.

PR #3982, opened 2023-03-05
Add EVs status into PKRS icon for routes and dungeons

Kasci:
If going for 100 completition, there is little to no information about, how much EVs are needed. Player needs to wait to see what pokemon are still being catched and check them one by one in Pokedex. Now it gives at least some overview, of how much is needed, it does not give it away, but also it gives valuable information.


src/components/dungeonView.html
```diff
@@ -119,7 +119,8 @@ <h2 class="pageItemFooter" style="display: block; font-size: 1rem;">
 
                   <!--Pokérus image-->
                   <knockout data-bind="if: RouteHelper.minPokerusCheck(DungeonRunner.dungeon.allAvailablePokemon())">
-                      <img data-bind="attr:{ src: 'assets/images/breeding/pokerus/' + GameConstants.Pokerus[RouteHelper.minPokerus(DungeonRunner.dungeon.allAvailablePokemon())] + '.png'}"
+                      <img data-bind="attr:{ src: 'assets/images/breeding/pokerus/' + GameConstants.Pokerus[RouteHelper.minPokerus(DungeonRunner.dungeon.allAvailablePokemon())] + '.png',
+                            title: RouteHelper.dungeonPokerusEVs(DungeonRunner.dungeon)}"
```
Jaaslet, line 123: Use the tooltip attribute we are using other places

src/scripts/wildBattle/RouteHelper.ts
```diff
@@ -38,6 +38,38 @@ class RouteHelper {
         return pokemonList;
     }
 
+    public static routePokerusEVs(route:number, region:GameConstants.Region): string {
+        const possiblePokemon: PokemonNameType[] = RouteHelper.getAvailablePokemonList(route, region);
+        if (this.minPokerus(possiblePokemon) == 3) {
```
Jaaslet, line 43: Use Gameconstant to make the code more readable

src/scripts/wildBattle/RouteHelper.ts
```diff
@@ -38,6 +38,38 @@ class RouteHelper {
...
+            return 'All Pokémon on this route are resistant!';
+        }
+        const currentEVs = this.getEvs(possiblePokemon);
+        return 'EVs until all Pokémon are resistant on this route: ${currentEVs} / ${50 * possiblePokemon.length}.';
```
Jaaslet, line 47: I think the automatic eslint might have ruined the strings. It needs to be ` and not ', as far as i remember

Jaaslet (changes requested):
I have not tested it, but i found some small things after looking through the code

src/scripts/wildBattle/RouteHelper.ts
Kasci: Done :)

src/scripts/wildBattle/RouteHelper.ts
Kasci: Done :)

src/components/dungeonView.html
Kasci: Tooltip is possible only in Button, IMG uses title, same as images above (pokeballs)

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3983, opened 2023-03-06
Additional Battle Cafe spins for shiny and resistant Alcremie sets

### LINKED ISSUE #3760: [SUGGESTION]: Add an additional spin at the Battle Cafe in Galar for every set of Alcremie we have gotten shiny.
### Current Behavior

Currently, the games adds an extra spin for every set of Alcremies we obtain, with the initial spins being 3 and having a max of 10 spins. 

### Improved Behavior

In addition to extra spins from obtaining all Alcremie in a set, I am proposing that making a set of Alcremie shiny should also give you an extra spin. The max amount of spins per day would increase to 17 if all Alcremie are shiny. Maybe 18 if you want to add shiny cheese into the mix even though it's not an Alcremie.

### Reason for change

This would speed up the ability to gain EVs for all Alcremie assuming you already have the berries needed. Currently, it would take about 57 days if you have an optimal Rowap farm with Power Bracers equipped (9 spins each). The math is probably wrong because we are not good at math. Anyway, it would take even longer with no boosts whatsoever regardless of if you have the required berries or not. Please note that I am not considering the RNG chance that one could obtain a shiny Alcremie while spinning. 

### Detailed Description

Making a set of Alcremie shiny should give an extra spin in the Battle Cafe, up to a max of 17 (or 18 if you add shiny cheese). This should speed things up a bit instead of taking months just to do the spins themselves. The time to do this does not consider how long it takes to farm berries.

CypherX:
Awards the player an additional Battle Cafe spin for each set of Alcremie that is shiny and each set that is resistant for a total of 24 spins.

Closes #3760


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3984, opened 2023-03-06
Made Vivillon Questline Not Auto-proc After Wulfric

EatPant2nd:
Since "The Great Vivillon Hunt!" Questline is now a Bulletin Board Questline, it no longer should auto-proc after reloading the game post-defeating Wulfric (8th Badge in Kalos). This gets rid of that line of code since it's unnecessary.


Jaaslet (approved):
I removed all the code related to that quest. Looks good. Thanks for the PR

PR #3986, opened 2023-03-06
 Kalos Route 6 now unlocks at the same time as Camphrier Town

LuchoDreamer:
Fixes the requirement of Route 6 since it was not modified when rivals were introduced.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3988, opened 2023-03-06
CustomQuest can now bring up a modal when the step is completed 

LuchoDreamer:
Added a new modal for customQuest. When a customQuest step is completed a modal will appear with the inputed dialogue, npc sprite and npc name.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #3990, opened 2023-03-06
Joey!

PixLSteam:
Joey.


CypherX:
Joey.

Jaaslet (changes requested):
The temp battle list in GameConstants is sorted, so the battle needs to be put up somewhere with the other johto battles. Other than that, it looks good

PixLSteam:
> The temp battle list in GameConstants is sorted, so the battle needs to be put up somewhere with the other johto battles. Other than that, it looks good

But Joey is the best, he deserves that spot :/

Jaaslet (approved):
Looks Joey! Thanks for the Joey!

PR #3992, opened 2023-03-07
Update Distortion World requirements

CypherX:
Currently it's possible for the Distortion World dungeon in Sinnoh to be selected as a clear dungeon quest when it wouldn't normally be accessible, a minor inconvenience. Access is restricted via the `visible` binding in `SinnohSVG.html` rather than through standard requirements, so this updates the dungeon requirements to match those bindings and prevent the dungeon from being selected when it is inaccessible.

Intended requirements:
(Cleared Spear Pillar and NOT cleared Distortion World) OR (Cleared Zero)


miaiam2006:
Wouldn’t the first of the intended requirements just mean that you get to do it once and then have to wait until you complete zero before you can finish the quest?

EatPant2nd:
You can already only clear the dungeon once (for the A New World questline) before it becomes invisible and have to wait for Zero's Ambition to access it again so nothing changes there.

miaiam2006:
Okay, fair, I could have worded that better. I know that’s how it already works. But isn’t intended requirements set 1 while you get the quest equally awkward to having it inaccessible while you get the quest, just with basically one fewer clear needed?

xslk:
> Okay, fair, I could have worded that better. I know that’s how it already works. But isn’t intended requirements set 1 while you get the quest equally awkward to having it inaccessible while you get the quest, just with basically one fewer clear needed?

Why would you attempt to complete a quest that you can only access once, before having to clear the rest of the region to unlock again, instead of refreshing the quest list?

Jaaslet (approved):
Looks good. Thanks for the PR

Issue #3960, opened 2023-02-24
[Bug]: [Real Evo] Evolved Pokemon are level 1 till a Pokémon is beaten


Razikd29:
### Confirmations

- [X] I have searched open and closed issues for any existing report of my issue

### What happened?

Started a new file with Real Evo activated and with Bulbasaur as my starter. I didn't capture any other Pokémon.
At level 16, Bulbasaur evolved into Ivysaur. However, its level was back to 1 instead of the actual level (16).

### What did you expect to happen?

I'd expect the level of my Pokemon to be directly the same as at the moment of the evolution.

### Steps to Reproduce

Start a new file with Real Evo. Take a starter (not Pikachu). Evolve it at lvl 16. You should see the fact that it is at lvl 1 for the duration of an encounter.

### What browsers are you seeing the problem on?

Chrome

### Additional Information

The behaviour was changed with #3439 but it seems like the moment between the evolution and the first encounter defeated was missed.

PixLSteam:
EXP is carried over but the level isn't recalculated before the next defeat, I also noticed this recently.

osherwott:
Fixed in #3993! Closed.

PR #3993, opened 2023-03-07
Copy base pokemon's level to evolved pokemon for real evolutions challenge mode

TheRealJoeFriel:
Quick fix for #3960, I saw that the exp for the pokemon was being carried over but the level was not. The level would be updated on subsequent gainExp calls since it would calculate the level from the level requirements and level type but until then it would remain at level 1.


Jaaslet:
I don't think every evolution requires the same amount of exp to level up, as their baseform

Razikd29:
Linked to #3960 ?

Qwertypop04:
> I don't think every evolution requires the same amount of exp to level up, as their baseform

No, all Pokemon within an eventually family have the same experience group.

PixLSteam:
Correct, that's what I read on Bulbapedia when I made the PR that copied the exp.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3999, opened 2023-03-10
Capitalized Region in MaxRegionRequirement

EatPant2nd:
Makes it so the Region mentioned in the hint of MaxRegionRequirement shows properly capitalized. Other requirements show the Region capitalized so this is for consistency's sake.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4005, opened 2023-03-11
Added shadow base code

Jaaslet:
The base code for shadow pokemon. It will not show anything in the game, but we can split the rest up in other PRs with this.
It can be saved on a party pokemon, and it can be set on a gym pokemon


src/scripts/party/Party.ts
```diff
@@ -40,21 +40,47 @@ class Party implements Feature {
...
+        this.gainPokemon(PokemonFactory.generatePartyPokemon(id, shiny, gender, shadow), suppressNotification);
     }
 
     gainPokemon(pokemon: PartyPokemon, suppressNotification = false) {
         PokemonHelper.incrementPokemonStatistics(pokemon.id, GameConstants.PokemonStatisticsType.Captured, pokemon.shiny, pokemon.gender);
+
+        if (pokemon.shadow) {
+            // Already caught (shadow)
+            if (this.alreadyCaughtPokemon(pokemon.id, false, true)) {
+                return;
```
Aegyo, line 62: If you don't have a pokemon shadow but not shiny, and then you catch a shiny shadow, this looks like it will prevent you getting shiny credit.

Aegyo (commented):
Dont have a laptop to run the code, but it looks fine other than the one comment I made

src/scripts/party/Party.ts
Jaaslet: Fixed

src/scripts/gym/GymPokemon.ts
```diff
@@ -3,9 +3,10 @@ class GymPokemon {
     maxHealth: number;
     level: number;
     shiny: boolean;
+    shadow: boolean; // Currently don't do anything. Will be added later
```
RedSparr0w, line 6: Should this not be a type of `ShadowStatus` instead of boolean?

RedSparr0w (approved):
Looks good to me, just left 1 comment, fine with whatever option you choose though 👍

Thanks for the PR!

src/scripts/gym/GymPokemon.ts
Jaaslet: It was a boolean because i did not think purified ever made sense for trainer mons. But we will have an icon, so it might make sense at some point... I have changed it

PR #4008, opened 2023-03-12
Fix Subregion Achievements being visible before the subregion is unlocked

Awec4:
I started a new game and noticed that I could see achievements for the Magikarp Jump subregion before I even left Kanto, so this should fix that. Included the Sevii Islands too for consistency although I think you unlock them before the achievement tracker.

Edit:
Reduced unnecessary filtering of the Achievements List on startup in order to fix the error with the Subregion Requirements.


DraKay:
The achievement tracker ks unlocked after get 110 unique pokemon... Not sure if that can be achieved before sevii, but most probably because it would only lock moltres and mewtwo

Jaaslet (changes requested):
The game does not currently load with this change

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4009, opened 2023-03-12
Fixes Crown Tundra map and Galar South Map

LuchoDreamer:
Last week I decided to fix some edges that were bothering me in the Crown Tundra map, and we found out that in the project the .tmx of Crown Tundra was an older version, so I replicated the actual Crown Tundra in the .tmx plus fixed the edges that were bothering me. From this I was told that it would be neat to actually connect the map with Galar South, so I made some changes in the mountains at north so now the maps of Crown Tundra and Galar South are continuous.

Because of this, I also changed a bit of Galar South so the trees in Crown Tundra show in Galar South, and also fixed some edges that were annoying me.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4010, opened 2023-03-12
Fix pokemon list attack

### LINKED ISSUE #511: Change Pokemon's Atk in the Pokemon List Depending on Region
In the Pokemon List, the Pokemon from Kanto should have their attack decreased while in the Johto region so that your Pokemon from Johto that are doing more while in Johto are higher up when sorted by the attack. This should be the same for Hoenn as well.

Awec4:
Fixes #511
The attack value displayed in the Pokemon List now changes depending on the Region.
Sorting in this list will also depend on the region.
Works with the Magikarp Jump subregion (all Pokemon except Magikarps will be displayed as 0).

### DIFF
--- a/src/components/nextRegionModal.html
+++ b/src/components/nextRegionModal.html
@@ -15,7 +15,7 @@
                    <strong>There are a few things you should know before you depart:</strong><br/>
                    You will gain <strong data-bind="text: App.game.breeding.queueSlotsGainedFromRegion(player.highestRegion())"></strong> queue slots in the Hatchery.<br/>
                    <!-- ko if: App.game.challenges.list.regionalAttackDebuff.active() -->
-                   When Pokémon are in a Region they're not native to, they will only retain <strong data-bind="text: App.game.party.getRegionAttackMultiplier(player.highestRegion() + 1).toLocaleString(undefined,{style: 'percent'})"></strong> of their total attack.<br/>
+                   When Pokémon are in a Region they're not native to, they will only retain <strong data-bind="text: Party.getRegionAttackMultiplier(player.highestRegion() + 1).toLocaleString(undefined,{style: 'percent'})"></strong> of their total attack.<br/>
                    <!-- /ko -->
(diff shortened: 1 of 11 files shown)


Farboo171:
There's been some discussion on the development team's side, and the consensus seems to be that we would prefer that this feature is togglable in the settings, and is defaulted as off. That way it provides the functionality for those who want it, but it's not defaulted and confusing for newbies or the people who don't.

src/components/nextRegionModal.html
```diff
@@ -15,7 +15,7 @@ <h5 class="modal-title" data-bind="text: `Travel to ${GameConstants.camelCaseToS
                    <strong>There are a few things you should know before you depart:</strong><br/>
                    You will gain <strong data-bind="text: App.game.breeding.queueSlotsGainedFromRegion(player.highestRegion())"></strong> queue slots in the Hatchery.<br/>
                    <!-- ko if: App.game.challenges.list.regionalAttackDebuff.active() -->
-                   When Pokémon are in a Region they're not native to, they will only retain <strong data-bind="text: App.game.party.getRegionAttackMultiplier(player.highestRegion() + 1).toLocaleString(undefined,{style: 'percent'})"></strong> of their total attack.<br/>
+                   When Pokémon are in a Region they're not native to, they will only retain <strong data-bind="text: Party.getRegionAttackMultiplier(player.highestRegion() + 1).toLocaleString(undefined,{style: 'percent'})"></strong> of their total attack.<br/>
```
Jaaslet, line 18: I don't like this refactor. Party should not have any static methods. Either use `App.game.party` or move it to a helper class

Awec4: I have moved this code into more appropriate places and removed all static methods from Party

Awec4:
> 

I have added a setting to enable/disable this. Though I'm not sure why it would be more confusing for newbies than the existing functionality

Jaaslet:
I am closing this for the same reason as #4046

Awec4:
Not impressed by this. I spent a long time putting this and #4046 together and made changes as requested. I don't think I will be contributing any more code to this project if my time isn't respected.

PR #4011, opened 2023-03-12
Orre map

LuchoDreamer:
This PR basically puts the Orre map from the Colosseum and XD games into the assets. It also adds the start of the Orre subregion into HoennSVG so it can be used to start working on the new mechanics of the region.


Jaaslet (changes requested):
It looks good! Just have some comments:
- Missing credits (at least i think some of the new assets are not from the games)
- Any chance you could add one town to TownList.ts? And add it as the starting town, just so it's easier to add stuff while developing? Or do we not know yet what should be a dungeon, and what should be a town?

LuchoDreamer:
Changes requested are done. Also Realgam Tower is in the map, so the map is fully complete unless extra changes are requested at a later date o/.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4012, opened 2023-03-12
Players can now auto restart in BF as well

### LINKED ISSUE #3711: [SUGGESTION]: Autostart Battle Frontier Button
### Current Behavior

Currently, you have to start each Battle Frontier pass manually.

### Improved Behavior

Adds a buyable repeat button to the Battle Frontier Shop for x BP.  Buying this adds the repeat button to the Battle Frontier Start button.

### Reason for change

I believe that there are many people that aren't able to be at their computers all the time, and having such a button would help those out that aren't able to spend a lot of time at their computers or want to do something else, or would like to go to sleep at some point in time.

### Detailed Description

This suggestion would add an unlock repeat start to the Battle Frontier Shop for a certain amount of Battle Points.  Once bought, this adds a repeat button to the start button of the Battle Frontier, which allows you to autorun the Battle Frontier, **however**, you only earn *half* of the BP that you normally would if you did it manually.

The cost of the unlock of the button can be something like 500-1000K BP, or the other route can be gone and it could only cost like 200K BP.  I'm not really sure what to set the price at for the unlock.

RegisCoaxans:
Closes #3711.
I have arbitrarily set the price at 10% of the last earnings, following my base idea, but it can be changed to whatever amount, or with whatever formula.
I have also set that this would not be available before having cleared up to stage 100 once as it only takes 20 minutes in...


Someonealive-QN:
I feel like it should be free. It isn't overpowered and it only makes it quicker to restart a run without other benefits(i think). And if it were to cost something why not money instead of bp? Taking 10% of bp someone has spent hours trying to get doesn't seem fair for something that takes like 1-2s to make it's action

RegisCoaxans:
I do not feel like locking that behind such a wall. While it looks like auto start on gyms, it can not work the same.
Auto restarting a run allows for time and profits optimisation for people on long runs, sure, but also avoids micro management for players on short runs. It should quickly be available; that is more about a QOL feature than auto restarting a gym. Thus, if available shortly after some runs in the BF, people should pay for it.

However, the price may need to scale differently.
When designing this, I also thought about it costing the (highest ?) amount of stages beaten. The cost would be high on short runs but would scales down very quickly after all, compared to the earnings.

Razorflamekun:
I think you mistook my original suggestion for something else.  This button is not to be used to _restart_ a current BF run, it is meant to be able to have you automatically start a new one once you finish the previous one.  If you quit out of the BF, the autostart button will not trigger, because you didn't finish the run.  That's how I envisioned my original suggestion working.

Everyone in the #general of poke-crew agreed that it should be unlocked after achieving 25,000 total stages cleared.  If you would like to discuss it further, please join us there (if you cannot see that channel, ask for the CC role).

RegisCoaxans:
Come on. I have not mistaken your suggestion. What I have coded actually starts a new run once the previous one is lost.

Razorflamekun:
Ok, my bad.  Apologies.

RegisCoaxans:
It should unlock as soon as grinding in the BF becomes usefull and / or required, especially because it is when that feature happens to be used more frequently thanks to runs being short.

Stage 1000 is mid Galar difficulty, I think. You can have a real use for BP in Kalos already via Furfrou and megas so I would say around 600 should work.
Does it feel fair ?

RegisCoaxans:
What about the cost, then ?

Razorflamekun:
I'll leave it up to you since you seem to have a strong opinion of it costing some amount of BP in order to autostart.  10% seems fine to me, to be honest.  If we extrapolate that out, I currently get 34K BP a run, a 10% penalty would only cost me 3.4K BP, which I'd be more than happy to pay.

RegisCoaxans:
I have set the price to 5%. I will wait for some more news about that and change again if required.

Razorflamekun:
After talking it over in the general chat of the Poke Crew, we felt like a price of 10-20% would be more appropriate in addition to the current unlock requirement.  Thanks again for all your work on this!

Farboo171:
That was the crew consensus, but devs will have the final say.

Razorflamekun:
Yes, of course.  I did not mean to imply that.

RegisCoaxans:
Could the related issue be removed from the pool of approved suggestions, then ?

Aegyo:
> This will remove the only disadvantage it has

BF also has the disadvantage of not being able to catch pokemon, so not DT or EVs. I think that is a way bigger disadvantage than having to click a button every x hours.

RegisCoaxans:
I should also mention that this PR was part of a much bigger picture from a suggested [balance rework](https://github.com/pokeclicker/pokeclicker/issues/4078).
The master goal behind this being to ease pushing runs forward, even when AFK.

That PR was not meant to help on the BP grind, though. It aims to incentivize people in actually reaching the ordinary and expected end of their run, with an end-of-the-run-only feature (for you do not benefit from it if you actively shorten your run).
Keystones may have been the reason for Razorflame to suggest that change, although they just meant to me the appropriate moment to unlock an auto-start feature as a QOL improvement.

About eggsteps, I would rather ask for another PR to globally rebalance how they are computed. In and out of the BF.
Beware, though : the average eggstep amounts are not that good on a complete run... Not to mention again that the double speed, of whom I want to get rid, made the BF that overpowered, did the complete opposite of its original purpose (which is that of my changes as well), and can easily be gone for good.

If that autostart feature needs to be nerfed, I am all about it.
If the cost has to be something else, I am open for suggestions. This costing BP was the most consistent I could think of when designing the thing.

Razorflamekun:
Actually, I just wanted an in-game way of doing more than one BF run at a time because I didn't want to have to use a Macro to do it.  The main reason why I was wanting a cost was so that there was some drawback for the advantage that it gave you.  I also agree that there are much bigger disadvantages to using the BF than being able to AFK the BF.  The main point that Regis makes above is that this feature **WOULD ONLY WORK IF YOU DO NOT QUIT THE RUN IN THE MIDDLE OF IT AND YOU COMPLETELY FINISH THE RUN.**  

I find the fact that this **approved suggestion** was rejected quite concerning, honestly, because if you wanted to reject it @Jaaslet **then why was it approved in the first place?**

PR #4018, opened 2023-03-14
maxHealth formula improvement

umbralOptimatum:
Simplified the maxHealth formula with the power of algebra. Double-check my math but it should be algebraically equivalent to the old formula, while much more readable and minus two redundant function calls.

The +/- 10% comment would benefit from an edit for clarity—the variable health can't hit a full -10% but it could theoretically exceed +10%.


Jaaslet (approved):
The numbers checks out. Thanks for the PR

PR #4021, opened 2023-03-14
Pokemon List can now be sorted by Attack Bonus

### LINKED ISSUE #3731: [SUGGESTION]: Allow sorting by Attack Bonus
### Current Behavior

Currently, the game allows you to display attack bonus, but does not allow sorting by this value.

### Improved Behavior

I'd like to be able to sort by the Attack Bonus value that can already be displayed.

### Reason for change

The value is already calculated, and it would be useful to be able to sort by it for cases like afking to go eat and you just fill up your queue with the highest attack bonus pokemon.

### Detailed Description

Ideally, in this picture, I'd want to sort by attack bonus, which would cause Manaphy to come before Jirachi.

image

LuchoDreamer:
This is mainly for the hatchery, so people can sort by Attack bonus affected by regional debuff so they can make more informed decisions.

Closes #3731


PixLSteam:
I believe we need an update function if we don't want to mess everybody's sorting up.
This would affect the following settings:
- partySort
- hatcherySort
- vitaminSort
- heldItemSort

As well as the hatchery helpers.

LuchoDreamer:
Added update so all the sorting possibilities remain the same after adding this one in the middle of the list, thanks PixL

Jaaslet (approved):
I have moved the update function. Looks good. Thanks for the PR

PR #4027, opened 2023-03-17
Hoenn Starter Mega Stones

Farboo171:
Adds the ability to Purchase Hoenn starter mega stones in Fallarbor city. The stones get added to the Gem shop below the flutes.

Gem Deals are modified such that it is now possible to make items no longer visible.

The temporary battle that unlocks purchase of the mega stones is early in Delta Episode, and players will get the mega stone of their starter for free. If players did not select a starter, they will get a free keystone instead.


Jaaslet (approved):
Looks good. Thanks for the PR

I am considering if the stones are too cheap. But let's merge this, and discuss the price on discord. We can open a new PR, if we want to change the price

PR #4031, opened 2023-03-19
Bulletin Board Deoxys Questline

Qwertypop04:
Makes "Mystery of Deoxys" a bulletin board questline, seeing as it is no longer required to access the Battle Frontier, and is now just a regular legendary quest


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4032, opened 2023-03-20
Improve the Pokedex with 3 new features

DreamNya:
### Changelog

- Add 3 new filter features to the Pokedex, now we can search pokemons by `ID`, `Obtain Method` and `Mega Stone`.

- Change the pokedex layout to show more pokemon blocks when the Filters active.

- Move the `enum` `PokemonLocationType` from private of `PokemonHelper.ts` to global of `GameConstants.ts`.

- Rewrite the `PokemonHelper.getPokemonLocations`, now you can specify the obtain method argument if you only want to calculate a few of the methods before calculating all the methods, which can speed up the calc code. It's useful in the new pokedex filter.
also rewite the `PokemonHelper.hasEvableLocations`.

### Previews
layout
obtainmethod
ID
megastone

### Others(Murmur?)
I think the current Pokedex is not good enough, and it lacks something helpful for new players, the three new features will be useful in my mind so I code them.  

It's my first time to code TypeSctipt with unskillful coding, poor English.  
Feel free to point the mistakes if any.  

I change a lot of codes this time. I don't know if my codes is reasonable and if it affects others.  

If the PR is merged I will also work this on the Hatchery filters.

### DIFF
--- a/src/components/pokedex.html
+++ b/src/components/pokedex.html
@@ -29,15 +29,15 @@ aria-labelledby="pokedexModalLabel">
                     <div class="collapse" id="pokedex-filter-container" style="margin-bottom: 0.5rem;">
                         <div class="container-fluid py-3">
                             <div class="form-row text-left">
-                                <!--Name-->
-                                <div class="form-group col-md-6 col-6">
-                                    <label>Name</label>
-                                    <input id="nameFilter" class="form-control" placeholder="Bulbasaur" value=""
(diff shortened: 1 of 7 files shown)


Jaaslet:
I have merged the big pokedex layout PR we had lying around. So you have a ton of conflicts (sorry!).
But that one added the mega stone filter, and we don't want the obtained method filter.
Searching for ID and some of the code cleanup would be good to have

PR #4036, opened 2023-03-22
Pokedex treats regional forms as normal

osherwott:
### Proof:

Galarian Zen:
-No hide alt toggle:
image
-With hide alt toggle:
image

Pokedex:
- Before Galar:
before

- After Galar:
after


Jaaslet:
Can you give a more detailed description of what the intended effects are?

osherwott:
This lets all pokemon that are required for dex completion be known to the player upon entering the region those pokemon are native to. It also prevents regional forms that are treated as part of the Alola and Galar dex filters from being hidden via the form toggle, as there may be cases where you'd want to hide forms like Minior or Alcremie but not the regional variants

Jaaslet (changes requested):
I started a new save, and got this pokedex:
image

I think the alternative form slider should still hide alternative forms, as they are alternative forms.
But i don't mind them showing up as uncaught, when you reach Alola. I think it should be solved by using something like `&& basePokemon.nativeRegion == pokemon.nativeRegion` for the `// Alternate forms that we haven't caught yet` part of the code, and then remove the other change

Qwertypop04:
That's also showing alt forms that aren't obtainable. That probably needs fixed in some way?

osherwott:
1. Oh geez, yeah that's not supposed to happen...
2. Regional forms are quite distinct from regular alts in ways that affect gameplay. They have unique typings, stats, and different regional debuffs, which is why I opted to not hide them like regular alts even after capture

osherwott (commented):
Added the specification of not having a native region or having a native region of none to hide uncaught forms (since a native region is optional, non-regional alts have an undefined native region, which is what caused other forms to appear in testing)

Also added clause to prevent regionals from hiding uncaught base forms

Jaaslet:
I wrote some code:
`            // Alternate forms that we haven't caught yet
            if (!alreadyCaught && pokemon.id != Math.floor(pokemon.id)) {
                const hasBaseFormInSameRegion = pokemonList.some((p) => Math.floor(p.id) == Math.floor(pokemon.id) && p.id < pokemon.id && PokemonHelper.calcNativeRegion(p.name) == nativeRegion);
                if (hasBaseFormInSameRegion) {
                    return false;
                }
            }`

And you also need to change
`if (nativeRegion > GameConstants.MAX_AVAILABLE_REGION || nativeRegion == GameConstants.Region.none && !alreadyCaught) {`
to
`if (nativeRegion > player.highestRegion() || nativeRegion == GameConstants.Region.none && !alreadyCaught) {`

This looks more clean, and solves the problems i found.
And for the hide regional forms button: Players wanted it because they thought there was too much clutter in the pokedex. So i think we want to keep it as it is

osherwott:
Your code have been integrated successfully!

As for regionals, they will not be hidden, for the reasons stated above. They offer different gameplay services from regular alts, and cannot be considered clutter in the same regard as them. We can change this later if needed, but this PR is ready to merge otherwise.

Jaaslet (changes requested):
Can you rewrite the two other if's to match the code i gave you? it's a lot more clean, and future proof.
If not, i can help you

Jaaslet (approved):
I cleaned up the code, and did some optimizations. Now it's only running through the pokemon list, when necessary, and not for every pokemon.
Looks good. Thanks for the PR

PR #4037, opened 2023-03-22
Match hatchery dropdown for Pokemon ID #

### LINKED ISSUE #2121: [SUGGESTION] Hatchery Sort Improvements
**Current Behavior**

The hatchery sort has a long list of sorting options, but some are a bit misleading and one of them feel useless.

- [ ] **Level** feels useless because only Pokémon at Lv100 show up. I don't know why it's an option.
- [ ] **Pokémon #** mean the same as "Pokédex ID", I'm assuming. Would be nice if the text is consistent between both "Display Value" and "Sort" for this one.
- [x] **Egg Steps** sorting is wrong.
- [x] **Times Hatched** sorting is wrong.
- [x] **Category** sorting is wrong.
- [x] **Proteins Used** sorting is wrong.

**Improved Behavior**

1. Remove "Level" from the Sort dropdown.
2. Change "Pokémon #" to "Pokédex ID".
3. Fix the sorting for **Egg Steps**. If the calculation for this is different than the "Display Value" option for "Egg Steps", then it needs clarification. In the screenshots below, if the sort for **Egg Steps** is High to Low, I expect to see Pokémon with highest "Egg Steps" value to be first in the list, not the opposite.
4. Fix the sorting for **Times Hatched**. Similar to No.3's explanation.
5. Fix the sorting for **Category**. When I sort by **Category**, High to Low, why are Pokémon without categories shown first? When sorting Low to High, the list orders the Pokémon with the latest category (bottom option in the "Category" dropdown) first, then works backwards. Again, if this is logic is different, it should be clarified or be named differently because it's confusing when they don't match up with other options with the same name "Category".
6. Fix the sorting for **Proteins Used**. Similar to No.3's explanation.

**Reason for change**

I thought of creating this as a bug report, but I'm not sure if the functionality is correct, but the options' names are misleading because they match up with other options in the Filters section. I figure I'll just call it an Improvement to either clarify the sorting options and if they are bugs, let me know and I can create a new issue tagged as Bug.

It's an improvement because some of the sort options behave differently than the other options when it comes to the arrows sorting them in reverse order, at least it feels that way to me. This causes confusion when planning out the Helpers' sorting because I don't know which arrow is correct when the sort option in the Pokémon tab feels wrong.

**Detailed Description**

While the sorting bug is not a big deal for the Pokémon tab of the Hatchery popup because I can see the sort and thusly, I can toggle the arrows, but it is awkward and unintuitive for some of the sort options to behave differently. This has larger impacts for the Hatchery Helpers' sort because I don't know how they work or if they're consistent with the sorting used in the Pokémon tab. If I want to sort by "Egg Cycles", which arrow direction is correct? Consistency with text would be nice "Egg Cycles" vs "Egg Steps". If they are different, then having some way to clarify the difference in-game or in the wiki would be helpful.

**Screenshots:**

I'll only screenshot the **Egg Steps** one but the others I've mentioned are similar.
05-072349_hsNjW6t8IL
05-072401_cIbrHlNoAy

osherwott:
The dropdown selections in the hatchery now match and both say `Pokemon ID #` instead of one saying `Pokedex ID` and the other `Pokemon #`

Closes #2121

### DIFF
--- a/src/modules/settings/SortOptions.ts
+++ b/src/modules/settings/SortOptions.ts
@@ -27,7 +27,7 @@ export type SortOptionConfig = {
 
 export const SortOptionConfigs: Record<SortOptions, SortOptionConfig> = {
     [SortOptions.id]: {
-        text: 'Pokémon #',
+        text: 'Pokémon ID #',
         getValue: (p) => p.id,
     },
 
--- a/src/modules/settings/index.ts
+++ b/src/modules/settings/index.ts
@@ -244,7 +244,7 @@ Settings.add(new Setting<string>('breedingDisplayFilter', 'breedingDisplayFilter
         new SettingOption('Times Hatched', 'timesHatched'),
         new SettingOption('Breeding Efficiency', 'breedingEfficiency'),
         new SettingOption('Steps per Attack Bonus', 'stepsPerAttack'),
-        new SettingOption('Pokedex ID', 'dexId'),
+        new SettingOption('Pokémon ID #', 'dexId'),
         new SettingOption('Vitamins used', 'vitamins'),
         new SettingOption('EVs', 'evs'),
     ],


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #3727, opened 2023-01-07
[SUGGESTION]: Monotype Challenge


Crobat4:
### Current Behavior

No proper monotype challenge.

### Improved Behavior

Add a monotype challenge compatible with other challenges, catching-related quests, etc.

### Reason for change

Just a challenge for those who decided to go with a self imposed monotype challenge and want the game to ignore Pidgey's damage.

### Detailed Description

Just suggesting a proper monotype challenge that lets you select a type and only Pokémon that contains the said type are able to attack, similar to how Magikarp Jump subregion works. If it gets approved, I can provide a PR about this.

The changes I made were to try to tackle all possible softlocks that can occur depending on the selected type.

Changes compared to a normal run:
* Only Pokémon that contains the selected type can deal damage.
* Badge level cap removed to being able to use the hatchery to avoid any damage-related softlocks.
* Hatchery unlocked after completing Viridian Forest to avoid Electric monotype softlock on Brock.
* Quests unlocked after completing Viridian Forest to being able to purchase Mystery Eggs.
* Town Map unlocked after catching the starter to avoid Pidgey softlock in Ground monotype.
* Added a Safari-esque feature in every starting town that increase egg steps to avoid hatchery softlock if all Pokémon that can deal damage are in there (assuming no click attack challenge).
* Kanto original starters are replaced by a single starter depending on the selected type.

Other features/restrictions:
* Dark-type can't be selected.
* Challenge can't be disabled until the player gets the Earth Badge to avoid any possible unintended behavior regarding Badge level cap removal and hatchery unlocking earlier.
* Party list shows all Pokémon that can deal damage on the top of the list and the ones who can't are greyed out (This also applies in Magikarp Jump).
* On older saves, the monotype starter is given if the challenge is enabled.
* Starter can't be skipped anymore (Can be restricted to Monotype challenge if other self imposed challenges needs the skip but this was made to avoid another softlock).

Some screenshots:
<details><summary>Pics</summary>

![image](https://user-images.githubusercontent.com/104547700/211128800-b76e7aa8-641a-4aed-972f-2e28b13a1b09.png)
![image](https://user-images.githubusercontent.com/104547700/211128822-62e45a69-cdc8-4d9c-a39f-37319c301243.png)
![image](https://user-images.githubusercontent.com/104547700/211128943-f17b5676-cccf-4155-9960-125262dacbf9.png)
[...]

PR #4038, opened 2023-03-22
Monotype Challenge

Crobat4:
Closes https://github.com/pokeclicker/pokeclicker/issues/3727
This PR adds a Monotype challenge as a special challenge, different to the standard ones, also a proper layout to separate standard and special challenges.
<details>
<summary>Challenges layout</summary>

image
</details>

For the challenge, you need to select a type first and you'll get a starter of that type, then some things will change compared to a normal gameplay:
* You can catch any Pokémon, but only Pokémon that contains the selected type will deal damage, count for the click damage and yield Dungeon Tokens when caught. Magikarp Jump will still work only with Magikarps but they won't yield DT (unless Water monotype).
* To ease the DT farm a bit, a toggle to catch only Pokémon of the selected type was added.
<details><summary>Poké Ball toggle</summary>

image
</details>

* Town Map unlocks after the starter fight.
* "Catch 5 Pidgey" quest will change into "Catch 5 [starter]" with the challenge enabled.
* The starter is added in one of the routes before the first dungeon of each region.
* Hatchery will be unlocked after completing "Catch 5 [starter]" quest and it will work with level capped Pokémon instead lv 100.
* Normal quests will be unlocked after completing Viridian Forest.
* "Catch X-type Pokémon" normal quest will always ask for Pokémon of the selected type.
* Johto - Galar starters were also changed to match the selected type (Johto and Kalos starters are skippable with Dragon and Ground respectively since the only Pokémon of those types are evolved forms and/or legendary Pokémon).
* Monotype egg added with its own type pools. For the pools, I tried to give priority to evolved forms that gain a type the previous form doesn't have (ex. Cloyster in the ice pool, Poliwrath in the fighting pool, etc.) and they aren't found in the wild in the region where they were introduced. 
    * Typed and Mystery eggs pools are unaffected, Monotype eggs for the already available typed eggs have the same pool as those but with some Pokémon added following the previously mentioned criteria.
* Safari-esque feature was added to the starting town of each region that increase egg steps in the hatchery to avoid softlock if all Pokémon that can deal damage were added in there and click attack is disabled
<details><summary>Walk around [starting town]</summary>

image
</details>

* Tutorial skip temporary battle: Made this battle grabbing some stuff from the [RGB Old Man glitch](https://bulbapedia.bulbagarden.net/wiki/Old_man_glitch).
    * To enable it, you need to change your profile name to "OLD MAN", then, the battle will appear next to Cinnabar and you fight a random Pokémon. Winning the fight will complete all quests from the "Tutorial Quests" questline and giving you the quests rewards.
* Code to release non-selected type Pokémon: I heard releasing Pokémon is not wanted but I decided to try anyway, it can be reverted if anything. 
    * The code will release all Pokémon whose types don't match with the one the player selected. 
    * To use the code, you need the Monotype challenge to be enabled and the Require Complete Pokédex completion challenge to be disabled. 
    * It's repeatable.
<details>
<summary>Code is FREE-TRIP-TO-POKEMON-TOWER</summary>
Text grabbed from Sonic Adventure.

image
</details>

That's it, I think, hopefully I didn't forget anything. I'll add starters names tomorrow since in the code they are by index numbers.
Stuff grabbed from the issue and the Arduous spreadsheet which is also in the issue.

Tell me if I need to split something.


Crobat4:
> Will be solved in #3837

I know, this is just a small solution in the meantime, can be removed if is not wanted.

> Why?

Ground starter is Cubone and it will be softlocked by Pidgey, the map that early allows the player to move to Pallet Town to reroll.
Just wanted to add a pure ground starter, if anything, it can be changed to Rhyhorn, Onix or Geodude.

> I think the cleaner solution would be to just skip the tutorial, when the challenge is active. People should understand the game, if they select that challenge. And if not, we could add so the challenge can only be started, if you have at least one other save... Might be a good idea anyways.

I agree with the cleaner solution since I had to re-add the quest to change the Pokémon, this was made in case someone doesn't want to use the tutorial skip for some reason but if you prefer I can revert the quest thing and just use the tutorial skip.

> This seams weird and like it makes the start of the game easier. Is it nessesary?

Softlocks basically, for example, Dratini alone can't clear Viridian Forest without damage boost or click attack, Electric needs the hatchery before Brock because immunities.
I know the shiny code exists, but I tried to make the challenge without taking it in account.

> You can still catch the other types, right?

Yes, but the change was made in case someone doesn't want to catch Pokémon of the type they didn't select.

> I think we should just skip those starters instead.

If you prefer, I can remove them and skip them all.

> I think this was solved by adding mons to the mystery egg pool? Can't remember the PR...

Yes, I just wanted to give a bit more options plus let the players get Pokémon that gain the selected type on evolutions in case they didn't want to catch their preevolutions because their types don't match with the one the player selected.

> Reference to missing no. will lead to way more questions about the pokemon on the discord. And we already have too many of those.

If the skip itself is fine, MissingNo. can be removed from the pool and the overworld icon can be replaced with something else.

> With these restrictions, i think the code is fine. But it needs an "are you sure?" popup, if it does not already have one.

The popup is the one I showed in the first post:
<details>
<summary>Pic</summary>

image

</details>

> Overall, it seems like there are a lot of stuff to make it easier/do-able. I am not sure we need all of it (if any). I think we should add just the basic stuff, locked behind a DevelopmentRequirement, let the crew test it, see if some of the types are impossible, and then slowly add stuff to make it easier. This also does so i don't need to take a couple of days off work to review the PR. Generally, if you can split it up, it would help us a lot.

Sure, I'll try to split it.

> And a sidenote: sorry for not reviewing your Pokedex PR. Have been busy with a new job. Hoping to review it this week (will solve conflicts, unless you have time for it)

No worries. I'll try to solve the conflicts within the week.

Jaaslet:
Sorry for the lack of response. I will try to answer them now:

> I know, this is just a small solution in the meantime, can be removed if is not wanted.

Hopefully it won't be long before we get the filters. So let's wait with them.

> Ground starter is Cubone and it will be softlocked by Pidgey, the map that early allows the player to move to Pallet Town to reroll.
Just wanted to add a pure ground starter, if anything, it can be changed to Rhyhorn, Onix or Geodude.

Let's change it to Geodude. Skipping a route is sorta weird

> I agree with the cleaner solution since I had to re-add the quest to change the Pokémon, this was made in case someone doesn't want to use the tutorial skip for some reason but if you prefer I can revert the quest thing and just use the tutorial skip.

Hopefully no one wants to do mono-type and still insists on doing the tutorial. Let's add the skip.

> Softlocks basically, for example, Dratini alone can't clear Viridian Forest without damage boost or click attack, even if the boss is in the first square (iirc it needs 63 seconds or so because 8 damage per second), Electric needs the hatchery before Brock because immunities.

I don't really like the solution. Can we change the starters or something instead?

> If you prefer, I can remove them and skip them all.

I do prefer that

> If the skip itself is fine, MissingNo. can be removed from the pool and the overworld icon can be replaced with something else.

Let's just use a simple code to skip the tutorial. The other idea seems too complicated for something players just want to skip

PR #4043, opened 2023-03-25
More Pokemon +

Qwertypop04:
#3273 added more Pokemon into canon locations. Some of these Pokemon were added to only one location when they are available in multiple locations. This PR adds them to the other locations, and adds a couple extra Pokemon to canon locations

Also changes the Floette in the Tapu Ruins from all being Red form. This isn't canon, but they aren't canon spawns at all in these locations


LuchoDreamer:
Everything except the Swanna is correct according to canon, even the weathers.

xslk:
swanna spawns on marvelous bridge between route 15 and 16 in unvoa, since marvelous bridge doesnt exist in clicker the options are 15 or 16 or both

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4044, opened 2023-03-25
Longer name fields, Pokémon list always visible (mobile) in Pokédex/Hatchery and some fixes

Crobat4:
As requested in https://github.com/pokeclicker/pokeclicker/pull/3551#pullrequestreview-1354066486, this PR makes some changes to the Pokédex/Hatchery layout:
* Name fields width increased
* Pokémon list always visible in mobile in both modals (removed the collapse button but can be re-added if wanted)
* Hatchery: Changed Category select to a Bootstrap dropdown
* `Edit category` button moved to Category dropdown
* Reorganized filters positions

Fixes:
* Category/Region dropdown text being white on hover on dark themes
* Category dropdown text being overflowed
* Category still being selected even after being deleted (added code about this in `Update.ts`)
* Double scroll in both modals
* Neither filters/Pokémon list and collapsible buttons appearing in certain screen sizes

Also, an idea came up about adding a third option to the Mega filter to show Pokémon with available Mega Evolution but without Mega Stone, mostly for the Hatchery, so if you want, I can try to add it here or in another PR.

image


Jaaslet (approved):
Looks good. Thanks for the PR

And for your suggestion for the filter: I like the suggestion. I have already considered adding it

PR #4045, opened 2023-03-25
Fix: Evolution items modal text colors

Crobat4:
Items/Pokémon names were unreadable in some themes due text having the same or similar color as the background, this PR fix those.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4046, opened 2023-03-25
Multiselect Types in Breeding Modal

Awec4:
Adds the ability to select multiple types in each of the types filters in the Breeding Modal, as was requested in the Discord.

### DIFF
--- a/src/modules/settings/BreedingFilters.ts
+++ b/src/modules/settings/BreedingFilters.ts
@@ -24,25 +24,17 @@ const BreedingFilters: Record<string, FilterOption> = {
             new SettingOption('Shiny', '1'),
         ],
     ),
-    type1: new FilterOption<number>(
+    type1: new FilterOption<number[]>(
         'Type 1',
-        ko.observable(-2).extend({ numeric: 0 }),
+        ko.observable([]),
         'breedingTypeFilter1',
-        [
-            new SettingOption('All', '-2'),
-            ...Settings.enumToSettingOptionArray(PokemonType, (t) => t !== 'None'),
-            new SettingOption('None', '-1'),
-        ],
+        Settings.enumToSettingOptionArray(PokemonType, (t) => t !== 'None'),
     ),
-    type2: new FilterOption<number>(
+    type2: new FilterOption<number[]>(
         'Type 2',
-        ko.observable(-2).extend({ numeric: 0 }),
+        ko.observable([]),
         'breedingTypeFilter2',
-        [
-            new SettingOption('All', '-2'),
-            ...Settings.enumToSettingOptionArray(PokemonType, (t) => t !== 'None'),
-            new SettingOption('None', '-1'),
-        ],
+        Settings.enumToSettingOptionArray(PokemonType, (t) => t !== 'None'),
     ),
     region: new FilterOption<number>(
         'Region',
--- a/src/modules/settings/MultiSetting.ts
+++ b/src/modules/settings/MultiSetting.ts
@@ -1,0 +1,42 @@
+import BaseSetting from './BaseSetting';
+import SettingOption from './SettingOption';
+
+export default class MultiSetting<T> extends BaseSetting<T[]> {
+    private optionValues;
+
+    constructor(
+        name: string,
+        displayName: string,
+        options: SettingOption<T>[],
+        defaultValue: T[],
+    ) {
+        super(
+            name,
+            displayName,
+            options,
+            defaultValue,
+        );
+        this.optionValues = options.map(o => o.value);
+    }
+
+    validValue(value: T[]): boolean {
+        if (!this.isUnlocked(value)) {
+            return false;
+        }
+
+        if (value.length === 0) {
+            return true;
+        }
+        if (this.options.length === 0) {
+            return false;
+        }
+
+        for (let i = 0; i < this.value.length; i += 1) {
+            if (!this.optionValues.includes(value[i])) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+}
--- a/src/scripts/party/PartyController.ts
+++ b/src/scripts/party/PartyController.ts
@@ -206,5 +206,36 @@ class PartyController {
         };
     }
 
+    public static filterByTypes(partyPokemon: PartyPokemon, filterTypes1: PokemonType[], filterTypes2: PokemonType[]) {
+        const numTypes = GameHelper.enumLength(PokemonType) - 1;
+        const { type: [type1, type2] } = pokemonMap[partyPokemon.name];
 
+        // Check if all types are selected or if the first type is included in the first filter.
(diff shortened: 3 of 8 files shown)


src/modules/settings/BreedingFilters.ts
```diff
@@ -24,25 +24,17 @@ const BreedingFilters: Record<string, FilterOption> = {
             new SettingOption('Shiny', '1'),
         ],
     ),
-    type1: new FilterOption<number>(
+    type1: new FilterOption<number[]>(
         'Type 1',
-        ko.observable(-2).extend({ numeric: 0 }),
+        ko.observable([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17]),
```
Aegyo, line 29: This should be created from PokemonType somehow, something like `GameHelper.enumNumbers(PokemonType).filter(v => v !== PokemonType.None)`

src/modules/settings/MultiSetting.ts
```diff
@@ -0,0 +1,30 @@
...
+
+        if (value.length === 0) {
+            return true;
+        }
+        if (this.options.length === 0) {
+            return false;
+        }
+
+        if (this.optionValues === undefined) {
+            this.optionValues = this.options.map((o) => o.value);
```
Aegyo, line 19: Would be cleaner to create this in a constructor, and probably better to use a Set since we are just using to check if values exists in there.

src/modules/settings/MultiSetting.ts
```diff
@@ -0,0 +1,30 @@
...
+        if (!this.isUnlocked(value)) {
+            return false;
+        }
+
+        if (value.length === 0) {
+            return true;
+        }
+        if (this.options.length === 0) {
+            return false;
+        }
```
Aegyo, line 16: These checks don't seem needed, both get covered by the "are all values included in optionValues" bit

src/modules/settings/Settings.ts
```diff
@@ -25,6 +25,11 @@ export default class Settings {
         return this.list.find((setting) => setting.name === name) || null;
     }
 
+    static isContainedInMultiSetting(name: string, value: any) {
+        const setting = Settings.getSetting(name);
+        return setting !== undefined && Array.isArray(setting.value) && setting.value.includes(value);
+    }
+
```
Aegyo, line 32: This feels like it should just be an override of `isSelected` in the MultiSetting class

src/scripts/party/PartyController.ts
```diff
@@ -206,5 +206,35 @@ class PartyController {
         };
     }
 
-
+    public static filterByTypes(partyPokemon: PartyPokemon, filterTypes1: PokemonType[], filterTypes2: PokemonType[]) {
+        const numTypes = GameHelper.enumLength(PokemonType) - 1;
+        const { type: [type1, type2] } = pokemonMap[partyPokemon.name];
```
Aegyo, line 211: why `PartyPokemon`? I think this should just take the pokemon name, for when we inevitably want to use it in the pokedex filtering too

Aegyo (changes requested):
I think the UI needs to look and behave the same as the region multiselect (and possibly make the Region multiselect use your new setting type, maybe for another PR though)

src/modules/settings/MultiSetting.ts
Awec4: I tried doing it in the Constructor at first but I kept having errors where it wasn't being set and I have no idea why. This was the only thing that seemed to work

Aegyo: The options are passed into the constructor so I don't see any reason it wouldn't be able to set this there?

Can you share how you were doing it?

src/modules/settings/MultiSetting.ts
Awec4: Removed

src/modules/settings/Settings.ts
Awec4: Changed to be an override of isSelected

src/scripts/party/PartyController.ts
Awec4: Changed

src/modules/settings/MultiSetting.ts
Awec4: The base constructor calls validValue on the default value, but we can't set the optionValues before this as we are not allowed to access 'this' before calling 'super'
The only alternatives I can think of are to remove the call to validValue for the default value or to create a new empty function called in the base constructor which sets the optionValues in an override

src/modules/settings/BreedingFilters.ts
Awec4: Changed

src/modules/settings/MultiSetting.ts
Aegyo: Guess we can just leave it for now then

Aegyo:
I've made a PR to your fork for some UI changes to this Awec4/pokeclicker#1

CypherX:
We recently merged a multi-select control template in #4342, would you be able to work it into these changes?

The template, example usage, and documentation can be found in `src\components\templates\multiSelectDropdownTemplate.html`

Awec4:
> We recently merged a multi-select control template in #4342, would you be able to work it into these changes?
> 
> The template, example usage, and documentation can be found in `src\components\templates\multiSelectDropdownTemplate.html`
> 
> You may need to add a property to the template to allow passing in an `onChange` function, or alternatively subscribe to the breeding filter observables elsewhere to handle calling `Settings.setSettingByName()`

Reworked to use the Template.

src/translations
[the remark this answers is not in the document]
Aegyo: don't worry about, release process will overwrite that anyway.

Jaaslet (changes requested):
It's a ton of changes for a "change one select" description. It has a lot of refactors without any explanations. And it affects all of our settings code, and also some specific settings (which makes it hard to test, because it's a long list of stuff i need to test, and it's not easy for me to figure out what stuff i need to test).

- Why does the selects change color on hover (makes them unreadable in Darkly)?
- I don't think the type boxes needs to be that big
- It takes for ever to select one type (which is the drop downs primary use)

These are the things i have found so far

Jaaslet:
I am going to close this PR.
It's a very small thing, but the PR changes 17 files. And none of the changes are simple. It will take a long time to review and test, to make sure stuff still works. There is no reasoning for why all the refactoring are there, and i can't see a reason myself. And we don't have time for refactoring, just for refactoring sake. The old code was good enough, i think.
Sorry for waiting this long. I have looked at it many times, and each time i just have not had the time to go through the PR. There are just too much i need to review, which takes way less time and has way more impact.
I am not likely to ever review this PR (and it looks like no one else is either), so it does not make sense to just stay here, ignored. Therefor I think it's best to just close it.

PR #4047, opened 2023-03-26
Fix mine.skipsRemaining TypeError

umbralOptimatum:


CypherX:
What is there to fix? There is nothing wrong with that code. Your change also causes a build error.

PR #4048, opened 2023-03-26
Growlithe part of Grandpa's Bill Quest now gives correctly a fire stone

LuchoDreamer:
Fixed it so now it gives a fire stone instead of a water stone.
Also changed a small dialogue to make the Eevee reward less confusing.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4053, opened 2023-03-27
Updating Caught Pokemon Selector to catch caught shinies

morcousw:


Jaaslet:
The pokeball filter is almost done, which makes this obsolete. I will close it. Sorry

PR #4056, opened 2023-03-27
Updates tooltip of Freeze Mulch

LuchoDreamer:
Because people ask about it every other day.


LuchoDreamer:
Improved the tooltip of Rich Mulch too

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4057, opened 2023-03-27
Fix Farming replant chances

Acegelica:
will fix the bug that causes replant chances over 100% to result in <100% chance of replanting


Acegelica:
i think this is correct now?

Aegyo (approved):

PR #4059, opened 2023-03-27
Fix Soul Dew and Quick Powder region

Qwertypop04:
Fix Soul Dew and Quick Powder region


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4060, opened 2023-03-27
Near Space Dungeon Fixes

Farboo171:
Locks Deoxys alt forms behind already having obtained them

Adds a Mythic tier chest item: Carbos

Changes base difficulty to Kalos difficulty.


Razorflamekun:
Personally, I don't believe that the dungeon itself should be Kalos difficulty if it is in Hoenn.  Having it Kalos Difficulty HP-wise, is fine, but I don't agree with making everything else about it Kalos difficulty.

Jaaslet:
It's the same we did for sevii4567

Razorflamekun:
Ok, if that's the way it has to be, then I guess I have to accept it.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4065, opened 2023-03-29
Fixup default settings overriding queue size

RedSparr0w:
This PR fixes the issue with `breedingQueueSizeSetting` setting being saved to the default settings with a higher value than a new save would be allowed, allow you to queue Pokémon in Kanto or more Pokémon than normal in other regions.

Refer Discord bug thread:
https://discord.com/channels/450412847017754644/1089811501126258748


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4066, opened 2023-03-31
Fix transparent background in Mossdeep Space Center

Farboo171:
Fixes the transparent background in the Mossdeep Space Center image, that let you see the battle image by accident.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4067, opened 2023-03-31
Show breeding pokemon in vitamin list

CypherX:
Currently pokemon that are in the hatchery or queue will be hidden from the vitamin list. This sometimes leads to confusion and players wondering why they can't see a certain pokemon.

This change will show these pokemon in the vitamin list (in both vitamin modals) with muted text and display a warning notification when trying to add/remove vitamins.

image
image


Jaaslet (approved):
I did some small code cleanup. Looks good. Thanks for the PR

PR #4068, opened 2023-03-31
Fix images for Rampaging Zacian and Zamazenta

Dobbykroket:
imageName pointed at a nonexistent image, leading to the default image being shown.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4069, opened 2023-03-31
Improved GymRunner and DungeonRunner code consistency

umbralOptimatum:
I noticed that GymRunner and DungeonRunner use a mixture of their class names and `this` to access properties. Both are static classes that should never be instantiated or subclassed, so I changed all uses of `this` to the relevant class name for consistency.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4074, opened 2023-04-04
Use different dungeon chest sprites for each loot tier

### LINKED ISSUE #3647: [SUGGESTION]: Different chest sprites for different loot tiers
### Current Behavior

Chests in dungeons are all the same, no matter the loot they give on opening

### Improved Behavior

5 different chest sprites, one for each loot tier: common, rare, epic, legendary, mythic.

### Reason for change

- Clarity: It is common in games to have different chests depending on the rarity of the loot in them
- Specific loot farming: For example in the case of Zen (Galarian) Darmanitan, if there's a chest of mythic rarity it allows the player to change the Pokéball accordingly for EV farming

### Detailed Description

Both the sprite in the battle screen and the sprite on the map would be different depending on the rarity, just changing its colour or adding glows/"rarity gems" to them. 
image

I'll also use this part as "reason not for change" and mention that the loot is generated at chest opening rather than at the moment the dungeon is generated (as of 10.4).  I personally don't think this change is worth changing the whole code, but I feel like we could have a change like this if the way dungeons work is ever changed in the game.

Aegyo:
Closes #3647

Old chest sprite still used for the chest counter and mimic indicator on dungeon mons.

Example image


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4079, opened 2023-04-05
Remove Deoxys Quest auto start

Qwertypop04:
Removes check and fix auto starting deoxys quest, as this is no longer intended. Also removes update function to the same effect, although this would only effect very old saves


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4081, opened 2023-04-06
Add linting rules for css files

Aegyo:
Planning to add some more rules to catch things like low performance animations, this one is just adding stylelint with the recommended rules and fixing what it complains about (or disabling the rule)

Also adds a couple of package scripts (`npm run stylelint` and `npm run stylelint-fix`) similar to the eslint ones, and includes style linting in the test command we have.

<hr>

Note: this does remove all vendor prefixes - if these are needed for some browser we want to support, then we should be configuring autoprefixer (used in gulpfile.babel.js) to handle prefixing for that version.

Oh, and you can install the Stylelint VSCode plugin to get warnings in the editor


Jaaslet (approved):
More linting rules are cool. And it seems like the CSS still works.
Looks good. Thanks for the PR

PR #4087, opened 2023-04-07
Limit UsePokeballQuest to purchasable pokeballs

CypherX:
Getting a `Use xxx Ultraballs` quest immediately after unlocking quests in Kanto isn't fun. This change prevents the `UsePokeballQuest` from selecting Greatballs and Ultraballs until they are available to be purchased. These balls can be obtained earlier in dungeons so I couldn't set the requirements on the balls themselves or they would be locked/hidden in the selector.


src/scripts/quests/questTypes/UsePokeballQuest.ts
```diff
@@ -11,7 +11,17 @@ class UsePokeballQuest extends Quest implements QuestInterface {
     }
 
     public static generateData(): any[] {
-        const possiblePokeballs = [GameConstants.Pokeball.Pokeball, GameConstants.Pokeball.Greatball, GameConstants.Pokeball.Ultraball];
+        const possiblePokeballs = [GameConstants.Pokeball.Pokeball];
+        if (player.highestRegion() > GameConstants.Region.kanto) {
+            possiblePokeballs.push(GameConstants.Pokeball.Greatball, GameConstants.Pokeball.Ultraball);
```
Jaaslet, line 16: Lavender town and Fuchsia city will always be unlocked when highestRegion is above kanto, right? We should just be able to remove this, and only use the code below?

CypherX: Yes, that would be true and we could do that, I included it as a small optimization.

Jaaslet: It's a very small optimization, compared to all the other code running all the time. It's not worth making the code more complex for

CypherX: Scrapped it 👍

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4089, opened 2023-04-07
Can now snatch shadow pokemon from trainers

Jaaslet:
You can now steal shadow pokemon from trainers. They are just normal catchable pokemon, if they are shadow.
Right now, the shiny filter also takes account for shadow. So what ever you have in uncaught shinies will also be used for uncaught shadow. The plan is to just use this, until the advance filters are done. If the advance filters are not done before Orre is released, i will make a PR with UI changes, to make this clear.

Catching shadow mons in gym will require a lot more code, so that will be in an other PR


Aegyo:
I'm not 100% sure, but it looks like if the last pokemon a trainer has is shadow, the battle ends without showing the catch animation. (Testing with viridian dungeon trainers)

Jaaslet:
> I'm not 100% sure, but it looks like if the last pokemon a trainer has is shadow, the battle ends without showing the catch animation. (Testing with viridian dungeon trainers)

It looks fine, i think? I have tried both with dungeon bosses and normal trainers, and it works fine

Aegyo (approved):

PR #4091, opened 2023-04-07
Filter for Pokémon without Mega Evolution

Crobat4:
This PR adds another filter to the Unique Forms dropdown for Pokémon with available Mega Evolution but without Mega Stone.
This also hides the filter if the player didn't reach Kalos yet.

image


Jaaslet:
Oh, i misread your suggestion. I thought the filter would show pokemon with mega evolutions, where you did not have the evolution yet. So you only breed the pokemon that you needed.
I am not sure what the usecase is for this filter?

Crobat4:
No, you are right, I made it wrong, my bad.
Should work fine now.
image

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4092, opened 2023-04-07
Removed Fog Check for Castform (Hoenn)

EatPant2nd:
Removed the Fog weather check for Castform (Snowy) since there is never Fog weather in Hoenn.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4097, opened 2023-04-07
Fixed Category deletion

RegisCoaxans:
image
I used Crobat's fix and made it so deleting a category flowingly updates filters and CSS.


Aegyo:
I feel like this would be better solved by giving each category a unique id, and using that for identification in filter/sorting/css vars etc instead of the using the index in a list and having to deal with the mess whenever the list changes.

RegisCoaxans:
(╯°□°)╯︵ ┻━┻
It should work as you suggested !

src/modules/party/Category.ts
```diff
@@ -11,6 +10,7 @@ import PokedexFilters from '../settings/PokedexFilters';
 import Settings from '../settings/Settings';
 
 export type PokemonCategory = {
+    id: KnockoutObservable<number>,
```
Aegyo, line 13: Why observable? It shouldn't ever change, right?

src/scripts/Update.ts
```diff
@@ -2122,6 +2130,8 @@ class Update implements Saveable {
             if (saveData.statistics.temporaryBattleDefeated[111]) {
                 Update.startQuestLine(saveData, 'A Beautiful World');
             }
+
+
```
Aegyo, line 2134: Blank lines added here. I wonder if this empty suggestion will delete them :thinking: 
```suggestion
```

src/modules/party/Category.ts
RegisCoaxans: Probably not useful, indeed.

RegisCoaxans: Made it number.

src/modules/party/Category.ts
```diff
@@ -33,55 +33,74 @@ export default class PokemonCategories implements Saveable {
...
-        while (categoryIndex-- > 0) {
-            this.removeCategory(categoryIndex, true);
+        let max = 0;
+        PokemonCategories.categories().forEach(c => {
+            max = Math.max(max, c.id);
+        });
+        while (max >= 0) {
+            PokemonCategories.removeCategory(max, true);
+            max--;
         }
```
Aegyo, line 43: This is kind of odd, why don't we just do:
```suggestion
        PokemonCategories.categories().forEach(
            c => PokemonCategories.removeCategory(c.id, true),
        );
```

RegisCoaxans: Sure, I can optimise that code. I do not know why I stuck with that.
However, `forEach` does not work because of the use of `splice`.

Aegyo: you can take a copy with `[...PokemonCategories.categories()]` and forEach over that

RegisCoaxans: Done. Should be fine.

Aegyo (approved):

PR #3985, opened 2023-03-06
Clear Silph twice during Rocket Quest

Qwertypop04:
This PR changes the Team Rocket quest so that you have to clear Silph Co. twice to progress
- First with a Scientist as the boss
- Then you must defeat Blue 5
- Then a second time with Giovanni as the boss.

This serves two purposes:
- To introduce the concept of clearing dungeons multiple times during a story quest (for use in #3978)
- Match canon more accurately, with Blue being fought before Giovanni

PR #4099, opened 2023-04-08
Fix Team Rocket quest soft lock

CypherX:
In #3985 a step was added to the Team Rocket quest to defeat Blue 5 in Silph Co.

Players can become soft locked at this quest step if they have previously defeated Blue 5 or if they have previously cleared Silph Co. causing Blue 5 to appear and proceed to defeat him prior to the required quest step.

Related thread on discord: https://discord.com/channels/450412847017754644/1094093055277137920


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4102, opened 2023-04-08
Fixed evo stone crash on multiple stone use

### LINKED ISSUE #3827: [Bug]: (Real Evolution) Trying to use multiple stones will eat the second and throw an error
### Confirmations

### What happened?

Attempted to batch use an evolution stone, with Real Evolutions active, on a pokemon I didn't have the evolution of yet.
2 stones were used, but since the first stone removed the pokemon from my party this threw an error (and wasted that second stone)

Only one item used, maybe a notification about the pokemon no longer available to use stones on.

### Steps to Reproduce

1. Make sure Real Evolution challenge is active
2. Get a pokemon that evolves by stone, but you don't have the evolution of yet
3. Get at least two of the evolution stone needed
4. Set the evolution item modal to use 10 stones at once, and try to evolve the pokemon

_No response_

Firefox

### Relevant log output

```shell
Uncaught TypeError: partyPokemon is undefined
    use http://localhost:3000/scripts/script.min.js?v=0.10.9:20312
    useStones webpack://pokeclicker/./src/modules/items/ItemHandler.ts?:64
    click http://localhost:3000/libs/knockout-latest.js line 74 > Function:3
    init http://localhost:3000/libs/knockout-latest.js:100
```

### Additional Information

_No response_

RegisCoaxans:
Closes #3827.
If the evo is new, it just calls to stop, as it does when the evo comes shiny.
I also removed all `this` in ItemHandler.


src/scripts/party/evolutions/EvolutionHandler.ts
```diff
@@ -55,10 +56,11 @@ class EvolutionHandler {
                 evolvedPartyPokemon.heldItem = basePartyPokemon.heldItem;
             }
             App.game.party.removePokemonByName(data.basePokemon);
+            baseRemoved = true;
         }
 
         // EVs
         evolvedPartyPokemon.effortPoints += App.game.party.calculateEffortPoints(evolvedPartyPokemon, shiny, GameConstants.STONE_EP_YIELD);
-        return shiny;
+        return shiny || baseRemoved;
```
Aegyo, line 64: I don't like this. The callers of this function can no longer know if a `true` response means shiny.
Instead, I think you should be returning a more detailed value like
```ts
return {
    shiny,
    baseRemoved,
}
```
and update anywhere calling this function to handle the response however they need to.
Additionally the `return false` guards at the top can probably just return `null` to signal no evolution was done.

I would probably pass this return type out of PartyPokemon.useStone and EvolutionItem.use as isdirectly, and handle the logic up in ItemHandler.useStones

Aegyo (changes requested):

src/scripts/party/evolutions/EvolutionHandler.ts
RegisCoaxans: Would it look better if I just test if the mon is still in the party in [ItemHandler](https://github.com/pokeclicker/pokeclicker/blob/develop/src/modules/items/ItemHandler.ts#L87), then ? So everything else remains untouched and preserved.

Aegyo: Yeah that would be fine, getPokemon and getPokemonByName both use a Map lookup so it isn't an expensive operation to be doing

Aegyo (approved):

PR #4105, opened 2023-04-08
Adding Orre Achievements category

Farboo171:
Adds Orrechievements category.

Should auto-populate with dungeons at least, and we can add the rest as we decide on them.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4110, opened 2023-04-09
Purify chamber

Jaaslet:
Fix your shadow pokemon!
Still needs to figure out how to limit how often you can use it. But it works for testing, for now.


Jaaslet:
I still need to think about how I want to implement how much exp you need to purify a mon. But I will set it ready for review, both in case someone wants it to test something they are working on, and just because it might as well be two PRs, to make it easier to review

Aegyo (approved):

PR #4111, opened 2023-04-09
Colosseum locations and dungeons

LuchoDreamer:
This PR creates all the locations from Colosseum in the SVG, as well as all the towns and dungeons.

Most of the dungeons are not in the map because they are part of the town, for most of them you need to go to the town first, and then the dungeon.

Things to do in other PRs: 
1. Images for all the towns and dungeons.
2. A way you can go from a town to another. Some cities are not in the map because they are only accessible to go there through another city. (This one is important)
3. Adding the dock, farm, UG, etc..
4. Populate the towns and the dungeons.

### DIFF
--- a/src/components/HoennSVG.html
+++ b/src/components/HoennSVG.html
@@ -1183,44 +1183,8 @@
         <!--Mt. Battle-->
         <%- include('templates/mapTown',
             { name: "Mt. Battle"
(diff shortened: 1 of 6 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4113, opened 2023-04-09
Move the rest of underground into modules

Aegyo:


Jaaslet (changes requested):
The game goes into an infinite loop, when trying to generate a new layer.
The survey no longer shows all items

Aegyo:
Inifnite loop should be fixed now, I think it broke when I was fixing those eslint warnings.
I couldn't see any issue with the survey, maybe just a side effect of that infinite loop bug? Let me know if it is still happening and I'll look closer.

PR #4114, opened 2023-04-10
Added shadow and purify achievements

Jaaslet:
Adding achievements for catching/purifying 1/10/50/131 shadow pokemon. Most of the names are ideas from discord.
I tried to make the achievement bonus somewhat high, as it's the main goal/challenge of the subregion


Aegyo (commented):
These look like huge achievement weights compared to anything else in the general category, was that on purpose?

Qwertypop04:
> These look like huge achievement weights compared to anything else in the general category, was that on purpose?

"I tried to make the achievement bonus somewhat high, as it's the main goal/challenge of the subregion" - going by what Jaaslet said, I would presume yes

Aegyo:
Yep, I clearly didn't read

Aegyo (approved):

PR #4120, opened 2023-04-11
Made Balls 0.25s Faster

EatPant2nd:
Made all Balls 0.25s faster. Made half of special balls, those at Poké Ball speed (1.25s before), match new Great Ball speed (0.75s, before it was 1s). This speeds up grinding in early Kanto. Also helps with late goals like EV-ing Pokémon and getting DT for the dungeon achievements.

In other words:
Poké Ball changed from 1.25s -> 1s.
Great Ball changed from 1s -> 0.75s.
Ultra Ball changed from 0.75s -> 0.5s.
Master Ball changed from 0.5s -> 0.25s.
As for special balls. Most are now 0.75s (before, 1.25s and 1s) except Fast Ball that has 0.25s (before, 0.5s) to match new Master Ball speed, and Luxury Ball that has 1s (before, 1.25s) to match new Poké Ball speed.

### DIFF
--- a/src/scripts/pokeballs/Pokeballs.ts
+++ b/src/scripts/pokeballs/Pokeballs.ts
@@ -25,11 +25,11 @@ class Pokeballs implements Feature {
 
     constructor() {
         this.pokeballs = [
-            new Pokeball(GameConstants.Pokeball.Pokeball, () => 0, 1250, 'A standard Poké Ball', undefined, 25),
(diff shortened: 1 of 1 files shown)


Jaaslet:
I will close this PR. It's a big change, and we are already soon getting the advanced catch filters, which already will make you spend less time on ball animations

xslk:
What about upping the speed of only the special balls that are at 1.25s catch time to make them slightly more appealing?

PR #4129, opened 2023-04-12
Comment out Beige Shards

Qwertypop04:
So as to prevent unnecessarily breaking daily chains before Hisui is released


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4132, opened 2023-04-13
Add test comparing daily deal against static value

Aegyo:
This should catch any accidental changes to daily deals.

Adding new items should make the test fail.
If deals being different is an intended change, `npm run vitest -- -u` can be used to update the snapshots automatically, and this will also be visible to anyone reviewing the PR.

It still might be possible for something to change deals in a small way that leaves our test snapshots alone, if we happen to be unlucky in the rng rolls we are testing. To reduce the risk of this, I'm testing 3 days with a maxDeals = 10, giving us 60 random item rolls that need to remain the same.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4147, opened 2023-04-14
Pokemon with only level evolutions now shows up in friend safari

Jaaslet:
Found a clean solution to check if a pokemon has any stone evolutions or not in `PokemonHelper.hasEvableLocations()`, so now we are closer to the correct list


src/scripts/pokemons/PokemonHelper.ts
```diff
@@ -437,7 +437,7 @@ class PokemonHelper extends TmpPokemonHelper {
         return locations[PokemonLocationType.Dungeon] ||
             locations[PokemonLocationType.DungeonBoss] ||
             locations[PokemonLocationType.DungeonChest] ||
-            locations[PokemonLocationType.Evolution] ||
+            (locations[PokemonLocationType.Evolution] && PokemonHelper.getPokemonStonePrevolution(pokemonName)) || // Only stone evolutions gives EVs
```
Aegyo, line 440: This searches through the pokemonList looking for a stone prevo, but we should already have the info on locations[PokemonLocationType.Evolution] to use
```suggestion
            (locations[PokemonLocationType.Evolution] as EvoData[])?.some((evo) => evo.trigger === EvoTrigger.STONE) || // Only stone evolutions gives EVs
```

Jaaslet: That works and is more effective. Thanks!

Aegyo (approved):

PR #4152, opened 2023-04-15
Creates function moveToTown

LuchoDreamer:
This creates a function that you can use to create a button to move to a town in the same way we have one to move you to a dungeon.

When used you can move to a town from another without using the map, useful when the town is not in the map for whatever reason, in the case of Orre there is a town _underneath_ another town.

Thanks Jaas for this!


Aegyo (approved):
I think it's a little difficult to tell the difference between town and dungeon buttons, but the MoveToTown ~~function~~ class appears to work fine :+1:

PR #4153, opened 2023-04-15
Quest items can now stay in inventory for multiple quests

Jaaslet:
Hisui will have a quest, where you gain a questitem in a quest, but don't need it until a later quest. This PR adds support for that


src/modules/items/QuestItem.ts
```diff
@@ -2,11 +2,13 @@ import QuestLineState from '../quests/QuestLineState';
...
-    constructor(name: string, displayName : string, description : string, private questlineName : string) {
+    constructor(name: string, displayName : string, description : string, private questlineName : string, private endQuestlineName : string = undefined) {
         super(name, undefined, undefined, undefined, displayName, description, 'quest');
+        endQuestlineName = endQuestlineName ?? questlineName;
     }
 
     public isActive() : boolean {
-        return App.game.quests.getQuestLine(this.questlineName).state() == QuestLineState.started;
+        return App.game.quests.getQuestLine(this.questlineName).state() > QuestLineState.inactive &&
+            App.game.quests.getQuestLine(this.questlineName).state() < QuestLineState.ended;
```
Aegyo, line 12: Shouldn't this be using endQuestlineName?
```suggestion
        return App.game.quests.getQuestLine(this.questlineName).state() > QuestLineState.inactive &&
            App.game.quests.getQuestLine(this.endQuestlineName).state() < QuestLineState.ended;
```

src/modules/items/QuestItem.ts
```diff
@@ -2,11 +2,13 @@ import QuestLineState from '../quests/QuestLineState';
 import Item from './Item';
 
 export default class QuestItem extends Item {
-    constructor(name: string, displayName : string, description : string, private questlineName : string) {
+    constructor(name: string, displayName : string, description : string, private questlineName : string, private endQuestlineName : string = undefined) {
         super(name, undefined, undefined, undefined, displayName, description, 'quest');
+        endQuestlineName = endQuestlineName ?? questlineName;
```
Aegyo, line 7: This trying to default as questlineName doesn't actually work, because you are setting a local variable instead of this.endQuestLineName.

I think you could also just do this in the param list:
```suggestion
    constructor(name: string, displayName : string, description : string, private questlineName : string, private endQuestlineName = questlineName) {
        super(name, undefined, undefined, undefined, displayName, description, 'quest');
```

Aegyo (changes requested):

src/modules/items/QuestItem.ts
Jaaslet: Fixed. Thanks

src/modules/items/QuestItem.ts
Jaaslet: Looks fun. Will test it later

Jaaslet:
I have submitted and tested both your change suggestions, and i am happy with them

Aegyo (approved):

PR #4141, opened 2023-04-14
Primal Reversion

Farboo171:
Adds a Primal Reversion quest in the middle of the Delta Episode. This quest:

1. Is very weather related
2. Is set to be late-kalos difficulty, spun out of Delta episode right before the end
3. will allow for catching of Primals in the correct weather after the quest is completed
4. treats primals the same as megas from an Evo perspective

PR #4155, opened 2023-04-15
Mega Stone refactor

CypherX:
The goal of this was to decouple mega stones from the `PartyPokemon` and turn them into regular items. This also fixes the real evolution issue where mega stones can be lost and allows for the X/Y stones to exist as separate entities rather than being merged into one stone.

Changes:
- Mega stones are now regular items rather than a property on `PartyPokemon`
- Renamed mega stone images to use actual names instead of pokemon IDs
- Pokedex and pokemon list entries support showing both mega stones (really only affects Mewtwo and Charizard)
- Update function to give player any mega stones they have already obtained, even if it was "lost" due to real evolution

There should be no noticeable changes in mega stone functionality for the player. Planning to add mega stones to the item bag in another PR when/if this gets merged.


src/modules/requirements/MegaEvolveRequirement.ts
```diff
@@ -1,16 +1,29 @@
...
+
     hint(): string {
-        return App.game.party.getPokemonByName(this.name)?.megaStone?.getTooltipText() || `${this.name} holds no Mega Stone.`;
+        const attackRequired = pokemonMap[this.name].attack * this.baseAttackMultiplier;
+        if (this.getProgress()) {
+            return 'Use a Key Stone to Mega Evolve.';
+        } else if (App.game.party.getPokemonByName(this.name).attack < attackRequired) {
+            return `Needs at least ${attackRequired.toLocaleString('en-US')} attack to Mega Evolve.`;
+        } else {
+            return `${this.name} holds no Mega Stone.`;
```
Aegyo, line 26: It's probably better to name the mega stone needed here? Something like
```ts
`${this.name} needs the ${MegaStoneType[this.megaStone]} Mega Stone.`
```

Also I think it should either be shown at the same time as the attack requirement, or override it when the player doesn't have the stone yet.

src/modules/items/MegaStoneItem.ts
```diff
@@ -27,17 +27,17 @@ export default class MegaStoneItem extends Item {
...
+        return super.isAvailable() && App.game.party.alreadyCaughtPokemonByName(this.basePokemon);
     }
 
     get image(): string {
-        return `assets/images/megaStone/${pokemonMap[this.pokemon].id}.png`;
+        return `assets/images/megaStone/${MegaStoneType[this.megaStone]}.png`;
     }
     isSoldOut(): boolean {
-        return App.game.party.alreadyCaughtPokemonByName(this.pokemon) && App.game.party.getPokemonByName(this.pokemon).megaStone;
+        return player.itemList[this.name]();
```
Aegyo, line 41:
```suggestion
        return player.hasMegaStone(this.megaStone);
```

src/modules/requirements/MegaEvolveRequirement.ts
CypherX: Updated

CypherX:
Updated to fix the conflicts/issues from #4141 being merged

Jaaslet:
Why have we removed the tooltip from the pokedex and pokemon list?

Jaaslet (changes requested):
`Red_Orb` and `Blue_Orb` shows up in the tooltip like that.

I am going to merge this, as soon as you have fixed the two problems. So you can just clean up the update function now

CypherX:
Fixed tooltip and cleaned up the update function.

> Why have we removed the tooltip from the pokedex and pokemon list?

Two reason:

- I couldn't find a decent way to handle it that wasn't messy (particularly the pokedex tooltip for the two pokemon that have two mega stones)
- The evo item modal tooltip has all of that information now.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4156, opened 2023-04-16
Fix: Hatchery scrollbar in Hatchery Helpers and Help tabs

Crobat4:
Previous fix broke the scrollbar in the last two tabs, this PR should fix that and the double scrollbar in the Pokémon list tab.


Jaaslet (changes requested):
For the pokemon list, some window sizes now have two scroll bars
image

Crobat4:
Had some outdated breakpoints, should work fine now.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4157, opened 2023-04-16
Images for Colosseum towns and dungeons

LuchoDreamer:
Puts images for all the Colosseum towns and dungeons.

They can be changed at a later time but this gives something to start with.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4159, opened 2023-04-16
Updated Egg Colors

EatPant2nd:
Updated Fire, Dragon, and Bug egg colors so they better reflect their typings and they don't get confused with other typings. Some examples:

Bug and Grass are too similar (left is Bug, right is Grass):
bug-grass

Considering Bug looks like this in the Dex:
bug-poison-pokemon

I made Bug look a bit more like that (left is Bug, right is Grass):
new-bug-grass

Secondary type Dragon and secondary type Dark use the same black color (hex 111):
grass-dark-dragon

Considering Dragon looks like this:
grass-dragon

I changed secondary type Dragon to be similar to the Dex color (left is Dark, right is Dragon):
new-grass-dark-dragon

If you are wondering if it gets confused with secondary type Poison now, they are distinct enough (left is Poison, right is Dragon):
comparison-with-poison-2

Also, secondary type Fire was yellow and it made Electric / Fire blend in (left is Electric Fire; right is pure Electric):
electric-fire-electric

So, I made Fire a shade of orange (left is Electric / Fire; right is pure Electric):
new-electric-fire-electric

Since secondary type Fighting also uses a shade of orange, I had to make sure they didn't look too similar (top left is Electric / Fire; top right is pure Electric; bottom left is pure Fire, and bottom right is Water / Fighting): 
new-electric-fire-fire

Finally, to make sure a theoretical Fighting / Fire doesn't blend in, I poorly colored one to check how they would look. The pattern can be seen so I think it's good enough (left is pure Fire, right is a theoretical Fighting / Fire):
new-fire-test


Jaaslet:
Do we know if the egg colors are based on anything in canon?

EatPant2nd:
The original ones? No idea how the colors were originally chosen for them. Also, it's my understanding that base games are all the same? Perhaps some in manga / anime have specific colors? But I don't Pokémon so I don't know. I only changed them because I wanted them to be more visually distinct.

xslk:
Canon in majority of games has all eggs as white with green spots. Anime and a few other sources have eggs colored and having patterns similar to the pokemon it will hatch into.

[Bulbapedia egg page](https://m.bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_Egg#:~:text=For%20the%20Egg%2Drelated%20glitches,Glitch%20Egg%20and%20Bad%20Egg.&text=A%20Pok%C3%A9mon%20Egg%20(Japanese%3A%20%E3%83%9D%E3%82%B1%E3%83%A2%E3%83%B3,this%20is%20not%20the%20case.)

Someonealive-QN:
I like that you used the colours from the dex, it makes it way easier to discover what is the pokémon type is that way. Also the similarity of bug and grass was stupid damn. Anyways, using the yellow and green spots eggs would be boring af and also why does it **have** to have a relation with canon stuff? Those are just egg designs who cares if they're not like canon. All that it matter is that they are distinguishable enough from the others so people who can't see colors very well don't have to spend 50k qp in kanto for some mon cus they can't save scum for it.

Jaaslet (approved):
I have no idea what the old colors are based on, so don't mind changing them.
Looks good. Thanks for the PR

Issue #2042, opened 2022-04-19
[SUGGESTION] Improved Pokemon List


Malchron:
**Current Behavior**

Right now when you mouse over your pokemon on the "Pokemon List" which is by default located on the left side of the game screen, nothing happens.

**Improved Behavior**

It would be neat if (this could be a toggle on/off feature via options) when you mouse over the picture an enlarged picture of the pokemon would appear. Could include other features.

**Reason for change**

I like to look at the pokemon sometimes briefly. I could open the pokedex but I already always have the list open on the left anyways.
It's not a huge deal to open the ingame pokedex and look around there but if this was easy to implement it would be a nice feature I think.

**Detailed Description**

Edit- Basically when you open the pokedex and click a pokemon, I want that to be doable through the pokemon list itself without having to open the dex.

**Screenshots:**

![image](https://user-images.githubusercontent.com/76578407/164055245-32e9bcad-37ee-42e7-bcb8-85b44f921cda.png)

osherwott:
Approved. Clicking on a pokemon in the list and having its Statistics page show up is open to be worked on. 👍 

We have concerns about slowdown and lag, so a hover option may not be the best, but opening the modal upon click can be good enough.

PR #4165, opened 2023-04-17
Easing the use for Statistics and Profile

RegisCoaxans:
- A new button in StatisticsModal directly sends the shown pokémon's data to the profile (shiny + gender form depending on PartyPokémon settings), to avoid browsing in there.
- An additionnal row in StatisticsModal shows pokémon attack at level 100 if the pokémon is still leveling up (the row disappear once the pokémon reaches level 100). There was nowhere to read the full attack of a pokémon, if we were quickly and constantly breeding it, without stopping.
- Fixed strip pattern being broken in StatisticsModal because of rare hold item row being emptied instead of not being computed when no rare hold item to display.
- Clicking a pokémon in PokémonList directly opens StatisticsModal for the clicked pokémon, closes [#2042](https://github.com/pokeclicker/pokeclicker/issues/2042) as described when approved.

(Do not mind the party name, lol.)
I put the button near shiny and gender form toggles for it makes use of these. I can move it anywhere else.
Moreover, this button could open ProfileModal as a shortcut as well so we can see its changes in action but I was not sure about that...
image


RegisCoaxans:
You know, some weirdos very often change their Discord profile picture... So maybe they do on PokéClicker too !
Anyway, I agree it added to the narrowness and might need some better place if it has to come back one day.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4167, opened 2023-04-17
Console warning

CypherX:
Missingno. is not an easter egg. Some sites/videos claim otherwise which leads to people injecting code via the console. This displays a warning to not do that. Will it help? Meh.

Also removes the `Game saved` log to avoid pushing the warning off screen.

image


Jaaslet (approved):
When i first learned about the console 13 years ago, people thought you where a wizard when you used it. Now everyone uses it, no matter if they know what it is or not...
What i am trying to say is: hopefully this will work. About removing the saving log is fine, as it's less useful, now that the game automatically saves on close.
Looks good. Thanks for the PR

PR #4170, opened 2023-04-19
Micle repel aura buff

LuchoDreamer:
Buffs Micle repel aura slightly to not change interaction with the farm (you still need 4 Micles to get the full repel effect), but this buff to the repel aura allows for a new interaction with Lum, allowing you use to use 3 Micles and a Lum for the same effect (with the previous numbers the interaction between Micle and Lum didn't serve any purpose).

This allows for more flexibility in high end farm setups in niche challenge playthroughs.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4171, opened 2023-04-19
Changes Fighting Dojo defeated dialogue to explain how typed eggs work

LuchoDreamer:
Let's make use of the first instance in the game of getting rewarded a typed egg to explain a little bit what their use is!


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4173, opened 2023-04-19
Purify chamber now needs exp to purify

Jaaslet:
Each time a pokemon in your party, that is not yet level 100, gets exp, that exp will be added as "flow" to your purify chamber (divided by 1000, to make the numbers smaller).
When you have enough flow, you can purify one pokemon, and the flow resets.
(draft because i still need a formular. Working on it)


Jaaslet:
Formula added! It looks stupid...
What i tried to do:
First ~10 only takes a couple of minutes of farming in hoenn.
Then it goes slowly up until like 120, where it takes 4 hours per mon in galar.
Last couple of mons takes almost a day in Galar.
I don't understand the formula. Got a friend to help me...
My hope is that the crew will discuss the optimal farming method, do some calculations and then tell if it's too stupid or not.

Aegyo:
No graph? 😞

Jaaslet:
> No graph? 😞

I am not sure I want to see it...
But it's y will just be some big numbers, so not sure if it's super useful?

Jaaslet:
I tried to make a graph that makes sense... It's hard...

Jaaslet:
I was hoping for it to be a bit more smooth around... But yeah, this is what i asked for.
I wanted the last couple to take a long time, without others taking too long.
But the total is 131

Farboo171:
Image added for record keeping's sake

Farboo171:
To codify what was said in discord: I'm down with merging this as-is and balancing before actual release

Jaaslet:
I have changed the formula to Eugenes more smooth curve. I don't think it's correct, but i think it will be a lot easier to discuss and find the correct formula, when the crew starts testing Orre.
So (like Eugene said) i think we should merge this PR and then change the formula before release of Orre

Aegyo (approved):

PR #4174, opened 2023-04-19
Fixes tiles_custom.png for Tiled

LuchoDreamer:
My previous PR for the ships worked with an old and wrong tiles_custom.png. This image for Tiled has actually been here for a while, but since the maps with sprites from this png had not been touched, this went unnoticed until today.

This PR fixes this png file, and with that all the images for all the maps that got messed up because of it, either because of the last PR or by anything else.

From now on it will be possible to work in any map from Tiled without having to worry about tiles_custom.png.

P.S.: Fixing the file was torture...


LuchoDreamer:
Just finished fixing everything, this is ready for review.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #3755, opened 2023-01-10
BerryDex Preview Mode

osherwott:
### Problems that this solves: quickly gathering information on a berry
As adding filters to the berrydex has proven difficult, I have done my best to streamline the process of showing relevant info in the form of showing important parts of the berrydex popup modal within the Berrydex Tab of the Farm. This includes Growth Time and Current Berry Amount, as well as other future implementations, such as Flavors or the upcoming Size and Firmness elements. Filters would still be welcome if someone else wants to help, but this is the best solution I have in the absence of them. My resulting layout was made through the intention of displaying information that was planned to be filterable. A good example of this philosophy is reflected in my addition of Aura highlights.
___
### TASKS: Split PR into smaller ones:
- [x] Berry Aura info #4181 
- [ ] Create row for Parasite status
- [x] Create rows for Wanderers #4422
- [x] CSS variable list setting #4182 
- [x] Change order of information #4266 
- Change how that information is displayed
  - [x] Add "Growth and Berry Stage Length" Row that splits if the time for each is uneven #4265 
  - [ ] Remove aura and parasite information from Description? (maybe have it as a tooltip on the Row?)
- [ ] Add colors and settings to change them #4304 
- [x] Create Berrydex Preview (this PR)
___
_~~To do (also detailed at the bottom)~~ Completed:_
- _clicking again on berry image brings up modal (accessibility)_
- _make sure new modal lines have proper ARIA labels (accessibility)_
- _setting to turn off CSS animations_
- _get feedback/assistance for flexbox/modal sizing_
- _figure out what to do with Aura "multiplier" text for Micle_

### Additions
- **Aura information in Berrydex**
- Tricked out Berrydex with neat CSS stuff (pictured below)
Flavor colors are based upon that of their corresponding condition
https://bulbapedia.bulbagarden.net/wiki/Category:Contest_color_templates
### Changes
- Farm modal **Berry Tab now shows Berrydex preview** with a button for more details on its side
- **Renamed information in the Berrydex Modal** to be clearer to the reader:
Growth time and Berry stage length are named as such
"Total time until Stage" added to Growth Stages section above the times
- **Added information to Berrydex Modal**:
Aura type, if a berry has one
"Aura multiplier at Stage" to Growth Stages section
Aura multiplier numbers, also in the same section
- **Moved rows in the Berrydex modal**:
Hint row to the bottom
Grouped non-time Statistics together underneath Flavor section
### Notes
- Size and Firmness are TBD because **This is to be merged after** #3822
[...]

PR #4181, opened 2023-04-23
Berrydex aura

osherwott:
Splits up part of #3755 
### Adds Aura info to berryDexModal.html

- Adds row for AuraType
image

###
- Includes intensity of aura at each growth stage
- Adds extra row to label time until growth stage to prevent confusion
- Can differentiate between multiplicative and additive auras (`auraDisplay` in Farming.ts)
image image


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4182, opened 2023-04-23
Create CssVariableListSetting.ts

osherwott:
Splits up part of #3755 
Adds ability to change css values from a list of options


Jaaslet:
- The PR has a warning
- Can you explain what this will be used for?

osherwott:
- The warning has been resolved

- Explanation:
  - This file allows for manipulating CSS values from a set of finite options, as opposed to the current `CSSVariableSetting.ts` which is restricted to a color picker.
  - In this specific PR's case, it will be used to customize the intensity of the colors in the Aura and Flavor rows (including disabling them). These will be submitted later in a separate PR with more details.

Jaaslet:
I have now looked at what you need this class for, and we will not be needing this class (so i will close the PR).
The solutions should be:
- Add a setting with options like `text='Medium'` and `value='aura-highlight-medium'`
- Set the class of the html object to the value of the setting
- Add the 3 classes to the css files and set the transparency there
Then we can change the values based on themes, if the values does not work for some themes. And then we can also adjust the values, without having an update function

PR #4184, opened 2023-04-23
Move Special Event Dungeon Boss data

osherwott:
Continues work from #4150


osherwott:
Added some notes in SpecialEvents.ts so we can still keep track of what is added to each event wile everything is being scattered across different files. They can be cleaned up and listed better later but it's good to have them somewhere in the file for now.

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4185, opened 2023-04-24
Battle environment overrides

Aegyo:
Battle environment overrides for places that don't want to just use the town environment. Blue 1 temp battle and Kanto Elite 4 used as examples.

- Adds environment optional arg to Gym and TemporaryBattle
- Moves gym displayName and imageName into new optionalArgs
- Gym rewardFunction and gymFlags should probably also be moved into optionalArgs, but I didn't want the hassle of making sure that all still works


src/scripts/towns/TownList.ts
```diff
@@ -5,7 +5,7 @@
 ///<reference path="../quests/BulletinBoard.ts"/>
 ///<reference path="BattleCafe.ts"/>
 ///<reference path="../../declarations/requirements/MultiRequirement.d.ts"/>
-///<reference path="../Safari/SafariTownContent.ts"/>
+///<reference path="../safari/SafariTownContent.ts"/>
```
osherwott, line 8: This should probably be its own PR...

Jaaslet:
Why are the values in `optionalGymArgs` not just a part of `gymFlags`?

Aegyo:
Because flag means a boolean value, which the things in optional args are not

Jaaslet (approved):
> Because flag means a boolean value, which the things in optional args are not

You are correct, and I don't like it...

Looks good. Thanks for the PR

Issue #4054, opened 2023-03-27
[SUGGESTION]: Remove gender sign from nidoran name in battle


Oyne:
### Current Behavior

![image](https://user-images.githubusercontent.com/91478447/227929542-7ce664ba-6d98-43f5-819e-e219f85984c7.png)

### Improved Behavior

There will be only Nidoran and color gender sing like with all other pokemon

### Reason for change

Cuz it`s strange to see two gender signs in a row

### Detailed Description

We can remove last char of Nidoran name when displaying it

PR #4189, opened 2023-04-26
Hide gender character if Pokémon have it already in the name

Crobat4:
Closes https://github.com/pokeclicker/pokeclicker/issues/4054

Nidoran
image
Chinese Nidoran (Different names for both genders so no icon)
image
image

It also works with Indeedee.


Jaaslet (approved):
I did some small code cleanup.
Looks good. Thanks for the PR

PR #4194, opened 2023-04-26
Fixed Statistic JSON cancelling 0s

RegisCoaxans:
That was almost bound to happen, lol.
Values of 0 in statistic objects are not stored anymore, as first expected.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4195, opened 2023-04-27
Move Special event Quest data

osherwott:
The last work for #4150 I'm capable of easily doing.
Glad to have helped! 👍


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #4006, opened 2023-03-11
[SUGGESTION]: Add a weather app shortcut


Vodovik:
### Current Behavior

You're done EVing galar, except you're missing some weather reliant forms. You're grinding EVs in another region. You'd like to know easily if galar has the right weather or not for the form you're looking for.

### Improved Behavior

A shortcut somewhere that gives access to an app on the rotomphone or whatever that displays a weather app that tells us for each region what the weather is at the moment. Also maybe a weather prevision for the next cycle (that would be wrong half the time ofc /j .. or not!)

### Reason for change

 Going to galar and back to check the weather is a hassle (maybe because still no dock shortcut /j), also losing 4 seconds of route grinding when switching hurts my soul.

### Detailed Description

add a button in shortcuts or a key item (maybe) that when clicked opens a weather app that displays current (and maybe future) weather for each region. Either by checking a list or by choosing/navigating through regions with the weather displayed.

osherwott:
Approved. Everything in this suggestion is open to be worked on. 👍 

Some ideas to consider (but not necessary):
- unlock the shortcut after clearing the Weather Institute in Hoenn
- see week-long forecasts in the Weather Institute and limit the shortcut modal to the current weather
- other Weather Institute or map-related integration
- expand the current Dock modal to include weather for each region

PR #4196, opened 2023-04-27
Weather App

Crobat4:
I planned to split this in two PR:
* Weather app modal (this PR, closes https://github.com/pokeclicker/pokeclicker/issues/4006).
* Weather notifications.

This PR contains the first point, adds a modal to see the weather forecasts for all regions as well as expands the dock modal to show the current weather for all regions.

How many days will the app show depends if the modal is opened from the shortcut module (7 days) or from the Weather Institute (30 days).

The region list depends on the max region unlocked, you shouldn't be able to see all regions right away.

The Weather App unlocks after a Weather Institute dungeon clear.
<details><summary>Weather App modal</summary>

image
</details>
<details><summary>Dock modal</summary>

image
</details>

Special thanks to the one/s who made [this](https://docs.google.com/spreadsheets/d/1HFK5gS4XDmqXvZPP8stBYGC9ZF77i-Wug2J_gcmJ5rw/edit#gid=651749561) spreadsheet which helped me to understand a bit how the `SeededRand` works in this case.

_____
As for the second point, if approved, you would be able to click a specific weather forecast and when that time comes, it will show a notification.
<details><summary>Weather App modal with notifications</summary>

image
</details>


Farboo171:
We had some discussion in the dev chat and here's the feedback:

- For now, can you remove the shortcut? We're trying to come up with a unified plan for shortcuts,and its better to add one later than take one away.
-  Please cut the forecast time down to one week. 30 days is just a big scroll bar that is unlikely to be used by most players
- the NPC has a "!." typo in their text
- for now, let's hold off on the notifications, still discussing that

Thanks!

Crobat4:
Made the requested changes.
I kept the `GameHelper` functions just in case as they may be useful, they can be removed if they are not wanted.

Jaaslet (approved):
I have removed the NPC (outdated information) and changed the default region to hoenn (the institute is in hoenn).
Looks good. Thanks for the PR

PR #4201, opened 2023-04-28
Use translation for logbook labels

Aegyo:


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4203, opened 2023-04-29
Initial Dungeon populations for Orre

Farboo171:
Populates most of the Orre dungeons so we can get working on the quest, balancing, and other Orre stuff. Notably does not include:

1. Shadow Togetic
2. Under Colosseum Battles (no shadows in Colo)
3. Orre Colosseum Battles (No shadows in Colo)
4. Specific trainer classes, because the sprites are not done yet and that would be broken images. Everyone is a cool trainer for now.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4206, opened 2023-04-29
Added tooltip for dungeon total pokerus icon

morcousw:
When about to enter a dungeon, the pokerus icon at the top of the screen now shows a tooltip with the number of pokemon needed to EV. It uses the same message that is shown on the tooltip for pokerus at the bottom of the "inside dungeon" screen


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4208, opened 2023-04-30
New Friend Safari Town image

osherwott:
## Description

Replaced the Town Image of the Friend Safari

## Motivation and Context

The old image was from the anime, which was not only from the wrong generation but also heavily jarring as it does not match the style of the rest of Kalos' towns. Since the area is a place in the games, it goes without saying that an image sourced from canon is preferable.

## How Has This Been Tested?

This was tested by running the game and going to the Friend Safari.

## Screenshots (optional):

image

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [x] Aesthetic enhancement


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4210, opened 2023-05-01
Mega Stones tab in items modal

CypherX:
## Description

Adds a Mega Stones tab to the item modal that lists all acquired mega stones. Tab is not visible until the first mega stone is obtained.

## How Has This Been Tested?

Tab and all acquired stones are visible on an end-game file.
Tab is not visible on a file in Kanto that does not have any mega stones.
Tab becomes visible upon obtaining a mega stone for the first time.

## Screenshots (optional):

image


Jaaslet:
I think they should be added to the item bag tab instead of their own tab.
We can't really have too many tabs, so we need to find a better solution. But until we do, let's just gather them in the item bag

CypherX:
Moved to the bottom of the item bag tab. Went with a new tab because that's what was said in discord originally.

Jaaslet (approved):
We might want to add the mega stone tooltip here, but it's not a very big change, so won't hold the PR back for that. You can add it in a new PR, if you want to

PR #4212, opened 2023-05-03
pokdexPokerusFilter -> pokedexPokerusFilter

LegitSi:
## Description
I added a character inside the hapax legomenon variable of pokdexPokerusFilter to be more consistent.

## Motivation and Context
In my search to fix another bug, I found this. That's it.

## How Has This Been Tested?
I opened inspect element in client with CTRL+SHIFT+I and found it at the very top.

## Types of changes

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/src/modules/settings/PokedexFilters.ts
+++ b/src/modules/settings/PokedexFilters.ts
@@ -64,7 +64,7 @@ const PokedexFilters: Record<string, FilterOption> = {
     statusPokerus: new FilterOption<number>(
         'Pokerus',
         ko.observable(-1).extend({ numeric: 0 }),
-        'pokdexPokerusFilter',
+        'pokedexPokerusFilter',
         [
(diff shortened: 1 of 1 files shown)


osherwott:
Thanks for using the form! Devs are very busy irl so it helps to be concise!

CypherX:
As I mentioned on discord, the `pokdexPokerusFilter` value you are changing here is what's used to track the selected value for the pokedex pokerus filter in the save file. Changing this without an update function to also change the string in the save file causes the filter value to reset to 'All'.

LegitSi:
I'll investigate how to do that tomorrow when I get back. Thanks for letting me know.

Jaaslet:
I will close this, as it's a lot of work for a small refactor, and it's the PR is not finished

PR #4215, opened 2023-05-03
Added Primals to the Pokédex/Hatchery Mega filter

Crobat4:
## Description

Added Primal Reversion forms to the Unique Forms filter in the Pokédex/Hatchery.

They were added along with Megas instead new dropdown options as they are just two.
If needed, separated dropdown options can be made for them.

I also changed the options name to include "Primal Reversion".

## Motivation and Context

Primals are considered as Megas internally but they weren't showing in the filter because, currently, the filter only check if the Pokémon name starts with `Mega `.

## How Has This Been Tested?

* Added Primal Groudon to the party through console
* Selected "Mega Evolution/Primal Reversion available" filter: Both Groudon and Kyogre appeared
* Selected "Unobtained Mega Evolution/Primal Reversion" filter: Kyogre appeared but not Groudon due first point
* Selected "Obtained Mega Evolution/Primal Reversion" filter: Primal Groudon appeared but no Primal Kyogre as it wasn't added to the party


Qwertypop04:
> I don't think we want to say "Mega/Primal" in the filters. We kinda just pretend like they are megas in the rest of the game

I don't think that's quite correct. If we wanted that we would have changed their names. We just use the Key Stone to evolve them because it's easier

Jaaslet:
> > I don't think we want to say "Mega/Primal" in the filters. We kinda just pretend like they are megas in the rest of the game
> 
> I don't think that's quite correct. If we wanted that we would have changed their names. We just use the Key Stone to evolve them because it's easier

You might be right... I need to think about it...

Jaaslet (changes requested):
Just remembered something... Can we just use `PokemonHelper.hasMegaEvolution()` for the filters?

Crobat4:
Made the requested changes.
As for the Mega/Primal thing, if stating Primals is wanted but not in the dropdown options, it could be a tooltip like "Rare Hold Item" checkbox has that says "Include Primal Reversion forms" or something like that.

Jaaslet (approved):

And for the primal text: i think your first solution was the best. Just add primal to the dropdown values. Should lead to less confusion. Merging this PR now, as i have tested the changes in it

PR #4217, opened 2023-05-03
Pokemon type catch filter

Aegyo:
## Description

Adds a new option to catch filters for Pokemon Type.

You can select one type, and the filter will then match any pokemon that has that type.
PokemonType.None is left as an option mostly because it is a pain to deal with in the Test tab, selecting None effectively makes the filter only catch monotypes. Not exactly useful, but it's an option.

## Motivation and Context

This filter option is useful for:
- Monotype runs
- "Defeat/Catch <type> pokemon" quests
- Players who want more of a certain type of damage can farm EVs on just those pokemon

## How Has This Been Tested?

1. Created a filter with only the pokemon Type option, set to Flying
2. Placed between two other filters "catch shiny", "catch everything"
3. Observed what happens on Kanto route 1 - Pidgey used the type filter, Rattata used "catch everything"
4. Changed filter to None - now Rattata matches, Pidgey does not

## Screenshots (optional):

Screenshot_2023-05-03_23-59-16

## Types of changes

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/src/components/pokeballSelectorAdvModal.html
+++ b/src/components/pokeballSelectorAdvModal.html
@@ -162,7 +162,7 @@
                             <p>Configure these options to describe a Pokémon you want to test:</p>
                             <table>
                                 <tbody data-bind="foreach: Object.entries(App.game.pokeballFilters.testSettings)
-                                                                .filter(([k]) => pokeballFilterOptions[k].canUse())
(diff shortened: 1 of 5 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4219, opened 2023-05-04
Fix: Wandering Pokémon icon broken in notifications

Crobat4:
## Description

Pokémon icon in wandering notification is broken since `PokemonHelper.getImage()` gets the Pokémon name instead the ID.

## How Has This Been Tested?

* Gets wandering notification

## Screenshots (optional):

image

## Types of changes

* Bug fix


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4222, opened 2023-05-04
Improved berry growth-time-left display rounding

umbralOptimatum:
Very minor issue, but rounding up before accounting for growth multipliers meant the displayed growth time would sometimes be off by 1.


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4227, opened 2023-05-04
Added special Kanto map for Flying Pikachu event

Jaaslet:
## Description

- Adds a new version of kanto, that is only shown during flying pikachu event (map made by soul)
- Subregions can now have alternative maps based on a requirement (please don't overuse it)
- Moved all maps to nice folders, so it looks less chaotic

## Motivation and Context

Players like the halloween event, because you notice it more. So i want to try to change kanto for one event, and see if players likes it. If they do, we can change more regions for more events.
Code is made generally, in case someone gets a stupid idea, like adding snow to all of hoenn in the middle of a questline

## How Has This Been Tested?

Time traveled to see that it works with the event

## Types of changes

- New feature


Jaaslet:
Draft because i need to update all subregions in all svg.html to support the new way of getting the map

PR #4228, opened 2023-05-05
Fix Temporary Battle imageNames that reference pokemon forms

osherwott:
## Summary

Some Temp Battles against wild pokemon display the error image instead of that of the pokemon.

## Description

Changed the imageName reference for all erroneous ids in TemporaryBattleList.ts

## Motivation and Context

This was changed as per requested by a playtester. Errors are not supposed to be part of the game.

## How Has This Been Tested?

1. Run a save that has not completed a temporary battle and test to see if the problem exists.
2. Change the imageName reference to the correct spelling of the pokemon id for all temp battles.
3. Run the save again and see if that change has fixed another similar temporary battle.

## Screenshots (optional):

A completion modal before the fix:
galarian not moltres fled
A different completion modal (but against the same subject) after the fix:
galarian moltres fled
## Types of changes

- Bug fix
- UI improvement


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4237, opened 2023-05-06
DNA Splicer quest rename

### LINKED ISSUE #3317: [SUGGESTION] A better name for the DNA Splicer quest
**Current Behavior**
The name of the quest is simply "A quest for the DNA Splicer". That is a pretty boring name for a quest which gives you access to 2 legendary forms

**Improved Behavior**
The quest could get a rename like "Splicing Evil from Good" or something more creative

**Reason for change**
The quest  name is a pretty boring name for a quest which gives you access to 2 legendary forms

**Detailed Description**
It would look nicer instead of the name implying what you're gonna get by progressing which mostly of the other quests don't do that

osherwott:
## Summary

- Renames "A quest for the DNA Splicer" into "Hollow Truth and Ideals", referencing the themes of Unova and Kyurem.
- Closes #3317 

Nobody else was doing this so I decided to help chip away old suggestions. If there are better names for this quest feel free to comment.

## Description

Replaced all instances of the old name with the new one. This change is more or less cosmetic.

## Motivation and Context

We are closing old suggestions. This brings some life to an otherwise boring quest name.

## How Has This Been Tested?

This has not been tested, but it should. Past quest renames have been referenced so hopefully this should be good even if it's not tested.
If anyone feels like contributing, they are welcome to do so :)

### To do:
- [x] Include/Change update function in Update.ts

## Types of changes

- Enhanced immersion


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4241, opened 2023-05-06
Hide Contest Shop

Farboo171:
## Description
Hides the contest shop from view, beyond Dev requirement. This way, no one will see uncaught map colors or a shop name that they cannot access.

## Motivation and Context
To prevent an unavailable bit of content from being seen.

## How Has This Been Tested?
Ran lint tests, loaded a save file and the Contest Shop does not appear in Lilycove City anymore.

## Screenshots (optional):
N/A

## Types of changes
- Bug fix, kinda


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4243, opened 2023-05-07
Catch Rate Fix

Qwertypop04:
## Description
Fixes catch rates of some alt forms that did not match their base form, fixed Eevee's catch rate, and updated stats of Hisuian Pokemon

## Motivation and Context
Pokemon stats should match canon stats

## Types of changes
- Bug fix


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4244, opened 2023-05-07
Vivillon Dummy Evos

Qwertypop04:
## Description
Vivillon to being Dummy Evolutions instead of not evolving from Spewpa and having their egg steps changed manually

## Motivation and Context
To match the current way such evolutions are done


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4251, opened 2023-05-07
Field mutation now supports more than one berry

Jaaslet:
## Description

- The field mutation now takes a list of berries instead.
- Using this mutation for the hopo (hisui) berry, based on Qwerties instructions

## Motivation and Context

Qwerty asked nicely

## How Has This Been Tested?

Changed the mutation to use simple berries, and removed the requirement, so i could mutate it

## Types of changes

- New feature


src/scripts/farming/mutation/mutationTypes/FieldMutation.ts
```diff
@@ -44,7 +45,7 @@ class FieldMutation extends GrowMutation {
      */
     get unlocked(): boolean {
         // Check for Berry requirements
-        if (!App.game.farming.unlockedBerries[this.fieldBerry]()) {
+        if (this.fieldBerries.some((fb) => !App.game.farming.unlockedBerries[fb.berry]())) {
```
Aegyo, line 48: Why only some and not every?

Jaaslet: If any if the required berries are not unlocked, this berry cannot be unlocked. That is how it should work, right?

Aegyo: yeah, I misread the code. seems fine

Aegyo (approved):

PR #4252, opened 2023-05-07
Pokedex filter now filters on party name again

Jaaslet:
You used to be able to filter pokemon in pokedex/hatchery based on party name. It broke. Can't figure out why it broke, but i have readded it


src/scripts/pokedex/PokedexHelper.ts
```diff
@@ -91,7 +91,8 @@ class PokedexHelper {
             // Check if the englishName or displayName contains the string
             const displayName = PokemonHelper.displayName(pokemon.name)();
             const filterName = PokedexFilters.name.value();
-            if (!filterName.test(displayName) && !filterName.test(pokemon.name)) {
+            const partyName = App.game.party.getPokemonByName(pokemon.name)?.displayName;
+            if (!filterName.test(displayName) && !filterName.test(pokemon.name) && (partyName != undefined && !filterName.test(partyName))) {
```
Aegyo, line 95: Uncaught pokemon are always shown with this, try this instead
```suggestion
            if (!filterName.test(displayName) && !filterName.test(pokemon.name) && !(partyName != undefined && filterName.test(partyName))) {
```

Aegyo (changes requested):

src/scripts/pokedex/PokedexHelper.ts
Jaaslet: Committed this and fixed it for hatchery too. Thanks!

Aegyo (approved):

PR #4260, opened 2023-05-07
Re-add mega evolution hint to pokedex entry

CypherX:
## Description
Re-adds the mega evolution hint to the pokedex entry that was removed in the mega stone refactor.

## Motivation and Context
Saw some mentions of how useful it was after the update that removed it.

## How Has This Been Tested?
Checked the tooltip text at all 4 possible states: No mega evolution, attack requirement not met, mega stone not obtained, ready to mega evolve.

## Screenshots (optional):
image
image

## Types of changes
- UI improvement


src/components/pokedex.html
```diff
@@ -250,7 +250,14 @@ <h4 data-bind="text: `Showing ${PokedexHelper.filteredList().length} Pokémon`">
...
-                                                            : ''),
+                                                            : '') +
+                                                        (PokemonHelper.hasMegaEvolution($data.name) ?
+                                                            App.game.party.getPokemon($data.id).attack < pokemonMap[$data.name].attack * GameConstants.MEGA_REQUIRED_ATTACK_MULTIPLIER ?
+                                                                `<br/>Needs at least ${(pokemonMap[$data.name].attack * GameConstants.MEGA_REQUIRED_ATTACK_MULTIPLIER).toLocaleString('en-US')} attack to Mega Evolve.`
+                                                                : PokemonHelper.getMegaStones($data.name).some(m => player.hasMegaStone(m.megaStone)) ?
+                                                                    '<br/>Use a Key Stone to Mega Evolve.'
+                                                                    : `<br/>Obtain the Mega Stone to Mega Evolve ${$data.name}.`
```
Aegyo, line 259: Can this not just use the hint from the mega evolution instead?

CypherX: The hint is on the MegaEvolveRequirement so it's not the cleanest but I guess it could be done like this:
`pokemonMap['Pidgeot'].evolutions.flatMap(e => e.restrictions).find(r => r instanceof MegaEvolveRequirement).hint()`.
The only caveat is for Charizard and Mewtwo, who have two mega evos, it would only ever mention the first stone.

Aegyo: Maybe for each mega evolution, there could be a header with the mega stone name and the evo requirement hint underneath?

CypherX: Alright, tweaked it to do just that. It's not pretty but it works. Also added the tooltip back to the pokemon list icon. 

image
image

Aegyo (approved):

PR #4261, opened 2023-05-07
Fix access of Primal Reversion questline to be before mega-evolving Rayquaza

Farboo171:
## Description
As title, accidentally put the access step too late

## Motivation and Context
Return to intended behavior

## How Has This Been Tested?
It hasn't, but it's one number changed

## Types of changes

- Bug fix


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4263, opened 2023-05-07
Change Kanto Safari Zone to TownContent

CypherX:
## Description
Changes the Kanto Safari Zone to TownContent to work with map color overlays, allow shops/NPCs, and be consistent with the Friend Safari. Originally also added the code necessary for map color overlays but Jaas took care of that in #4249.

## Motivation and Context
Map color overlays are neat.

## How Has This Been Tested?
Verified everything still worked as intended with the safari.

## Types of changes
- Refactor


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4265, opened 2023-05-07
Clearer berry Growth times

osherwott:
## Summary

Splits up part of #3755 
Growth Time display now includes Berry Stage length

## Description

Growth Time row is renamed "Growth and Berry Stage Length" for clarification, and splits into "Growth Length" and "Berry Stage Length" if they are unequal, like in Lum.

## Motivation and Context

Some players are confused as to what Growth Time refers to. This helps emphasize that it's the length from seed until a berry has ripened.
Berry Stage Length helps more hardcore players plan around berries with longer ripe stages, as it is easier to retain information at a glance than subtract the Growth Time from the Total Life Span.

## How Has This Been Tested?

This has been tested in #3755 and copied from there.

## Screenshots (optional):

Table for Cheri vs Lum
image image
## Types of changes

- UI improvement


Jaaslet (changes requested):
I think it's confusing that they sometimes split up. Consistency makes UI easier to use. So they should always be split up

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4267, opened 2023-05-08
Fix MegaEvolveRequirement hint causing error on wiki

CypherX:
## Description
MegaEvolveRequirement hint uses `App.game.party.getPokemonByName` which always returns `undefined` on the wiki and isn't being handled properly, causing display issues on the wiki pages of mega pokemon.

## How Has This Been Tested?
Tested that the restriction still works correctly in-game.

## Types of changes
- Wiki bug fix


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4274, opened 2023-05-09
Blank plate

Qwertypop04:
## Description
Adds the Blank Plate to the underground, and as loot in Hall of Origin, Snowpoint Temple and Reflection Cave. There's probably more good places to put it, but it's a start

Also uncomments the Beige Shard, and adds the Stale Shard for Paldea, so hopefully daily deals will only have to change once

Also added Peat Block and Black Augurite to the underground

## Motivation and Context
This plate is canon as of PLA, and it completes the set of plates and will be used for the first Arceus questline in Hisui

## How Has This Been Tested?
Game loads correctly, and item can be found in the underground

## Types of changes
- New item

## Notes
Opened as draft as we don't want to mess with daily chains until necessary


Qwertypop04:
Uncommented the Beige Shard and added the Stale Shard. Stale is the Shard that will be used in Paldea. Adding it with these other 2 items will add it to daily deals, and hopefully mean they don't have to change again for Paldea

osherwott:
Are you sure it's supposed to be "Stale" Shard? It looks like a typo of slate, which is an actual color, albeit more grey than the uncharacteristically aubergine item png (the jagged underground image is more on point though).

Qwertypop04:
> Are you sure it's supposed to be "Stale" Shard? It looks like a typo of slate, which is an actual color, albeit more grey than the uncharacteristically aubergine item png (the jagged underground image is more on point though).

I don't know for certain. I didn't name any of the shards I've added. Both of the image names I had to go from called it "Stale" but I've long found that weird myself lol.

RegisCoaxans:
You would need to add the itemList counterparts for the shards and the plate, as they live in the `player.itemList` too now. Also to change the name passed to the UGItem constructor to their saveName in itemList.

Qwertypop04:
Ah, didn't notice that had changed. Thanks

src/modules/underground/UndergroundItems.ts
```diff
@@ -85,6 +85,7 @@ UndergroundItems.addItem(new UndergroundGemItem(113, 'Stone_plate', [[1, 1, 1, 1
 UndergroundItems.addItem(new UndergroundGemItem(114, 'Toxic_plate', [[1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1]], PokemonType.Poison));
 UndergroundItems.addItem(new UndergroundGemItem(115, 'Zap_plate', [[1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1]], PokemonType.Electric));
 UndergroundItems.addItem(new UndergroundGemItem(116, 'Pixie_plate', [[1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1]], PokemonType.Fairy));
+UndergroundItems.addItem(new UndergroundGemItem(117, 'Blank Plate', [[1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1]], PokemonType.Normal));
```
CypherX, line 88:
```suggestion
UndergroundItems.addItem(new UndergroundGemItem(117, 'Blank_plate', [[1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1]], PokemonType.Normal));
```

CypherX (approved):

PR #4275, opened 2023-05-09
Added Description to MegaStoneItem

EatPant2nd:
## Description

Added a description to MegaStoneItem that's the same used for the Hoenn Mega Stones. This makes a description show up for the Wiki instead of being blank.

## Motivation and Context

Adding a description to the items makes it so the description column in the Wiki doesn't show blank, which looked bad and didn't help.

## How Has This Been Tested?

Opened the console and checked some Mega Stones to see if they displayed the correct text.


Jaaslet (approved):
The PR might be more useful than you think...
image


Thanks for the PR!

PR #4279, opened 2023-05-09
New NPC in Battle Frontier explaining that some Pokémon that you see there might not be available to catch yet

LuchoDreamer:
## Description
New NPC in Battle Frontier, completely made up.

## Motivation and Context
We get questions daily about Pokémon that appear in BF that cannot be caught yet. They are a nice teaser, so this NPC just tells you that some Pokémon might not be avilable to catch yet with a bit of flavor text. If somebody reads it, it might reduce questions.

## How Has This Been Tested?
I went to BF and the NPC was there

## Types of changes
- New NPC


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #4128, opened 2023-04-12
[TASK]: Hisui


Qwertypop04:
### Task Description

Add Hisui to the game as the 9th region

### Notes

[ ] Make the map of Hisui (Soul)
[ ] Create the questlines for region progression (Qwertypop04)
[ ] Create the postgame questlines for Arceus forms (Qwertypop04) 
[ ] Create or source sprites for all Hisuian trainers (unclaimed)

### Issues Required For This Task

None

llorens4:
any news?

stygmaskyzoo:
hi, is this still being worked on? i'm up for contributing if the project has come to a halt

PR #4280, opened 2023-05-09
Hisui Dynamic Background Sprites

Qwertypop04:
## Description
Adds Dynamic Background sprites for Hisuian Pokemon

Thanks to Soul for resizing these so we could use them!

## Motivation and Context
Required for #4128 

## How Has This Been Tested
Game runs correctly


src/modules/background/DynamicBackground.ts
```diff
@@ -82,6 +82,7 @@ export default class DynamicBackground {
         662, 663, 666, 691, 707, 714, 715, 717,
         738, 742, 743, 746, 801,
         822, 823, 826, 841, 845, 854, 855, 873, 885, 886, 887, 890,
+        642,
```
osherwott, line 85:
```suggestion
        905,
```

Qwertypop04: Er, thanks. Not sure why I did that

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4127, opened 2023-04-12
Updated Plasma Frigate unlock requirements

EatPant2nd:
Plasma Frigate dungeon becomes hidden after first clear as part of the Quest for the DNA Splicers quest line. However, it was possible to use + or - to access it anyway. Now it properly becomes inaccessible until said Quest Line has been completed which was how it became unhidden before anyway.

PR #4281, opened 2023-05-10
Bug fix and Tweak Plasma Frigate to be more canon 

osherwott:
## Summary

Plasma Frigate now actually moves!
- ~All~ Most Plasma climax battles have been moved to the Frigate from the Giant Chasm.
- Fixes the bug of returning players not being able to complete it while keeping the fix from #4127

## Description

**Changed as per canon:**
- Removed unnecessary requirements from first Frigate unlock
  - _Unova route 22 does not need to be completed to access it in the games_
  - _Removed bug-causing requirement_
- ~All~ Most Plasma Giant Chasm content has been moved to the Frigate
- The Frigate now remains a dungeon when above the Giant Chasm, rather than an extension of the Chasm

**Code-wise:**
- Added new requirements so that access is mostly dependent on the status of the questline
- The Frigate has been renamed and made into a Plasma Frigate extension instead of one for Giant Chasm in UnovaSVG.html

## Motivation and Context

There was a bug where returning players couldn't progress in the quest because of the requirement introduced in #4127 that made it so that you could only access it if you haven't beaten it, to prevent the autogeneration of quests that would send you to a dungeon not accessible on the map.

In canon, the Frigate itself is accessible from the Giant Chasm at that point in the story, and all the dialogue and battles that happen in the Giant Chasm in Pokeclicker actually happen on the Frigate.

## How Has This Been Tested?

The bugged save provided via discord was used and played through. No problems were detected.

## Types of changes

- Bug fix
- Enhanced immersion


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4283, opened 2023-05-10
Fix Base64 encoder

### LINKED ISSUE #4272: [Bug]: Adding Emoji to the catch filters prevent the user from downloading the file
### Confirmations

### What happened?

I added emoji to the catch filters because I saw it was possible to do so. Later, I realized the game prevented me from downloading the file and, since that was the only thing I did to the file, I removed them to check. This let me download the file again.

Be able to add emoji to the catch filters and the game having no issue downloading the file since it doesn't break the save in any way.

### Steps to Reproduce

1. Open the ball selector filters.
2. Add some emoji to any of them.
3. Go to menu -> save.
4. Try to download and be stopped by the game.

Windows 10

Desktop Client

### Relevant log output

```shell
Error trying to download save DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
    at Save.download (https://www.pokeclicker.com/scripts/script.min.js?v=0.10.11:2303:94)
    at HTMLButtonElement.onclick (https://www.pokeclicker.com/:47298:102)
```

### Additional Information

While the emoji doesn't let me download the save at all, the file works just fine. I have CTRL+R it multiple times to see if it would prevent me from entering the file again but no error. I assume this is similar to the Discord Rich issue?

DreamNya:
## Description

Fix Base64 encoder for save files, use `encodeURI` to encode more characters than Latin1 before `btoa`.

## Motivation and Context

Fixes #4272 
The new catch filters allow players type characters out of the Latin1 range, and will be exported when players download the save files.
However the vanilla Base64 encoder `window.btoa`  only supports to encode the Latin1 characters, otherwise it will throw an error.

## How Has This Been Tested?

Just change the name of new catch filters with non Latin1 characters like emojis or other languages.

## Screenshots (optional):

save

## Types of changes

- Bug fix


DreamNya:
image

The codes passed local test but failed with GitHub Action. Might just be a server error.

Aegyo:
Github actions had an internal error, I retried the build for you.

Aegyo:
There is a minor edge case where things that look uri encoded will get decoded on importing the save - eg name a catch filter "hello%20world", and then download and reimport the save to find it has changed to "hello world".

I think I'd be fine with that as I doubt it would really ever occur, but the btoa mdn page suggests something else which seems better: https://developer.mozilla.org/en-US/docs/Web/API/btoa - Running the string through a Uint16Array and then Uint8Array separates those multibyte characters into something btoa can handle, and you can do the reverse to reconstruct them. Can you see if that works for us?

Aegyo:
Though the example string on that page is entirely multibyte characters, I'm not sure if it works with a mixed string.

Aegyo:
> Though the example string on that page is entirely multibyte characters, I'm not sure if it works with a mixed string.

A quick copy and paste into dev tools later, it does appear to work fine with a mixture of single and multi byte characters

DreamNya:
> > Though the example string on that page is entirely multibyte characters, I'm not sure if it works with a mixed string.
> 
> A quick copy and paste into dev tools later, it does appear to work fine with a mixture of single and multi byte characters
image

MDN's solution does work, but there are some problems, such as increasing the size of the save file.

Aegyo:
Yeah, I think I prefer the accidental conversions over double file size

DreamNya:
image

I have no idea which is better.
The `replace+encodeURI` is the smaller but has the problem you pointed out

Aegyo:
The replace method has a bigger issue which is more likely to happen I think - if you have a % character not being uriEncoded, and not followed by valid characters to make a uri encoded sequence, the decodeUri fails and the save file is not able to be imported.
Eg, name a catch filter "100% shiny", download the save and try importing

Aegyo:
I think you can just get around that by including % characters in your replace though

Aegyo:
Oh, and would that then fix the %20 thing I brought up initially?

src/scripts/Save.ts
```diff
@@ -226,7 +226,7 @@ class Save {
     public static convert() {
         const base64 = $('#convertTextArea').val().toString();
         try {
-            const json = atob(base64);
+            const json = decodeURI(atob(base64));
```
Aegyo, line 229: This shouldn't be needed, `convert` is only used to import shinies from the legacy pokeclicker version, which won't have your uri encoding.

DreamNya:
> I think you can just get around that by including % characters in your replace though

I'm searching the developer docs which characters are URI strings, I think the replace method will be fine if the URI strings can be replaced together

Aegyo:
Could you add something like this to modules.test.ts ? (and try to configure your replace so that all pass)
```ts
describe('SaveSelector', () => {
    describe('encoding', () => {
        it('encodes and decodes chinese characters', () => {
            const input = '妙蛙种子';
            const decoded = decodeURI(atob(SaveSelector.btoa(input)));
            expect(decoded).toBe(input);
        });

        it('encodes and decodes emoji', () => {
            const input = '🙂';
            const decoded = decodeURI(atob(SaveSelector.btoa(input)));
            expect(decoded).toBe(input);
        });

        it('encodes and decodes uri encoded strings', () => {
            const input = 'hello %20 world';
            const decoded = decodeURI(atob(SaveSelector.btoa(input)));
            expect(decoded).toBe(input);
        });

        it('encodes and decodes non-uri % characters', () => {
            const input = '% % % %';
            const decoded = decodeURI(atob(SaveSelector.btoa(input)));
            expect(decoded).toBe(input);
        });
    });
});
```
Would also be nice for SaveSelector to have "encode" and "decode" methods it uses, so this test doesn't need to know so much about how it works.

DreamNya:
> I think you can just get around that by including % characters in your replace though

You are right, all the encoded characters have '%' and numbers, so just replace '%' together.

I tested the encode by the code
image

Aegyo:
Sorry, just getting back to this.
There is one more issue, we are currently handling this kind of thing by doing a encode/decodeURI in each place we expect user input. I think this is fine for saves created on the version with your changes, as there will be a double layer of encodeURI to undo, but on existing downloaded save files there is only a single layer of encodeURI to undo.

eg, on the live site use the pokedex to rename your Rattata to "top% rattata" and download the file. Attempting to load this file on your branch, the "% " will make the decodeURI in PartyPokemon throw an error, and the game will be prevented from fully loading.

I think with your changes we can remove all of the extra encode/decodeURI. I don't think you will need an update function, but we should still test all of these places to check.

DreamNya:
I pushed two commits.

The first is remove the nesting `decodeURI` and `encodeURI`.
So, we don't need to add paired `decodeURI` and `encodeURI` for every input.

The second is a compatible function.
Due to the lack of escape for catch filters in old save function, if the player input `%` in the catch filters, it will be accidentally escaped in the new function to prevent importing savefile. 
So, I add a `try...catch` block to return the raw decodeString if `decodeURI` failed. In this way player may import a savefile with unparsed input names. I think this only affects those who enter % in the catch filter, which should be very few.

src/modules/SaveSelector.ts
```diff
@@ -87,6 +87,19 @@ export default class SaveSelector {
...
+    }
+
+    static atob(encodeString: string):string {
+        const decodeString = atob(encodeString);
+        try {
+            return decodeURI(decodeString);
+        } catch {
+            return decodeString;
+        }
+    }
```
Aegyo, line 101: I haven't had a chance to test it, but I think there is an issue with this. If we have an old save with a % in a catch filter and also a pokemon nickname, the nickname one will have been uri encoded but not the catch filter, so we error and drop into the catch block, and just return the decodeString. This leaves us with all of the nicknames etc still uri encoded.

I think we will need to always return decodeString here based on the save version in decodeString, and make an update function to handle decoding everything that was being handled individually before this version.

DreamNya: This is indeed a flaw, because it is difficult to distinguish whether % is encoded or not.
So I chose a compromise solution, returning an undecoded save string to ensure that the save file can be imported successfully. This does have an effect, but I think it's very slight, only in 0.10.11 → 0.10.12, and only affects those who enter % in filters and elsewhere.

DreamNya: If you really want a update function, I added it in the latest commit.
It's just a way to handle it and I'm not sure if the nesting `try...catch` block does anything to future version.
Feel free to modify my code.

If you need a save file to test, you can try the attachment in live version, prev `try...catch` version and current nesting `try...catch` version.
[[v0.10.11] PokeClicker 2023-05-22 11_39_08 .txt](https://github.com/pokeclicker/pokeclicker/files/11583075/v0.10.11.PokeClicker.2023-05-22.11_39_08.txt)

src/scripts/Update.ts
```diff
@@ -2305,10 +2305,10 @@ class Update implements Saveable {
...
         const button = document.createElement('a');
         try {
-            button.href = `data:text/plain;charset=utf-8,${encodeURIComponent(btoa(backupSaveData))}`;
+            button.href = `data:text/plain;charset=utf-8,${encodeURIComponent(SaveSelector.btoa(backupSaveData))}`;
             button.className = 'btn btn-block btn-warning';
             button.innerText = 'Click to Backup Save!';
-            const filename = settingsData.saveFilename ? decodeURI(settingsData.saveFilename) : Settings.getSetting('saveFilename').defaultValue;
+            const filename = settingsData.saveFilename || Settings.getSetting('saveFilename').defaultValue;
             const datestr = GameConstants.formatDate(new Date());
             button.setAttribute('download', GameHelper.saveFileName(filename, {'{date}' : datestr, '{version}' : this.saveVersion, '{name}' : decodeURI(saveData.profile.name)}, true));
```
Aegyo, line 2313: I think this decodeURI needs to go as well, otherwise looks all good :+1: 
```suggestion
            button.setAttribute('download', GameHelper.saveFileName(filename, {'{date}' : datestr, '{version}' : this.saveVersion, '{name}' : saveData.profile.name}, true));
```

Aegyo (approved):

PR #4286, opened 2023-05-10
Give Rainbow Rocket Legendaries the same HP

Qwertypop04:
## Description
Gives each Leader with 2 legendaries, 2 legendaries with the same HP, instead of one being much stronger than the other

## Motivation and Context
It does not make sense that one of them is stronger than the other, is likely an oversight

## Types of changes
- HP fix


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4292, opened 2023-05-12
Fixed map colorization of Officer Jenny Pinkan Trade Shop

morcousw:
## Description

The code that figures out what to color each route/town/dungeon was previously not checking Berry Master Trades. This only made a difference at Officer Jenny's Pinkan Trade Shop in Sevii Islands. It was always showing as complete whether all the pokemon had been acquired, shinied, or EV'd.

I added a section that would check if the town is a Berry Master, and if so check all of its PokemonItems for Shiny/EV.

## Motivation and Context

Motivation is that it was annoying me since I'm trying to Shiny/EV Sevii right now

## How Has This Been Tested?

I tested the coloration of all Berry Masters (Johto, Hoenn, Sinnoh and Sevii). Specifically, I checked out the displayed color when:
- Johto at Goldenrod City when the gym and/or the Radio Tower achievements are not obtained yet, as well as when the Dungeon has 0 clears. And then after both are cleared.
- Hoenn at Mauville City when the gym is not yet cleared  + after it is cleared
- Sinnoh at Hearthome City when the gym is not yet cleared + after it is cleared
- Sevii at Officer Jenny's Pinkan Trade Shop when there is an unobtained Pinkan pokemon, when there is a non-shiny pinkan pokemon, when there is a non-resistant pinkan pokemon, when all are captured, shiny, and resistant.

The colors displayed as expected based on my Settings.

## Screenshots (optional):

image

## Types of changes

- Bug fix


Jaaslet (changes requested):
This code should be in `BerryMasterShop.ts:area Status()`. Look at how it's done in Shop.ts

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4296, opened 2023-05-13
Fixed broken image in Detective Pikachu Quest Line

EatPant2nd:
## Description

Aipom Alley.png wasn't showing correctly in the appropriate quest step since it was missing the imageName in the temporary battle. This fixes that.

## Motivation and Context

Image was broken and that's not good.

## How Has This Been Tested?

Redid the quest line until I hit the spot where the image should show, and it did.

## Types of changes

- Bug fix


osherwott (approved):

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4298, opened 2023-05-13
Additional Held Item filters

### LINKED ISSUE #4191: [SUGGESTION]: Additional filter for held items
### Current Behavior

Currently the game filters held items based on the common filter options like stats, breeding, id etc.

image

### Improved Behavior

I would like an additional filter like is done when applying vitamines to filter based on region.

image

### Reason for change

The reason is that me (and I assume many others) only know about a few generations of pokemon so as soon as you enter into unfamiliar regions it is very difficult to figure out which pokemons are best to use your limited amount of held items on for the region you are in. 

### Detailed Description

By not being able to filter regions it makes it difficult for me to figure out which pokemon are from the region I am in and do not get debuffed attack.

As it is now I have to go through the trouble of first sorting by pokemon id, write down the names (or have another tab open on the internet with all pokemon from a region) and then filter on breeding and look for the right names. A region filter would save players a lot of trouble.

Basically I would just like the list to not be cluttered with pokemon from all regions but be able to select pokemon from 1 region.

CypherX:
## Description
Adds additional filters to the held item modal and a few small UI adjustments (mainly spacing).

## Motivation and Context
Improves functionality and makes managing held items a bit easier/quicker. There have been a few requests for these additional filters.

Closes #4191 

## How Has This Been Tested?
Tested that the filters work correctly.

## Screenshots (optional):
image

## Types of changes
- New feature
- UI improvement

### DIFF
--- a/src/components/heldItemModal.html
+++ b/src/components/heldItemModal.html
@@ -1,5 +1,5 @@
 <div class="modal noselect fade" id="heldItemModal" tabindex="-1" role="dialog" aria-labelledby="heldItemModal">
-    <div class="modal-dialog modal-dialog-scrollable modal-dialog-centered modal-sm" role="document">
+    <div class="modal-dialog modal-dialog-scrollable modal-sm" role="document">
         <div class="modal-content">
             <!-- ko if: modalUtils.observableState['heldItemModal'] === 'show' -->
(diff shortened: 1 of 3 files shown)


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4302, opened 2023-05-13
Add quests for catching shadow pokemon

Farboo171:
## Description
Adds the tracked statistics need for, and the actual function of a "Catch X Shadow Pokemon" quest

## Motivation and Context
This will fill out Orre content and add a new quest to the daily rotation

## How Has This Been Tested?
Tested that the quest does not show up until one shadowmon has been caught.
Tested that catching a shadowmon completes the quest

## Types of changes

- New feature


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4307, opened 2023-05-14
Added Safari Level

### LINKED ISSUE #3812: [TASK]: Implement Safari XP
### Task Description

Create a global XP statistic that is gained by catching Pokémon in the Safari Zone.

A higher Safari level will allow for:
- Better Pokémon
- Better rocks/bait/whatever Safari upgrades
- Better loot (if/when) implemented

### Notes

- Benefits of a higher level don't need to be implemented in the first PR
- Exact xp gains and level formula TBD
- Safari level and xp progress in current level should be visible somewhere, like quest xp/level is

### Issues Required For This Task

None

Jaaslet:
## Description

Safari level!
This PR will add safari level, where you gain exp by throwing bait/rocks and catching safari pokemon.
Rocks will become better with each level.
~~Different berries can be used as bait, depending on your level (the berries and their balance will be in an other PR).~~ Bait code is weird and will require a bit more. So that will also be in a new PR
This will effect animation speed, somehow. But that is a bigger project (and i am not exactly sure what to do), so that will be added in an other PR. But the current idea is that we have 2-3 items, that increases the animation speed, which are unlocked at specific levels.

## Motivation and Context

Closes #3812
Hopefully this makes the safari zone more fun to do, so we can release friend safari without feeling bad

## How Has This Been Tested?

Played through and used the different features to see that the EXP goes up

## Types of changes

- New feature

### DIFF
--- a/src/components/safariModal.html
+++ b/src/components/safariModal.html
@@ -26,6 +26,15 @@
                         />
                     </span>
                 </h4>
+                <div class='col-6'>
+                    <div>Lvl. <knockout data-bind='text: Safari.safariLevel()'></knockout></div>
+                    <div class="progress" style='height: 5px'>
+                        <div class="progress-bar bg-info" role="progressbar"
+                             data-bind="attr:{ style: 'width:' + Safari.percentToNextSafariLevel() + '%' }"
+                             aria-valuemin="0" aria-valuemax="100">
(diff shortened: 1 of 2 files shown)


Jaaslet:
Draft because i am not done yet

PR #4314, opened 2023-05-15
Adds Poké Spots to Orre

LuchoDreamer:
## Description
Added the three Poké Spots from Orre as routes for Orre with their canon pokémon.

## Motivation and Context
Eugene asked me nicely to add them

## How Has This Been Tested?
I go to Orre and the three poké stops are there and I can go there (but they don't work, Jaas is gonna check what's happening)

## Extra Notes
HP is a kind of random value since I don't know exactly the balance Eugene want to give these three routes. Requirement right now is DevRequirement for the same reason.

### DIFF
--- a/src/components/HoennSVG.html
+++ b/src/components/HoennSVG.html
@@ -1113,6 +1113,35 @@
             href:'assets/images/orre.png?v=$VERSION'
         }"></image>
 
+    <!-- ------ -->
+    <!-- ROUTES -->
+    <!-- ------ -->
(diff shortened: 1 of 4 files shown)


Jaaslet (approved):
I changed the hp to fit the end of the region (where i think they will be used/useful). Don't know if it's correct, but i think it's better, in case we forget the balance and just releases it without changing the values.
Looks good. Thanks for the PR

PR #4320, opened 2023-05-16
Berry Master Vitamins

### LINKED ISSUE #3924: [SUGGESTION]: Sinnoh Berry Master trades
### Current Behavior

The berry master from sinnoh can only trade masterballs and proteins.

### Improved Behavior

I believe the berry master should also trade carbos and calcium as they're both vitamins

### Reason for change

Only 1 of the vitamins is present in the sinnoh berry master pool, adding the other 2 would make it feel more complete

### Detailed Description

Only masterballs and proteins are in the sinnoh berry masters trades, if the other 2 were added then maybe it would become less useless as the vitamin grind is now bigger it would make it atleast less boring and waste less time to grind enough to efficiently max all of the pokémons (i've seen a few people trying to do this but it costs too much so this maybe would help).

osherwott:
Adds Calcium and Carbos to the Sinnoh Berry Master if you have the Basic Badge
Closes #3924 

## Description

As stated above

## Motivation and Context

Closes the aforementioned suggestion

## How Has This Been Tested?

Loaded a save file before and after Unova, and the new trades appeared.

## Types of changes

- New feature


Jaaslet:
Any thoughts behind the price?
I am considering if they should be more expensive for consistency (or make proteins cheaper, if this makes them stupidly expensive)

osherwott:
After some feedback, prices have been adjusted to (hopefully) incentivize more players to trade. These change include:
- Limiting fifth gen berries only those that can be grown in under a day
- Adjusting the cost of Protein to reflect its new use
- Making the cost of Calcium and Carbos to be less demanding
If there are any foreseeable problems with these changes, please feel free to comment!

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4324, opened 2023-05-17
Performance Optimizations

CypherX:
## Description
Identified a couple areas that could be optimized while doing some performance profiling.

### calculateTownCssClass
This function is called A LOT and is a primary source of lag/stutters. I changed a few things (mostly related to dungeons) to optimize it which drastically increased it's overall performance:
- Dungeon `isThereQuestAtLocation` - Changed this to a pureComputed on the dungeon object (there is also a similar function for gyms but it barely registered in my profiling so I didn't change it)
- `DungeonRunner.dungeonCompleted` - This is used to determine if all pokemon available in a dungeon are caught. It's called 3 times per dungeon, I optimized it down to one time per dungeon.
- Dungeon `getCaughtMimics()` - This is used by the `dungeonCompleted` function in the previous bullet point and also happens to be the slowest part of that call as shown by my profiling. I tweaked it to store a list of mimics on the dungeon object (populated in the dungeon constructor) that is then used by `getCaughtMimics()` rather than iterating over the entire loot table and doing a bunch of `pokemonMap` lookups.

### Hatchery / Helpers
When a hatchery helper has an open egg slot it sorts the caught pokemon list then traverses the sorted list to find the first pokemon that meets the criteria to be hatched. This sorting happens every time they need an egg.

~~The first change is to make that sorted list a pureComputed property on the helper so that the sorting happens only when necessary, such as the sorting option/direction changes, or a new pokemon is caught.~~
Modified how helpers select pokemon again after a suggestion by Jessica, see discussion below.

The second change is to move `BreedingController.visible()` to a pureComputed property (`isHatchable`) on the `PartyPokemon`. This is used by both the hatchery modal and helpers when looking for their next pokemon to hatch.

Benchmarking numbers for helpers as they currently are vs. with my changes. This is the total time, in milliseconds, for 3 hired helpers to search for a pokemon to hatch. Filters were set up in a way so that no pokemon would be found therefore all 3 would sort (on live) and traverse my entire caught list of 1304 pokemon.

image

As the numbers indicate there is a significant performance gain with the above changes.

## How Has This Been Tested?
Played normally to verify everything functions the same but feels more smooth overall.
Had several crew members test and also affirm that things felt improved and had no noticeable issues.

Link to test: https://cypherx.github.io/pokeclicker/

## Types of changes

- Performance


Aegyo:
If the HatcheryHelper sortedList is only used to get the next pokemon to hatch, then we don't need to sort it. We can just run through the list, ignore non hatchables, and keep track of the best match so far according to the comparison function.
That should be O(n*log(n)) -> O(n), probably a significant improvement at max caught list

Aegyo:
Also would be nice to see what you did to get those numbers, so anyone can repeat the testing theirself

CypherX:
> If the HatcheryHelper sortedList is only used to get the next pokemon to hatch, then we don't need to sort it. We can just run through the list once, ignore non hatchables, and keep track of the best match so far according to the comparison function. That should be O(n*log(n)) -> O(n), probably a significant improvement at max caught list

Can you give an example of what you mean? I'm kind of following but not quite.

> Also would be nice to see what you did to get those numbers, so anyone can repeat the testing theirself

In Breeding.ts:
```
const startTime = performance.now();
this.hatcheryHelpers.addSteps(amount, this.multiplier);
const endTime = performance.now();
console.log(`addSteps - ${endTime - startTime}ms`);
```
Quite an advanced technique, I know

Aegyo:
instead of having its own sorted list, the hatchery helper can get the next pokemon by
```ts
const compare = PartyController.compareBy(this.sortOption, this.otherThing, regionDebuffObs);

const pokemon = App.game.party.reduce((best, pokemon) => {
    if (!pokemon.isHatchable()) {
        return best;
    }
    if (best === null) {
        return pokemon;
    }
    return compare(best, pokemon) >= 0 ? best : pokemon;
}, null);
```
I'm typing on mobile and I forget if compare > 0 means they need swapping or the order is good, so that might need tweaking to work properly, but that's the general idea I have.

Only needs to run through the list once, this should be far less comparisons needed than sorting the full list. Also I don't think a computed is worth using here unless we want to display the next pokemon somewhere. We put the pokemon in an egg, so a computed would need to recalculate at least every time this code runs anyway - might as well do the calculation right here and not worry about other things triggering a recalc when we don't need it.

Aegyo:
actually a pure computed holding the next pick would be useful when all options are being filtered out and the helper is waiting for a new mon, so you don't have to look through the list every time you gain eggsteps

CypherX:
Something like this on the helper?

```ts
this.nextPokemon = ko.pureComputed(() => {
    if (!this.hired()) {
        return null;
    }

    const compare = PartyController.compareBy(this.sortOption(), this.sortDirection(),
        +Settings.getSetting('breedingRegionalAttackDebuffSetting').observableValue());

    return App.game.party.caughtPokemon.reduce((best, pokemon) => {
        if (!pokemon.isHatchable()) {
            return best;
        }
        if (best === null) {
            return pokemon;
        }
        return compare(best, pokemon) <= 0 ? best : pokemon;
    }, null);
});
```

From a quick test it seems about the same speed or very slightly faster than looking through the list each time egg steps are gained (I'm talking like maybe 1-2ms average difference)

Aegyo:
Yeah something like that. If you aren't seeing much difference having it in a computed it probably isn't worth doing that.

Is there a noticeable difference between this and the sorting?

CypherX:
Yeah, both ways are faster than what I had. Did some more testing, including in a different browser, and the pureComputed actually seems just very slightly slower.

image

CypherX:
Pushed changes w/ Jess' suggestion

Aegyo (approved):

Issue #4326, opened 2023-05-17
[SUGGESTION]: Weather icon Missaligned in the dock modal


Someonealive-QN:
### Current Behavior

The weather icon is missaligned in the dock modal(see screenshot)

### Improved Behavior

stretch a bit the weather icon so it matches the current region selection button

### Reason for change

it's a little bit unsatisfying with it being a little smaller than the button

### Detailed Description

the weather icon that shows the weather in the region in the dock modal is missaligned with the select region button and it's quite unsatisfying to see it

LuchoDreamer:
It is perfect on Cyborg, could you list the themes where it's misaligned?

Someonealive-QN:
> It is perfect on Cyborg, could you list the themes where it's misaligned?

sure, lemme make a quick list really quick

Someonealive-QN:
Cerulean, Cosmo, Darkly, Flatly, Journal, Litera, Lux, Materia, Minty, Sketchy, Slate, Solar, Spacelab, Superhero, United and Yeti

LuchoDreamer:
Those are a lot, that is even misaligned in default is... weird.

CypherX:
The height for the weather icon/button is hardcoded to 36px, that's why it's off in some themes that have different sized buttons.

PR #4328, opened 2023-05-18
Fix: Weather button misaligned in dock modal

Crobat4:
## Description

Closes https://github.com/pokeclicker/pokeclicker/issues/4326
Weather button in the dock modal is misaligned in some themes due having a fixed height, this PR removes the fixed height on said button.
I also tried to emulate the Lumen theme button effect on it.

In some themes, the weather button may look bigger than the region button but that's because the region button may have a border color that blends with the background. If you prefer, I can give the same border color to both buttons.

## Motivation and Context

Forgot to check other themes...

## How Has This Been Tested?

* Checked every theme to see if both buttons have the same height (through inspect element)

## Screenshots (optional):

Bigger button to see if the button is aligned:
image

Lumen effect:
image

## Types of changes

- Bug fix


Jaaslet (approved):
Looks good. Thanks for the PR

PR #4329, opened 2023-05-18
Event calendar can now be bought and can start events early

Jaaslet:
## Description

- The event calendar can be bought in Hammerlocke (might change). Each time we release a new region, it should be moved to that region.
- If you have the event calendar, and an event is not active, you can start the event from the Event-menu in the start menu. I have considered moving this up, now that it's more important, but this PR is getting big already, and it's a nice-to-have thing
- To activate an event, it costs `number_of_days_left_to_event*500` QP. It's intended for lategame players, with nothing else to do, so it's expensive
- When you activate it, it's activated for 24 hours (countdown will stop when you are offline).

## Motivation and Context

People have asked for it, and i promised it like 8 months ago...

## How Has This Been Tested?

I have started events, seen that they are saved and the timer ticks down. I have checked that the Let's go event roamers shows up

## Types of changes

- New feature


Jaaslet:
Draft because it needs crew test and Halloween has not yet been refactored (so we need to figure out what to do with that event)

Jaaslet:
I have added a parameter to hide the activate button from the calendar. Used for Halloween, so we don't need to wait for the halloween refactor, to merge the event calendar. I also want to use that parameter for when/if we add new events, so people with money cannot try the event before everyone else

Jaaslet:
I have a list of stuff i want to do. But they are all nice to have, and not required for the event calendar to work. And because i want this PR to be finished, and not abandoned (like the first one) and i want it to be easier to review, they will be added in PRs after this one is merged. The list is (people are allowed to help, if they want to):
- Refactor notifications to use ticks instead of setting the timer
- Refactor the SpecialEvents class to not have anything static methods/variables
- Add a setting to disable special events (i promissed someone to do it, after i closed their PR attempting to do it)
- Change the halloween event to fit the new code, so that can be used by the event calendar too
- Make keyitems clickable (event calendar should open events and gemcase should open gems)
- Move the event button up in the start menu
- Add an enum with event names, and use that instead of the title, as identifier, in the code
- Add a warning, if you are starting more than one event at a time

Jaaslet:
Crew likes it. It's ready for review

src/modules/specialEvents/SpecialEvent.ts
```diff
@@ -59,6 +66,47 @@ export default class SpecialEvent {
...
+            Notifier.notify({
+                title: 'Cannot afford',
+                message: `This costs ${price} QP.`,
+                type: NotificationOption.danger,
+            });
+            return;
+        }
+        Notifier.confirm({
+            title: 'Do you want to start this event early?',
+            message: `Starting '${this.title}' early will cost you ${price} QP.`,
```
Aegyo, line 101:
```suggestion
            message: `Starting '${this.title}' early will cost you ${price} QP for 24 hours of event time..`,
```

Aegyo (approved):

PR #4331, opened 2023-05-19
Category Improvements (part 1) - category management

### LINKED ISSUE #4063: [SUGGESTION]: Ability to reorganize Hatchery categories
### Current Behavior

Currently, there is no way to re-organize already created Hatchery categories.

### Improved Behavior

With this suggestion, I hope to be able to drag a category from where it is and drop it where I want it to be in the list.

### Reason for change

The main reason for this change is that it would be a major quality of life change, because as it is right now, I'd have to delete all of the categories below the new one that I want to create and then re-organize all of the Pokemon that were in the categories below the one I wanted to create and put them back in their original categories.

### Detailed Description

There really isn't much of a detailed description that I can give as I simply want to be able to, in the Edit Categories window, simply be able to drag a category to where I want it in the list of categories in the Hatchery, and then drop it by releasing the left mouse button so that it is where I want it to be in the category list.  Another idea would be the ability to move all Pokemon from one category to another via some sort of button or by some means.

Thanks for taking the time to read this suggestion!

CypherX:
## Description
The first step of my plan to improve categories and make them more useful.

- Categories can now be re-ordered in the Edit Categories modal, this order will be reflected everywhere they are displayed.
- Ability to move all pokemon from one category to another category.
- Remove Category button hidden for the None & Favorite categories as we do not allow the player to remove those.

(Thanks Jessica for fixing the sortable binding code)

## Motivation and Context
Categories are neat but also limited in their usefulness. This aims to improve that usefulness and overall management of categories.

Closes #4063 

## How Has This Been Tested?
Re-ordered categories & reloaded the page to verify the order persisted.
Moved pokemon between categories several times.

## Screenshots (optional):
image
image

## Types of changes
- Feature update
- UI improvement


CypherX:
Additional plans for categories (disclaimer: none of these are guaranteed):

- Assign a pokemon to multiple categories
- Filter by multiple categories in the pokedex/hatchery
- Ability to assign hatchery helpers to a specific category
- Use icons/shapes to differentiate categories in addition to color?
- Category pokeball filter?

src/modules/party/Category.ts
```diff
@@ -106,10 +105,11 @@ export default class PokemonCategories implements Saveable {
     }
 
     fromJSON(json: Record<string, any>): void {
-        if (!json) {
+        if (!json || !json.categories) {
```
Jaaslet, line 108:
```suggestion
        if (!json?.categories) {
```

Jaaslet (changes requested):
- If you switch the two bottom categories, and then deletes the bottom, they both go away
Other than that, it looks good

CypherX:
> * If you switch the two bottom categories, and then deletes the bottom, they both go away
>       Other than that, it looks good

Fixed (thanks Jess for the ko sortable fix. Note that the change in `pokeballSelectorAdvModal.html` is related to this)

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4335, opened 2023-05-19
Shadow Stats Fix

Farboo171:
## Description
When I added shadow stats in #4302 , they didn't work correctly. Now they should
Also removed encounter stats, because they were meaningless and could not be incremented.

## Motivation and Context
Fixing a bug I created

## How Has This Been Tested?
I tested all the stats and made sure they increment when they should, and don't when they don't

## Types of changes

- Bug fix


Jaaslet (approved):
Looks good. Thanks for the PR

Issue #3303, opened 2022-11-04
[SUGGESTION] Allow for Attack value display to be toggled to true attack against individual Pokemon


SteelWulf1:
**Current Behavior**
The game currently displays "Pokemon Attack" as a sum total of all the base attack values of the Pokemon you own. If you would like to see the exact damage you're doing against any given Pokemon you're fighting, you can hover over the number and it will show it for you.

**Improved Behavior**
Allow a setting in the options menu for this to become a toggle, so the mouse doesn't have to remain parked on the Attack value at all times for people like myself who would rather see that number.

**Reason for change**
Minor convenience/QOL change.

**Screenshots:**
![image](https://user-images.githubusercontent.com/30699627/200017371-c5ce15c1-e924-422c-8844-a879c247972c.png)
Attack display on mouse-hover, suggestion to become a setting toggle.

Farboo171:
After an internal discussion, it was determined this was not worth implementing, as the information is already easily available in hover text. Closing.

DraKay:
Then hover text should be hinted, because people don't know it and discover it in later regions when they are said in discord

DaveYognaught:
To clarify, is it "not worth" implementing or "will not" be implemented. It seems like a rather minor QOL change, so while I appreciate it's not the highest priority if someone were to submit a PR for it anyway don't see any reason why it would be rejected.

Farboo171:
The wording that was voted on was "Do not add the setting, it being in hover is good enough"
If someone wants to make a FAQ entry to NPC talking about the hover text, that would probably be fine. If they felt really passionate about it.

PR #4346, opened 2023-05-21
Add damage info to the FAQ

DraKay:
Here I am, changing the Faq again... Wait a surprise right? 

## Description
Added info about how damage is calculated to the faq + where to check the total damage

## Motivation and Context
This info wasn't in faq and is basic. Check comments in #3303 

## How Has This Been Tested?
Eeeh..... 

## Types of changes

- UI improvement

### DIFF
--- a/src/components/FAQModal.html
+++ b/src/components/FAQModal.html
@@ -25,6 +25,12 @@
             <p><strong>How do I get Dungeon Tokens?</strong><br/>
             <span class="text-muted">Read the in-game Tutorial.</span><br/>
             <span class="text-muted">Set the &quot;Caught&quot; selector in the top left to Poké Ball to capture Pokémon you have already caught.</span></p>
+            <p><strong>How is damage calculated?</strong><br/>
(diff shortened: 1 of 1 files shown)


Farboo171:
My suggested edit, slightly re-ordering and adding punctuation, plus a few minor edits:
--

The damage you do against an enemy Pokémon is the sum of the attack of all Pokémon the player has, after type effectiveness, type weakness, and other multipliers are applied.

For enemy Pokémon with two types, effectiveness and weakness is applied once to each type.

For enemy Pokémon with one type, effectiveness and weakness is applied twice to that single type.

For players' Pokémon with two types, only the most efficient type is applied.

Other factors that will multiply damage value include flutes, battle items, and weather. Gems can be used to change the effectiveness and weakness multipliers as well.

Players can check the attack against a specific Pokémon they are fighting by **hovering over Pokémon attack value**.

--
"Effectiveness and weakness" is the term used by a lot of official pokemon sources.

DraKay:
Changes requested commited

Jaaslet (approved):
Looks good. Thanks for the PR

PR #4352, opened 2023-05-23
Orre Achievement Percentage to 75%

Farboo171:
## Description
Changes Orre Achievement Percentage to 75%

## Motivation and Context
We voted and this was the end result

## How Has This Been Tested?
I didn't test it, but it was just 2 figures changing

## Types of changes

- Balance change


Jaaslet (approved):
Just for future context: Magikarp jump was 25% because we expected hisui to have 75%. But now that it's a region, it will have 100%.
And Magikarp jump is small, while Orre spans two regions. So this split made sense.

Looks good. Thanks for the PR

PR #4357, opened 2023-05-25
Reduce time for underground generation

RedSparr0w:
## Description

Reduce the time taken to generate layers in the underground

## How Has This Been Tested?

Changing the mine to get the list of available items first did almost half the execution time for me.

Was running this script to test the timings:
```js
function measureExecutionTime(fn, numRuns) {
  let totalExecutionTime = 0;

  for (let i = 0; i < numRuns; i++) {
    const startTime = performance.now();
    fn();
    const endTime = performance.now();

    const executionTime = endTime - startTime;
    totalExecutionTime += executionTime;
  }

  const averageExecutionTime = totalExecutionTime / numRuns;
  return averageExecutionTime;
}
console.log('Average completion time:', measureExecutionTime(() => Mine.loadMine(), 5000))
Mine.completed()
```
Current code was giving me `1.9ms` to `2.2ms`
Getting the list first, shuffling it and then popping items out of it gave me `1.1ms` to `1.3ms`

## Types of changes

- Optimization


Aegyo:
This code looks like we no longer take into account the item weight, so items which are supposed to be rare (I think just the mega stones?) appear with the same frequency as everything else.

RedSparr0w:
Nice catch, forgot they were weighted tbh

RedSparr0w:
Should be using weighted shuffle now,
slight increase, takes about `1.2ms` to `1.4ms` now

Aegyo (approved):

PR #4361, opened 2023-05-26
Fix negative quest amounts

### LINKED ISSUE #3123: [BUG] Pixie Plate Negativ
**Info**
 - OS :  Windows 10
 - Browser : Chrome and Huawei Browser (on phone)
 - Version : v0.9.16

**Description:**
To complete the Team Rocket's quest to earn 1000 gems, I sell my pixie plates, but the count of pixie plates went negativ (so the quest doesn't complete). 

**Steps to Reproduce:**

1. Go to 'Collect Fairy Gems and Pixie Plates Quest'
2. Sell your Pixie Plates
3. The Gain Pixie Plates will be negativ (but I gained it)

**Screenshots:**

image

**Additional Information:**

RedSparr0w:
## Description
Fixes quests going into negative amounts

Alternative solution to PR #4358 - Felt a little too complex to me, not sure which solution works better though tbh

## Motivation and Context
Some quest that require collecting items will go negative if you sell those items

## How Has This Been Tested?
Selling items that are required for quests, gaining more

## Types of changes
- Bug fix

closes #3123 - bug report
closes #4358 - alt PR


RedSparr0w:
Oh, some of the quest items aren't observable 🤔

RedSparr0w:
Should be fixed now 🤞

DaveYognaught:
Yup, seems to be fixed! Reloading works now too and selling all of your inventory doesn't disrupt it either. Does what my PR does and more so i'd be inclined towards this as a more complete (and cleaner) fix. 


The only thing of note here is that the _focus.subscribe seems to hook into a lot more than the current quests.
For instance, it'll run twice everytime you defeat a pokemon on route.

Not too sure if that's of concern? I'm not too familiar so correct me if i'm wrong, but would that mean it's disposing and starting a new subscribe each time an observable increases? ie, for defeating a pokemon on route, it's the number of defeated and I guess cash? I was going to suggest if we could dispose on completion also, but not sure if that matters.

RedSparr0w:
> it'll run twice everytime you defeat a pokemon on route.

Do you have multiple quests active that are tracking defeated pokemon?

> but would that mean it's disposing and starting a new subscribe each time an observable increases?

It only disposes of it and recreates it whenever the focus changes (if we are tracking something else instead of the current quest), so basically shouldn't change.

> I was going to suggest if we could dispose on completion also, but not sure if that matters.

Yeah probably a good idea, no harm in adding it 👍(added)

DaveYognaught:
So here's a screencap of when this is running on a new account. Seems to be more than I thought!
Obviously i've got this quest to defeat pokemon on Route 2, but I've highlighted that quest when it ran in the console. 

Other observables of note appear to be showing up are total defeated, dungeon tokens, total cash, etc. 
Plus some I don't even recognise. No clue. 

Observables

I noticed your code runs outside of this check that it used previously: 
`this.initial() !== null`

I think that was being used to limit the scope. If it doesn't have an initial, then we don't need to track it right?
So out of curiousity, I tried copying it inside it like so:


            if (this.initial() !== null) {
                // Dispose of our old subscriber if one exists
                this.focusSub?.dispose?.();
                // Subscribe to the new focus
                this.focusValue = this._focus();
                this.focusSub = this._focus.subscribe?.((newValue) => {
                    // If the focus goes down, adjust our initial value
                    if (newValue < this.focusValue) {
                        this.initial(this.initial() - (this.focusValue - newValue));
                    }
                    this.focusValue = newValue;
                    console.log('v2 focusValue: ' + newValue);
                });
                return Math.min(1, ( this.focus() - this.initial()) / this.amount);
            } else {
                return 0;
            }
 
This results in a new account doing the same thing getting the following console logs: 

ObservablesV2

Way less now, but still some extra fluff. However, this fluff also exists in my own PR and live. So it's good as far as i'm concerned.

Not sure if there's a better way to limit when your code runs? If not, just checking for an initial does the job it seems. 

I've also tested the Pinkan quest once more just to make sure it didn't impact how that functions, but they always meet the criteria to run. So no impact on that front.

RedSparr0w:
Yeah, it looks like every custom quest has a focus value set upon loading the game, so this is setting `.progress` and `.progressText` to track every time any of those values change too, definitely seems like we need to rework quest loading, but that's seems to be a whole different issue.

DaveYognaught:
> definitely seems like we need to rework quest loading, but that's seems to be a whole different issue.

Yeah I agree. 
However, just to get this fix over the line, what is your thoughts on just moving your code inside the `this.initial() !== null` check and then pushing it out as is? Would bring it in line with current behaviour, not doing or tracking anything excess it isn't already beyond the scope of fixing quest tracking. With a TODO to look at improving this later down the line.

I think this PR is sufficiently important enough to push this out with those changes.

Jaaslet (approved):

PR #4364, opened 2023-05-27
Orre Colosseum Quest

Farboo171:
## Description
Adds the main story for Pokemon Colosseum to Orre. This is just the main plot, walking the character through where Shadow pokemon come from and stopping the evil Team Cipher

## Motivation and Context
This quest is the first half of the total plot for Orre, to be done post-Hoenn E4 and thru Sinnoh difficulty.
The quest itself is internally balanced, but some of the locations might get compressed down.
Location unlocks are not included yet, I am adding this quest first as a spine to attach the unlocks to, then add some NPCs to explain what everything is supposed to do.

## How Has This Been Tested?
I have run through the quest to make sure all the steps work.
There's no point in crew testing right now because all the location unlocks are still a WIP. That can happen when everything but images is ready.

## Types of changes
- New feature


src/scripts/quests/QuestLineHelper.ts
```diff
@@ -1035,6 +1035,120 @@ class QuestLineHelper {
         App.game.quests.questLines().push(metaGroudonQuestLine);
     }
 
+    //Orre Questlines
+    public static createOrreColosseumQuestLine() {
+        const orreColosseumQuestLine = new QuestLine('Shadows in the Desert', 'Explore Orre and uncover the origin of Shadow Pokémon.', new GymBadgeRequirement(BadgeEnums.Elite_HoennChampion), GameConstants.BulletinBoards.Hoenn);
```
Aegyo, line 1040: Does this need a DevelopmentRequirement?

Farboo171: Good call, added that commit

Aegyo (approved):

PR #4377, opened 2023-05-29
Safari Level

Farboo171:
## Description
Adds Safari Level to the Safari Zones, formula in images

## Motivation and Context
Allows locking things behind safari levels, and improvements to various safari elements based on level. Not in this PR, this is just the bones of the Level

## How Has This Been Tested?
Checked XP being gained for rock throwing, bait throwing, ball throwing, and mon catching. Checked that level does increase and the bar resets on a level.

## Screenshots (optional):
image

## Types of changes

- New feature

### DIFF
--- a/src/scripts/safari/Safari.ts
+++ b/src/scripts/safari/Safari.ts
@@ -20,8 +20,8 @@ class Safari {
     static maxSafariLevel = 40;
     static safariExp: KnockoutComputed<number> = ko.pureComputed(() => {
         return App.game.statistics.safariRocksThrown() * 10 +
-            App.game.statistics.safariBaitThrown() * 10 + 
-            App.game.statistics.safariBallsThrown() * 5 + 
+            App.game.statistics.safariBaitThrown() * 10 +
+            App.game.statistics.safariBallsThrown() * 5 +
             App.game.statistics.safariPokemonCaptured() * 50;
     });
     static safariLevel: KnockoutComputed<number> = ko.pureComputed(() => {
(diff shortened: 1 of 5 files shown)


src/scripts/safari/Safari.ts
RegisCoaxans, line 1: I would recommend decreasing XP from baits and increasing XP from balls thrown since you can basically throw baits forever and for free while preventing the mon to escape too soon (and the keybind makes it AFK-compatible) whereas balls are limited per run.
image

RegisCoaxans (approved):

src/scripts/safari/Safari.ts
Farboo171: Swapped ball and bait XP

src/scripts/safari/SafariPokemonList.ts
```diff
@@ -38,7 +38,7 @@ class SafariPokemonList {
         const pokemon : SafariType[] = SeededRand.shuffleArray(App.game.party.caughtPokemon.map((p) => p.name)
             .filter((p) => !PokemonHelper.hasEvableLocations(p) && Object.keys(PokemonHelper.getPokemonLocations(p)).length)
             .map((p) => {
-                return {name: p, weight: 3};
+                return {name: p, weight: 10};
```
Aegyo, line 41: Was this change on purpose? It looks unrelated.

Aegyo (commented):
~~I have some concerns with the safariLevel computed, but I'll need to actually measure the performance before deciding if it is worth complicating it.~~ Seems ok after testing

src/scripts/safari/SafariPokemonList.ts
Farboo171: This change was intentional, I should have mentioned it initially.
Filler mons in the Friend Safari were put there to keep it from breaking on strange challenge runs, but they were showing up way more frequently than felt good. This change was made to make the intended mons (Only EVable in the Friend Safari) much more common, and the filler mons be a side note.

Aegyo (approved):

PR #4380, opened 2023-05-30
Updated the shiny chances in the FAQ

EatPant2nd:
## Description

Noticed that the shiny chance for the Safari Zone was outdated so I fixed it. Also added the Reward and BF chance since they were missing.

## Motivation and Context

Outdated information can be confusing so updating it is for the better.

## How Has This Been Tested?

I opened the game and checked the FAQ section.

## Types of changes

- FAQ update


Farboo171 (approved):
Verified the odds and tested it, looks good!

Aegyo (approved):

Issue #4378, opened 2023-05-30
[Bug]: Castform app greying entire days out for no reason


RegisCoaxans:
### Confirmations

- [X] I have searched open and closed issues for any existing report of my issue

### What happened?

Castform app greys days out as if they already passed but they did not.

Here, 30 / 05, 2am, breaks on June the first :
![image](https://github.com/pokeclicker/pokeclicker/assets/68825215/59c9f0c1-0ac1-4e8c-8e05-4d5cd508e60a)

Also, 04 / 05, The red line indicates the time of the day :
![image](https://github.com/pokeclicker/pokeclicker/assets/68825215/b6d5723c-3c34-46e6-8a73-2c1ba7bccb3d)

### What did you expect to happen?

Cells should be greyed if they are past the current time of the day.

### Steps to Reproduce

This bug is difficult to reproduce as it highly depends on timezones.
Players in France and Spain experienced it, for instance, but players in USA did not.

### What browsers are you seeing the problem on?

Firefox

### Additional Information

[Discord thread with issue report](https://discord.com/channels/450412847017754644/1112859346167140413)

PR #4382, opened 2023-05-30
Fix: Castform App showing future dates as 'passed'

Crobat4:
## Description

Fixes https://github.com/pokeclicker/pokeclicker/issues/4378
I was able to reproduce the bug only in Firefox, not even other Gecko-based browsers.

The issue was about Firefox reading the dates as `DD/MM/YYYY` when the dates were saved in the observable as `MM/DD/YYYY` (the default format), ~~so I changed the format so the month is the word instead the number.~~

~~Before: `5/30/2023, 12:00:00 AM`~~
~~Now: `Tuesday, May 30, 2023  00:00`~~

Date object is stored now.

## How Has This Been Tested?

* Check in Firefox and Edge if the bug is fixed
* Change the hour to see if something broke with the changes

## Screenshots
<details><summary>Firefox</summary>

image

</details>

<details><summary>Edge</summary>

image

</details>

## Types of changes

- Bug fix

### DIFF
--- a/src/modules/weather/WeatherApp.ts
+++ b/src/modules/weather/WeatherApp.ts
@@ -34,14 +34,13 @@ export default class WeatherApp {
      */
     public static generateRegionalForecast(region: Region, dateRange: number = WeatherApp.defaultDateRange, date: Date = new Date()) {
         const weatherForecastList = [];
-        const options: any = { dateStyle: 'full', timeStyle: 'short', hourCycle: 'h23' };
         // Creates forecasts for X hour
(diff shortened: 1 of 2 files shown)


AevitasDragonkin:
Is there a reason it's being localised first instead of just passing the date object itself, and only using toLocale when displaying the value?

Crobat4:
Honestly... I don't remember... Maybe I had some issues when I was making it.
I changed it so it stores the date object instead the string in the observable, it didn't break anything it seems so I'm leaving it like that.

Aegyo (approved):

PR #4392, opened 2023-06-04
Add challenge info to FAQ

DraKay:
All you wanted to know about the challenges and nobody told you! 

## Description
Added basic info in the faq about the challenge modes. And yes, even if it surprise you, yes, we have a faq! 

## Motivation and Context
There is a lot of doubts about some challenge modes as if you need alt forms to go to next regions or how the regional debuff works (actually this was even a mess for crew) 

## How Has This Been Tested?
E

## Types of changes

- UI improvement


Jaaslet (approved):
I cleaned up some of the text. Looks good

PR #4359, opened 2023-05-26
Safari Item List Population

Farboo171:
## Description
Populates the Safari Zone items lists with lore-friendly loot

## Motivation and Context
The new mechanic was made, and I put in content.
Weights are based around evo stones, such that after the upcoming changes tying item spawning to catching instead of encounter resolution, a player should get ~1 evo stone per safari run.

Megastones added because Pinsir and Scyther are iconic safari mons, and they needed a mechanic.

Single Exeggutor added to encourage old players to try the safari once improvements are implemented

Some items are Safari Level restricted, in the tradition of locking megastones behind mechanic progress.

## How Has This Been Tested?
Ran through Kanto Safari Zone and made sure I could get a broad distribution of items, including the new Exeggcute

Tested that SafariLevelRequirment worked with single execute.

## Types of changes

- New feature

PR #4400, opened 2023-06-10
Implemented Safari level effects

RegisCoaxans:
## Description
- We use `(level - 1) / 50` as a modifier which we then adds to each individual escape / catch modifiers. (Thanks Eugene.)
- The chances for an item to drop after catching a pokémon are now increased with Safari levels.
- +1 to eggsteps per step every 10 Safari levels.
Cumulated max bonuses look overkill but remember level 40 is hard to reach, and some pokémons in Friend Safari will need modifiers that high...

## Motivation and Context
Safari zones will be bearable with those changes.

## How Has This Been Tested?
Did some runs and tested that the numbers looked as expected.
We can only test the item changes in the Friend Safari until #4359 is merged.

## Types of changes
- New feature


Farboo171 (approved):
Code looks as discussed. Tested and it works between levels 2 and 3 for values changing as anticipated.

src/scripts/safari/Safari.ts
```diff
@@ -347,7 +347,9 @@ class Safari {
     }
 
     public static spawnItemCheck() {
-        if (Rand.boolean()) {
+        const baseChance = 0.4;
+        const levelModifier = (Safari.safariLevel() - 1) / 100;
+        if (Rand.chance(baseChance + levelModifier)) {
```
Nydaleclya, line 352: Starting with Level 62 this would reduce your chance for an item compared to the previous Level

RegisCoaxans: Maximum level is 40, though. I do not think we plan on adding more levels as that one is already long to reach.

RegisCoaxans:
Hopefully made the names less confusing.

Jaaslet (approved):

PR #4405, opened 2023-06-12
Fixed table stripes being broken in held item modal

RegisCoaxans:
## Description
Same issue as fixed in #4165. Some rows being processed but then hidden were breaking the pattern. I applied the same fix as in the linked PR.

## Motivation and Context
Looked weird.

## How Has This Been Tested?
Hid rows using checkboxes and verified the pattern was still correctly displayed.

## Screenshots (optional):
How it looked like.
image

## Types of changes
- Bug fix


Jaaslet (changes requested):
This solution adds some lag to the filtering, because adding and removing elements requires more than hiding and showing.
I am not sure if we have a better solution, and if the lag is bad enough that we don't think it's worth it.
Use Muscle Band as an example, as it shows all the pokemon, and then type in the name filter

Jaaslet:
I am going to close this PR.
I don't like the lag it introduces, and after doing some research, i know we can fix it, only using CSS. I just need to figure out which solution to use.
You can ask CSS to ignore hidden rows. The question is how we implement it. I have 4 ideas (ordered by how much i like them):
- Open a PR at the bootswatch github repository. I don't know if it will get accepted, but it would be the cleanest solution
- Have all bootstrap files locally in our repository. Then we are less dependent on an external website. Then we can also move the theme specific fixes to that file (which might make the code cleaner)
- Upgrade to bootstrap 5. As far as i can see, the colors are defined as variables there. Then it will be easy to make some CSS that overrides it for all themes
- Have an override for each theme

PR #4408, opened 2023-06-14
DT and P$ quest rebalance attempt

RegisCoaxans:
## Description
I used JK's idea in #3585 and twisted it a bit so it acts smoother, and better under specific circumstances.

At the end of Kanto, the quest rewards between 80 and 240 QP for P$, 200 and 600 for DT.
At the end of Galar, the quest rewards between 80 and 800 QP for P$, 200 and 2000 for DT.

Required amounts scale respectively with the most rewarding gym defeated and the most expansive dungeon cleared.
On new files, first dungeon and first gym in Kanto are used.
Moreover, the upper bound of the range increases with highest region.

## Motivation and Context
The Great Rebalance motivated me. :^)

## How Has This Been Tested?
Made files in each region and tested different scenarios.

## Types of changes
- New feature


Jaaslet (approved):
The values seems doable, but not too easy and not too hard. The rewards are worth something, but they are not broken at all. There is a chance the rewards might want to be buffed a bit, but I think it's a conversation for after this has been released.

PR #4410, opened 2023-06-14
Startmenu: Moved up events

Jaaslet:
## Description

The Events button is moved up, as it now does something, when you own the event calendar. I moved it right above all the items, that does not open something that affects the game.

## Motivation and Context

Makes the start menu easier to use

## How Has This Been Tested?

I opened the game and saw the achievement button was gone, and the event button was moved up

## Screenshots (optional):

image

## Types of changes

- UI improvement


DraKay:
Special files have less than 110 pokes, example pant

CypherX:
The achievements module is locked until the player has captured 110 unique pokemon, so removing it from the start menu will prevent players doing certain types of challenge runs (monotypes mainly) from accessing achievements completely.

Perhaps we can hide the start menu item after the module is unlocked, or lower the amount of unique pokemon required (or remove that requirement altogether).

Qwertypop04:
Perhaps instead of having it be unlocked by number of Pokémon, have it be unlocked by progress? (Like gem case, for example)

Jaaslet:
Oh, then it makes sense that it's there. Yeah, unlock it earlier/in some other way, might make sense. I will remove it from this PR, as it will require some discussion

RegisCoaxans (approved):

RedSparr0w (approved):
Sounds good, thanks for the PR! 👍

PR #4412, opened 2023-06-14
Made SpecialEvents.ts fully non-static

Jaaslet:
## Description

SpecialEvents.ts has some static stuff. This is now all non-static

## Motivation and Context

It's a refactor. The properties had no reason to be static. It might also make the class easier to move to modules, some day

## How Has This Been Tested?

I started an event using the event calendar, to see stuff still works

## Types of changes

- Refactor


RedSparr0w (approved):

PR #4414, opened 2023-06-15
Uncaught Shadow Pokémon

RegisCoaxans:
## Description
New map filter for uncaught shadow Pokémon. This way, players will know where they miss a Shadow Pokémon.
I picked the default colour from shiny XD001 sprite, a bluish dark red.

## Motivation and Context
It helps keep track of shadow Pokémons, since we are introducing shadow-related achievements, it makes sense,

## How Has This Been Tested?
Changed the colour in settings to be sure it displays correctly, caught every shadow from a dungeon and made sure it was correctly updating.

## Types of changes
- New feature
- UI improvement

PR #4418, opened 2023-06-17
Map Filter improvement

RegisCoaxans:
## Description
Following [this comment from #4414](https://github.com/pokeclicker/pokeclicker/pull/4414#issuecomment-1594236408), I added a way to ignore map filters until they are needed or should be released. Hence, I applied it on both Resistant and Shadow.
- Settings can now handle requirements.
- Shadow is locked behind the quest line for Orre and Resistant has the same requirement as the Pokérus key item.
- For instance, the map does not process missing resistant anymore until Pokérus is unlocked.
- The setting modal does not display pokérus or shadow related map filters as long as they are irrelevant to the player.
- The map legend is now dynamic, and only includes relevant filters at any time.

I also refactored the area status `unlockedUnfinished` into `incomplete` as its corresponding setting states it that way, for consistency and easy matching.

## Motivation and Context
Orre is likely not coming out in the next update so we need to hide Shadow map filter.

Also the map legend and setting modal start to be overwhelming in Kanto.
Hiding later features helps the player focus on what they should already get access to.

## How Has This Been Tested?
I unlocked Shadow or Pokérus with a save file so I could see map filters and setting modal newly taking those features into account.

## Screenshots (optional):
Orre unlocked, pokérus locked :
image

The map shows Kanto r1 is complete even though no Pokémon is resistant there (I changed completed to an obvious green) :
image

## Notes
However, I think Missing Resistant should now have a visible default colour in order to make players notice it before stumbling upon it whenever they open the setting modal.
Also, since we can catch those shadow mons, should we add them in the calculation for missing resistant, now ? And uncaught pokémons ?

## Types of changes
- New feature
- UI improvement


Jaaslet:
> However, I think Missing Resistant should now have a visible default colour in order to make players notice it before stumbling upon it whenever they open the setting modal.

I can't remember why we did not add a default color... Maybe because most players won't care about it until super late game?

> Also, since we can catch those shadow mons, should we add them in the calculation for missing resistant, now ? And uncaught pokémons ?

We could do that. I have not done it personally, because i think it's a lot of weird code, for something that is not relevant very often. But if you want to do it, go for it!

RegisCoaxans:
> I can't remember why we did not add a default color... Maybe because most players won't care about it until super late game?

There probably was no default colour because otherwise the missing resistant would shows up on routes or dungeons the players already completed so far (every achievements, every shiny pokémons) despite pokérus not being unlocked.
> We could do that. I have not done it personally, because i think it's a lot of weird code, for something that is not relevant very 
often. But if you want to do it, go for it!

Actually, it just needs two array.concats, not big of a deal. I may do that tomorrow then.

Thank you for the quick feedback.

src/scripts/worldmap/MapHelper.ts
```diff
@@ -16,6 +16,30 @@ enum areaStatus {
...
 class MapHelper {
 
+    public static generateMapPopoverContent(): string {
+        const rows = [];
+        GameHelper.enumStrings(areaStatus).map(status => Settings.getSetting(`--${status}`))
+            .filter(setting => setting.isUnlocked()).forEach(setting => {
+                rows.push(
+                    `<div class='col-10 legend-item'>
+                        <span>${setting.displayName}:</span>
+                    </div>
```
Jaaslet, line 26: I don't really like HTML code in the typescript files. This generated code can easily be done with knockout in the html file

src/modules/settings/CssVariableSetting.ts
```diff
@@ -1,14 +1,16 @@
 import Setting from './Setting';
+import Requirement from '../requirements/Requirement';
 
 export default class CssVariableSetting extends Setting<string> {
-    constructor(name: string, displayName: string, options = [], defaultValue = '') {
+    constructor(name: string, displayName: string, options = [], defaultValue = '', requirements: Requirement[] = []) {
```
Jaaslet, line 5: Requirements should just be a requirement instead of a list of requirements.
I know we have some lists of requirements out there, but that is a mix of them being added before MultiRequirement, and people (like me) not knowing what the standard should be

Jaaslet (changes requested):
I have not tested it yet, but I have some comments for the code

Jaaslet (approved):

PR #4419, opened 2023-06-18
Friend Safari pokemon list generation optimizations

CypherX:
## Description
Improves two areas of the friend/kalos safari pokemon list generation:

1. It's quite slow when the player has a large party (2.2s w/ 1304 pokemon)

Current: The entire caught pokemon list is filtered and shuffled, then the first five pokemon are selected.

Optimized: The caught pokemon list will now be shuffled first then the first five pokemon that meet the friend safari requirements will be selected.

2. The list only needs to be generated once per day. Due to the computed nature it was being unnecessarily recalced when changing regions, when the weather changes, when a new pokemon is caught, and possibly at other times.

Safari pokemon lists are no longer computed and `generateKalosSafariList()` will be called when the day changes to refresh the list only when necessary.

## Motivation and Context
Noticeable lag caused by the old generation method.

## How Has This Been Tested?
Verified safari lists generate properly and that the kalos list refreshes when the day changes.

## Types of changes
- Optimization

### DIFF
--- a/src/scripts/Game.ts
+++ b/src/scripts/Game.ts
@@ -445,6 +445,8 @@ class Game {
                 BattleCafeController.spinsLeft(BattleCafeController.spinsPerDay());
                 // Generate the weather forecast
                 WeatherApp.initialize();
+                // Refresh Friend Safari Pokemon List
+                SafariPokemonList.generateKalosSafariList();
 
                 DayOfWeekRequirement.date(now.getDay());
             }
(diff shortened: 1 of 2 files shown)


Jaaslet (approved):
It's hard for me to test, as I am personally not experiencing the lag. But it looks good in theory, and works, so I will merge it, so it can be tested by more people experiencing the lag.

Only weird thing with it, is if you catch a new pokemon, the list will not update, until either midnight or you restart the game. This can maybe be used tactical to reset the pool in a more controlled way. And I at least think that is fine

osherwott:
The lag is gone 👍

PR #4421, opened 2023-06-18
Typo in real evolutions challenge mode text

### LINKED ISSUE #4417: [SUGGESTION]: Fix typo in the real evolution challenge mode description
### Current Behavior

The text that appears when hovering over the real evolutions challenge in the challenge mode menu says  "Your Pokémon go away, when they evolve"

Example 2

### Improved Behavior

Remove the comma in the text and make it say: "Your Pokémon go away when they evolve"

### Reason for change

This is grammatically incorrect.

### Detailed Description

Example

Here is an image I took using devtools to alter the text and make it correct. This is what it would look like if the change was made.

Basic-Person:
## Description

Removing a small grammatical typo.

## Motivation and Context

Closes #4417 
Solves a grammatical error.

## How Has This Been Tested?

I didn't test it, but I doubt removing a single character from a string will break anything (just to be clear I know very little about coding). Also this seems like the correct place where the text listed for the description occurs.

## Screenshots (optional):

Example
This is taken from using inspect element so this is not from actual testing.

## Types of changes

- UI improvement

### DIFF
--- a/src/modules/challenges/Challenges.ts
+++ b/src/modules/challenges/Challenges.ts
@@ -16,7 +16,7 @@ export default class Challenges implements Saveable {
         disableGems: new Challenge('No Gem', 'Disables the usage of Gems to increase damage multipliers'),
         disableVitamins: new Challenge('No Vitamins', 'Disables the usage of Vitamins'),
         slowEVs: new Challenge('Slow EVs', 'Gain EVs 10x slower'),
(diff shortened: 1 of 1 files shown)


osherwott:
Nice. Also could you change your title to reflect the change?

Jaaslet (approved):

PR #4423, opened 2023-06-18
Added method "isInLiveVersion" for wiki and bot

Jaaslet:
## Description

Added the helper and method `ExternalHelper.isInLiveVersion()` which takes stuff like towns and questlines as parameter and says if it's released or not.
The idea is to use this method in the wiki or the bot to automatically tell if something is available in the game or not. I don't know if we want to hide unobtainable stuff, add it to it's own page or just mark it. But with this method, all the options are there!
I have made some caching to help performance. And because the wiki never reloads the page, the caching should stay while people browses around on the wiki.
I have also added `ExternalHelper.pokemonIsInLiveVersion()`. It's not done, but I am keeping this PR as a non-draft, as I don't know if i personally will finish that method, and I think we want to start using what we already have

## Motivation and Context

Making the wiki and bot less confusing for people, without needing to do all the stuff manually.
I have added the method here, instead of on the wiki, because it would be awful to code, without typescripts typechecking

## How Has This Been Tested?

I have tested some examples using the console, but I think we need to release it and implement it on the wiki, before we really can test it.

## Types of changes

- Wiki


RedSparr0w (approved):
Looks good, thanks for the changes! 👍

PR #4435, opened 2023-06-22
Quest items will now be sold out after first buy

Jaaslet:
## Description

When you buy a quest item, it will now show as sold out. Shops had not support for sold out, but trades did, so added it for trades too.
Changed keyitems to using sold out too, just in case we want some text difference, some day

## Motivation and Context

Right now, you can buy more than one of each memory in the Silvally questline, which is weird. And as far as i know, all planned uses for questitems only requires one of each. That might change, but if that happens, we will write some more intelligent code

## How Has This Been Tested?

Tested that quest items works as i wanted, and checked that keyitems are still sold out

## Types of changes

- Bug fix
- New feature
- UI improvement


osherwott (commented):
It works! 👍
But it also works if the only change is adding this in QuestItem.ts 🤔 
```
     isAvailable(): boolean {
         return player.itemList[this.name]() === 0;
     }
```
I'm not savvy enough to say if you should simplify it or not, so my only feedback is that it works as intended.

Jaaslet:
> It works! 👍 But it also works if the only change is adding this in QuestItem.ts 🤔
> 
> ```
>      isAvailable(): boolean {
>          return player.itemList[this.name]() === 0;
>      }
> ```
> 
> I'm not savvy enough to say if you should simplify it or not, so my only feedback is that it works as intended.

"Sold out" was in a weird state of being used by trades and mega stones, but not shops and key items. I cleaned that part up in this PR, so make the code less confusing

PR #4444, opened 2023-06-25
Multiquests now gives rewards for each step

Jaaslet:
## Description

In the Silvally questline, there is a quest where you need to talk to 4 different people. They give you a memory each.
Multiquest's subquest's rewards did not work, so they where all given at the end. This PR fixes that problem, so you get the item, each time you talk to a person.

I discovered a problem with TalkToNPC quests not being saved, when they are a part of MultiQuests and you reload the game. I will try to fix that in an other PR. But i am not sure if we should wait for that fix, before releasing the questline, as talking to 4 NPCs is a quick quest

## Motivation and Context

It feels more natural for the questline to work like this

## How Has This Been Tested?

I went through that part of the quest and saw that i got the items

## Types of changes

- Bug fix


RedSparr0w (approved):

Issue #4464, opened 2023-07-01
[Bug]: Typing DELETE and then clicking X in the "Delete save" window deletes the file anyway


EatPant2nd:
### Confirmations

- [X] I have searched open and closed issues for any existing report of my issue

### What happened?

I opened the "Delete save" option in the save menu to delete a save file. I typed DELETE but then decided to not delete my file. I tried clicking outside but that doesn't close the window. So, I clicked the X thinking it would go back to the previous window. My save got deleted anyway.

### What did you expect to happen?

My save to not get deleted and to return to the previous window.

### Steps to Reproduce

1. Go to Menu -> Save.
2. Press "Delete Save".
3. Type "DELETE".
4. Click the "X".
5. Watch your file disappear anyway.

### What operating system are you using?

Windows 11

### What browsers are you seeing the problem on?

Desktop Client

PR #4465, opened 2023-07-01
Fix: Delete file modal deleting the file, regardless if pressing submit or close button

Crobat4:
## Description

When deleting a file, typing `DELETE` in the delete prompt and closing the modal will delete the file, regardless if the player clicked Submit, pressed the Enter key or closed the modal by pressing `X`. This PR fixes it so it will delete the file only when pressing enter or clicking Submit.

## Motivation and Context

Closes https://github.com/pokeclicker/pokeclicker/issues/4464.

## How Has This Been Tested?

* Right click on a file and select Delete
* Typing anything but `DELETE`: Nothing happens
* Typing `DELETE`:
  * Clicking Submit: Deletes the file
  * Pressing Enter: Deletes the file
  * Clicking the `X` button: Nothing happens

## Types of changes

- Bug fix


CypherX:
The modal remains in the DOM when using the `x` button to cancel & close whereas before it was always removed when closed. Could you be sure it gets removed even when using the close button? Thanks!

It might be more simple to just add a click event to the close button that then sets the input to a blank value.

Crobat4:
You are right, it's more simple this way.
I changed it, it should still work fine and the modal is removed from the DOM.
I also added code so the Esc key doesn't delete the file neither.

RedSparr0w (approved):

PR #4468, opened 2023-07-02
Safari Zone Tweaks

Farboo171:
## Description
Adds a mild catch boost based on Safari Level, capping out just under 14% at level 40
Adds a Catch Achievement in the Safari Zone
Adds XP gains from picking up Safari Items

## Motivation and Context
Improves Safari Experience for all involved, asked to add these by Dev.

## How Has This Been Tested?
Checked that Achievements show up and increment, and that the boost is present for catch rate

## Types of changes

- New feature

### DIFF
--- a/src/modules/requirements/SafariCatchRequirement.ts
+++ b/src/modules/requirements/SafariCatchRequirement.ts
@@ -1,0 +1,16 @@
+import * as GameConstants from '../GameConstants';
+import AchievementRequirement from './AchievementRequirement';
+
+export default class SafariCatchRequirement extends AchievementRequirement {
+    constructor(value: number, option: GameConstants.AchievementOption = GameConstants.AchievementOption.more) {
+        super(value, option, GameConstants.AchievementType.Safari);
+    }
+
+    public getProgress() {
+        return Math.min(App.game.statistics.safariPokemonCaptured(), this.requiredValue);
+    }
+
+    public hint(): string {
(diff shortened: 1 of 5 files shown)


RegisCoaxans:
About the catch rate bonus, is such a value not overkill ? How did you get that ?
On the Pokémon with the lowest catch rate (Giratina, 0.5%), you throw Nanab + rock then it can break throught 100% at level 28 with MB; it reaches 90% at level 40 with no MB. In comparison, it can only reach 66% at level 40 on live version with MB.
Could it be around +8% instead (so modifier × 10) ?

Farboo171:
> About the catch rate bonus, is such a value not overkill ? How did you get that ? On the Pokémon with the lowest catch rate (Giratina, 0.5%), you throw Nanab + rock then it can break throught 100% at level 28 with MB; it reaches 90% at level 40 with no MB. In comparison, it can only reach 66% at level 40 on live version with MB. Could it be around +8% instead (so modifier × 10) ?

In my testing, berry and rock effects do not stack but instead overwrite. This prevents the values from getting too crazy. Is this generally true, or was my testing an outlier?

RegisCoaxans:
Berry effects and rocks do stack. You first throw a Nanab then the rock, it overwrites the bait effect but keeps the berry effect. the catch rate skyrockets : we cumulatively get ×(1.5 + modifier) from the berry then ×(2 + modifier) from the rock.

Jaaslet (approved):

PR #4469, opened 2023-07-02
Safari items no longer shows up after game over

Jaaslet:
## Description

If you caught a safari pokemon with the last ball, there was a chance that an item showed up on the screen, for a couple of seconds. This PR prevents that

## Motivation and Context

It feels bad to see an item, and then be told it's game over

## How Has This Been Tested?

I did a couple of runs with spawnchance set to 100% and saw it did not show up. It might have shown up beneath the battle window each time. But it's a small fix, so it's hard to mess up.
Also tested that items still spawn normally

## Types of changes

- Bug fix


CypherX (approved):

RedSparr0w (approved):
Nice change, thanks for the PR! 👍

Issue #3556, opened 2022-12-11
[SUGGESTION]: Implement Z moves


Farboo171:
### Current Behavior

Z moves are not a thing in Pokeclicker, though Z-Crystals are used as Alola badges to some extent

### Improved Behavior

Z-moves should be unlocked in Alola as part of gameplay, with the 12 Z crystals currently in the game unlocked as they currently are, and the remaining type-related Z crystals unlocked somehow, likely as a quest reward.

Z crystals would be accessible in a modal somewhere, likely the main screen like flutes and battle items. When a Z-crystal is activated, it would have the following effects:

1. Damage of that type is boosted by some percentage (50% is what I propose) for 60 seconds
2. After 60 seconds, all Pokemon of that type are reduced to level 1
3. All Z crystals are under a cooldown for some duration (5 minutes?)
4. Only one Z crystal can be active at a time

### Reason for change

Z moves are a big unique mechanic unlocked in the Gen 7 games, and it would be neat to see them in Pokeclicker!

### Detailed Description

60 seconds for the duration of the Z move effect is long enough to let it simulate lasting the whole battle.
The cooldown of 5 minutes prevents people from just always having a Z move active
Reducing every Pokemon of the move type to level 1 makes this mechanic one that should be used tactically

For challenges, it would be possible to implement Short Z moves (30 or even 10 seconds)

MahePoulet:
maybe z moves could be related to click attack ? like you unleash a very powerful click that does X times the regular click damage output with a typing, and then it goes into a cooldown

jk13pclick:
Tentatively approved. Needs more meat to the feature, but approved for planning at least. I have concerns with the UI, but if there is a small clickable icon somewhere that opens the modal (display transparent/active Crystal?) that could work.

Veta333:
Maybe it can be done an special questline about the specific pokemon z-cristals like the Eevium Z. In this one you´ll get to obtain those, maiking the player travel across alola  getting gems  of the tipe of the pokemon of the cristal, capturing or hatching the pokemon and defeating a strong trainer that uses that tipe before each z cristal.The z-cristal would be use as an held item or maybe an instan buff to that pokemon total power.

Farboo171:
Closing, as this is well underway

PR #4472, opened 2023-07-02
Implement Z-Moves

RegisCoaxans:
## Description
Added Z-Moves and Z-Crystals into the game the way they have been suggested in [#3556](https://github.com/pokeclicker/pokeclicker/issues/3556).
They are battle items the player can collect once and for all. Activating a crystal makes the Pokémon of the same type 50% stronger for a minute, then 25% weaker for four. Only one crystal can be active at a time.
I have not implemented any way to get the said items, since it is highly incomplete for now.
The UI is hidden until the player get one crystal so no need for dev requirement.

## Motivation and Context
I have been peer-pressured into coding it !

## How Has This Been Tested?
Gave myself the items, activated them, checked the values were correct, verified the display was fine.

## Screenshots (optional):
image image

## Types of changes
- New feature

## Additional notes
Some more things need to be done after this PR is merged :
- Sprites for the Z-Crystals which are not already Alola's badges, I put filler for now.
- Obtaining methods for those said Z-Crystals.
- Once this is done, changes on reward functions to give the crystals-badges as item as well, and making the update function retroactively give the items for already obtained crystals-badges.


src/scripts/party/Party.ts
```diff
@@ -172,7 +172,7 @@ class Party implements Feature {
         return Math.round(attack * bonus);
     }
 
-    public calculateOnePokemonAttack(pokemon: PartyPokemon, type1: PokemonType = PokemonType.None, type2: PokemonType = PokemonType.None, region: GameConstants.Region = player.region, ignoreRegionMultiplier = false, includeBreeding = false, useBaseAttack = false, overrideWeather: WeatherType, ignoreLevel = false, includeFlute = true): number {
+    public calculateOnePokemonAttack(pokemon: PartyPokemon, type1: PokemonType = PokemonType.None, type2: PokemonType = PokemonType.None, region: GameConstants.Region = player.region, ignoreRegionMultiplier = false, includeBreeding = false, useBaseAttack = false, overrideWeather: WeatherType, ignoreLevel = false, includeFluteZMoves = true): number {
```
Jaaslet, line 175: Could we call `includeFluteZMoves` something like `includeTempBonuses` or something, so we don't end up with a long name, if we add more? And it's also more clear what it should be used for.
(at least that is what i think it does. If i am wrong, ignore me)

RegisCoaxans: You are absolutely right. I changed it. I also changed it few lines above as I should have done before...

Jaaslet (approved):

PR #4484, opened 2023-07-04
Safari Animations Speed Up with Level

EatPant2nd:
## Description

With the introduction of Safari levels making the Safari experience less agonizing, animations now will also speed up. Safari now has 4 tiers at levels 10, 20, 30, and 40. Each of these tiers will speed up all interactions by 90%, 75%, 57%, 45% of the original speed, respectively. So, for example, level 30 uses 57% of the original speed value, making interactions take almost half the time than animations at level 1 (default). Some values linked to the ball cannot be tampered with as they completely break the animations so, sadly, it couldn't be properly sped up. (If anyone can speed them up, be my guest).

As a bonus, once the player hits level 20, the walking animation is also faster from 250 to 175. I didn't want to make the value too low since I'm aware that Safari Zone gives egg steps. This speed seemed ok to me.

## Motivation and Context

Safari is slow. People don't like it. I made it a bit faster.

## How Has This Been Tested?

~~With my eyes.~~ Went to Safari Zone at levels 1, 16 (so tier 10), 20, and 40 to make sure the speed up wasn't breaking anything.

## Types of changes

- New feature


Jaaslet (changes requested):
The increased speed for walking animation does so the player no longer stays in the grid (which looks weird). The rest looks good and balanced

EatPant2nd:
Ok, removed the increased walking animation for now while I check if it's even possible to make it look nice. Thanks for the feedback.

Jaaslet (approved):

PR #4493, opened 2023-07-06
Friend Safari fixed rotation

CypherX:
## Description
Per discussion on discord this changes the friend safari rotation from random based on the players caught pokemon to a daily fixed rotation (seeded by trainer id & shuffled) of all available non-EVable pokemon. Currently there are 126 (127 on dev) pokemon that can appear in the friend safari so 26 days for everything to appear.

Added the ability for pokemon to be "locked" in the safari until caught elsewhere. As the friend safari is no longer based on caught pokemon, it's possible for pokemon to be selected that have not been caught by the player yet. These pokemon will be in the pool but cannot be encountered.

Updated the Safari Ranger NPC to handle the previously mentioned "locked" pokemon - they will appear as a silhouette and also added dialog to indicate they must first be caught elsewhere. The friend safari NPC will also now display the pokerus status icon below each pokemon.

Removed the `PokemonHelper.hasEvableLocations()` function in favor of one called `PokemonHelper.isObtainableAndNotEvable()` to speed up the `pokemonList` filter and so we aren't calling `getPokemonLocations` twice for every pokemon. Will look into speeding it up even more in a different PR, it takes about 1.1s to run on my machine.

## Motivation and Context
Removes the completely RNG element that could result in waiting months, or longer, for some pokemon to appear in the pool.

## How Has This Been Tested?
Went to friend safari, removed one of the daily pokemon from my party and verified it no longer appeared as an encounter and was shown as a silhouette in the NPC list.

Time traveled to verify the pool changed at midnight and it correctly rotates through the pool.

## Screenshots (optional):
image

## Types of changes
- Feature change

### DIFF
--- a/src/modules/GameConstants.ts
+++ b/src/modules/GameConstants.ts
@@ -275,6 +275,8 @@ export const LEGAL_WALK_BLOCKS = [
 
 export const SAFARI_OUT_OF_BALLS = 'Game Over!<br>You have run out of safari balls to use.';
 
+export const FRIEND_SAFARI_POKEMON = 5;
+
 // Quests
 
 // Numbers calculated by Dimava assumes ability to 1 shot on high routes and some use oak items,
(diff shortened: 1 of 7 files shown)


Jaaslet (changes requested):
We need an observable somewhere. The ranger NPC does not update when you catch the pokemon.
But other than that, it seems like it works

CypherX:
It doesn't update live while the dialog modal is open but does update upon re-opening. In normal use that modal should never be open when a pokemon is caught.

Jaaslet:
You are right

Jaaslet (approved):

PR #4495, opened 2023-07-06
Rare Candy

### LINKED ISSUE #1407: [SUGGESTION] Rare candy as breeding skip
**Current Behavior**
rare candy has no use

**Improved Behavior**
The rare candy could be used to give an pokemon +25% attackbonus from base attack (1x breeding)

**Reason for change**
Gives rare candy a use without being overpowered. Just saves some time and buffs Legendary by doing so.

**Detailed Description**
The usage is very simple. It skips 1x breeding and gives you the bonus instantly.
It's pretty wothless if you only have one and if your Pokemon didn't have any proteins so far.
But if you use them wisely they could give you an huge instant boost in terms of saving time.

The rare Candy could go with the price of masterballs so you can't just buy a bunch and power up any pokemon.
You can buy it but it should not be worth to do that in general (except you have too much money).

Instead reward it for milestones. Like finishing a region, getting the highest stage of an Achievement, High Stagelevel reward for battle frontier and/or other stuff like that. 

The best usage would be on high base attack but low breeding effency pokemon - also known as most of the legendarys. So it would buff these Pokemon too.

RegisCoaxans:
## Description
Rare Candy reintroduced.
As we talked about it, it applies to a pokémon and gives as much attack as a breeding cycle does (without vitamins).
Not purchasable. Locations to find it need TBD precisely, I can add them in a future PR.
For now No Vitamins Challenge prevents players from using Rare Candies : maybe that should change.

## Motivation and Context
Closes #1407, finally.

## How Has This Been Tested?
Gave some candies to my pals and checked their attack.
Opened all vitamin-related menus to see if something broke.

## Types of changes
- New feature


Jaaslet:
You can add them in the safari zone. It was the plan from the start

RegisCoaxans:
That works.

Jaaslet (changes requested):
Looking at the code, i don't think they should be vitamin type. They might be in the game, but the code starts getting messy, when almost all places needs exceptions for the rare candy.

If we want to keep it as a vitamin, we could add a subclass for the 3 other vitamins, that implements the vitamin class, and then referrer to that class, instead of having the exceptions

Jaaslet:
Oh, an other thing! Rare candy is in the official drop pool of one of the dream orbs. You can find it on Bulbapedia. If you don't mind, can you add it there? Else I can do it after this is merged

RegisCoaxans:
I will put it there too.
I am looking for a solution to Rare Candies. Is it fine if I make a new item category for "consumables" instead ? I do not know what else would fit in there (blocks ?) but Rare Candy is the perfect first candidate to that. If it can, it should not be a vitamin anymore.

Jaaslet:
Yeah, that sounds fine

src/scripts/party/PartyPokemon.ts
```diff
@@ -283,6 +283,34 @@ class PartyPokemon implements Saveable {
...
+        if (!player.itemList[itemName]()) {
+            return Notifier.notify({
+                message : `You do not have any more ${ItemList[itemName].displayName}`,
+                type : NotificationConstants.NotificationOption.danger,
+            });
+        }
+        switch (type) {
+            case GameConstants.ConsumableType.RareCandy : amount = Math.min(amount, player.itemList[itemName]());
+                const curAttack = this.calculateAttack(true);
+                GameHelper.incrementObservable(this._attackBonusPercent, 25 * amount);
```
Jaaslet, line 297: Can we move this code to the rare candy code? Make an overrideable method in consumables called `use` or something.
Will make it easier to see what rare candy does

src/modules/items/ItemList.ts
```diff
@@ -201,3 +200,6 @@ ItemList.Carbos   = new Vitamin(VitaminType.Carbos, 1e5, Currency.money, {
     saveName: `${VitaminType[VitaminType.Carbos]}|${Currency[Currency.money]}`,
     visible: new MaxRegionRequirement(Region.unova),
 }, undefined, 'Reduces steps required when hatching');
+
+// Consumables
+ItemList.RareCandy = new Consumable(ConsumableType.RareCandy, Infinity, undefined, undefined, 'Rare Candy', 'A rare-to-find Candy<br/>Keep away from children!<br/>Increases the attack of a Pokémon as if bred');
```
Jaaslet: `Rare_Candy` instead of `RareCandy` would follow the style of the rest of the item names

src/scripts/safari/SafariItem.ts
```diff
@@ -12,34 +12,36 @@ type SafariItemWeighed = {
...
+            {item: {id: 'Sitrus', type: ItemType.berry}, weight: 1.95},
+            {item: {id: 'Oran', type: ItemType.berry}, weight: 1.95},
             {item: {id: 'Leaf Stone', type: ItemType.underground}, weight: 0.25},
             {item: {id: 'Fire Stone', type: ItemType.underground}, weight: 0.25},
             {item: {id: 'Thunder Stone', type: ItemType.underground}, weight: 0.25},
             {item: {id: 'Water Stone', type: ItemType.underground}, weight: 0.25},
             {item: {id: 'MediumRestore', type: ItemType.item}, weight: 2.25},
             {item: {id: 'LargeRestore', type: ItemType.item}, weight: 1.25},
             {item: {id: 'Protein', type: ItemType.item}, weight: 0.25},
+            {item: {id: 'RareCandy', type : ItemType.item}, weight : 0.25},
```
Jaaslet, line 27: This weight might be a bit too low, as it's not something gamebreaking, and we kinda made it to make the safari loot a bit more fun. The same for dream orbs.
Talk with Eugene about the weights. If you don't feel like doing it, i can just do it after it's merged

src/components/consumableModal.html
```diff
@@ -0,0 +1,81 @@
...
+                            data-bind="checked: Settings.getSetting('consumableHideShinyPokemon').observableValue()"
+                            onchange="Settings.setSettingByName('consumableHideShinyPokemon', this.checked)" />
+                        <label for="consumable-shiny-filter" class="form-check-label" data-bind="text : Settings.getSetting('consumableHideShinyPokemon').displayName"></label>
+                    </div>
+                  </div>
+                </div>
+                <table class="table table-striped table-hover table-bordered table-sm m-0">
+                  <thead>
+                      <tr class="text-light">
+                          <th class="text-left bg-dark align-middle">Pokémon:</th>
```
Jaaslet, line 68: Can we add `attack` as a column? Just so players can see it makes a difference

src/scripts/party/PartyPokemon.ts
```diff
@@ -283,6 +283,34 @@ class PartyPokemon implements Saveable {
...
+                Notifier.notify({
+                    message : `Your Pokémon gained ${this.calculateAttack(true) - curAttack} attack points`,
```
Jaaslet, line 299: Can we replace "Your Pokémon" with the name of the pokemon?

src/scripts/party/PartyPokemon.ts
Jaaslet: I am not totally sure if this is a good idea or not... It sounds good in theory, but it might look messy, if we move it...

Jaaslet (changes requested):
It works!
I just have some small comments

src/scripts/party/PartyPokemon.ts
RegisCoaxans: You know what ? I did it that way first lol. But I did not like it and thought you would not neither. I passed the function through the constructor. Although it needed to import much more things which had nothing to do with models. I thought it would be messy for no reason and ended up putting that `useConsumable` in whom we can stack the processes in a proper way.

src/scripts/safari/SafariItem.ts
Farboo171: ! think making Rare Candy weight 1 in both safaris and taking the weights from the berries works well. As Jaas says, they're not very powerful

RegisCoaxans:
All issues should be addressed now.

Jaaslet (approved):

PR #4505, opened 2023-07-08
Friend safari rotation fixes

CypherX:
## Description
Realized that the new friend safari fixed rotation didn't "wrap" around on the last day, meaning that when there are 126 non-EVable pokemon the last day in the rotation would only have two in the list rather than the intended five.

This is fixed by duplicating the list of possible pokemon by repeating it 5 times (or whatever the amount of daily pokemon is in the event it gets changed). This ensures the list is evenly divisible by 5, every day has a full pool, and the list wraps correctly when the end is reached.

(thanks to pumpkinking0192 in discord for the idea)

## Motivation and Context
Fixes an unintended issue.

## How Has This Been Tested?
Changed my PC time to verify the pool changes daily and properly wraps when reaching the end of the list.

## Types of changes
- Bug fix


RedSparr0w (approved):

PR #4509, opened 2023-07-10
Improvements to safari item & pokemon spawning

CypherX:
## Description
When spawning an item or pokemon on the map in the safari zone a random position is generated. If this location is blocked by terrain, the player, or something already spawned on that position the player loses out on that item/pokemon spawn. This can make the item spawn chance appear lower than it is.

I've changed it to now make 20 attempts to generate a valid position before giving up.

It was also noticed that when selecting a position to spawn the item/pokemon the current size of the window was being used to generate a position rather than the safari grid size. This is fine unless the window size is changed after generation and could result in unintended placement or invalid positions being selected more often.

That has been changed to now use the grid size rather than window size to generate spawn positions.

## Motivation and Context
Fixing things that appear unintended but were never noticed before. Item spawn chance is already RNG, so when an item *should* spawn but doesn't it sucks.

## How Has This Been Tested?
Spawning items and pokemon manually and during normal gameplay.
Generating safari and changing window size then validating item/pokemon spawn positions.

## Types of changes
- Safari improvement
- Bug fix


src/scripts/safari/Safari.ts
```diff
@@ -369,34 +370,45 @@ class Safari {
...
+            result = this.canPlaceAtPosition(x, y);
+        }
+
+        return result ? {x: x, y: y} : null;
+    }
+
+    private static canPlaceAtPosition(x: number, y: number) {
+        return !(!this.canMove(x, y) || (x == this.playerXY.x && y == this.playerXY.y) ||
+            this.pokemonGrid().find(p => p.x === x && p.y === y) ||
+            this.itemGrid().find(i => i.x === x && i.y === y));
```
RedSparr0w, line 411: I feel like this logic makes it more confusing to read, could we reverse the logic here?
canmove && !player && !pokemon && !item

RedSparr0w (changes requested):

Just one change request plz

src/scripts/safari/Safari.ts
CypherX: Sure can. Updated!

RedSparr0w (approved):
Thank you, looks good 👍

PR #4514, opened 2023-07-12
Fix: Broken Safari fence tiles

Crobat4:
## Description

In the Safari, when fence tiles are generated, some of them are removed to create an entrance, if the removed tiles are either from the left or the right, the fence tiles at the corners look broken:

image

Fixed tiles:

image

## Motivation and Context

Chopped fence looked weird

## How Has This Been Tested?

* Changed the code to generate only fences
* Create several Safari runs to check if the tiles look correctly

## Types of changes

- Graphic fix

### DIFF
--- a/src/scripts/safari/SafariBody.ts
+++ b/src/scripts/safari/SafariBody.ts
@@ -231,15 +231,43 @@ class FenceBody extends SandBody {
     }
 
     private openFence() {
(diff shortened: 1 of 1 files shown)


RedSparr0w (approved):
The fix I didn't know I needed, looks great, thanks for the PR! 👍

PR #4537, opened 2023-07-16
Gem master and shard trader now supports map color

### LINKED ISSUE #4529: [Bug]: missing inidcators at parfum Palace
### Confirmations

### What happened?

the furfrou shard trader at parfum palace in kalos does not display caught and pokerus indicators while the gem trader does. 
additionally Parfum Palace doesnt highlight on the map for non caught Pokemon like other traders. ex furfrou friseur on the same map 
gem
shard
map

i expected both traders to show caught and pokerus indicators similar to other trader locations, and for the location on map to highlight based on map settings.

### Steps to Reproduce

view traders in parfum palace in kalos

_No response_

Desktop Client

### Relevant log output

_No response_

### Additional Information

im assuming that the indicator on the gem trader was an unintended consequence of #4486

Jaaslet:
## Description

Just copied code from berry master, so the other trades supports map colors

## Motivation and Context

closes #4529
Makes the Silvally questline a bit better

## How Has This Been Tested?

Opened the game and saw that the colors was correct in my game

## Types of changes

- Bug fix
- New feature
- UI improvement


RegisCoaxans:
There is a (probable) bug with Shop map colour, by the way. As shops were meant to almost always stay visible, their unlocked status is ignored. This means locked pokémon shops in unlocked location will show up with uncaughPokémon status instead of whatever  from the other contents.
You can try it in Sevi4567 : Officer Jenny trade shop, locked behind a quest but still counting for the map colour calculation.

Farboo171 (approved):
Works on all the gem and shard shops, including the pinkan reserve not having an overlay when the quest is not completed. Looks good, thanks for the PR!

PR #4538, opened 2023-07-16
Gigantamax

Jaaslet:
## Description

This PR implements the base code for how to obtain gigantamax forms. I will require a lot of stuff to be filled in, some NPCs and some balancing. Either Qwerty or Eugene will do that, in an other PR. Everything is hidden behind a development requirement, so this can safely be merged.

The solution is very close to Qwerty's idea. Some stuff have been changed, so it could be implemented using existing features, with small tweaks. I expect some of the details to change, after someone else takes over, so i have not spend too much time on stuff, like images and text. The solution is:

- After galar champ, you get a questline.
- When you accept this questline, raids start showing up
- Raids are temp battles, where only a select few are visible each day. All temp battles reset at mightnight, so we can reuse raids
- When you complete a raid, you gain a Wishing Piece
- The first step in the questline requires you to collect 1 piece. The next requires you to collect 2 (numbers will change)
- Each time you complete a quest step, a new boss shows up in the new dungeon "Max Lair"
- Which boss it is, is random with the trainer id as seed

Missing (will be solved in other PRs):

- [ ] Add way more raids
- [ ] Figure out how many raids should show up each day
- [ ] Balance the raids
- [ ] Add images for raids and wishing piece
- [ ] Change quest title + description (and other texts)
- [ ] Add all forms to Max Lair
- [ ] Balance that dungeon + add loot
- [ ] Figure out how many wishing pieces are needed for each step
- [ ] Add a NPC to the start of the quest, to explain what is going on
- [ ] Add a steps in between existing steps, so it looks like you are turning in the pieces (the same NPC can be reused)
- [ ] Testing + tweak the solution
- [x] Fix `SeededDateRand.shuffleArray`. The results does not look random at all
- [ ] Move the dungeon (I added it to "Tunnel to the top" because i am bad at reading. But it's a fine placeholder, so i don't have to ruin the map)

## Motivation and Context

We have wanted this solution for a while

## How Has This Been Tested?

Went through the questline, and saw it worked as i wanted.
Timetravelled, to see that the temp battles gets reset

## Types of changes

- New feature


RedSparr0w (approved):

PR #4550, opened 2023-07-20
Hide uncaught base forms from Pokedex if alternate non-regional form is caught

CypherX:
## Description
Hides uncaught base form pokemon from the pokedex after obtaining an alternate form with the same native region.

This appears to already be intended but was not working correctly.

## Motivation and Context
This was previously fixed at one point then was unintentionally busted again.
Frequent questions and confusion regarding Unown (A) appearing as uncaught when players are trying to complete the Johto pokedex while having caught a different Unown.

## How Has This Been Tested?
Unown (A) no longer shows as uncaught after capturing any other Unown.
Alolan & Galarian regional forms (still) appear independently from their earlier forms when uncaught.

## Types of changes
- Bug fix

### DIFF
--- a/src/scripts/pokedex/PokedexHelper.ts
+++ b/src/scripts/pokedex/PokedexHelper.ts
@@ -119,7 +119,9 @@ class PokedexHelper {
                 return false;
             }
             // Hide uncaught base forms if alternate non-regional form is caught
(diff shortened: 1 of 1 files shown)


Jaaslet:
I can't seam to replicate the bug? This is how my pokedex looks without your change, after have removed Unown A:
image

CypherX:
Your pokerus status filter is set to Contagious

Jaaslet (approved):
Sorry. Yeah, it was because of the filter.
I was just confused, because i feel like i have merged this exact PR a couple of times before (why does it keep breaking?).
But it works! Thanks!

PR #4555, opened 2023-07-21
Add a "Disable Auto Save" feature

PaddyTwohill:
## Description

Adds a "Disable Auto Save" setting that when toggled disables auto saves

## Motivation and Context

SRing for specific mons in challenge runs is painful, so this aims to help by allowing auto saves to be disabled.

## How Has This Been Tested?

Tested the game with the setting on, and it doesn't automatically save. Manual save via Shift+S still possible
Tested the game with the setting off, and it does automatically save

## Types of changes

- New feature


Jaaslet:
I am going to close this PR.
I am afraid that we will get a lot of bug reports from players enabling this setting, and then not understanding why their game is not saving.
A little bit of history. Back when i joined, the game did not autosave on game close. It only saved every 10th sec. And it let to a lot of bug reports, from players who did not understand this. That is why it got added.
I know that this will lead to a lot less bug reports, but it's only there to play the game in an unintended way. I don't mind players using SR, but it's a bit weird to have official support for it. The game is not balanced with SR in mind, but it will be harder to ignore, if we help players do it.

And we are slowly working on making monotype runs doable, while not officially supporting it. Rare candy and the code should help some, but we are looking at the rest of the problems.

Sorry

PR #4558, opened 2023-07-21
Initial XD Content

Farboo171:
## Description
Adds in towns, dungeons, temporary battles, dummy images, and new trainers that will eventually be Post-Unova-E4 difficulty content.

Attempts were made to make all XD content hidden or locked, such that Colosseum content can be brought out of development and launched, with XD content not interfering.

## Motivation and Context
XD content in Orre is needed, this is one manageable chunk of the larger project

## How Has This Been Tested?
Tested each dungeon
Checked the console for broken image links (fixing these, then I'll bring it out of draft)
Checked that no temporary battles were shifted around (index check)
Checked that all dungeon clears were as expected (index check)

## Types of changes

- New feature


src/scripts/quests/QuestLineHelper.ts
```diff
@@ -1495,6 +1495,15 @@ class QuestLineHelper {
 
         App.game.quests.questLines().push(plasmaUnovaQuestLine);
     }
+    // XD Questline, available after Unova E4
+    public static createOrreXDQuestLine() {
+        const orreXDQuestLine = new QuestLine('Gale of Darkness', 'Team Cipher has returned to Orre. Stop their new evil plan!', new DevelopmentRequirement(), GameConstants.BulletinBoards.Hoenn);
```
Jaaslet, line 1500: I think the questline should be in the unova board. Makes it easier for players to discover it.
Also, i think we want the players to complete the colosseum questline first, right?

Jaaslet (approved):
I will merge this PR, because it's fine. Just have some small comments, but those can be added later

src/scripts/quests/QuestLineHelper.ts
Farboo171: Yes to both. The quest line in this PR is just a placeholder now, next PR will incorporate both of those changes

PR #4559, opened 2023-07-21
Add desert environments

osherwott:
Adds 2 new environments based on deserts

## Description
Adds two new environments based on tropical and temperate deserts.
- Tropical desert background is taken from Haina Desert in gen 7 (Sun & Moon, Alola)
  - used in areas that are sufficiently hot and representative of subtropical or tropical deserts:
    - Hoenn 111, Orre Rock Poke Spot, Outskirt Stand, Unova 4, Desert Resort, Haina Desert
- Temperate desert background is taken from Potbottom Desert in gen 8 (Sword & Shield, Galar)
  - used in dry, somewhat rocky areas of more temperate climate that aren't outright deserts, or deserts with not as harsh conditions as the above (this also replaces Reversal Mountain's battle background to better match the red hue of its town image):
    - Pyrite Town, Pyrite Town Battles, Sinnoh 228, Reversal Mountain, Kalos 13, Galar 6, Galar 8, Dusty Bowl, Potbottom Desert, Stow-on-Side
    - Also added to Lentimas Town in Unova, because despite having no battles, the town image sometimes takes a split second to load and the red of the desert matches the town image better than the default environment

_Disclaimer: These images were taken as screenshots from the games themselves with minimal editing to remove any obstructions. Maybe another person can find better ways to source these pictures, but this is what I could do with my limited knowledge and resources._

## Motivation and Context
Part of long-term plan to expand the variety of battle backgrounds. Feedback was taken from players and developers to add both these environments, as opposed to just one.

_Closes #3297_

## How Has This Been Tested?
The game was run and every location that uses these new environments was tested. The backgrounds appeared without fault.

## Screenshots:
desert sandshrew
desert sandaconda

## Types of changes
- New feature
- UI improvement


osherwott:
One desert has been chosen for now. A different PR will be made to more freely choose CSS backgrounds independent of environments, as to prevent unnecessary duplicates or multiples of similar ones. The other image will be added later because of this.

Jaaslet (changes requested):
Can we use the other images? The one in use right now, has very strong colors compared to the other backgrounds (which is a bit hard on the eyes)

DraKay:
IMG_20230804_090853
Dunno if it helps, took the image, reduced saturation and light and raised contrast a bit. Seems less harmful than before dunno if enough

Jaaslet:
> Dunno if it helps, took the image, reduced saturation and light and raised contrast a bit. Seems less harmful than before dunno if enough

Looks a lot better. Both solutions would be fine imo

Farboo171:
The following items are a mis-match between the description and the code, in that they are not desert:
Pyrite Town Battles, Reversal Mountain, Dusty Bowl

Is this intentional, or something from changing from two deserts to one and these didn't make the cut?

osherwott:
Good catch!
I had preemptively removed them until the other desert was in, in case of eye strain (you can idle on routes but dungeons require active screen-watching). But the image has been color-balanced and that is no longer a problem, so I have added them back in :)

Farboo171 (approved):

PR #4575, opened 2023-07-24
Ability to pause & resume bulletin board quests

CypherX:
## Description
Adds the ability for players to pause and resume bulletin board quests.

Also a minor language tweak ("Progress" -> "Step") in the quest line log Active Quest Lines table.

## Motivation and Context
Challenge files stuck with quests that were previously automatically triggered (e.g. Mystery of Deoxys)
Quest prioritization and reducing the amount of space being used on the main UI 

## How Has This Been Tested?
Loaded a save with several quest lines in progress. Paused them at various steps and verified progress was not gained. Reloaded the game, resumed a quest line and verified it continued at the same step it was paused on.

## Screenshots (optional):
image
image
image

## Types of changes
- New feature


CypherX:
Good catches.

> The change does not work with requirements. If you are right before a temp battle, and you pause it, you can still defeat the temp battle. Which will hardlock that quest for you

Updated QuestLineStepCompletedRequirement to handle suspended quests. I did not see any other requirements where changes were needed. Tested by progressing the Zero's Ambition quest to a step with a temp battle, paused the quest line, and the temp battle was no longer accessible.

> If you have completed a part of a multi quest, it resets the whole quest on resume.

Completed sub-quests of multi quest steps will be preserved now. Any progress in incomplete steps will be lost (e.g. having 5 of 10 pixie plates would get reset to zero upon resuming). I could not find a good way to preserve partial progress without reworking how quests track progress. There is a warning on the pause button tooltip about this progress being lost.

Jaaslet:
> I could not find a good way to preserve partial progress without reworking how quests track progress

I think we have quests where you need to defeat more temp battles in the same quest? Maybe the meltan questline? It will be possible to hardlock that one

CypherX:
Right, I see where that could cause an issue. I'll try to figure something out.

CypherX:
Updated to preserve all progress which actually ended up simplifying some things.

Tested with both single quest and multi quest steps in Zero's Ambition, including a temp battle.
Tested with the first step in `Defeat Rainbow Rocket` that requires battling two temp battles in one step.

Jaaslet (approved):

Issue #4576, opened 2023-07-25
[SUGGESTION]: breedingModal performance tweak


umbralOptimatum:
### Current Behavior

I've noticed that the breeding modal can take a while to load all the available-to-breed pokemon, and the menu is often pretty laggy on my computer. I noticed that the modal currently binds elements for every single party pokemon, even if only a tiny fraction of them actually need to be shown. Testing with a KO `if` binding instead of a `hidden` binding, I found that the pokemon list loads virtually instantaneously and is more performant. The only downside I've noticed is occasional flickering when elements are created for newly-available pokemon.

As someone outside the development team, I don't know the specifics of why it was implemented this way; has the `if` approach been considered?

PR #4577, opened 2023-07-25
breedingModal performance improvements

umbralOptimatum:
Edit: Substantially overhauled instead, see later comments.

Changes the breedingModal pokemon list from `hidden` to `if` bindings to improve performance. This way, it doesn't have to create and bind menu items for the entire party, making it load the pokemon actually matching the breeding filters much more quickly. See #4576.

Tested on desktop by editing the bindings pre-game-load and experimenting with various combinations of filters, adding pokemon to the hatchery, allowing filtered-for pokemon to reach lv100 while the menu was open, etc. The only issue I noticed was occasional flickering around the edges of pokemon elements when inserted into the menu.

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -2412,7 +2412,7 @@ class Update implements Saveable {
                     helper.sortOption++;
                 }
             });
-          
+
             //Replace Blaze Cassette with Magma Stone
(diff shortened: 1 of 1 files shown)


Jaaslet:
My experience is that this makes changing the filters a lot slower, as it has to readd all the html. How many mons are you testing with?

PR #4578, opened 2023-07-25
Safari: Surf

Crobat4:
## Description

This PR adds the ability to use Surf on water tiles in the Safari, Pokémon can also appear in there. 
Player sprites and battle background changes accordingly.

https://github.com/pokeclicker/pokeclicker/assets/104547700/d5b864f3-0f41-4c36-b550-f3b0f5c8e7ba

Also adds an environment attribute to `SafariPokemon` to choose which Pokémon can appear in water tiles, grass tiles or both.
Items can't spawn on water tiles.

Safari pool was modified to add Pokémon that can appear in water following some FRLG encounters, Dratini and Dragonair require to be caught before, like most egg exclusives.
As for Friend Safari, I just added Lapras.
Friend Safari special Pokémon are grass-only.

## Motivation and Context

Started as a workaround for inaccessible items/Pokémon due water tiles and fence but decided to try  to make it a bit more like the main games and add wild Pokémon in there.

## How Has This Been Tested?

* Tested several runs, walking on water and trigger battles. No grass-only Pokémon in water and vice versa.
* Filled the map with Pokémon spawns, all of them appeared in tiles according their environment.
* Filled the map with items, no items on water.

## Types of changes

- New Safari feature


src/scripts/safari/SafariPokemonList.ts
```diff
@@ -40,6 +42,14 @@ class SafariPokemonList {
             new SafariEncounter('Cubone', 10),
             new SafariEncounter('Marowak', 5),
             new SafariEncounter('Tangela', 4),
+            // Water
```
Farboo171, line 45: Magikarp is missing from the encounter list.

But, if we're going slightly off canon, should the dragons be in here? It would make the dragon egg obsolete.

Crobat4: I'll add Magikarp.

As for Dratini and Dragonair you still need to catch them before they can appear (same condition as Friend Safari special Pokémon) so I don't think it would make the egg obsolete.
I can remove them if you prefer, however.

Farboo171: Serves me right for commenting before fully understanding.

Adding Magikarp and keeping the dragons locked behind already being obtained should be totally fine, thanks!

src/scripts/safari/SafariPokemonList.ts
```diff
@@ -63,7 +74,7 @@ class SafariPokemonList {
         const endIndex = startIndex + GameConstants.FRIEND_SAFARI_POKEMON;
 
         const pokemon: SafariEncounter[] = shuffledPokemon.slice(startIndex, endIndex).map((p) => {
-            return new SafariEncounter(p, 10, true);
+            return new SafariEncounter(p, 10, [SafariEnvironments.Grass], true);
```
Jaaslet, line 77: We could consider checking if the pokemon is water type or not? But i am not sure if we can make a general rule for that? Should they all only be in water, or should they all only be in grass?
But it's not a problem with this PR, so i will merge this, and then we can make the change in a new PR, if we want to

Jaaslet (approved):

src/scripts/safari/SafariPokemonList.ts
Aegyo: imo pure water (and maybe Water/Ice) should be Water only encounter, Water/Other should be both, everything else grass only

Jaaslet: That is smart

Aegyo: Mentioned on discord that if some pokemon will be water exclusive, we should make sure every safari has water tiles

PR #4584, opened 2023-07-27
Gift NPC

CypherX:
## Description
Implements the ability for NPCs to give gifts/rewards/whatever that the player can choose to accept. Hell, we could even have a NPC scam the player (hint hint eugene).

Example usage:
```
const GiftTest = new GiftNPC('Shady Bloke', [
    'Oi bruv, I found this thing, you want it?',
], () => {
    App.game.party.gainPokemonByName('Lapras');
}, 'assets/images/pokemon/131.png', { saveKey: 'uniquestring', image: 'assets/images/npcs/Professor Oak.png' });
```
**Important Note: ** ALL gift/reward NPCs must have a unique `saveKey` string passed in or they will not work properly.

Other Changes:
- Reworked some of the NPC code. Created NPCController, got rid of the jquery used to populate the NPC modal and replaced it with knockout bindings.
- Ability to track when a player has talked to a NPC. This has to be enabled by passing a `saveKey` to the NPC.
- Moved the `hash()` function out of `RedeemableCodes` and into `GameHelper`
    - Statistics would not allow string keys so we hash the `saveKey` string and use that as the statistic key.

Things to come in future PRs once this is merged:
- Moving Lapras and Eternatus out of mandatory temp battle rewards
- Change `TalkToNPCQuest` to check the statistic rather than the weird way it works now?
- More fun things?

## How Has This Been Tested?
Created a test NPC that gifts the player a Lapras. Upon accepting the gift the NPC is hidden and the statistic is updated according.
Checked various other NPCs to verify they display correctly.

## Screenshots (optional):
image

## Types of changes
- New feature


src/scripts/towns/NPC.ts
```diff
@@ -1,16 +1,23 @@
 type NPCOptionalArgument = {
     requirement?: Requirement | MultiRequirement | OneFromManyRequirement,
     image?: string,
+    saveKey?: string,
 };
 
 class NPC {
     public talkedTo = ko.observable<boolean>(false); // Used for custom quests
```
Aegyo, line 8: Am I blind or is this no longer updated by anything

CypherX: You right. Forgot to re-add it after toying with something. Fixed.

Jaaslet (approved):

PR #4586, opened 2023-07-27
No more forced Pokemon

Qwertypop04:
## Description
Makes Lapras and Eternatus optional using the new GiftNPCs

## Motivation and Context
Means there are now no specific Pokemon the player is forced to catch

## How Has This Been Tested?
Game runs, NPCs work

## Types of changes
- Player option improvement

### DIFF
--- a/src/scripts/temporaryBattle/TemporaryBattleList.ts
+++ b/src/scripts/temporaryBattle/TemporaryBattleList.ts
@@ -170,9 +170,6 @@ TemporaryBattleList['Blue 5'] = new TemporaryBattle(
     {
         displayName: 'Rival Blue',
         imageName: 'Blue2',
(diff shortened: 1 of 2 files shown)


CypherX:
Should we hide these from players who have already captured Lapras/Eternatus?

Qwertypop04:
> Should we hide these from players who have already captured Lapras/Eternatus?

Could do I suppose, though I don't think there's any real need to

CypherX:
Could be done with a requirement or an update function to mark the NPCs as talked to. But I agree, no real need to, just throwing it out there.

Jaaslet:
I think we might get bug reports about it not giving EVs or something, if we don't hide it. So a requirement sounds good

Jaaslet (approved):

PR #4591, opened 2023-07-28
Remove Scent held item on purify

CypherX:
## Description
Scent held items can only be used by shadow pokemon. This will remove the scent upon being purified if one is equipped.

## Motivation and Context
Bug

## How Has This Been Tested?
Purified a pokemon with no held item.
Purified a pokemon with a power brace - it was not removed.
Purified a pokemon with a joy scent - it was removed.

## Types of changes
- Bug fix

### DIFF
--- a/src/scripts/party/PartyPokemon.ts
+++ b/src/scripts/party/PartyPokemon.ts
@@ -22,6 +22,7 @@ class PartyPokemon implements Saveable {
     public exp = 0;
     public evs: KnockoutComputed<number>;
     _attack: KnockoutComputed<number>;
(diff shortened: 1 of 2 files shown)


Jaaslet:
Could we make a more general solution? Maybe turn requirement into a purecomputed and subscribe to it? Don't think it will cause lag

CypherX:
Updated to have a computed and subscription on the partypokemon to automatically remove a held item if the pokemon no longer meets the requirements to equip it.

Jaaslet (approved):

PR #4593, opened 2023-07-28
Dungeon : fix and clean for prevention of boss tile from being replaced by entrance tile

RegisCoaxans:
## Description
This was expected to be prevented by two different pieces of code. I fixed the broken one and removed the second.
However, the second part was removing either a chest or an encounter in order to spawn the entrance.
Dungeons now have their correct amount of special tiles.

## Motivation and Context
I was looking at the map generation code and noticed two statements were supposed to do the same thing.
Before the change, up to 60% of the dungeon maps with even size lacked either a chest or an encounter.

## How Has This Been Tested?
I run a script to generate many maps and check all tiles were fine.
```js
var res = [], size = 6;
for (var i = 0; i < 5000; i++) {
	var map = new DungeonMap(size, () => {return {}});
	var tiles = GameHelper.enumNumbers(GameConstants.DungeonTile).reduce((acc, t) => {acc[t] = 0; return acc;}, {});
	map.board().flat(5).forEach(t => tiles[t.type()]++);
	if (tiles[GameConstants.DungeonTile.entrance] !== 1 || tiles[GameConstants.DungeonTile.chest] !== size || tiles[GameConstants.DungeonTile.enemy] !== size * 2 + 3 || tiles[GameConstants.DungeonTile.boss] !== 1)
		res.push(map);
}
res.length;
```

## Screenshots (optional):
Before change.
image
After change.
image

## Types of changes
- Bug fix
- Code optimization
- Make it harder


RegisCoaxans:
New algorithm.
Instead of pushing all empty tiles until `size ^ 2`, it goes up to `size ^ 2 - 1`, and shuffles, then insert the empty tile right where it should be.
This way, shuffling goes down to one iteration.

Jaaslet (approved):

PR #4594, opened 2023-07-28
Added Fish Police to Magikarp Jump

EatPant2nd:
## Description

Added "The Fish Police" NPC to Magikarp Jump subregion in case... somehow, people make it there with no Magikarp in their party. I also hid the other NPC with a Magikarp Requirement (I made sure to consider both available before this point) so they won't appear until they get a Pokémon that can do damage in the region. Once a relevant Pokémon is acquired, The Fish Police will go away and the usual NPC will show up normally.

## Motivation and Context

Since Magikarp Jump subregion can be reached after becoming the Alola Champion regardless if the player has a Magikarp or not to even accept the quest line that grants access to said place; I thought it would be funny to have this NPC informing the player that **only** Magikarp works there.

## How Has This Been Tested?

I went with a file with no Magikarp and was stopped by the Fish Police. Went and captured a Magikarp and the Fish Police left me alone. Tested with both Magikarp.

## Screenshots (optional):

fish-police

## Types of changes

- New feature


Jaaslet:
Could we just add the Magikarp requirement to the quest itself?

CypherX:
> Could we just add the Magikarp requirement to the quest itself?

That's no fun!

EatPant2nd:
Consider it a nice Easter Egg. Besides, it does have a purpose, in a way. If I lock the quest line with a Magikarp, people would be unaware of this place existing since Bulletin Board doesn't exactly show you unavailable quests. Like this, however, if, for some reason, someone didn't have a Magikarp this late into the game, it would be nice to let them know that the region exists so they can work towards it.

Jaaslet (approved):
I don't like the solution, but it's too funny to not merge

PR #4598, opened 2023-07-29
Added a simple implementation for "startContest"

Jaaslet:
## Description

A step towards contests being finished!
This PR only focuses on giving rewards for completing a contest. Splitting stuff into smaller PRs, so it's easier to review.
You gain contest tokens for each style point you gain.
Your rank is based on your style points. Currently, it only increments some statistics. I have made a requirement based on those statistics.
My current plan is to base rewards on requirements, so that is why i have created one. Also: we want to create some achievements.
The feature might need some animations or something. I will need to think about that, so it will not be in this PR. I also want this one merged, so it's easier to work on the other parts of the feature.
It's currently not consuming berries. I think it's boring if it just consumes one of each. So i will have to think about that one.
I intentionally left the `lastEnteredDate` unsaved, to make it easier to test contests

## Motivation and Context

I want contests released!

## How Has This Been Tested?

I completed a contests and created a new requirement using the console.

## Types of changes

- New feature


Farboo171:
Works as described, tested multiple times across multiple saves. Correctly doesn't let players enter twice, except on reload (which you state is intended for testing and has comments)

Looks good, thanks for the PR, as they say.

Farboo171 (approved):

PR #4600, opened 2023-07-29
Purify chamber is now based on all shadow mons

Jaaslet:
## Description

You gain purify chamber juice each time you gain exp for shadow mons, no matter their level.
Eugene will adjust the formula to the new values.

## Motivation and Context

I tried to be too clever with my old solution, and it turned into another thing mostly based around breeding.
This solution makes exp held items do something useful, and it rewards the player for getting more shadow mons

## How Has This Been Tested?

I opened the game and saw that EXP still works

## Types of changes

- New feature


Farboo171 (approved):
Tested, behaves as anticipated with both shadow and purified mons

Aegyo (approved):

PR #4604, opened 2023-07-29
Contest prizes

Jaaslet:
## Description

Added a prizewall to the contest. The idea is that you can claim them once, after they are unlocked. Unlocked in based on requirements, so we have a lot of freedom with them. And with the requirement from #4598, Eugene should be able to add all the prizes the way he wants

## Motivation and Context

Want to finish the contests

## How Has This Been Tested?

Opened and claimed stuff. Reloaded and saw it stayed claimed

## Types of changes

- New feature


Farboo171:
Buttons to claim a prize that one can claim (Altaria stone in this case) and a prize one cannot (Rare candy) are both the same color and both clickable.

The obtainable one resolves as expected, but there is no indication that the unclaimable one isn't claimable except that it doesn't work.
image

If that's to be resolved later, everything else looks good. Tested several random items to make sure the changes to ItemNameType didn't affect them.

Farboo171 (approved):
Tested, the new buttons work great! Looks good, thanks for the PR!

PR #4625, opened 2023-08-05
Mewtwo Strikes Back Rework

Farboo171:
## Description
Re-works the Mewtwo Strikes Back event in the method requested by the powers that be.

## Motivation and Context
Makes the event more interesting. Adds Clone Pikachu.

## How Has This Been Tested?
Tested with the Event Calendar.

## Types of changes

- New feature


src/scripts/dungeons/Dungeon.ts
```diff
@@ -1306,6 +1306,59 @@ dungeonList['Berry Forest'] = new Dungeon('Berry Forest',
     ],
     1750, 29);
 
+dungeonList['New Island'] = new Dungeon('New Island',
+    [
+        new DungeonTrainer('Armored Mewtwo',
```
osherwott, line 1311:
```suggestion
        new DungeonTrainer('Armored',
```
Do this and change the filename to "Armored". No clue why it's not showing up, but this works. Maybe it's confusing the name with the pokemon name?

Jaaslet (approved):

PR #4626, opened 2023-08-05
Quest pausing fixes / tweaks

CypherX:
## Description
Fixes two things found during testing:

1.  Pausing a quest while engaging in a temp battle or dungeon locked behind said quest could result in a softlock or accessing a dungeon boss that shouldn't be available.

Players may now only pause quests while in a town or route, this should prevent abusing the pausing ability or any softlocks that might occur. Updated the pause button tooltip to reflect this.

2. Quest lines that start and rely on progress of other quest lines (there are currently only two that I'm aware of) could be paused and softlocked by making progress in the second quest line while the first is paused.

Added a parameter to explicitly disable pausing for certain bulletin board quests and applied it to these two quests.

## How Has This Been Tested?
Progressed to a temp battle step in a quest line and attempted to pause the quest while doing the battle, button was disabled.

Entered a dungeon and attempted to pause a quest, button was disabled.

Pause option was disabled for the two quests it was specifically disabled for.

## Types of changes
- Bug fix


CypherX:
#4630 and #4631 both cover edge cases where pausing could be used to exploit or access an area or pokemon that should not be available while the related quest is paused.

Jaaslet (approved):
The two questlines with exceptions seems like one of those you want to be able to pause. Monotype runs that starts them by accident most likely want to remove them from the list.

But I don't really know a better solution, so this is good enough for now

PR #4631, opened 2023-08-06
Update Vivillon quest line to use requirements

CypherX:
## Description
Updates the Vivillon quest line to use requirements rather than pushing to the dungeon boss list or BF reward list for the quest step.

## Motivation and Context
It's better. And works properly with quest pausing.

## How Has This Been Tested?
Played through the quest line and verified everything appeared and disappeared at the correct stages.

Activated the Lunar New Year event and the Vivillon's re-appeared as dungeon bosses.

## Types of changes
- Quest Line


src/scripts/battleFrontier/BattleFrontierMilestones.ts
```diff
@@ -113,6 +113,10 @@ BattleFrontierMilestones.addMilestone(new BattleFrontierMilestoneItem(550, 'Elec
 BattleFrontierMilestones.addMilestone(new BattleFrontierMilestoneItem(575, 'Protector', 15, new MaxRegionRequirement(GameConstants.Region.sinnoh)));
 BattleFrontierMilestones.addMilestone(new BattleFrontierMilestoneItem(600, 'Reaper_cloth', 15, new MaxRegionRequirement(GameConstants.Region.sinnoh)));
 BattleFrontierMilestones.addMilestone(new BattleFrontierMilestoneItem(660, 'Sachet', 15, new MaxRegionRequirement(GameConstants.Region.kalos)));
+BattleFrontierMilestones.addMilestone(new BattleFrontierMilestonePokemon(666, 'Vivillon (Poké Ball)', new MultiRequirement([
+    new QuestLineStepCompletedRequirement('The Great Vivillon Hunt!', 34),
+    new QuestLineCompletedRequirement('The Great Vivillon Hunt!', GameConstants.AchievementOption.less),
```
Jaaslet, line 118: I don't think we need the `less` requirement, as the milestone goes away after you achieve it the first time?

CypherX: That's true going forward. Because of how it was added and removed to the reward list the milestone wasn't being tracked in the save, so players who have already completed the quest and obtained it will see it re-appear in the reward list.

I suppose adding it to the save in the update function and getting rid of that extra requirement would be better then.

CypherX: Updated to remove the extra requirement and added an update function that will add it to the obtained milestones if the quest is completed.

Jaaslet (approved):

PR #4650, opened 2023-08-09
Safari shiny now gives extra exp

### LINKED ISSUE #4503: [SUGGESTION]: Shiny and legendary in safari could gives more XP
### Current Behavior

Shiny and legendary pokemon when captured gives the same amount of safari XP than the other pokemon

### Improved Behavior

it could gives twice as much XP so 100
and a legendary shiny could give 200

### Reason for change

it would give an incentive to capture shiny and legendary of pokemon you already have

### Detailed Description

by giving a bonus of XP it would rebalance hard pokemon like legendary to be more worth capturing, even tho it's more difficult
and shiny is just for "fun"

Jaaslet:
## Description

We now start tracking how many shinies the player catches in the safari zone.
Shinies gives 5x exp.
Added an achievement for catching a single shiny in the safari zone

## Motivation and Context

Closes  #4503.
Shines gives bonus for everything else. 5x seems worth it, but not broken.
The achievement is only for a single, because i am not too mean

## How Has This Been Tested?

Caught a shiny in the safari zone, and saw the exp go up by what i expected. Saw i got the achievement

## Types of changes

- Feature improvement


Farboo171 (approved):
Looks good! Changed grammar in the description.

PR #4651, opened 2023-08-09
TemporaryBattleQuest

Nydaleclya:
## Description

Added a Defeat Temporary Battle Quest class.

## Motivation and Context

Making it simpler to add Temporary Battles in Questlines.

## How Has This Been Tested?

Run through some Questlines in different Regions and nothing seems broken.

## Types of changes

- Small Code Refactor


Jaaslet (approved):
I have wanted to do this for a while. Thanks!

PR #4652, opened 2023-08-09
Mega Stone achievements

CypherX:
## Description
Adds 3 achievements for obtaining 1, half, and all available mega stones. All mega stones are not implemented yet so it calculates the half and total from those available to prevent impossible achievements.

Feel free to come up with better names or change the bonus percentages.

## Motivation and Context
More achievements!

## How Has This Been Tested?
Loaded a save with all current mega stones obtained, achievements were completed.
Loaded a save with no mega stones, achievements were incomplete. Obtained one mega stone and the first achievement completed correctly.

## Screenshots (optional):
image

## Types of changes
- Achievements


RegisCoaxans:
We are almost done implementing the megas. I think the 50% can already be set to 25 (out of 50).
Additionally, do you mind a Power Rangers reference with "Go, Go, Mega Force !" ? :^)

EatPant2nd:
I'm pretty sure it's "y'all" and not "ya'll", by the way.

RegisCoaxans:
I would say second one.

CypherX:
- Updated the second/half achievement to require 25 and changed the name to `Go, Go, Mega Force!`
- Fixed `Ya'll`

Jaaslet (approved):
I moved the achievements from global to Kalos, as they are a Kalos specific mechanic

PR #4654, opened 2023-08-10
Halloween event refactor

CypherX:
## Description
Refactors the Halloween event to work with the Event Calendar.  Content of the event does not change.

Might be worth splitting the Random requirement to it's own thing and using a MultiRequirement when adding them to the routes?

## How Has This Been Tested?
Activated the event and checked that the spoopy pokemon were added to routes. Loaded the live version and performed magic to where the event would normally be active. Compared the routes and they matched.

## Types of changes
- Event


src/modules/requirements/SpecialEventRandomRequirement.ts
```diff
@@ -0,0 +1,15 @@
...
+
+    constructor(specialEventName: string) {
+        super(specialEventName);
+        this.isAvailable = SeededRand.boolean();
+    }
+
+    public getProgress(): number {
+        return +(this.isAvailable && super.getProgress());
+    }
+}
```
RedSparr0w, line 15: I think it would be worth adding special hint text for this as it's being added to every route with a requirement, maybe something along the lines of `The Pokémon seems to be avoiding the area this year` (just as an example).

Also curious as I haven't actually looked over the event calendar code, why does the previous code not work anymore?
As that only added it to routes based on the boolean rather than locking behind a req

CypherX: Something like that?

And yeah - event pokemon were moved to requirements to work with the event calendar. Halloween was the last hold out because of the unique distribution based on year

CypherX:
Alternatively, we can skip the `SpecialEventRandomRequirement` entirely and just do something like:
```
SeededRand.seed(new Date().getFullYear());
Routes.getRoutesByRegion(Region.kanto).forEach((route) => {
    if (SeededRand.boolean()) {
        route.pokemon.special.push(new SpecialRoutePokemon(['Spooky Bulbasaur'], new SpecialEventRequirement('Halloween!')));
    }
    if (SeededRand.boolean()) {
        route.pokemon.special.push(new SpecialRoutePokemon(['Gastly'], new SpecialEventRequirement('Halloween!')));
    }
});
```

RedSparr0w:
Only problem with that is on new years needing the game to be refreshed, dunno if worth it or not 🤔

RedSparr0w:
Probably easier to just keep it as you've already done, shouldn't have much of an impact as the boolean is set on loading of the event

CypherX:
> Probably easier to just keep it as you've already done, shouldn't have much of an impact as the boolean is set on loading of the game/event req

Works for me if it's good with you.

Farboo171 (approved):

PR #4655, opened 2023-08-10
Route and dungeon cycling does not prevents zooming in and out anymore

RegisCoaxans:
## Description
Changing the zoom level was unavailable in routes or dungeon towns because of the cycling keybinds cancelling default event handling. This PR prevents cycling if Ctrl is also held (as it is often used for zoom keybind combination). This does not fix zoom being kind of broken on client, though.

## How Has This Been Tested?
Opened the game, zoomed in then out, closed it.

## Types of changes
- Bug fix
- UI improvement


RedSparr0w (approved):

PR #4660, opened 2023-08-12
Vitamin, held item, & consumable modal performance improvements

CypherX:
## Description
Optimizes the vitamin, held item, and consumable modals to improve overall performance.

The primary change is altering how the pokemon list is rendered and filtered for each modal. Previously the full caught pokemon list was displayed then knockout visible/hidden bindings were used to show/hide each pokemon entry based on the filters. Now the filtering is done independently on the caught pokemon list itself which results in significantly fewer dom/style updates and bindings.

Added lazy loading to the held item and consumable modals.

Removed the `ko if: observable state` from each modal as it's not needed with lazy loading (resets to first page upon close) and it caused lag when closing the modals.

## Motivation and Context
These modals have noticeable performance issues causing lasting lag, stutters, and memory usage spikes. Players frequently complain and report these.

## How Has This Been Tested?
Tested that all modals still work as expected. Asked several players to test the changes and all reported noticeable improvements including significantly less lag, no stutters when opening/closing the modals, and no memory usage spikes.

## Types of changes
- Performance


RegisCoaxans (approved):
Looks good to me.
It even fixes the issue I wanted to address in #4405.

Farboo171 (approved):

PR #4661, opened 2023-08-12
Kanto now has increasing max levels again

Jaaslet:
## Description

In #4600 i wanted the level function to calculate the exp, even if the pokemon was at max level, for the purify chamber. I forgot about the badge max level, so that went away. This PR fixes that

## Motivation and Context

Fixing a bug in dev

## How Has This Been Tested?

Started a new game. Stood on a route for a while, to make sure level 20 was max level. Then i beat brock and saw that my pokemon could level up again.

## Types of changes

- Bug fix


Farboo171 (approved):
Verified the level cap from badges works again, and that the Flow calculations for shadowmons still work. Thanks for the PR!

PR #2460, opened 2022-06-15
Small fix

Nydaleclya:
Knockout just expects a 6 digit hex code not a 3 digit one.

image

PR #4665, opened 2023-08-14
Fixing category name text colour and None category default colour

### LINKED ISSUE #4663: [SUGGESTION]: Modify the color of the category names for darker category colors
### Current Behavior

At the moment, when viewing the category dropdown in the Pokedex or hatchery all names of categories are listed in black. This means any darker color requires straining your eyes to read the name. My 'Mega Evo' category in particular is much harder to read on my computer than this screenshot suggests.

image

### Improved Behavior

If it is not incredibly difficult to do so, this could be resolved by a system of writing darker colors of category in white.

### Reason for change

Improves a feature visually.

### Detailed Description

I would suggest that any color below a certain threshold of Red, Green and Blue on the color settings automatically causes the text to be written in white.

While I am aware people can simply choose their own lighter colors if they have a problem, I believe a lot of the problem is caused by the default 'None' category being in black, the color affected worst by this feature. Therefore an alternative, simpler solution could be to change the default color of the 'None' category to something lighter, such as a paler shade of grey.

RegisCoaxans:
## Description
I added function `isColorLight` which uses hexadecimal colour and tells if it rather is bright or dark. Used it on texts for category names in category dropdowns (used in hatchery and pokédex) to avoid writing bright on bright or dark on dark.
I also completed #2460 and fixed invalid colour format stored in the save for None category.

## Motivation and Context
Closes #4663 and improves readability.

## How Has This Been Tested?
Opened the game, played around with categories.

## Types of changes
- Bug fix
- UI improvement


Farboo171 (approved):
I went in and resolved the conflicts with the update function. Looks good, thanks for the PR!

PR #4666, opened 2023-08-14
Replaced santa snorlax with Snorlax (Snowman)

Jaaslet:
## Description

Santa Snorlax have been replaced with Snorlax (Snowman) from the pikachu short "Snorlax Snowman". It is now also ice type.
As it's just a snorlax for another snorlax, the old one is just replaced, with the same EV, breeding and so on.
As Elf Munchlax is no longer really a baby version of it, i have added it as a roamer. If players like this change, i have an idea for replacing that too, so i am keeping it there as a temp solution (not really that temp, as it's a fine solution).
New snorlax is created by Drakay and the creator of the old snorlax is okay with me removing it.

## Motivation and Context

It would be cool to remove at least some of the non canon forms, and this seems like an easy way to start, as it's mostly just a clean change.
When i say "non canon", i mean that "canon" is anything officially licensed. So from the games, from any sidegames (including PoGo), from the trading card game or from the anime. Officially licensed toys are also technically "canon" with my definition.
I have plans for replacing most of the non-canon forms. Some more drastically than others. Depending on the response from this change, i will replace between 0 and all of the non-canon event mons, while finishing my planned event rework.
I am okay with people not liking this change. I just want to try.

## How Has This Been Tested?

Opened my save and saw i had the new snorlax.
Saw that elf munchlax now was a roamer, with the event calendar active

## Types of changes

- Bug fix


src/modules/pokemons/PokemonList.ts
```diff
@@ -14062,7 +14062,6 @@ export const pokemonList = createPokemonArray(
         'levelType': LevelType.slow,
         'exp': 78,
         'catchRate': 50,
-        'evolutions': [StoneEvolution('Elf Munchlax', 'Santa Snorlax', StoneType.Soothe_bell)],
```
RegisCoaxans, line 14065: Can we then hardcode the computed 32 EC for Elf Munchlax until it gets killed too? Its eggsteps were going down from 1600 to 1280 thanks to baby evolution.

Farboo171:
> I am okay with people not liking this change. I just want to try.

Internal discussion was conducted to see what reaction to replacing existing Fakemons would cause.

Based on this discussion, there were too many questions of consistency with partial replacement, and lots of concerns about specific replacement. 

Unfortunately, existing Fakemons are here to stay, all we can do is not add new ones. That said, please do not replace non-canon mons in your upcoming event rework.

osherwott:
> Based on this discussion, there were too many questions of consistency with partial replacement, and lots of concerns about specific replacement.

That does not mean that people were opposed to this PR specifically.

Jaas it would probably help if you opened a larger PR or Suggestion that laid out all your plans for the replacements.

osherwott:
Also reopening this cause it has been less than a day and not everyone has had the time to share their opinion on it.

Jaaslet:
Turns out there is an official version of santa snorlax, that fits my definition of official. I still have plans for snow snorlax, but that will be a part of my christmas rework.
(link: https://www.pokemoncenter.com/product/741-09481/snorlax-holiday-knit-sweater-adult)

PR #4667, opened 2023-08-14
Achievements Fixes

CypherX:
## Description
The existence of future content in the code was causing some unobtainable achievements to appear.

- Shadow & Purified Achievements
  - Updated the third achievement to be 51 instead of 50. There will be a total of 51 shadow/purified pokemon included in the Orre part 1 content. This will allow players to track their progress to completion.
  - Commented out the 131 achievements until Orre part 2 is released.
- Gyms & Dungeons Achievements
  - Orre gym achievements were missing entirely so I added those.
  - Added code to exclude creating gym & dungeon achievements for locations with a `DevelopmentRequirement`. Some locations exist for future Orre part 2 content (and Max Lair for gigantamax) and were creating unobtainable achievements.
  - Added `DevelopmentRequirement` to future content so they would be excluded from achievements.
- Mega Stone Achivements
  - The Altarianite mega stone was uncommented to be used as a contest reward. This was causing the obtain all mega stones achievement to have one extra that was not obtainable. I re-commented the item (and contest reward) out.

## Motivation and Context
Unobtainable achievements bad and confusing for players.

## How Has This Been Tested?
Loaded a save and verified the impossible achievements are no longer in the list.

## Types of changes
- Achievements


Farboo171 (approved):
Tested them all and they work! Looks good, thanks for the PR!

PR #4673, opened 2023-08-16
Hatchery menu rework for fast loading

### LINKED ISSUE #4576: [SUGGESTION]: breedingModal performance tweak

umbralOptimatum:
Overhaul to the breedingModal pokemon list to improve load time and performance.

Currently the breedingModal list creates elements for every party pokemon regardless of breeding eligibility, with ineligible mons hidden but still part of the DOM. This dramatically slows loading and performance on queries that only needed to display a small number of pokemon, the page instead unnecessarily loading the entire party — and as the rendered list doesn't update when filters reset, it instead tries to render up to all 1400-plus pokemon currently in the game.

I rewrote the breedingModal to only render a filtered list of pokemon matching the breeding filters, similar to the pokedex approach, and to properly reset the rendered list when filters change. Pokemon matching the filters but currently ineligible for breeding are included but hidden as before. Because most of the filter criteria are static, the filtered list rarely needs to be rebuilt from pokemon changes: newly caught pokemon and shiny/pokerus changes are rare occurrences while the menu is open, and the list isn't updated while the menu is closed. Category changes are the only common change, but are unlikely to happen faster than the 500 ms rate limit.

I tested all of this through both regular gameplay and in-depth checking many combinations of all the breeding filters, changing species' eligibility through regular means and direct editing of game statistics. In all cases the filters behave as expected. The only visible difference is the speed at which the list loads. On develop, filter/sort direction combos that begin with many ineligible pokemon (like Galar and sort by ID) can take multiple seconds to display, as the LazyLoader gives a maximum of 600 entries per second. Using this PR's version, the list loads nigh-instantly as the LazyLoader only sees matching entries.

Supersedes #4577, resolves #4576.

### DIFF
--- a/src/components/townView.html
+++ b/src/components/townView.html
@@ -5,40 +5,55 @@
         <div class="col no-gutters">
             <button type="button" class="btn btn-outline-info" style="float: right;"
                 data-bind="visible: player.town() instanceof DungeonTown && !QuestLineHelper.isQuestLineCompleted('Tutorial Quests'), tooltip: { title: 'Capture Pokémon to gain Dungeon Tokens in order to battle dungeons.', trigger: 'hover', placement:'left' }">?</button>
-            <div class="pageItemTitle" data-bind="using: { isDungeon: player.town() instanceof DungeonTown && player.town().dungeon }">
-              <div style="font-size: 1.8rem;" data-bind="text: player.town().name, css: { 'mx-5': $data.isDungeon }">Town Name</div>
-              <knockout data-bind="if: $data.isDungeon">
-                <div class="d-flex flex-column position-absolute justify-content-center ml-1" style="left: 0; top: 0; height: 100%;">
-                    <!--If all Pokémon in the dungeon are caught-->
-                    <div data-bind="if: (!DungeonRunner.dungeonCompleted(player.town().dungeon, true) && DungeonRunner.dungeonCompleted(player.town().dungeon, false))" style="font-size: 0;">
-                        <img title="You have captured all Pokémon in this dungeon!" class="pokeball-smallest ml-0" src="assets/images/pokeball/Pokeball.svg" />
-                    </div>
 
-                    <!--If all Pokémon in the dungeon are caught shiny-->
-                    <div data-bind="if: DungeonRunner.dungeonCompleted(player.town().dungeon, true)" style="font-size: 0;">
-                        <img title="You have captured all Pokémon shiny in this dungeon!" class="pokeball-smallest ml-0" src="assets/images/pokeball/Pokeball-shiny.svg"/>
-                    </div>
+            <div class="pageItemTitle d-flex justify-content-center" data-bind="let: { isDungeon: player.town() instanceof DungeonTown && player.town().dungeon }">
+                <div class="title-left" data-bind="if: isDungeon">
+                    <div class="d-flex align-items-center ml-1 mr-auto">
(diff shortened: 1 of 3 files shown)


CypherX:
This has been in crew testing for a few weeks with mixed results. Players with weaker machines say it feels better, but players with more powerful machines report that it feels worse.

There has been one big issue found (still happens after your recent changes) - in chromium based browsers after changing a filter several times then switching back to All (for instance I did it with one of the Type filters) it would load the entire pokemon list and slow the game to a crawl until the hatchery was closed, upon re-opening the hatchery the lag would resume and the game would become unresponsive again. Have not been able to replicate this in firefox.

CypherX:
> Re: feeling worse on more powerful computers, worse how? The menu loading slowly, lag with the menu open, lag with the menu closed? The changes I just made might help with those, especially the initial load time. Inevitably there will be some compromise between responsiveness and lag since I'm not aware of a way to forcibly skip the rateLimit delay.

Here's my experience (high end desktop, firefox)

Opening/closing hatchery - your version is slightly more smooth
Searching - a bit faster in your version
Filtering - this is where your version feels worse for me, filtering seems to take longer or at least has a lag time and appears unresponsive before anything happens in the pokemon list, whereas in the current version changing a filter has an immediate response. Filtering times also seem a bit quicker on the current version (opened hatchery then timed changing a type filter from All -> Fire)

I'll ask the people who tested and said it felt worse to test again with the latest changes and leave their comments here

CypherX:
Summary of comments from testers thus far:

- No significant difference between previous test and latest updates
- Filtering is slower than live (significantly slower for some people) while sorting seems to be a little faster
- The previously discussed lazyload/freezing issue is making it hard to test for some

CypherX:
Feedback from the latest changes is looking good, those who previously had issues are all saying those issues have been addressed and that it performs better

src/modules/utilities/LazyLoader.ts
```diff
@@ -1,24 +1,28 @@
...
+    let currentlyLoading = false;
 
     const loadMore = () => {
-        if (visible) {
+        if (visible && !currentlyLoading && !doneLoading.status) {
+            currentlyLoading = true;
             GameHelper.incrementObservable(page);
 
             // keep loading more in case we don't push the loader off screen
             // @ts-ignore
```
Aegyo, line 16: Unrelated to your changes, but I think this ts-ignore can be removed

Aegyo (approved):
No major problems I could find.

I did notice though that if you spam click the bottom of the scroll bar, the loader can get "stuck" - at least until you scroll back up a bit and then go down again. Seems to be because we don't get an intersection event if you make the loader show appear again so quickly after it got pushed down.

CypherX:
Everything seems to check out so I'm going to go ahead and merge this now.

Works great, thanks!

PR #4669, opened 2023-08-15
fix RoamerNPC dialogue stuck

DreamNya:
## Description

Fix an issue where RoamerNPC dialogue would stuck the game when the player had never encountered a roaming Pokémon before.

## Motivation and Context

When the value of  `App.game.statistics.pokemonSeen[roamer.pokemon.id]` changes, `dialogHTML()` of `RoamerNPC` will be triggered, and then the value will be changed repeatedly, falling into an infinite loop.
In order to break this loop, the code needs to add a restriction statement.

## How Has This Been Tested?

- Start a new game with the old code, then talk to Wendy(RoamerNPC), if you haven't met Mew before, the game will be stuck slightly and the devtools will throw some errors.
Here is a newgame savedata for test if you need
[[v0.10.14] PokeClicker 2023-08-15 06_31_42.txt](https://github.com/pokeclicker/pokeclicker/files/12341898/v0.10.14.PokeClicker.2023-08-15.06_31_42.txt)

- Start the Mewtwo event with the old code, and gain the clone Pikachu, then talk to Wendy or Jessie and James, the game will be stuck for a long time

## Screenshots (optional):

bug
bug2

## Types of changes

- Bug fix

PR #4677, opened 2023-08-17
Roamer NPC - Use roamer group name

CypherX:
## Description
Changes the Roamer NPC to use the roamer group name rather than region name when there are currently no roamers. This allows it display the correct area name in subregions, such as 'Orre' instead of 'Hoenn' and 'Alola - Magikarp Jump' instead of 'Alola'.

## Motivation and Context
Accurancy

## How Has This Been Tested?
Talked to the roamer NPC in Orre and saw it said Orre instead of Hoenn.

## Screenshots
image

## Types of changes
- Text

### DIFF
--- a/src/scripts/towns/RoamerNPC.ts
+++ b/src/scripts/towns/RoamerNPC.ts
@@ -17,7 +17,9 @@ class RoamerNPC extends NPC {
 
         // If no roaming Pokemon yet
         if (!roamers.length) {
-            return `There hasn't been any reports of roaming Pokémon around ${GameConstants.camelCaseToString(GameConstants.Region[this.region])} lately.`;
+            const regionName = RoamingPokemonList.roamerGroups[this.region]?.[this.subRegionRoamerGroup]?.name
+                ?? GameConstants.camelCaseToString(GameConstants.Region[this.region]);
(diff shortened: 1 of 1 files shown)


src/scripts/towns/RoamerNPC.ts
```diff
@@ -17,7 +17,9 @@ class RoamerNPC extends NPC {
 
         // If no roaming Pokemon yet
         if (!roamers.length) {
-            return `There hasn't been any reports of roaming Pokémon around ${GameConstants.camelCaseToString(GameConstants.Region[this.region])} lately.`;
+            const regionName = RoamingPokemonList.roamerGroups[this.region]?.[this.subRegionRoamerGroup]?.name
+                ?? GameConstants.camelCaseToString(GameConstants.Region[this.region]);
+            return `There hasn't been any reports of roaming Pokémon around ${regionName} lately.`;
```
osherwott, line 22:
```suggestion
            return `There haven't been any reports of roaming Pokémon around ${regionName} lately.`;
```
Might as well fix this while you're at it.

Farboo171 (approved):
Other than seeing #4669's bug show up, looks good. Thanks for the PR!

PR #4681, opened 2023-08-18
Add pokemon location function for Gift NPC rewards

CypherX:
## Description
Adds a location function to `PokemonHelper` to check if the pokemon is gained through a Gift NPC.

Eternatus was moved from a temp battle reward to a NPC gift to make it optional. This resulted in it no longer having any valid locations thus it was being excluded from the friend safari rotation.

## How Has This Been Tested?
See screenshots below.

## Screenshots (optional):
before:
image

after:
image


Farboo171 (approved):

PR #4683, opened 2023-08-18
Rebalance of contest store

Jaaslet:
## Description

Readded the contest store and balanced the prices.
The store now follows the requirement of the contest itself. Makes for a smoother release. And because of a PR i did recently, it will now be hidden as long as it has the development requirement.
The mons have increasing cost, based on how strong i think they are, without looking up anything.
Prices are based on the idea that you will get ~700 points per day. Note that there are plans for another contest in the future.
There are 3 mons with low enough cost that you can get one mon per day, the first 3 days. I feel like that gives players enough rewards quickly, that the rest can be a bit harder.
With 700 points per day, it will take around 10 days to get them all. So 90 days for EVs? I think that is fair, as it's something you can do each day, starting in hoenn.
Added random items, so you have something to buy, after you resist them all. I wanted to add a master ball, but did not feel like dealing with the math, right now.

Also added the last non-anime furfrou! It will take you like 20 days to get. So half a year for full EVs? 3 months when the bug contest comes out? Might be a bit high, but it is a furfrou afterall!

## Motivation and Context

Want to finish contest

## How Has This Been Tested?

Opened the game and saw that the mons where there

## Screenshots (optional):

image

## Types of changes

- New feature


Farboo171 (approved):
We can probably tweak the non-mon loot in the future, but this looks good for the mons! Thanks for the PR!

PR #4685, opened 2023-08-19
Fix Red sometimes not challenging you on update day

RegisCoaxans:
## Description
The update function now resets the tempbattle as well.

## Motivation and Context
Some players might not be able to encounter Red if they happen to play then update their game in the same day.

## How Has This Been Tested?
Loaded an affected save before and after me changing the update function. Looked for Red not appearing on first save, appearing on fixed save.

## Types of changes
- Bug fix


Farboo171 (approved):
Looks Red, thanks for the PR!

PR #4686, opened 2023-08-19
Anomaly Mewtwo trainer from Mega Mewtwo questline does not show up anymore

RegisCoaxans:
## Description
Provided support for technically trainers whose we do not want to show names and sprites in dungeon, and applied it to the two Anomaly Mewtwo in Pokémon Village.

## Motivation and Context
In that dungeon, Mewtwo was showing twice on the screen, as a trainer and as its own pokémon.

## How Has This Been Tested?
I verified only one Mewtwo was showing there. I also checked an actual trainer was still showing as expected.

## Types of changes
- Bug fix
- New feature
- UI improvement

### DIFF
--- a/src/components/dungeonView.html
+++ b/src/components/dungeonView.html
@@ -18,7 +18,7 @@
             </knockout>
 
             <knockout class="left">
-                <!-- ko if: DungeonBattle.trainer() -->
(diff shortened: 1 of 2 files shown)


Farboo171 (approved):

PR #4687, opened 2023-08-19
Updated Master Farmer Achievement

EatPant2nd:
## Description

Updated Master Farmer achievement with the new current max (69).

## Motivation and Context

Max changed from 68 to 69 so it was necessary to fix this achievement. Achievement percentage went up accordingly as well.

## How Has This Been Tested?

I opened the game, got the new berry, achievement procced correctly.

## Types of changes

- Bug fix


amative1:
Would this be better as a pointer to the berry list, rather than a hardcoded number that needs to be remembered to be changed?

EatPant2nd:
Honestly? Yes, that sounds like a wonderful idea. I don't know how to do that though. I can close this then and someone else can put the better fix.

CypherX:
Berries that are not available yet (Hopo) would need to be excluded from the count, somehow.

EatPant2nd:
Is there a way to make Hopo have a DevReq then? Because I can't think of anything.

amative1:
I think leaving this change in place, and then working on a dev req followed by a second (permanent) change to this bit of code would probably be the smoothest, at least fow now.

Farboo171 (approved):
Looks good enough for now, thanks for the PR!

Future improvements welcome as discussed.

PR #4692, opened 2023-08-20
Pokédex Shadow filters

UnicornSnuggler:
Added `Caught Not Shadow`, `Caught Shadow`, and `Caught Purified` as Pokédex filters. Also adjusted the Shadow and Purified SVG locations on the Pokédex entries to be slightly lower so there's not an overlap with the mega stone icons.


Farboo171 (approved):

PR #4694, opened 2023-08-20
Pokerus resistant notification sound

### LINKED ISSUE #4112: Pokérus notification sound [SUGGESTION]:
### Current Behavior

Currently, there is no sound effect playing when a pokémon becomes resistant to the pokérus.

### Improved Behavior

With a sound effect, people would know if one of their pokémon became resistant to the pokérus.

### Reason for change

When afk'ing, we can't know if one of our pokémon became resistant to the pokérus, as we most probably aren't looking at the screen, making us lose time that we could have spend resisting other pokémon. So, with a notification sound, we will be able to know that information.

### Detailed Description

When a pokémon becomes resistant to the pokérus, a sound effect will play. Players can deactivate the sound if they want to, just like every other sound effect.
I have recorded myself coughing, so a sound effect has been created.

https://user-images.githubusercontent.com/98485129/230795388-39edc909-c029-4638-a2d4-fbf14fd50ded.mp4

RealPendragon:
## Description
I added a sound queue to the notification that pops up whenever a Pokémon becomes resistant to pokerus.
_Note however that for the time being I used an already existant sound that I felt was most appropiate for this kind of notifcation, if need arises this might have to be swapped for another one._

**Edit:** I have now added a unique sound, the Toxic Sound from B/W, if that does not fit, or there are any other issues with it, tell me and I'll do something about it.

## Motivation and Context
Sometimes you want to idle on a route to get the last mon that appears on it resistant, but given that there's no sound cue for that, you would never know without checking the game. This solves this issue by adding a sound to the notification making it clear even when the game is not activly looked at.

Closes #4112  _4112 was seemingly deleted but I still thought it was a worthwhile inclusion._

## How Has This Been Tested?
1. Opened the settings to see if the new option is there, and made sure it was turned on.
2. Made a pokemon resistant to check if the sound played.
3. Turned the setting off, and repeated 2. to check if it no longer played.
4. Turned it back on, did step 2. a final time to make absolutly sure.
5. Caused a few other events that would play sounds to check if other sounds weren't affected-

## Types of changes
- Miscellaneous
- Sound

Find the strings for the translation files here:  [Translation PR](https://github.com/pokeclicker/pokeclicker-translations/pull/96)


RealPendragon:
> Rather than re-using the old sound, it's probably best to find a new sound.

And I fully agree with you there, however I didn't want to pick a sound without discussing it further, I will now look for a sound and if it's not to your liking it can always be changed.

Jaaslet:
Missing credits for the sound

RealPendragon:
> Missing credits for the sound

As I have mentioned the sound is the sound of "Toxic" from Pokémon Black/White, if that's an issue I'll find another one. Or are you referring to the person that extracted the sound? In that case I'll update it once I am back home.

Farboo171 (approved):
Sounds good, thanks for the PR!

PR #4695, opened 2023-08-20
Fix broken image URI

umbralOptimatum:


Farboo171:
I am confused

This is how it looks on dev:
image

Is this just a syntax change?

Farboo171 (approved):
Makes sense. Looks good, thanks for the PR!

PR #4699, opened 2023-08-20
Fix: Safari active environment variable not resetting

Crobat4:
## Description

`Safari.activeEnvironment()` doesn't reset when the Safari loads so if you finish the run on water and start a new run, the `Safari.activeEnvironment()` variable is set to water until you walk on land, this makes the player sprite changes to the surfing one while on land.

## Motivation and Context

Fix

## How Has This Been Tested?

* Finish Safari run on water and start a new one
* Walk 1 step and see the sprite changing is fixed

## Types of changes

- Bug fix


Farboo171 (approved):
Tested, and it fixes the issue. Thanks for the PR, and for the rapid response!

PR #4704, opened 2023-08-21
Developement requirement improvement

RegisCoaxans:
## Description
I improved DevelopmentRequirement so it works on top of another requirement instead of working next to it. Changed all pieces of code using it side by side with other requirements. Reworked the hint so it talks about developement if not in dev, talks about the actual requirement if in dev.
I made it so the current method still works, by the way.

## Motivation and Context
Orre update has shown that listing development requirement alongside normal requirements confuses players.
Also, this change makes it easier not to mess with requirements in dev version : no more multirequirement needed just to put a developement requirement on a content, for instance.

## How Has This Been Tested?
Messed with requirements in dev version.

## Types of changes
- Code ehancement
- Future bug fix prevention :^)


Farboo171 (approved):

PR #4709, opened 2023-08-21
Add shadow sprite toggle in Pokédex entry.

UnicornSnuggler:
## Description
I've added a toggle to the Pokédex entry modal that will allow users to view the shadow form of a relevant shadow Pokémon independently of their party default setting, akin to the shiny sprite toggle.

## Motivation and Context
As the code currently stands, when a user changes the default setting for the shadow form of a Pokémon, it also changes the sprite as viewed in the Pokédex entry modal. This is a deviation from the setting for the shiny form, which only changes the form in the list view *(and elsewhere)* while leaving the sprite in the entry modal as-is. This is a small update that seeks to unify the behavior of the setting and add a similar toggle specific to the entry modal for the sake of consistency.

## How Has This Been Tested?
I have validated the following scenarios:
* Toggling the shiny sprite view on and off while the shadow sprite view is on
* Toggling the shiny sprite view on and off while the shadow sprite view is off
* Toggling the shadow setting on and off while the shadow sprite view is on
* Toggling the shadow setting on and off while the shadow sprite view is off

## Other Notes
I noticed during the testing of these changes that the shiny setting doesn't have any impact on the sprite that appears on the little Pokédex cards while querying the Pokédex. I assume this is due to the `Hide shiny images` setting within the Pokédex modal, but it seems like an oversight to have the shiny setting for a given Pokémon ignored for these results. This also creates a behavior discrepancy between the shadow setting and the shiny setting. It would be easy enough to remedy this, but I figured I would bring it up first to discuss whether this is intended or not before making any modifications in that department.

## Screenshots
image
image
image

## Types of Changes
UI improvement

### DIFF
--- a/src/components/pokemonStatisticsModal.html
+++ b/src/components/pokemonStatisticsModal.html
@@ -9,6 +9,11 @@
         <h5 class="modal-title text-light ml-1" data-bind="if: App.game.party.getPokemon(pokemonMap[$data].id)">
           <img width="18px" src=""
             data-bind="attr: { src: `assets/images/pokeball/Pokeball${App.game.party.alreadyCaughtPokemon(pokemonMap[$data].id, true) && PokedexHelper.toggleStatisticShiny() ? '-shiny' : ''}.svg`}, click: () => PokedexHelper.toggleStatisticShiny(!PokedexHelper.toggleStatisticShiny())" />
+          
+            <!-- ko if: App.game.party.getPokemon($data)?.shadow > GameConstants.ShadowStatus.None -->
+              <img width="18px" src=""
+                data-bind="attr: { src: `assets/images/status/${PokedexHelper.toggleStatisticShadow() ? 'shadow' : 'purified'}.svg`}, click: () => PokedexHelper.toggleStatisticShadow(!PokedexHelper.toggleStatisticShadow())" />
+            <!-- /ko -->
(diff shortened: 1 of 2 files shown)


UnicornSnuggler:
To elaborate on what I've written beneath the `Other Notes` header...

Here is the default view for Ledian in the Pokédex with both of the form settings untoggled *(meaning that Ledian should show up as shiny and purified)*.
image

Here is what it looks like when the shadow setting is toggled on *(meaning that Ledian should show up as shiny and shadowy)*.
image

Here is what it looks like when the shiny setting is toggled on and the shadow setting is toggled off *(meaning that Ledian should show up as non-shiny and purified)*.
image

Notice the card in the Pokédex modal in the background in all three images. It changes for the shadow setting, but not for the shiny setting, defaulting to the shiny sprite regardless of the setting.

UnicornSnuggler:
If this were a business application, I'd argue that there's value in allowing the client to view various forms of the Pokémon independently of their rendered defaults in all other views...

...but it's not a business application and I'm inclined to agree with you here 😂 As proud of myself for these changes as I am, I can't deny that it's a little convoluted.

Farboo171:
I too am ok with this specific change, but will put it to an internal vote. Having two toggle areas does feel like bad design

PR #4710, opened 2023-08-22
Limit Pokedex region filter to players highest region

CypherX:
## Description
The pokedex region filter was displaying all main regions including future regions (Hisui, Paldea). This caused confusion with players asking how to access said regions. The filter will now display up to the players highest reached region, like every other region filter in the game.

## Motivation and Context
Displaying inaccessible things causes confusion.

## How Has This Been Tested?
Made a save with Paldea selected. Loaded it into my changed version. Checked pokedex, region filter was set to All and only listed up to Galar (my highest region).

## Types of changes
- UI improvement


Farboo171 (approved):
Looks filtered, thanks for the PR!

PR #4714, opened 2023-08-22
Safari code quality improvements

### LINKED ISSUE #4618: [Bug]: Safari Zone UI issues with small maps
### Confirmations

### What happened?

Reducing the window size to generate a small Safari Zone map and afterwards resizing the window produces several issues:
- When the window expands vertically, the modal does not change size, ending awkwardly at the bottom edge of the map
- When the window expands horizontally, the modal expands normally but adds empty grass on the right to fill the space, which appears to be part of the map but is unreachable
- When the window shrinks horizontally, the tile images compress and the player sprite is displaced

I expect the modal to resize along with the window, not size itself to the current map. If the map is smaller than the modal's current size, the map should be centered both vertically and horizontally without fake tiles. If the map is larger than the modal's current size, it should become scrollable horizontally rather than distort the map display.

### Steps to Reproduce

1. Resize game window
2. Enter safari zone
3. Resize window again

_No response_

Firefox, Chrome, Desktop Client

### Relevant log output

_No response_

### Additional Information

Small safari map:

Screen Shot 2023-08-03 at 11 00 28 PM

The same safari map in a larger window, with fake tiles on the right, awkwardly hovering at the top of the window:

Screen Shot 2023-08-03 at 11 00 50 PM

The same safari map with distorted tiles and displaced sprite:

Screen Shot 2023-08-03 at 11 01 15 PM

umbralOptimatum:
- Converted the rest of the SafariBattle animation speed magic numbers to constants
- Fixed inconsistent catch animation speeds where catching would sometimes be processed too early
- Fixed minor graphical issue with the rock throwing animation
- Added surfable water tiles to `GameConstants.SAFARI_LEGAL_WALK_BLOCKS`
- Added extra defensive checks
- Fixed `SafariBattle.delay()` not passing results through
- Fully converted SafariBattle methods to Promises for consistency and readability
- Made Safari static methods consistently use the class name instead of `this`

Tested with a lot of Safari exploration. Safari appears to run identically to before.

Fixes #4618 (the important parts, at least).


Farboo171:
Testing has shown an unexpected delay after a successful catch animation, could that be caused by ` SafariBattle.delayRoll`, or something associated?

CypherX:
Couple things I noticed while testing

- Items spawn don't appear to be working?
- At higher safari levels the 'pokemon has been caught' text displays before the catching animation finishes. It's slightly noticeable at the level 20-30 speeds and very noticeable at the level 30-40+ speeds

CypherX (approved):
Apparently the dock modal waves background image is one of the safari water tiles, who knew? Anyway, fixed that and the rest appears to be working as expected so I'll go ahead and merge this

PR #4720, opened 2023-08-23
Change gen 7 teacher filename to include subclass

osherwott:
## Description
The filename "Teacher-gen7" was incompatible with trainer code due to having a dash instead of a space. It has been changed to include a space and the Trainers' School teacher boss has been swapped out for it.

## Motivation and Context
We have a teacher sprite for gen 7 but it's only used as an npc and not in the dungeon.

## How Has This Been Tested?
See below:

## Screenshots:
image
image

## Types of changes
- Sprite


Farboo171 (approved):

PR #4723, opened 2023-08-24
Hoopa questline typo-fix

amative1:
Just a quick tweak to grammar (multiple instances of "it's" which is always a contraction of "it is", rather than the possessive that was used in the dialog)


Farboo171 (approved):
It's good, thanks for the PR

PR #4725, opened 2023-08-25
Improve Friend Safari pokemon list generation

CypherX:
## Description
Generating the friend safari pokemon list can take up to several seconds. This causes the game to take longer to load and a lag spike when the list is re-generated at midnight.

This changes the list to be generated asynchronously and to yield back to the main thread every 25ms, both decreasing initial load time and eliminating the lag spike when the list would be re-generated. A minimal side effect is the list is not fully generated until a second or two after the game is loaded.

One person I had test saw their initial load time go from 13s to 7s.

## Motivation and Context
Optimization. Faster load time. No lag spikes.

## How Has This Been Tested?
Loaded the game and verified the friend safari was populated correctly.
Changed my time and saw that the list was re-generated at midnight (with no lag).
Asked several players to test and all reported improvements.

## Types of changes
- Optimization


Farboo171:
This seems to have broken the SafariRanger NPC
image

CypherX:
How soon after loading did you talk to the NPC? There can be a slight delay after loading before the friend safari list is fully generated which will cause the NPC to show filler 'mon. I don't have any better ideas for it at the moment

Farboo171:
Yeah, checking the ranger right on load shows the filler mons, and waiting a minute shows the correct mons.

CypherX:
Idea:

image

wormania:
The underlying issue to the bad performance here is that `getPokemonLocations` is designed with fetching a single poke in mind, but is being run against every poke in succession, pushing the operation to effectively n² complexity. 
Rather than trawl every dungeon*, town, route, etc. for each input, the operation should be flipped and each location processed once building up a map of Id : Locations, which `isObtainableAndNotEvable` can then use.
I can help with the rewrite if you want to go this route

\* the dungeon list is iterated through _four_ times

CypherX:
Going to close this, improving the performance of the location functions is better than my lazier approach.

PR #4728, opened 2023-08-25
Fix pokedex entry column stretching

CypherX:
## Description
The 'Show party pokemon shiny' and 'Show party pokemon shadow' text was causing the table to stretch and overflow into the second column on some themes

## How Has This Been Tested?
With my eyes

## Screenshots (optional):
before:
image

after:
image

## Types of changes
- UI improvement


Farboo171 (approved):

PR #4733, opened 2023-08-27
Old man now shows up after red is defeated

Jaaslet:
## Description

Something i forgot, when making Red daily!
When Red is defeated, an old man shows up and explains he will be back. Also talks about the light ball, in case anyone misses it.
The old man will go away again, when Red comes back (he is hiding from him, i guess?)
I don't like my own writing, so if anyone can help me make it sound more human, please do so! Or atleast tell me it sounds human...

## Motivation and Context

Less confusion!

## How Has This Been Tested?

Opened the game. Red was there, but no old man. Defeated Red. Old man was there. Timetraveled and saw the old man ran and red came back

## Types of changes

- UI improvement


RealPendragon:
Perhaps this would sound a bit better:
```
'This young man comes here every day.',
'He can't take loosing very well, poor boy. But he's always back here the next day!',
'But even though he hates losing, he will always give his opponent a shiny ball, as a prize for winning.',
```
Man, because I feel like 'fellow' is a word an old man wouldn't use, the other's, are mainly just things I think sound slightly better. But I am not great at this either, so do with my opinion what you will.

Farboo171 (approved):
Sounds like a nice fellow, thanks for the PR!

PR #4734, opened 2023-08-27
Magikarp jump gymleaders now show up on defeated

Jaaslet:
## Description

When creating magikarp jump, i did so we could overwrite the image path for the gymleaders, as many of the magikarp jump gymleaders have the same image. In this change, i forgot the receiveBadgeModal. This PR fixes that.

## Motivation and Context

Fixing a bug i noticed

## How Has This Been Tested?

Loaded my game, removed the friend league badge using console. Defeated the league and saw that the trainer now shows up in the modal

## Types of changes

- Bug fix


Farboo171:
Will this also fix Giovanni? If so, awesome. If not, that's another easy PR.

Jaaslet:
> Will this also fix Giovanni? If so, awesome. If not, that's another easy PR.

Most likely. Which Giovanni is bugged?

Farboo171 (approved):
Tested with the Gio fight, fixes that too as expected

PR #4735, opened 2023-08-27
Clarified Dungeon Boss hints

osherwott:
Dungeon Boss hints now refer to the requirements, like the wiki does.
Credit to Crobat4 for referring to the `$data.lockMessage`.

## Description
- Update hints from being generic for all bosses to revealing the relevant requirements.
- Unhide mandatory Pokemon as of the recommended Pokedex Challenge. This includes:
  - Ho-oh, Lugia, and Celebi
  - Zacian, Zamazenta, and Calyrex
- Hide Pokemon that
  - the player unlocks through a forced questline or required dungeon clear for progression
  - are alternate forms
  - have an Obtained Pokemon requirement of themselves
  - are locked in a dungeon that isn't in their native region
  - are locked as a dungeon boss in their native region but are found elsewhere in the region
- Added Pikachu Valley NPCs to keep some kind of hint for Cap Pikachus. 
- Added requirements to bosses whose questlines are unlocked after the E4, or are otherwise not unlocked through the natural story.
- Cleaned up redundant requirements for some bosses:
  - Giratina (Altered) now refers to the quest step instead of the temp battle of that quest
  - Kyurem now refers to the quest instead of the temp battle battle that ends it
  - Volcanion doesn't have a questline requirement anymore because the Champion requirement supersedes it
- Added requirements to Necrozma both to reference the SM obtain method and to keep the UBs hidden while still telling the player which questline is needed to progress. If they weren't hidden they would overwhelm the player because they appear in many and multiple dungeons.
- Cleaned up the requirements for Minior pladeholder bosses in Lake of the Sunne and Moone to be opposite of the actual bosses.
- Also cleaned up some unneeded `hide: true` for regular dungeon trainers.

## Motivation and Context
Some Pokemon that are mandatory are locked behind questlines. This made it so that the current hint of progressing through the story was incorrect. To fix this, some bosses were hidden, but that meant we were now hiding necessary information from the player. This change makes it so that lock hints can be more helpful with our ever-evolving game complexity.

## How Has This Been Tested?
I gave some bosses new requirements for testing purposes and checked that they unlocked when completed.

## Types of changes
- UI improvement


Qwertypop04:
How is the player supposed to encounter Necrozma before it is unlocked?

osherwott:
Ah! I forgot the npc! Looker is supposed to tell you to go check out Ten Carat Hill! Will fix that right away!

RegisCoaxans:
Not a fan of redundant requirements.
Looking at Ho-oh, the player may search for the Beasts separately from getting the required quest.
If the hint is not clear enough with the quest name and its step. I would rather make the quest line appear locked in the Bulletin Board with its requirements hinted.
Also, multi requirements can be confusing if there is too much information to read, and this is even more confusing when stuffs are redundant.

CypherX:
> Not a fan of redundant requirements. Looking at Ho-oh, the player may search for the Beasts separately from getting the required quest. If the hint is not clear enough with the quest name and its step. I would rather make the quest line appear locked in the Bulletin Board with its requirements hinted. Also, multi requirements can be confusing if there is too much information to read, and this is even more confusing when stuffs are redundant.

Adding onto this, it also makes hints on the wiki kind of messy:

before:
image

after:
image

osherwott:
Fixed it. Thanks for checking! The last thing I want is a messy wiki!

Also locked Bulletin Board quests would definitely be better, yeah.

src/scripts/dungeons/Dungeon.ts
```diff
@@ -8453,11 +8453,11 @@ dungeonList['Giant Chasm'] = new Dungeon('Giant Chasm',
...
-        new DungeonBossPokemon('Tangrowth', 30000000, 100, {requirement: new TemporaryBattleRequirement('Ghetsis 2')}),
-        new DungeonBossPokemon('Audino', 32000000, 100, {requirement: new TemporaryBattleRequirement('Ghetsis 2')}),
-        new DungeonBossPokemon('Mamoswine', 32000000, 100, {requirement: new TemporaryBattleRequirement('Ghetsis 2')}),
+        new DungeonBossPokemon('Tangrowth', 30000000, 100, {hide: true, requirement: new TemporaryBattleRequirement('Ghetsis 2')}),
+        new DungeonBossPokemon('Audino', 32000000, 100, {hide: true, requirement: new TemporaryBattleRequirement('Ghetsis 2')}),
+        new DungeonBossPokemon('Mamoswine', 32000000, 100, {hide: true, requirement: new TemporaryBattleRequirement('Ghetsis 2')}),
         new DungeonBossPokemon('Kyurem', 35000000, 100, {requirement: new MultiRequirement([
-            new TemporaryBattleRequirement('Ghetsis 2'),
+            new QuestLineCompletedRequirement('Hollow Truth and Ideals'),
             new GymBadgeRequirement(BadgeEnums.Elite_UnovaChampion),
```
RegisCoaxans, line 8461: This multirequirement is redundant : The league is locked behing finishing Hollow Truth and Ideals.

src/scripts/dungeons/Dungeon.ts
```diff
@@ -9604,7 +9604,6 @@ dungeonList['Kalos Power Plant'] = new Dungeon('Kalos Power Plant',
                 requirement: new MultiRequirement([
                     new ClearDungeonRequirement(5, GameConstants.getDungeonIndex('Kalos Power Plant')),
                     new GymBadgeRequirement(BadgeEnums.Elite_KalosChampion),
-                    new QuestLineStepCompletedRequirement('A Beautiful World', 11),
```
RegisCoaxans, line 9607: A Beautiful World 11 requires the player to beat some Team Flare folks in here. Removing this makes old save able to progress the quest from defeating Volcanion instead. The quest should be turned into DefeatDungeonBoss before removing the requirement.

src/scripts/dungeons/Dungeon.ts
osherwott: This was included for cases for returning players who haven't engaged with Unova's questline yet but have beaten the league, and because it functions as Kyurem's associated questline, making it relevant information.

src/scripts/dungeons/Dungeon.ts
osherwott: I have commented a TODO in the associated questline step to turn it into a dungeon boss quest for when the quest refactor is merged, because the story questline isn't supposed to be relevant to the unlock requirement of Volcanion, and shouldn't be shown as a hint. For clarity's sake I will add another note in this file too. 👍

Farboo171 (approved):
Thanks everyone for testing and giving feedback!

PR #4736, opened 2023-08-27
Breeding Efficiency attack bonus toggle & include additional bonuses

CypherX:
## Description
Adds a setting (defaults to true) to control whether attack bonuses are included in the Breeding Efficiency calculation. Also updated the BE calculation to include the bonuses from held items and the shadow/purified status, if EVs are included then all attack modifiers should be for consistency. Vitamins are always included as they directly affect the attack gained through breeding rather than being an attack modifier.

## Motivation and Context
After the previous change to always include EVs in BE and removal of the toggle on the vitamin modal, many players expressed the desire to be able to toggle between the "base" BE and BE with bonuses included.

## How Has This Been Tested?
Checked the BE of various pokemon while changing the modifiers and toggling the setting.

## Screenshots (optional):
image

## Types of changes
- New feature


Farboo171 (approved):

PR #4739, opened 2023-08-28
Pokeball filter Category option

CypherX:
## Description
Adds a Category option to pokeball filters. To get this to work I had to change `Setting` to allow a `computed` options list.

Trying to delete a category that is currently selected by a filter will display a warning indicating that it must first be deselected in the filter.

## Motivation and Context
More filter customization.

## How Has This Been Tested?
Enabled the Category option for existing and new filters and tested that they worked as expected.
Attempted to remove a category selected in a filter - displayed the warning. Disabled the category option in the filter then again tried to delete the category - it was deleted.
Re-arranged categories and verified the order was updated in the filter dropdown.

## Screenshots (optional):
image

## Types of changes
- New pokeball filter type


Farboo171 (approved):
Tested, looks good! Thanks for the PR!

PR #4740, opened 2023-08-28
Added "Hoenn" to Orre Roamer Group

EatPant2nd:
## Description

Orre, unlike Magikarp Jump that has "Alola", didn't have "Hoenn" which made it look inconsistent. So, I added the corresponding main region.

## Motivation and Context

Roamer group names are used in the Roaming Pokémon page of the Wiki. Adding the "Hoenn" part to Orre helps people know where it's accessed from.

## How Has This Been Tested?

Didn't test but it's just a string.

## Types of changes

- Wiki


Farboo171 (approved):

PR #4742, opened 2023-08-29
Scents for Purified Pokémon

RealPendragon:
## Description
Scent Items can now be held by Shadow and Purified Mons

## Motivation and Context
Only having them work on Shadow Mons makes the items pretty much useless, and that is not good.

## How Has This Been Tested?
Loaded up one of my Saves, went to my Items and looked if my Purified Pokémon showed up when selecting Scents.


Jaaslet:
> Only having them work on Shadow Mons makes the items pretty much useless, and that is not good.

The idea is to make them be a purify item. The current design is intentional

RealPendragon:
> > Only having them work on Shadow Mons makes the items pretty much useless, and that is not good.
> 
> The idea is to make them be a purify item. The current design is intentional

Eugene said this would be a good simple change, that's why I made it. The message can be found here: [Link](https://discord.com/channels/450412847017754644/698525560640503878/1145867391222431805)

Farboo171:
Yeah, the current canon application is extremely boring so I wanted to change it to make the scents be actually useful in theory.

Farboo171 (approved):
Smells good, thanks for the PR!

PR #4744, opened 2023-08-29
Remove cutscene Cipher Peon Scrub

Qwertypop04:
## Description
Removes the duplicate of Cipher Peon Scrub from Relic Cave, as it is from an NPC on NPC cutscene fight

## Motivation and Context
Fights the player is not involved in should not be present

## How Has This Been Tested?
Hasn't been tested, but it just removes a trainer, should be fine

## Types of changes
- Bug fix


Farboo171:
Are there any filler fights that could be added to make this  trainer pool less sparse?

Qwertypop04:
Yes. There are a few more that could be added in XD too, if you'd like

Farboo171 (approved):
Looks good! There's a few changes that need to be done on XD content, we can include those in future PRs.

PR #4747, opened 2023-08-30
Added dream orbs to PokemonHelper.getPokemonLocations

PixLSteam:
## Description
Adds Dream Orbs to `PokemonHelper.getPokemonLocations`.

## Motivation and Context
This could be used to show the Dream Orb obtain method on a Pokémon's wiki page.

## How Has This Been Tested?
Checked the result of `PokemonHelper.getPokemonLocations("Shuckle")[PokemonLocationType.DreamOrb]` on a local dev instance.

## Screenshots (optional):
image

## Types of changes

- Wiki improvement

### DIFF
--- a/src/scripts/pokemons/PokemonHelper.ts
+++ b/src/scripts/pokemons/PokemonHelper.ts
@@ -531,7 +531,8 @@ class PokemonHelper extends TmpPokemonHelper {
             locations[PokemonLocationType.Safari] ||
             locations[PokemonLocationType.Shop] ||
             locations[PokemonLocationType.Wandering] ||
-            locations[PokemonLocationType.Trade];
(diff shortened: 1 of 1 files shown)


Farboo171:
This seems to change the Friend Safari list in some way. Is that expected?

Not a show-stopper since other changes will cause a FS list shuffle next update, just wanted to check first.

PixLSteam:
> This seems to change the Friend Safari list in some way. Is that expected?
> 
> Not a show-stopper since other changes will cause a FS list shuffle next update, just wanted to check first.

Good catch. This would've allowed the Therians to show up in there despite them being EVable. I had to adjust `isObtainableAndNotEvable`.

Farboo171 (approved):

PR #4758, opened 2023-09-01
Performance Fix: Friend Safari generation

wormania:
## Description
Added memoisation to `PokemonHelper` functions called by `GetPokemonLocations`, to avoid massive duplication of work when building the list of Friend Safari Pokemon on game load / midnight reset. Nearly a 100x speedup
Improved the string matching in `getPokemonTempBattleReward` so that it no longer returns that you can get Charizard and other last-stage starters from megastone traders. No gameplay change, but fixes incorrect data on the bot/wiki.
Generally made the `PokemonHelper` functions lie slightly less about what types they are handling/returning

## Motivation and Context

The current game takes multiple seconds to generate the Friend Safari list. This is done on game load and on midnight reset, causing the game to load slower than it should, and to lock up for a few seconds at midnight.
In addition, the current way of building the list will scale poorly as more Pokemon, dungeons, and methods of obtaining Pokemon are added.

## How Has This Been Tested?

Performance testing:
```
develop:
generateKalosSafariList took 3688.9 milliseconds.
generateKalosSafariList took 3520.4 milliseconds.
generateKalosSafariList took 3596.2 milliseconds.
generateKalosSafariList took 3523.0 milliseconds.
generateKalosSafariList took 3685.4 milliseconds.

PR:
generateKalosSafariList took 36.9 milliseconds.
generateKalosSafariList took 57.2 milliseconds.
generateKalosSafariList took 40.3 milliseconds.
generateKalosSafariList took 45.9 milliseconds.
generateKalosSafariList took 45.9 milliseconds.
```

Output of the Kalos friend safari (`SafariPokemonList.list[GameConstants.Region.kalos]()`) was confirmed to be the same before and after the changes in this PR
Output of `GetPokemonLocations` is the same except for the removal of the erroneous Mega-evolutions.

## Types of changes
- Performance improvement
- Bug fix


CypherX:
Looks good for the most part from my tests. Just a couple things I noticed:

In some cases a pokemon can be obtained in the same area under different circumstances, such as appearing on the same route during different weather or different trainers in a dungeon having the same shadow pokemon. These areas are now being duplicated in the results whereas before it would only return the location once.

Examples below from comparison between output from `copy(JSON.stringify(pokemonList.map(p => ({ pokemon: p.name, methods: PokemonHelper.getPokemonLocations(p.name) }))))`

Poliwrath is available as a shadow from two different trainers in this dungeon
image

On some Magikarp Jump routes (33 - 40) Magikarp is listed multiple times as way to weight it higher, each route is being returned multiple times for each time it's in the route data. For example, route 35 is listed 3 times in the results.

Base wanderers that are attracted to all berries are returning every berry rather than just 'Always'
image

Also, I was checking friend safari generation and the pokemon list is the same between current dev and your PR, but for some reason (and it may not even be related to this PR) I'm getting a different shuffle result which doesn't make much sense seeing as the list and seed are the same. Not sure if it's even worth looking into, the rotation will change anyway with the update.

wormania:
Duplicate routes/shadow pokemon/too many wanderer berries have been fixed, the only difference to develop now is that Gigantamax Pokémon Bosses show both the `QuestLineStepCompletedRequirement` and the `CaughtRequirement` versions

As for the ordering - I can't see any difference, the `shuffledPokemon` and the `SafariPokemonList.list[GameConstants.Region.kalos]` show the same between dev and PR for me

Farboo171 (approved):

PR #4761, opened 2023-09-02
Fix ExternalHelper+add dungeons

Jaaslet:
## Description

I finally got around to using my ExternalHelper for the wiki! I just found out that it did not work...
This PR fixes this.
Also adds support for dungeons.
And the plan is to finish the pokemon method (which is why it's still a draft)

## Motivation and Context

Want the wiki to be less confusing

## How Has This Been Tested?

Ran the method using the console

## Types of changes

- Bug fix

### DIFF
--- a/src/scripts/ExternalHelper.ts
+++ b/src/scripts/ExternalHelper.ts
@@ -5,42 +5,45 @@ class ExternalHelper {
     private static routeCache: {[name: string] : boolean} = {};
     private static gymCache: {[name: string] : boolean} = {};
 
-    public static isInLiveVersion(content: Town | QuestLine | TemporaryBattle | RegionRoute | Gym) {
+    public static isInDevelopment(content: Town | Dungeon | QuestLine | TemporaryBattle | RegionRoute | Gym) {
         if (content instanceof Town) {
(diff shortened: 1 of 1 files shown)


src/scripts/ExternalHelper.ts
```diff
@@ -100,17 +103,17 @@ class ExternalHelper {
...
 
-    public static pokemonIsInLiveVersion(name: PokemonNameType) {
+    public static pokemonIsInDevelopment(name: PokemonNameType) {
         if (PokemonHelper.calcNativeRegion(name) > GameConstants.MAX_AVAILABLE_REGION) {
-            return false;
+            return true;
         }
 
         const locations = PokemonHelper.getPokemonLocations(name);
         if (!locations) {
```
RegisCoaxans, line 112:
```suggestion
        if (!Object.keys(locations).length) {
```
`PokemonHelper.getPokemonLocations` always returns an object.
However, it will also return `true` Pokémon whose locations are not processed in the function yet (which should only happen to Alcremie, Milcery (Cheesy) and Exeggcute (Single), now).

PR #4766, opened 2023-09-02
Implements pokémon location finder for Safari Item and Battle Café

RegisCoaxans:
## Description
Added functions to locate and explicit obtaining for Safari Item and Battle Café Pokémon.

## Motivation and Context
Better Wiki.
Also, #4761 makes it required as it would treat Pokémon from those as dev contents.

## How Has This Been Tested?
Console injections to confirm the affected Pokémon behave as expected.

## Types of changes
- New feature


Farboo171 (approved):
Thanks, all on Discord for helping review.

PR #4768, opened 2023-09-03
Add Blu Coin to Blu Coin code

Farboo171:
## Description
Adds Dungeon Token reward to the Tutorial Skip code, to assist challenge runs, speed runs, and let people get a little boost on their file (similar to Farming Boost)

Also adds an NPC at the end of Kanto giving the code in-game.

## Motivation and Context
This code will allow for a quicker second playthrough for people if they play by in-game information only, and will allow some more monotypes and challenge runs to be possible.

EDIT: The amount of DT is very small, not enough to impact balance, but about how much a person might gain from the tutorial.

## How Has This Been Tested?
Tried the code at a usable time, and a non suable time. Both worked as intended.

## Types of changes

- New feature


src/modules/codes/RedeemableCodes.ts
```diff
@@ -200,6 +200,14 @@ export default class RedeemableCodes implements Saveable {
...
                     quest.curQuestObject().complete();
+                    App.game.wallet.gainDungeonTokens(15000);
+                    Notifier.notify({
+                        title: 'Tutorial Skip',
+                        message: 'You have skipped the tutorial, and found a stash of Dungeon Tokens.',
+                        type: NotificationConstants.NotificationOption.warning,
+                        timeout: 1e4,
+                    });
+                    return false;
                 }
```
CypherX, line 211: The `gainDungeonTokens` and notification should be after the while loop, not within it. Also don't need the `return false;`

```suggestion
                while (quest.curQuest() < quest.totalQuests) {
                    quest.curQuestObject().complete();
                }
                App.game.wallet.gainDungeonTokens(15000);
                Notifier.notify({
                    title: 'Tutorial Skip',
                    message: 'You have skipped the tutorial, and found a stash of Dungeon Tokens.',
                    type: NotificationConstants.NotificationOption.warning,
                    timeout: 1e4,
                });
```

RedSparr0w (approved):

PR #4770, opened 2023-09-03
Add statistic tracking contest Master results by style

CypherX:
## Description
Adds a statistic to track the amount of contest Master results by style. Also moves the ContestStyle enum to GameConstants.

## Motivation and Context
eugene wants it

## How Has This Been Tested?
Achieved master in several contests with different styles and verified the statistic was incrementing correctly. Refreshed and checked that the values persisted.

## Types of changes
- Statistics


Farboo171 (approved):
Looks good, thanks for the favor and the PR!

PR #4772, opened 2023-09-04
Fix bad request from Notifier

DreamNya:
## Description

Remove unnecessary default values in `Notifier.notify`, to prevent bad request.

## Motivation and Context

The old Notifier may open an bad request every time caused by the default value `image = null`.
If the `icon` option in `Notification` is `null`, the function may open a request with a path of `null`, then the website will response `404 (Not Found)`. While `undefined` won't.
And in this function the default value of `null` has no meaning, `undefined` is enough. So I remove all the ` = null`.

## How Has This Been Tested?

See screenshots.

## Screenshots (optional):

image

## Types of changes

- Bug fix


Farboo171 (approved):

PR #4774, opened 2023-09-04
Fix quest line pause button tooltip not working in client

CypherX:
## Description
This fixes the tooltip on the quest line pause button not working on the client when the button is disabled

## Motivation and Context
Bug

## How Has This Been Tested?
Checked that the tooltip now appears on the client when the pause button is disabled

## Types of changes
- Bug fix


Farboo171 (approved):

PR #4776, opened 2023-09-04
Fixed SeededDateSelectNRequirement

RegisCoaxans:
## Description
Fixed the requirement so raids are randomized as expected.

## Motivation and Context
Only the ten first raids were showing, no matter the shuffle. The reason was that `in` keyword uses keys, not values, and the keys of the 10-element suffled array always were numbers from 0 to 9.

## How Has This Been Tested?
I timetravelled and saw available raids change each day.

## Types of changes
- Bug fix


Farboo171 (approved):
Looks randomized, thanks for the PR!

PR #4778, opened 2023-09-04
Added Shadow Modifier to the FAQ

EatPant2nd:
## Description

Added Shadow Modifier to the FAQ in the EVs section since it was missing. Made Base Wanderer Yield say 0EV instead of 0xEV since it looked wrong when compared to the other values. Finally, put values with the same Yield / Modifier on the same line to make the list take less space.

## Motivation and Context

The other modifiers were there so an update seemed logical.

## How Has This Been Tested?

Opened the game, checked the FAQ, it shows fine.

## Types of changes

- FAQ update


Farboo171 (approved):

PR #4786, opened 2023-09-07
Changed one of the Marathon Runner Achievements' Name

EatPant2nd:
## Description

"Marathon Runner" was being used for both the "Hatch 100 eggs" achievement and the "Walk 10,000 steps in the Safari" one which caused issues when people selected one to track. Made the second one a sequel. Also fixed a typo in one of the quest steps from An Unrivaled Power.

## Motivation and Context

The 2 achievements having the same name was causing issues.

## How Has This Been Tested?

Opened the achievement tracker, checked both Marathon Runners.

## Types of changes

- Bug fix


Farboo171 (approved):

2.

PR #4788, opened 2023-09-08
Disable console i18next debug logs in Live & DevelopmentRequirement tweaks

CypherX:
## Description
- Disabled the i18next (translation) console debug logs in Live version. They will continue to appear in development builds
- Moved the code that determines if the game is a development build to a `GameHelper` function so it can be used outside `DevelopmentRequirement`
- Tweaked `DevelopmentRequirement` to be a singleton of sorts so each "normal" use of `new DevelopmentRequirement()` is the same instance rather than instantiating and keeping in memory 155 (at the time of this PR) separate objects

## How Has This Been Tested?
Loaded the game with development mode both enabled and disabled and verified everything still worked as expected.

## Screenshots (optional):
i18next console logs that will be hidden in Live:
image
image


Farboo171 (approved):
Works great! Thanks for the PR!

PR #4790, opened 2023-09-08
Hidden dungeon trainers get their name back

RegisCoaxans:
## Description
Reenabled trainer names in dungeon when the trainers are meant to be hidden.
Tempbattles with hidden trainers still show their names, so dungeon battles should as well.

## Motivation and Context
Fixes the thing I made in #4686.

## How Has This Been Tested?
Genocided every species in Pokémon Village until that Mewtwo showed up.

## Types of changes
- Bug fix


Farboo171 (approved):

PR #4794, opened 2023-09-09
Max Raid requirements and updated map sprites

Qwertypop04:
## Description
Added requirements to each Max Raid such that the route they are placed on must be completed and their return town unlocked before they can be fought. Changed Max Raid map sprites, making them larger, and adding a purple beam when activated

## Motivation and Context
Players shouldn't be able to fight max raids if they haven't reached the route they are on and unlocked the location they'll be sent to upon the fight ending. Map sprites changed to make them more visible

## How Has This Been Tested?
Runs, requirements untested

## Types of changes
- Map
- Gigantamax


Farboo171 (approved):
I didn't run a full test, but all the spot checks look good. Crew can hammer it out fully. Looks good, thanks for the PR!

PR #4797, opened 2023-09-10
Update mimic chest image in encounter list to reflect tier

CypherX:
## Description
The chest image used to indicate a mimic in the dungeon encounter list was using the old chest image. This updates it to use the new tier images and also adds a tooltip with the tier name.
## How Has This Been Tested?
With my eyes.
## Screenshots (optional):
image
## Types of changes
- UI improvement


Jaaslet:
I don't think the names of the tiers are displayed anywhere else in the game, so it might be confusing.
I know it helps the players who uses the wiki, but they have most likely looked up the image too

Qwertypop04:
When you are on a chest tile, it says "Open (rank) chest". So it should be fine

Jaaslet:
Missed that. Thanks

Farboo171 (approved):

PR #4798, opened 2023-09-10
Added requirements for every gym (kanto - galar) that didnt have one yet

RedzepR:
## Description

Added a lore-friendly requirement to every gym that didnt have one yet. The requirements are the same requirements you need for the town the gym is located in.

## Motivation and Context

During my playthrough I got a quest that I couldn't do because I havn't unlocked the gym yet. More infos here https://discord.com/channels/450412847017754644/1148880666688237580

## How Has This Been Tested?

Loaded up a savefile and refreshed the quests a bunch

## Types of changes

- Bug fix


osherwott (changes requested):
Trial "gyms" aren't really gyms, and they have a `quest: false,` argument to avoid showing up as quests for that reason. They are only used to award Z Crystal badges, so the extra requirements on them are superfluous.

RedzepR:
> Trial "gyms" aren't really gyms, and they have a `quest: false,` argument to avoid showing up as quests for that reason. They are only used to award Z Crystal badges, so the extra requirements on them are superfluous.

thanks for the hint

Farboo171 (approved):
Looks good for the Wiki! Thanks for the PR!

PR #4799, opened 2023-09-10
Made Pinkan less Sweet

EatPant2nd:
## Description

Made Pinkan be 35 Sweet (down from 50) to keep it more in line with the other berries as normally gen 3 are at max 30 and gen 5 are 40, so it's still a bit sweeter than other gen 3 due to the requirements attached to getting it.

## Motivation and Context

Pinkan at 50 was a bit excessive when nothing else came close to it, not even berries hard to get like Petaya. So 35 is a good compromise as it's a bit higher than usual gen 3 (like Belue at 30 Sour) but lower than the max flavor gen 5 (like Custap at 40 Sweet).

## How Has This Been Tested?

I opened the Berry Dex and checked that it showed the right flavor.

## Types of changes

- Balance

### DIFF
--- a/src/scripts/farming/Farming.ts
+++ b/src/scripts/farming/Farming.ts
@@ -621,7 +621,7 @@ class Farming implements Feature {
             0.1,
             2500,
             15,
-            [0, 0, 50, 0, 0],
+            [0, 0, 35, 0, 0],
             30,
             BerryColor.Pink,
             3.5,


Farboo171 (approved):
We've been meaning to get around to this and finally have a reason! Looks sweet, thanks for the PR!

PR #4802, opened 2023-09-12
Baby Kangaskhan

Qwertypop04:
## Description
Adds Baby Kangaskhan as a baby Pokemon of Kangaskhan, using 50% of Mega Kangaskhans stats as a reference to Parental Bond

## Motivation and Context
Adds it as a baby for Kangaskhan as an early introduction to baby Pokemon

## How Has This Been Tested?
Baby Kangaskhan is correctly obtained when Kanghaskhan is bred

## Types of changes
- New Pokemon


Shinesprite41:
Just a small note but in Gen 7 onwards, Parental Bond is only 25%, not 50%. That being said, a 50% power reduction makes a bit more sense in terms of baby Pokemon anyways.

RealPendragon:
> Just a small note but in Gen 7 onwards, Parental Bond is only 25%, not 50%. That being said, a 50% power reduction makes a bit more sense in terms of baby Pokemon anyways. 

I mean, that's mainly because Parental Bond was way to strong. Arguably still was after the nerf.

Shinesprite41:
> > Just a small note but in Gen 7 onwards, Parental Bond is only 25%, not 50%. That being said, a 50% power reduction makes a bit more sense in terms of baby Pokemon anyways.
> 
> I mean, that's mainly because Parental Bond was way to strong. Arguably still was after the nerf.

Yes. That doesn’t affect that the change occurred, and the 50% stats idea was based on parental bond’s effects. Like I said, I think the 50% makes more sense for the sake of being a baby mon. The lowest BST of a baby mon is Azurill at 190 and the highest is Munchlax at 390, so half (295) is right in the middle

CypherX (commented):
Hatching Kangaskhan doesn't appear to result in obtaining Baby Kangaskhan, I believe you need a DummyEvolution on Baby Kangaskhan for it -> Kangaskhan

Qwertypop04:
It appears the dummy evo in Baby Kanga went from regular Kanga to Mega Kanga for... some reason? Fixed

CypherX (approved):
Works as expected now 👍

PR #4809, opened 2023-09-16
Preventing quests going negative again

RegisCoaxans:
## Description
Rearranged the `focusSub` because focusValue was not updating anymore until the quest started, which led quests into going negative or jumping to completion on update.

## Motivation and Context
Bug reports said the quests go negative again. See [this thread](https://discord.com/channels/450412847017754644/1152070877157007422) for example.

## How Has This Been Tested?
Take Zero's Ambition, check `focusValue`, and obtain, say, 20 purple shards.
Then start the questline. Without a fix, `focusValue` is 20 units lower than `initial`, meaning you can go up to 20 in the negative from spending them. If you spent 20 shards instead of gaining them, the initial would be 20 lower and and update could then autocomplete the quest.
I loaded a file and verified `focusValue` was still updating despite the quest not being started.

## Types of changes
- Bug fix


CypherX:
Ran through a bunch of scenarios for testing and it mostly looks good but something still seems to be funky, possibly with just `MultipleQuestsQuest` and potentially unrelated to this fix.

In the Zero's Ambition quest line, during the Gain 500 Ghost Gems & 1 Spooky Plate step if I gain the Spooky plate then proceed to "sell" it the progress for that sub-quest will return to zero, or go negative if I continue selling more. Also, I'm not sure how I triggered it, but after doing various tests including gaining gems while the quest was paused, the step completed early after gaining a Spooky plate but the gem sub-quest was not completed.

RegisCoaxans:
Fixed `MultipleQuestQuest`. The `focusSub` of subQuest was disposing too early.

Could not replicate the bug we talked about. I have no idea where it could come from, maybe I fixed it too without noticing.

CypherX (approved):
Everything appears to be fully working and nothing I tried caused a quest to lose progress or go negative. I'll merge this, thanks! Fingers crossed it's the last we hear of negative or losing quest progress

Issue #4176, opened 2023-04-20
[SUGGESTION]: 


ElizaTezuna:
### Current Behavior

The pokedex allows the user to hide alternate forms of pokemons such as Pinkan and Valencian but not the hatchery.

### Improved Behavior

Allow hide alternate forms on the Unique Forms tab on the hatchery and add Sevii Region on the Regions tab since achievements has Sevii on the tab.

### Reason for change

To help users achieve shiny achievements on the pokedex faster.

### Detailed Description

![image](https://user-images.githubusercontent.com/131399273/233460667-485cd8d6-b861-465f-990a-b11734bdd7e0.png)
![image](https://user-images.githubusercontent.com/131399273/233460704-76979f88-fa50-45c6-af81-93a6a76034c2.png)
![image](https://user-images.githubusercontent.com/131399273/233460773-1574715d-342a-45b7-af5d-46ca7f2ba79e.png)
![image](https://user-images.githubusercontent.com/131399273/233460846-f785a80b-2619-47ee-9228-0c507d69c467.png)

Adding more options to the hatchery would not only allow the users to sort pokemons easier but it would also help with achievements.

PR #4812, opened 2023-09-19
Add hatchery filter to hide alternate forms

umbralOptimatum:
Adds a BreedingFilter to hide alternate forms in the hatchery, to help shiny hunt regional shiny achievements. Code copied and tweaked from the existing pokedex alternate filter, cursory test shows it behaves the same. Does not hide alternate forms native to other regions as those are part of the other regions' shiny completion.

Partially implements #4176.


EatPant2nd:
But alternate forms work for those shiny regional achievements? The game only checks the integer part of the ID so alternate Bulbasaur (like Clone one) works for the shiny achievement.

EatPant2nd:
I mean instead of just putting the alternates in a category after being done with that ID? I guess?

RefRec97:
this filter exists in the pokedex and personally i think it would be a nice addition to the hatchery.

CypherX (approved):
The filter is in the pokedex so I think it's fine to also have in the hatchery. I made a very small styling tweak to reduce the empty space between it and the filters below.

Looks good 👍

PR #4814, opened 2023-09-20
Change Alola z crystal badges to stamps and remove Trial gyms

osherwott:
## Description
- Remove faux gyms used for Trials
- Replace each island's gym badge with their stamp
  - Give Alola Champ its stamp too
- Move existing Z Crystal svgs to the zCrystal items folder
  - Add remaining crystal svgs too

## Motivation and Context
Alola's stamps correspond more to badges than the crystals, since they are rewarded upon the Grand Trial gyms and aren't related to dungeons. We are also in the process of using the crystals for their own thing, so it's important to make regional progress distinct from Z crystals for this reason.

The `DungeonGainGymBadge` code was left in, should we ever need to use it in the future. I figured with the amount of spin offs there are we might need it for something else down the line. (Interestingly enough it also works with temp battles.)

## How Has This Been Tested?
- Ran `npm run test` to check if anything was missed.
- Tested commits prior to this branch on my general test branch for Alola. The commits here were copied from there.

## Screenshots:
image
Fig 1. _A stamp for each island/gym. The Champion Stamp ended up nicely centered, too. They look very pretty :)_

## Types of changes
- UI improvement
- Canon


Qwertypop04 (approved):

PR #4825, opened 2023-09-26
RouteData Re-ordered by Pokémon instead of Weather

EatPant2nd:
## Description

Right now, RouteData has the weather encounters grouped by Weather which looks fine on paper but makes the Pokémon show up in multiple groups. Since the Wiki cares about Pokémon first, requirements later, it would be better to group all these Pokémon individually with their respective Weather requirements (unless some share the same requirements, then they are allowed to share the same array). By doing this, the Wiki will display the Pokémon just once in Special section of the Routes and, at the same time, it will also make Routes be displayed just once in the Pokémon pages, letting the tooltip take care of showing all the list of Weathers (and other Requirements if they exist) instead of multiple occurrences of the Pokémon / Route with one Weather type each.

This PR, then, seeks to do this reorder thus helping the Wiki display data without repeated entries.

## Motivation and Context

All the repeated entries made the Wiki display the weather encounters in a poor manner. With this, it will be easier to read and provide a better user experience.

## How Has This Been Tested?

Tested that the game loaded correctly. Went to check some routes. Didn't test all weather types in all routes for encounters but it's just a re-order so nothing should break.

## Types of changes

- Better Wiki


CypherX (approved):
👍

PR #4828, opened 2023-09-29
Pokedex LazyLoader performance improvement

umbralOptimatum:
Similar to the breedingModal's loading overzealousness (#4673), the Pokedex display currently does not reset its position in the pokemon list when filters change. Potentially loading many pages of data simultaneously defeats the point of the LazyLoader, so I've made it reset to the start of the list whenever a filter changes.


CypherX (approved):
👍

PR #4838, opened 2023-10-09
Fixing Johto Unlock Routes Near Ecruteak

Shinesprite41:
## Description

In Johto, before defeating Morty around Ecruteak, Mt. Mortar is unlocked, but Route 42, Mahogany Town, and Route 43 are not. In HGSS, it is possible to access Route 42 (namely it's grass for encounters), as well as everywhere else mentioned, in 2 ways. Either you can beat Morty and obtain Surf, or you can go through Mt. Mortar prior to beating Morty. You are not able to get to the Red Gyarados without Surf, so you still cannot start that questline. 

## Motivation and Context

If I'm honest, I haven't checked if base GSC is like this, but from what I understand, we are moving towards making the region based on HGSS. As such, it makes sense to alter the accessibility of the routes to be more accurate to the game. 

## How Has This Been Tested?

I ran the updated routedata repo and tested it with my Johto AC min egg file, which had access to 42 after beating Mt. Mortar, 43 after clearing 42, and Mahogany Town. I also tested it again after defeating Morty to ensure that it doesn't lock itself without clearing Mt. Mortar.

## Types of changes

- Minor adjustment
- Update to match base game


Qwertypop04 (approved):

PR #3536, opened 2022-12-06
Achievement chains in Achievement tracker

RegisCoaxans:
Closes [#1573](https://github.com/pokeclicker/pokeclicker/issues/1573).

Whenever the player unlocks the tracked achievement, the tracker goes up the chain and grabs the next difficulty for the very same achievement if there is any.

PR #4842, opened 2023-10-17
Fix : Achievement Tracker not correctly switching achievements on completion

RegisCoaxans:
## Description
#3536 extended and refined. Works with the achievements added since then (Contests, Customdex and Safari Shiny).
I also made it so it is easier for someone adding a new achievement requiement with parameters to make it work in the tracker.

## Motivation and Context
I remember a report on the Achievement Tracker not working sometimes.
I made ugly code...

## How Has This Been Tested?
Opened a save, completed few achievements and made sure it picked the next tiers of the corresponding achievements.
Run a script to look at achievement requirements' signatures.

## Types of changes
- Bug fix


CypherX (approved):
Works well and is definitely cleaner and easier to implement into new achievement requirements.

Issue #4845, opened 2023-10-20
[TASK]: Agile and Strong Scrolls


Qwertypop04:
### Task Description

Add 2 new Held Items: Agile Scroll and Strong Scroll. These are items that would represent the Agile and Strong Styles from Pokémon Legends: Arceus.
- The Agile Scroll, when held, would decrease the  Pokémons attack, but increase their contribution to click attack, and evolve Stantler into Wyrdeer at level 31.
- The Strong Scroll would decrease the Pokémon's contribution to click attack, but increase their attack, and evolve Hisuian Qwilfish into Overqwil at level 25.
Exact details on how much they would increase and decrease attack and click attack are still to be decided

Part of #4128 

### Notes

They would use the following sprites:
Agile Scroll: ![Agile Scroll](https://github.com/pokeclicker/pokeclicker/assets/63805905/c700c05b-4b8d-4464-9d8e-356b55091583)
Strong Scroll: ![Strong Scroll](https://github.com/pokeclicker/pokeclicker/assets/63805905/34b8fc18-1d42-4f53-9b1a-5afac1685fc9)

Is not absolutely required for Hisui to be released, but if it isn't, Wyrdeer will most likely not be obtainable at Hisui's launch

### Issues Required For This Task

None

CypherX:
I have this pretty much done, just need to figure out balancing.

PR #4846, opened 2023-10-21
Add Agile & Strong Scroll held items for Hisui

CypherX:
## Description
Adds the Hisui held items as described in #4845. Current modifier values are placeholders. Definitely needs testing to find a balance.

~~Tweaked how click attack is calculated to support modifying the contribution of a single pokemon. Benchmarking showed that this new method is faster than the original calculation.~~
Moved into separate PR

## How Has This Been Tested?
Equipped each held item and checked that the pokemon attack and click attack increased or decreased accordingly

## Screenshots (optional):
image
image

## Types of changes
- New Held Items


src/scripts/party/PartyPokemon.ts
```diff
@@ -117,6 +118,12 @@ class PartyPokemon implements Saveable {
                 this.addOrRemoveHeldItem(this.heldItem());
             }
         });
+        this.clickAttackBonus = ko.computed((): number => {
+            // Caught + Shiny + Resistant
+            const bonus = 1 + +this.shiny + +(this.pokerus >= GameConstants.Pokerus.Resistant);
+            const heldItemMultiplier = this.heldItem && this.heldItem() instanceof HybridAttackBonusHeldItem ? (this.heldItem() as HybridAttackBonusHeldItem).clickAttackBonus : 1;
```
RegisCoaxans, line 124:
```suggestion
            const heldItemMultiplier = this.heldItem() instanceof HybridAttackBonusHeldItem ? (this.heldItem() as HybridAttackBonusHeldItem).clickAttackBonus : 1;
```
First check is truish at all time due to `this.heldItem` pointing to the observable. However, it means it is not required as the second check sorts out `undefined` anyway.

CypherX: I copied it from the regular pokemon attack held item bonus check, lol. Thanks

src/scripts/party/PartyPokemon.ts
```diff
@@ -117,6 +118,12 @@ class PartyPokemon implements Saveable {
                 this.addOrRemoveHeldItem(this.heldItem());
             }
         });
+        this.clickAttackBonus = ko.computed((): number => {
+            // Caught + Shiny + Resistant
+            const bonus = 1 + +this.shiny + +(this.pokerus >= GameConstants.Pokerus.Resistant);
```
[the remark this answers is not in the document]
CypherX: Out of the 4 methods I benchmarked pureComputed was the slowest, including slowing than the original calculation, while computed was the quickest.

Qwertypop04 (approved):

PR #4848, opened 2023-10-22
Easter Rework Revival

Qwertypop04:
Revival of #4792 

## Description
Making easter event more fun!
Keeping the questline. Forcing players to redo it. I have changed the text to match the new event.
After the questline has been completed, instead of just getting a free pokemon, the real egg hunt starts!

Pikachu (Easter) will roam the dungeons of Kanto
Togepi (Flowering Crown) will roam the dungeons of Johto
Torchic (Egg) will roam the dungeons of Hoenn
When I say "roam the dungeons", I am using the code Jaaslet wrote for max raids. I have selected 6 semi-random dungeons in each of the regions. The pokemon will spawn as a dungeon boss in a random of them. Each time you defeat the Pokemon, it will move to a random dungeon on the list.

This PR also changes "Surprise Togepi" to "Togepi (Flowering Crown)". It's the only non-canon alt form that most of the crew agrees that is fine to replace, so far. The form uses the same ID, so it's a smooth transition.

All 3 pokemon is drawn by Drakay.

Pikachu is from the TCG, Togepi is from PoGo and Torchic is from some official merchandise

## Motivation and Context
Part two of Jaaslet's "make events more interesting, without needing to create new mechanics for them".
It feels like an easter hunt. You might still see the easter part of the event, even after the quest is completed.
Added two more, because a full event for just 1 pokemon is a bit boring. Was considering adding one of them as something else, but think this fits well to use the 3 regions the quest is in. So if we want more variation for the event, we can add 1-2 more mons in another PR.
Only one region per Pokemon, because else the hunt would be too hard/annoying. Hopefully this level of difficulty is fine. Else we can make sure they show up in 2 dungeons at a time. But I think that is a post-release discussion.

## How Has This Been Tested?
Tested that the quest still functions correctly, and the Pokemon shift dungeons every time they are defeated

## Types of changes
- Event Improvement


CypherX (approved):
Played through everything, looks good

src/scripts/dungeons/Dungeon.ts
```diff
@@ -3606,6 +3740,14 @@ dungeonList['Sealed Chamber'] = new Dungeon('Sealed Chamber',
         new DungeonBossPokemon('Regirock', 4500000, 20, {requirement: new QuestLineStepCompletedRequirement('The Three Golems', 8)}),
         new DungeonBossPokemon('Regice', 4500000, 20, {requirement: new QuestLineStepCompletedRequirement('The Three Golems', 8)}),
         new DungeonBossPokemon('Registeel', 4500000, 20, {requirement: new QuestLineStepCompletedRequirement('The Three Golems', 8)}),
+        new DungeonBossPokemon('Togepi (Flowering Crown)', 2700000, 23, {
```
DreamNya, line 3743: I think here should be 'Torchic (Egg)'

DreamNya: And it seems that `new PokemonDefeatedSelectNRequirement('Pikachu (Easter)', 4, 6, 1)` is missing, causing Pikachu to disappear in Kanto

PR #4849, opened 2023-10-23
Fix Hopo berry mutation being visible when not available

CypherX:
## Description
When the Hopo berry was added to the game code its mutation was added to the list before Kasib and Starf but no update function was added to shift the mutation index. This caused the Kasib `hintSeen` value to be transferred to Hopo resulting in Hopo being visible in the berrydex along with its mutation hint if the player had previously seen the hint for Kasib.

This PR does two things:
- Updates the 0.10.11 update function (when Hopo was added) to correctly shift the mutation indexes.
- Sets the `hintSeen` value for Hopo to false in the 0.10.16 update function.

Note that the index differs in the two changes - Snover berry was added in between.

## Motivation and Context
Hopo berry is not currently available so it should not be visible.

## How Has This Been Tested?
Loaded a save from 10.6 that has seen the Kasib hint, Hopo no longer appears in the berrydex.
Loaded a save with Hopo berry visible in the berrydex, after the update is applied it is no longer visible.

## Types of changes
- Bug fix


osherwott (approved):
Looks good

Qwertypop04 (approved):

PR #4850, opened 2023-10-23
Changed Thunderstorm Weight from 5 to 10

EatPant2nd:
## Description

Changed Thunderstorm Weight from 5 to 10 so it's more in line with the other weather types. 

## Motivation and Context

Thunderstorm being weight 5 was odd and it also caused issues with a quest line in Hoenn where, sometimes, people had to wait more than a week to complete it.

## How Has This Been Tested?

Checked the Weather Institute Forecast to see the changes.

## Screenshots (optional):

Before:
thunderstorm-2

With weight 10:
thunderstorm-1

## Types of changes

- I dunno, weather.


Aegyo (approved):
🌩

PR #4851, opened 2023-10-23
Fix for occasional Mine TypeError 

umbralOptimatum:
Fixed `Mine.loadSavedMine()` having the timeout call `Mine.checkCompleted()` in the global context, breaking its `this` references.


CypherX (approved):
Have never seen this error myself or reported anywhere but can confirm it exists and this fixes it

PR #4852, opened 2023-10-23
Click attack performance improvements

umbralOptimatum:
`Party.calculateClickAttack()` currently iterates over the entire party list every time it is called, up to 20 times per second. The party's base click attack changes far less often (a couple thousand times per save file) so I moved that code to a pureComputed. Performance testing shows this approximately doubles the speed of the method, going from ~2.7 ms to ~1.4 ms.


CypherX:
#4846 includes some changes to how click attack is calculated to support modifying the contribution of a single pokemon, those changes also increased the speed of the function. I also tested a pureComputed to hold the base click attack, like you did, on top of those changes but it didn't seem to increase performance any further

PR #4855, opened 2023-10-30
Added Ice Type to Santa Snorlax

EatPant2nd:
## Description

Since a few Event Pokémon have an added or modified second Type, I added Ice to Santa Snorlax due to Christmas, snow, and all that. Plus, it makes the Pokémon a bit more interesting like this.

## Motivation and Context

Other Event Pokémon also have gotten type changes like the Halloween Pokémon.

## How Has This Been Tested?

Checked the Pokédex, damage calculator, and the hatchery. Displays and works fine.

## Screenshots (optional):

santa-ice

## Types of changes

- Ice.


Qwertypop04 (approved):
Looks cool, thanks for the PR!

PR #4856, opened 2023-10-31
Added new SortOption: attack at level 100

umbralOptimatum:
Added a new party sort option to sort using each pokemon's attack at level 100 rather than its current level. Mainly this will improve hatchery performance by not resorting the entire party every time one pokemon levels up.

Spun off from #4673.


CypherX (approved):
Looks good, thanks 👍

PR #4857, opened 2023-10-31
Mega Evo and Gigantamax helper functions improvements/fixes

umbralOptimatum:
Added new functions to check if a pokemon has an unobtained mega evolution or is itself a mega evolution. Besides improving clarity, this fixes an upcoming bug where the "is this a mega evolution" checks would give false positives on other temporary forms of mega-able species, such as Gigantamax Venusaur.

Spun off of #4673.


src/modules/pokemons/PokemonHelper.ts
```diff
@@ -133,12 +133,32 @@ export function hasMegaEvolution(pokemonName: PokemonNameType): boolean {
...
+}
+
 export function getMegaStones(pokemonName: PokemonNameType): MegaStoneItem[] {
     return GameHelper.enumStrings(MegaStoneType)
         .filter(s => (ItemList[s] as MegaStoneItem)?.basePokemon == pokemonName)
         .map(s => ItemList[s] as MegaStoneItem);
 }
 
+export function hasGigantamaxForm(pokemonName: PokemonNameType): boolean {
+    return P.pokemonMap['Gigantamax ' + pokemonName].id > 0 || P.pokemonMap['Eternamax ' + pokemonName].id > 0;
```
CypherX, line 147: Can you change the string concatenation here and in `hasUncaughtGigantamaxForm` to template literals instead? ESLint is supposed to catch these, not sure why it isn't

src/modules/pokemons/PokemonHelper.ts
```diff
@@ -133,12 +133,32 @@ export function hasMegaEvolution(pokemonName: PokemonNameType): boolean {
...
+}
+
+export function hasUncaughtGigantamaxForm(pokemonName: PokemonNameType): boolean {
+    let gmaxForm = P.pokemonMap['Gigantamax ' + pokemonName];
+    if (gmaxForm.id <= 0) {
+        gmaxForm = P.pokemonMap['Eternamax ' + pokemonName];
+    }
+    return !!(gmaxForm.id > 0 && !App.game.party.alreadyCaughtPokemon(gmaxForm.id));
```
CypherX, line 155: I don't think the `!!` is needed here. It will always evaluate to a truthy or falsey value already, no?

CypherX (approved):
Looks good 👍

PR #4859, opened 2023-10-31
Changed Dungeon Size Decrease According To Poll By Red

RedzepR:
## Description

I changed the function for dungeon size reduction from every power of 10 to this formula (n(n+1)/2) * 50 in which n is the amount of reduction. After doing some math to solve for n the new formula for dungeon reduction is: (-1 + √(1 + 4x/25)) / 2 = n. 

## Motivation and Context

A poll Red created in april of last year about changing the function for dungeon size reduction. Its been about 1.5 years but the option with the most changes hasn't been implemented yet. 

## How Has This Been Tested?

Gone through a bunch of dungeons with different clears. Dungeon size was always correct

## Screenshots (optional):

image
image

## Types of changes

- Dungeon changes


EatPant2nd:
Which was relevant at Alola, but for regions above Alola, the poll that won is better.

PR #4864, opened 2023-11-03
Add fallback Pokéball filter UUID generation method

PixLSteam:
## Description
Adds a fallback Pokéball filter UUID generation method in case `crypto.randomUUID` is not available.

## Motivation and Context
A lot of bug reports on Discord, Reddit and GitHub.

## How Has This Been Tested?
Tested on version 1.1 of the official client.

## Types of changes

- Bug fix


CypherX (approved):
Works as expected on an old client version. We'll finally stop getting those bugs reports! 🎉

PR #4868, opened 2023-11-05
Ability to hide uncaught safari pokemon silhouette from ranger NPC

CypherX:
## Description
Adds the ability to hide uncaught pokemon from appearing as a silhouette in the ranger NPC dialogue.

## Motivation and Context
We often get questions from new players about being unable to find Dratini and Dragonair in the Kanto safari zone, unaware that they require being caught first. The confusion is caused by the ranger NPC displaying these pokemon as a silhouette but no explanation that they have to be caught elsewhere first.

Sure, we could simply add text to the NPC explaining this like the friend safari, but we all know how pokeclicker players and reading get along.

## How Has This Been Tested?
Checked the NPC while having both Dratini & Dragonair caught - both appeared.
Removed both from my party, checked the NPC again - neither appear.
Caught Dratini, checked the NPC again - Dratini appears

## Screenshots (optional):
before change:
image

after change:
image


osherwott (approved):
Looks good.

Qwertypop04 (approved):

PR #4872, opened 2023-11-07
Changed all gifted mons to GiftNPCs

RedzepR:
## Description

Went through all quests to see if a pokemon was gifted. Change every quest with a custom reward (gift pokemon) to a giftNPC.

## Motivation and Context

1. Transparency is key. Adding stuff to the players party feels wrong.
2. Before Eugene left he asked me to change gift mons to giftNPCs
3. If people have some challenge run going an unexpected gift mon could force someone to reset to an earlier save, making overall experience worst
4. _I_ lost about 1 day of progress because I was gifted a mon
5. We got giftNPC so why not use them?

## How Has This Been Tested?

Went through every quest atleast once (except Magearna). 
NPCs:

1. Showed up at the correct quest step
2. Progressed the quest once talked to them
3. Stayed if you declined the mon but could still progress with the quest
4. Vanished when the mon was accepted

## Types of changes

- QoL


Qwertypop04:
This seems unnecessary for Meta Groudon, Detective Raichu, Magikarp Saucy Blue and Magearna (Original Colour), since they are the only reason to do the quests in the first place

Qwertypop04:
Also, outside of mandatory fights, I don't see any problems with adding mons to players parties. The only people who have a problem with this are those doing weird challenges like monotype, and they should really know what they are doing anyway

Qwertypop04:
After discussion, it has been decided this is unnecessary. Closing

PR #4880, opened 2023-11-11
Protect player.gainItem and player.loseItem

AevitasDragonkin:
## Description

I'll start by stating that I am in no way proficient in this codebase, and would like others to verify that changing behaviour in this location does not break anything else.

Player.gainItem didn't perform any checks related to the maxAmount that was set for items in ItemList. This change makes it impossible for item count to go over the maxAmount set, or impossible to go below 0.

Edit:
Things to note:
* Already owning items with an item count above the limit will reset it to the limit next time you gain or lose some amount of this item.
* This also does not remove the option to gain/buy extra items, this will only prevent them from being added to your account.
* Once you have more than MAX_SAFE_INTEGER of an underground item, daily trades also become infinite, as you can trade smaller amounts which are insignificant, and thus just creating the second item while not losing any of the first.

## Motivation and Context

This protects items from going over their maxAmount limit, which by default is set to MAX_SAFE_INTEGER. The ability to go over this number affects Evolution stones in a way that you now have the option to have infinite evolution stones available, as subtracting 1 does not decrease the item count.
Downside (if this can be considered one??) would be that daily trades will also be capped at this number. 

## How Has This Been Tested?

1. Boot up the game
2. Open Evolution items tab
3. Open console
4. Run `ItemList['Fire_stone'].gain(1e20)` and verify it's capped at MAX_SAFE_INTEGER 9,007,199,254,740,991

## Types of changes

- Bug fix

### DIFF
--- a/src/scripts/Player.ts
+++ b/src/scripts/Player.ts
@@ -191,11 +191,11 @@ class Player {
     }
 
     public gainItem(itemName: string, amount: number) {
-        this._itemList[itemName](this._itemList[itemName]() + amount);
+        this._itemList[itemName](Math.min(this._itemList[itemName]() + amount, ItemList[itemName].maxAmount));
     }
 
     public loseItem(itemName: string, amount: number) {
-        this._itemList[itemName](this._itemList[itemName]() - amount);
+        this._itemList[itemName](Math.max(this._itemList[itemName]() - amount, 0));
     }
 
     public lowerItemMultipliers(multiplierDecreaser: MultiplierDecreaser, amount = 1) {


AevitasDragonkin:
Things to note:
* Already owning items with an item count above the limit will reset it to the limit next time you gain or lose some amount of this item.
* This also does not remove the option to gain/buy extra items, this will only prevent them from being added to your account.

Razorflamekun:
I'm not a fan of restricting players from getting too many of one item.  I think it's fine if players get billions of an item.  Also, quick tip: Update your branch to have it be up-to-date with the base branch.

RegisCoaxans:
It would only prevent players from getting too many of an item we set a maximum amount on — often 1. Most items would not be affected by this change. Plus, when not set, the maximum amount defaults to a high enough number.

AevitasDragonkin:
Just thought of another reason:
Once you have more than MAX_SAFE_INTEGER of an underground item, daily trades also become infinite, as you can trade smaller amounts which are insignificant, and thus just creating the second item while not losing any of the first.

CypherX:
Seems fine. My only real concern, and admittedly it's minor, is that the `maxAmount` property is defined as part of `ShopOptions` (I know it ends up as a base property of `Item` anyway). I don't think it will make a difference currently, but just wanted to note it.

I think we should also update the daily deal (and other trades/deals) `canUse` and `use` functions to also handle MAX_SAFE_INTEGER. That way we can prevent trades in the first place if the player is already at the cap and cap the amount traded if they attempt to go over to prevent losing items.

AevitasDragonkin:
Reason I did it on the player was to make that if any cases were missed, or future content forgot to take it into account, that it would still work.

I now see that it's indeed defined in the ShopOptions, but the Item constructor sets it to MAX_SAFE_INTEGER anyway if the option is not provided https://github.com/pokeclicker/pokeclicker/blob/7113b05f517d503b040e592c90711487353a2c1a/src/modules/items/Item.ts#L48C9-L48C9

AevitasDragonkin:
Closing this for now as this only fixes a small fraction of the game as not everything is stored on the player. It doesn't even fully fix underground trades as only evo stones are stored on the player's main inventory. Other items are stored on the mineInventory

PR #4884, opened 2023-11-14
Great Marsh

Qwertypop04:
## Description
Adds the Great Marsh as a Safari Zone in Sinnoh

## Motivation and Context
Its the only Safari other than Kanto that has any exclusive Pokemon, and having it removes the need for Carnivine and Skorupi being in eggs and shop due to it not existing

## Types of changes
- New feature


Shinesprite41:
I’ve always thought Kanto being the only safari zone was weird (now there’s friend safari but still)

It’s a feature in a heavy amount of the mainline games and with the new-ish safari zone achievements, I like the idea of adding more :) good pr

CypherX (approved):
Looks good 👍

PR #4889, opened 2023-11-19
Item Refactor : ported mineInventory into centralised itemList

RegisCoaxans:
## Description
Items from the underground had their own storage system, preventing them from being used as regular items anywhere else in the game.

## Motivation and Context
That refactor gives underground items the ability for multiple uses. For instance, few of them could now be turned into held items.
Mine inventory handling was poor design with different arrays crisscrossing for data matching.
Allows for a revival of #4880.

## How Has This Been Tested?
Gained mine items, used everything that could interact with old underground code.

## Types of changes
- Code Refactor


CypherX:
Couple things I noticed while testing:

- Fossils are being included in 'Sell all treasures'. I didn't try selling to see what happened but they are included in the overall item count
- Locked treasures to prevent selling did not carry over on update

Some other notes:
- Evo stones & mega stones appear in the underground treasure list now
- Underground items exist and need to be maintained in two locations now - ItemList and UndergroundItems, is there a better way to handle this?

RegisCoaxans:
> Underground items exist and need to be maintained in two locations now - ItemList and UndergroundItems, is there a better way to handle this?

That would need a bigger refactor, like `Item` itself being able to carry underground metadata... Not sure it is worth it.
For now, `ItemList` carries every actual items in the game, and `UndergroundItems` the metadata for the Underground to spawn some of those in the mine, as it did for Megastones and evo stones.

RegisCoaxans:
The issues are addressed, by the way.

RegisCoaxans:
Good catch, I removed duplicated code and simplified `gainMineItem` as MegaStoneItem already handles everything correctly.

CypherX (approved):
Looks good, thanks!

PR #4890, opened 2023-11-19
Fix incorrect Easter Roaming Dungeon Bosses

DreamNya:
## Description

Fix bug form #4848 which may cause roaming bosses in the Kanto and Hoenn not to appear.
The code of Pikachu (Easter) from #4792.

## Motivation and Context

In the original code, `Pikachu (Easter)` and `Torchic (Egg)` may disappear with continuous defeat.
In Kanto, `Pikachu (Easter)` only roams 5 dungeons, but the random seeds are set for 6.
In Hoenn, one of the dungeon bosses was incorrectly set to `Togepi (Flowering Crown)` which should only roam in Johto instead of `Torchic (Egg)`.

## How Has This Been Tested?

Just activate Easter event and keep defeating roaming dungeon bosses in the Kanto and Hoenn.
They should now appear correctly in each of the six dungeons.

## Screenshots (optional):

None.

## Types of changes

- Bug fix


CypherX (approved):
Nice catches, thanks

PR #4891, opened 2023-11-19
Update CapturePokemonTypesQuest

RedzepR:
## Description

Changed CapturePokemonTypesQuest to only suggest pokemon types you own

(I'm not too happy with the code but it does its job, I'm open and happy about suggestions)

## Motivation and Context

Having undoable quests for the player sucks so changing it that you can always breed away the quest is a good fix

## How Has This Been Tested?

Reloaded my quests on mono fairy 100 times and never got an undoable quest

## Types of changes

- Bug fix? maybe idk it wasnt really a bug to begin with


RegisCoaxans:
You could do something like this :
```js
const foundTypes = App.game.party.caughtPokemon.reduce((ftypes, p) => {
    const pp = pokemonMap[p.id];
    pp.type.reduce((ft, t) => {
        ft[t] = true;
        return ft;
    }, ftypes);
    return ftypes
}, {});
```
That gives you a `type→boolean` interface to multiply the weights against in one iteration over the Pokémon List.

umbralOptimatum (changes requested):
Instead of making a new array via `push()` and iterating over it for every type, it would be much faster to just iterate over the party once while counting types.

PR #4894, opened 2023-11-19
Resizeable UI modules

### LINKED ISSUE #4887: [SUGGESTION]: Allow UI box sizes be bigger
### Current Behavior

Currently, the UI boxes on the main screen are set to only be two sizes, minimized and open to a set size. For most boxes, such as the Battle Items, Quests, and Oak Items, their open size is big enough and can't really be expanded without stretching text or images. With boxes such as Poke Balls and Pokemon List however, once you reach enough Pokemon or filters to fill the open size, it puts a scroll bar on the side and you have to scroll to find what you want. This is especially true for the Pokemon List.

### Improved Behavior

To improve upon this, I suggest that the Poke Balls, Pokemon List, Hatchery, and Achievement Tracker UI be given a second size option upon clicking them that expands their "open" size to be bigger, or allow the bottom of the box to be grabbed and dragged down to the length that the user wants.

### Reason for change

It would allow players to more personally customize their PokeClicker layout and have the option to see more on the main screen so that they do not have to continue opening menus.

### Detailed Description

For the Pokemon List, I doubled the length of the Pokemon List to show more Pokemon which would allow players to view certain groups that they may have in full instead of scrolling. For me personally, I like to look at the attack of many of my Pokemon when I put them into specific categories, such as Pokemon that mega evolve, and be able to look at as many as I can without having to scroll a lot.

For the Poke Balls list, it would allow players to see all of the filters that they have made and sort through them instead of having to scroll. This one may not be useful for everyone, but I thought it would be nice since I changed Pokemon List.

For the Achievement Tracker box, I made multiple quests viewable which would be nice when trying to track multiple similar quests such as Safari Zone rocks, steps, and baits. I think this would be a nice add-on because I find it as bit annoying to have to click "List" on the Achievement Tracker or go through the start menu to see an achievement I don't have tracking.

For the Hatchery box, another thing I did is remove the scroll and just make a big box marked with arrows as to how the icons would move when a Pokemon is moved from the queue to hatching.

**Screenshots:**

Currently, a 5 Column layout can look like this, many UI boxes at the top, but a lot of empty space below
image

This is my suggestion/ideas of how this could be implemented. (This was done in MSPaint, so it doesn't look perfect)
UI_Suggestion

CypherX:
## Description
Adds the ability for UI modules to be resizable, specifically the pokeball selector and pokemon list, but can also be applied to other modules in the future that might benefit.

Closes #4887 

Had to add `background-attachment: fixed;` to the background image styling or it would get upset when a module was resized.
Also had to disable the collapse/expand transition animation for the resizable modules or things got wonky.

## How Has This Been Tested?
Resized the modules, checked that the size persisted through reloads.
Checked that collapsing/expanding still worked as expected and the size was peristed.

## Screenshots (optional):
image
image

## Types of changes
- UI improvement


RedSparr0w (approved):
Works well for me!

Re-thinking about it probably don't need the reset thing, people can still just put it where feels right?
Up to you if you want to leave it in or remove it though, I'm fine either way.

Thanks for the PR! 👍

CypherX:
> Re-thinking about it probably don't need the reset thing, people can still just put it where feels right? Up to you if you want to leave it in or remove it though, I'm fine either way.

I feel it's a good safety measure in the event something goes wrong and a module gets stuck at a small/massive size. Couldn't trigger any such issue while testing, but you never know.

Thanks for reviewing!

PR #4899, opened 2023-11-22
Oh oh oh, merry Christmas rework !

RegisCoaxans:
## Description
- The Snorlax (Snowman) hides in Seafoam Islands during Christmas event. It is an epic mimic (4 to 10%, because there is no other epic loot in here) and ignores the debuff.

- I added Santa Jynx and Christmas Presents : the presents give some units of a random currency (except BP and CT). Jynx may drop Christmas Presents on defeat.
Made a little scenario with daily steps :
1. You defeat a Santa Jynx tempbattle (Blue 2 difficulty). You then find Santa Jynx on routes until you drop 10 presents.
2. You fight against the same tempbattle but stronger. You can drop up to 15 presents before Santa Jynx stops appearing.
3. Tempbattle again, stronger again. Then you find Santa Jynx for 21 drops.
4. Each day from now on, the last tempbattle (about twice as strong as first one) allows you to collect up to 100 presents daily.

    If you did not collect every presents you could in due time, you just repeat the same step on next day.

- Changed Santa Snorlax to be Fairy instead of Ice, because it matches Christmas better.

- Made Elf Munchlax obtained through trading in Sandgem Town, Sinnoh. Turned it from Normal into Normal / Fairy.

- Reindeer Stantler now roams Johto.

- Additional functional changes.

## Motivation and Context
Elf Munchlax was boring as a Baby.
Snowman is cool.
Santa Jynx makes a neat alt form for a Pokémon which has none yet.
Same for Reindeer Stantler.

## How Has This Been Tested?
Started the event and chased Snorlax in the dungeon.
Defeated the Jynx battle, got few presents on routes and time traveled to do the other battles.
Traded a few Munchlax. Checked that Reindeer Stantler was roaming Johto.

## Types of changes
- Cool changes 😎

## Additional notes
- Soul working on the present sprite then this is ready for review.
- I put Munchlax in Sandgem for canon reasons (BDSP) but I feel like the Eterna Forest environment matches better, thoughts ?
- Feel free to suggest some text changes.


RedzepR:
Just a quick thought: Except for the sprite Santa Snorlax and Snorlax Snowman are practically the same mon. While not the first time that happens it would be nice if the two mons differentiate a bit. So while brainstorming we thought of Snorlax Snowman to be turned into a Ice/Fairy mon as its a magical snowman that came to live and as such its fitting it got the magical (fairy) type

RegisCoaxans:
I thought about that already.
I do not want to get rid of the normal type, though, as it is inherited from the base form; and ice fits Snowman better than fairy.
Santa Snorlax has turned normal / ice because it helps monotype. However, I might change it to normal / fairy in due time instead. Your challenge run will get its Mon too, do not worry.

RegisCoaxans:
Fixed the mimics not having lock when caught but not available at this time.
I also noticed the lock message tooltip could flicker due to the scrollbar misplacing it in front of the sprite to hover. I moved it to the left.
image

RegisCoaxans:
Santa Jynx and Christmas Present item introduced. Santa Snorlax changes from Normal / Ice to Normal / Fairy to avoid having two similar Snorlax alt forms.
I updated master post with a few more details.

Qwertypop04:
After discussion, it seems it would be preferable if Elf Munchlax is kept (as well as given the same Normal/Fairy type as Santa Snorlax), and that Snorlax (Snowman) should be an addition rather than a replacement.
It was also suggested that Raindeer Stantler could be added, perhaps as a roamer in Johto

RegisCoaxans:
I am already on it. I am changing the obtaining method for Munchlax, though.
Are there sprites for Stantler ?

Qwertypop04:
> Are there sprites for Stantler ?

Not as far as I'm aware

RegisCoaxans:
It's beginning to look a lot like Christmas rework.
Added back the bad boi. Changed its obtaining method.
Added Reindeer Stantler.
Updating master post accordingly...

src/scripts/temporaryBattle/TemporaryBattleList.ts
```diff
@@ -282,6 +282,126 @@ TemporaryBattleList['Blue 6'] = new TemporaryBattle(
...
+        new GymPokemon('Santa Jynx', 12240, 20),
+        new GymPokemon('Santa Jynx', 12240, 20),
+        new GymPokemon('Santa Jynx', 12240, 20),
+    ],
+    'Huh ? They are gone by the routes now.',
+    [
+        new SpecialEventRequirement('Merry Christmas!'),
+        new ItemRequirement(27, 'Christmas_present', GameConstants.AchievementOption.less),
+        new ItemRequirement(11, 'Christmas_present'),
+        new TemporaryBattleRequirement('Santa Jynx 1', 1, GameConstants.AchievementOption.less),
```
CypherX, line 327: The Santa Jynx 2, 3, and 4 temp battles don't appear because these previous temp battle requirements are set to `less` when they should be `more` (or left as default).

src/scripts/temporaryBattle/TemporaryBattleList.ts
```diff
@@ -282,6 +282,126 @@ TemporaryBattleList['Blue 6'] = new TemporaryBattle(
...
+// Kanto Christmas Temporary Battles, based on Blue 2
+TemporaryBattleList['Santa Jynx 1'] = new TemporaryBattle(
+    'Santa Jynx 1',
+    [
+        new GymPokemon('Santa Jynx', 12240, 20),
+        new GymPokemon('Santa Jynx', 12240, 20),
+        new GymPokemon('Santa Jynx', 12240, 20),
+        new GymPokemon('Santa Jynx', 12240, 20),
+    ],
+    'Huh ? They are gone by the routes now.',
```
CypherX, line 294: This doesn't really make sense, and maybe provide some additional direction:

`Huh? They ran off and are trying to blend it with the local Pokémon population. Find those Santa Jynx and collect more Christmas presents!`

src/modules/items/ChristmasPresent.ts
```diff
@@ -0,0 +1,37 @@
...
+            [Currency.money]: 50 * 3 ** scaling,
+            [Currency.questPoint]: 5 * (1 + scaling) ** 2,
+            [Currency.dungeonToken] : 30 * 2 ** scaling,
+            [Currency.diamond]: 1 + scaling * 2,
+            [Currency.farmPoint]: 15 * (1 + scaling) ** 2,
+        };
+        const cur = Number(Rand.fromArray(Object.keys(values)));
+        const { amount } = App.game.wallet.addAmount(new Amount(Math.round(values[cur] * this.size), cur), false);
+        Notifier.notify({
+            message: `You opened the gift and earnt <img src="/assets/images/currency/${Currency[cur]}.svg" height="24px"/> ${amount.toLocaleString('en-US')}.`,
```
CypherX, line 28:
```suggestion
            message: `You opened the gift and received <img src="/assets/images/currency/${Currency[cur]}.svg" height="24px"/> ${amount.toLocaleString('en-US')}.`,
```

earnt -> received

src/scripts/temporaryBattle/TemporaryBattleList.ts
RegisCoaxans: Those are daily tempbattles.
The tempbattle 2 unlocks on day 2 if you collected enough presents, tempbattle 3 on day 3 with more presents and tempbattle 4 on every next day after more presents have dropped. I therefore meant `less` to prevent a battle to unlock on the same day as the previous one.

CypherX: Ahhhh, okay. That makes sense. I thought they were supposed to become available one after another

src/scripts/temporaryBattle/TemporaryBattleList.ts
RegisCoaxans: Yeah, it looks way better. I put that sentence in. Thanks.

RegisCoaxans:
Setting this as ready for review. All sprites are here.

RegisCoaxans:
Added a NPC to appear and tell when the player can not find Santa Jynx on routes anymore.

CypherX (approved):
Looks good, thanks for taking care of all the changes

PR #4900, opened 2023-11-22
Fixes catching shiny shadow pokemons

AevitasDragonkin:
## Description

Fixes an issue where catching a shiny shadow pokemon would not gain the shiny status when it was non shiny non shadow before catching

## Motivation and Context

If you have a regular non-shiny, non-shadow pokemon and you catch a shiny-shadow it does not apply the the shiny status because of the `if (!pokemon.shiny && pokemon.shiny)` check.

## How Has This Been Tested?

Loaded a game where I had a non-shiny, non shadow pokemon. Forced every pokemon to be shiny and attempted to catch it. It applied both shiny and shadow correctly now.

## Types of changes

- Bug fix

### DIFF
--- a/src/scripts/party/Party.ts
+++ b/src/scripts/party/Party.ts
@@ -72,7 +72,7 @@ class Party implements Feature {
                 // Already caught (non shadow) we need to update the party pokemon directly
                 if (this.alreadyCaughtPokemon(pokemon.id, false, false)) {
                     this.getPokemon(pokemon.id).shadow = GameConstants.ShadowStatus.Shadow;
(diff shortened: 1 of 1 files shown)


AevitasDragonkin:
I wasn't quite sure on how it was handled and this seemed to be the easy fix to handle this particular bug. 

I do have another branch where I refactored this gainPokemon function so each status only handles itself. The only change that this would bring is the logging of the shiny when you gain a shiny shadow. The comment said we don't want to add a shiny log because the chance of this happening is pretty low. However, I think it should still log it.

I can create a PR for this, but I wasn't able to test it yet, nor do I really know what all the test cases would be and if there are already ways to easily test this, or if I should just alter my game via the console (I know how) and test it that way.

AevitasDragonkin:
Closed in favor of PR#4904

PR #4901, opened 2023-11-22
Removed Wormadam Changing into other Wormadam

EatPant2nd:
## Description

Removed Wormadam being able to change forms since it's not canon ~~and it causes a loop of constantly watching it change forms if you have a single copy of it~~.

## Motivation and Context

Canon!

## How Has This Been Tested?

Checked that Wormadam didn't become another Wormadam.

## Types of changes

- Bug fix


Qwertypop04 (approved):
Looks goof, thanks for the PR!

PR #4904, opened 2023-11-23
Cleanup gain pokemon

AevitasDragonkin:
## Description

This cleans up the gain pokemon function a bit. Each status will now handle it's own responsibility. No longer checking for shiny status when trying to add a shadow. This will prevent bugs like the shiny status not being applied when you catch a shiny shadow pokemon for the first time.
This closes the PR#4900

## Motivation and Context

There was a bug with catching a new shiny shadow pokemon when you only had the base version where the shiny status was not applied. Had a look into the code and it looked pretty messy by trying to handle multiple cases at once.

## How Has This Been Tested?

Used the party code to remove Qwilfish.
Used the party code gainPokemonById with all the combinations of regular / shiny / shadow.

## Types of changes

- Bug fix
- Code cleanup


CypherX (approved):
I made a minor performance tweak. Looks good, thanks 👍

PR #4905, opened 2023-11-23
Update The Legend Awakened (Genesect) quest step description

CypherX:
## Description
Changes the description of the digging quest step from
`The Red Genesect has crashed into the P2 Lab, leaving a pile of rubble in its wake. Dig through the rubble to learn more.`
to
`The Red Genesect has crashed into the P2 Lab and escaped underground. Start digging to learn more.`

## Motivation and Context
During testing the consensus was the current description was confusing, this change clarifies what needs to be done.

## Types of changes
- Text


Qwertypop04 (approved):
Looks good, thanks for the PR

PR #4911, opened 2023-11-26
Changed Gym Quest Logic to Only Generate For Cleared Gyms

EatPant2nd:
## Description

Currently, Defeat Gym Quests was looking at which gyms were unlocked to the player. This caused situations where the game would ask, for example, for the player to beat a Champion as soon as they were done with the Elite 4 thus making it an "impossible" quest for a bit. This changes the logic from isUnlocked() to clears() ensuring it only picks gyms the player has already cleared at least once. 

Also, makes Defeat Gym Quests only available after obtaining the first badge (Boulder).

## Motivation and Context

It is annoying to get quests you can't really "clear", so this ensures that doesn't happen.

## How Has This Been Tested?

Loaded some test files. Checked that the game generated Defeat Gym Quests accordingly.

## Types of changes

- Logic change


src/scripts/quests/questTypes/DefeatGymQuest.ts
```diff
@@ -13,13 +13,17 @@ class DefeatGymQuest extends Quest implements QuestInterface {
...
     }
 
+    // Only add Defeat Gym Quest if the player has defeated the first gym (Brock).
+    public static canComplete() {
+        return App.game.badgeCase.hasBadge(BadgeEnums.Boulder);
+    }
+
     public static generateData(): any[] {
         const amount = SeededRand.intBetween(5, 20);
         const region = SeededRand.intBetween(0, player.highestRegion());
```
wormania, line 23: Region would need to be restricted, if you roll your highest (non-kanto) region but do not have any gym clears in that region, `calcReward` fails because `possibleGyms` is an empty array so `gymTown` is `undefined`

image

EatPant2nd: Hm... So I guess I should have kept that part of the logic. I see now why it was there. Oops.

EatPant2nd:
So, thanks to Umbral, fixed the issue. Now it checks for first gym of Highest Region, if not there, then the range of regions that the code can pick goes from 0 to Highest Region -1.

CypherX (approved):
Looks good, thanks

PR #4912, opened 2023-11-27
Changed Rare Candy Chance from 1/2048 to 1/1024

EatPant2nd:
## Description

Changed Rare Hold Rare Candy item chance from 1/2048 to 1/1024. On average, it goes from around 4-5 candies after defeating a Pokémon that drops this item 10k times, to around 9 drops.

## Motivation and Context

Rarest Candy was too rare. Now it's Rarer Candy. ~~I dunno, lmao.~~ Seriously though, it was too rare and since mostly challenge runs take advantage of it, now it's less painful for them, and a tiny bit more useful for normal files.

## How Has This Been Tested?

Ran the function that generates Rare Hold item drops a few dozen times with different values to compare previous chance to improved chance.

## Types of changes

- "Balance"


Qwertypop04 (approved):

PR #4918, opened 2023-11-29
Orre map adjustment & remove Furfrou (Heart) from shop

CypherX:
## Description
- Updates the Orre map to reduce the shadow/darkness on Kaminko's Manor so it does not appear as locked (thanks Soul / someone)
- Comments out Furfrou (Heart) from the shop as the primary source of contest tokens is not yet implemented

## Motivation and Context
The shadow/darkness on Kaminko's Manor made it look locked when it was not.

After discussion we determined that while it is possible to obtain contest tokens from the bug catching contest, the Furfrou (Heart) price was balanced around full contests and it should not yet be available.

## How Has This Been Tested?
Loaded the game and checked the Orre map and Furfrou shop.


Qwertypop04 (approved):

PR #4922, opened 2023-11-30
Gigantamax typos and Genesect quest change

Qwertypop04:
## Description
Fixes typos in Peonia's dialogue, and switches around 2 steps in the Genesect quest as their current orientation doesn't make sense

## Motivation and Context
Typos bad, narrative logic good

## How Has This Been Tested?
Have not had the time, sorry

## Types of changes

- Typo


CypherX:
I think you need to also update the two `QuestLineStepCompletedRequirement` for the `Red Genesect 2` temporary battle since the step is one earlier now, otherwise the battle does not appear and the quest is stuck.

Qwertypop04:
Ah, so I do. Done!

PR #4933, opened 2023-12-02
Genesect quest ordering fix and Citadark typo

Qwertypop04:
## Description
Fixes the Genesect quests order because it was still wrong after I fixed it the first time. Also typo. Also better Max Lair town image

## Motivation and Context
Correct ordering good, typo bad

## How Has This Been Tested?
no

## Types of changes
- Typo


src/assets/images/towns/Max Lair.png
osherwott, line 1: Might want to compress this png, no?

Qwertypop04: Right, yeah. Done.

PR #4934, opened 2023-12-02
Fix friend safari rotation seeding

CypherX:
## Description
During load the friend safari pokemon list shuffling would result in an unexpected order, it appears something was causing the rand seed to change. Moving a couple things around in the generation has corrected the issue.

## Motivation and Context
Bug

## How Has This Been Tested?
Loaded a save that was encountering the issue on live and checked that the rotation was now correct. Also had several other players who reported the issue test.

## Types of changes
- Bug fix


wormania:
> it appears something was causing the rand seed to change

Specifically: getting the dungeon requirements for Gigantamax Pokemon calls their requirement function which calls `next()` on the SeededRand many times while `getPokemonLocations` is being run. This fix also future-proofs against any other re-seeds during the location finding

Qwertypop04 (approved):

PR #4935, opened 2023-12-02
Fix christmas present currency image not loading on client

CypherX:
## Description
Fixes the broken currency image in the christmas present notification on the client

## Motivation and Context
Bug

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #4939, opened 2023-12-03
Add ability to hide gyms

osherwott:
## Description
Can now hide gyms if they're not unlocked or have a visible requirement

## Motivation and Context
Better for paldea gyms whose towns are unlocked early
Useful for future optional content

## How Has This Been Tested?
Tested it on a regular gym and changed around the requirement.

## Types of changes
- UI improvement


Qwertypop04 (approved):

PR #4941, opened 2023-12-04
Collapsible menu performance improvements

umbralOptimatum:
Changed the pokemonList and breedingDisplay collapsible menus to not render their contents while collapsed for better performance. As part of this, I reworked modalUtils so the code can be reused for non-modal Bootstrap display elements, in this case `collapse`s.

I've thoroughly tested all the affected menus and all are working correctly. ~~The only problem is a visual issue with expanding the pokemon list. As the list contents aren't in the DOM until right after it's triggered, the collapsible element first expands to the height of the list header before snapping to the full height of the contents. This doesn't cause any problems with the game but is a bit disorienting for the player as the background changes size much more jerkily. I added a hacky workaround, but I'm not familiar enough with Bootstrap to know the right approach here—help would be much appreciated.~~


CypherX:
Pokedex is a bit borked, looks like just a rogue copy/paste or something left over from resolving a merge conflict

image

CypherX (approved):

PR #4944, opened 2023-12-04
Farm button to reverse plot locks

### LINKED ISSUE #4943: [SUGGESTION]: Farm button to invert plot locks
### Current Behavior

N/A

### Improved Behavior

A button to invert the lock status of all farm plots.

### Reason for change

Locking is very convenient for farm setups where some plots are regularly harvested and others are reserved for long-lived aura berries. When the boosting berries do need to be replanted, it's a little tedious to unlock all of those plots and replant them by hand. It would be much easier to click one button to invert all the plot locks, harvest and plant all, and then invert the locks a second time.

### Detailed Description

Straightforward enough.

umbralOptimatum:
Added a button to the farm that toggles all plot locks to the opposite state. Resolves #4943.


CypherX (approved):
Looks good, thanks

PR #4950, opened 2023-12-06
Fix Max Raid Cinderace's incorrect return town

wormania:
## Description

Fixing return town typo `Brawler's Cave` -> `Brawlers' Cave`

## Motivation and Context

Incorrect spelling was causing the return town of the Challenge Road Max Raid (Cinderace) to default to the dock town of Hulbury

## How Has This Been Tested?

Checked that you get correctly returned to the Cave

## Types of changes

- Bug fix


Qwertypop04 (approved):

PR #4952, opened 2023-12-06
Fix: dungeon "freeze" when restarted the frame lost on boss tile

VodkaFR:
## Description 
I don't think there is any reason to continue the tick() computation if dungeon already ended. So i propose to return the tick after handling dungeon win or loss. 

## Motivation and context
Made to fix a bug. 
https://discord.com/channels/450412847017754644/1182020200598610010
Loosing on a boss fight while holding spacebar restarts the dungeon but you are unable to move and end the dungeon. 
Require reloading to recover.

Bug cause : 
In the DungeonRunner, tick function doesn't end when dungeon is win or lost. 
This mean when you loose on a tileboss tick still handles playerinput this tick (and flute) 
In this case handle interaction to start boss fight again (because dungeonLost marks dungeonfight to false) 
Player is now considered in a boss fight when there is no fight, and thus is locked in the dungeon (since timer didn't start) 

## How Has This Been Tested?
Couldn't reproduce the bug and didn't see any impact on dungeon with a few runs.


CypherX (approved):
Seems to do the trick, thanks for the fix 👍

PR #4953, opened 2023-12-06
Changed No Active Quest Text + National Park

EatPant2nd:
## Description

Changed text in the quest display from "No active quest" to a more explicit "click List" one. Also, deleted the part of the NPC dialogue in National Park about "finding different bugs each day" as that's untrue.

## Motivation and Context

~~How Green Coin?~~ Making text clearer helps the player. And the Safari one was just wrong.

## How Has This Been Tested?

Checked the corresponding texts.

## Types of changes

- Text change


Qwertypop04 (approved):

PR #4954, opened 2023-12-07
Added an option to optimize vitamins to the vitamin menu

RedzepR:
## Description

Added a button to the vitamin menu (see screenshot) that allows on a single click to assign the optimal vitamins based on region to a mon.

## Motivation and Context

For almost all players having the optimal vitamins is the correct way of assinging vitamins. It's not data you can find in the game itself so you either need to calculate it yourself or use a 3rd party tool (companion or wiki) to actually figure it out. Afterwards you have to manually input the vitamins with the option of using 1, 5, 10 or max at one time. Yet most mons do not have optimal spreads that are easily input with these options. For Gyarados in galar you have to use 3x 10 Carbos, 2x 1 Carbos, switch to Calcium, either 10x or max Calcium. Thats a lot of clicks if done for a bunch of pokemon. To speed up the process I added the button.

## How Has This Been Tested?

File set in Galar: 
- Randomly assigned vitamins to gyarados, clicked the button, got correct spread
- Assigned no vitamins to decidueye, clicked the button, got correct spread
- Didn't have enough vitamins for optimal spread for magikarp, clicked the button, got a warning

File set in Kanto:
- Assigned 5 protein to gyarados, clicked the button, got all vitamins removed
- Assigned no vitamins to bulbasaur, clicked the button, got 5 proteins

File with no vitamin challenge:
- Clicked the button, got error

## Screenshots (optional):

image
image
image
https://github.com/pokeclicker/pokeclicker/assets/143606926/31218a61-e2a3-49e8-ab94-10b21483e568

## Types of changes

- New feature
- UI improvement


RegisCoaxans:
That does not look right to me.

The core idea behind adding more vitamins was to give the players some sort of customization, depending on Pokémon and needs.
If such option was to be added, then not only would designing the vitamins have been a waste of time — it could have been a rework of the attack formula instead — but, then, to the players, it would also mean there is only one way to properly use those.

At this point, returning to the one and true protein would be better than keeping a whole system with one button to ruin its [probably unnecessary] complexity. Whatever its cost, by the way.

CypherX:
Assigning vitamins is far from perfect but having a button to automatically assign optimal amounts is not a solution, it removes all engagement with the mechanic.

I do agree, and I mentioned it earlier in discord, that vitamins could benefit from (another) rework where more options are added that modify other parts of the game rather than just maximizing BE with every pokemon having a single optimal distribution. That is a much larger undertaking though.

RedzepR:
> For the UI, adding an entire column of identical buttons to the table looks pretty clunky.
I 100% agree. I'm no UI designer and I think it shows lmao. 
> I can open a PR for it on your fork or open an alternate PR here, whichever you'd prefer.
You can dm me on discord, that would be my prefered way. Unless you want your name on the code than PR to my fork.

> While I do like the automation, I feel like this should at least have some requirements to it, either by unlocking something, or by the optimisation having a cost.
Thats pretty reasonable. Everything in this game that does something automatically or saves the player time (helpers, autobattle) costs some kind of currency. I can't think one of that is a) fitting b) accessible enough and c) not so irrelevant that its not really a cost at all (looking at you, coins)

> The core idea behind adding more vitamins was to give the players some sort of customization, depending on Pokémon and needs.
There is exactly 2 ways of "customizing" of vitamins. Full carbos or optimized vits. There are maybe a handful player that play with a self imposed "no carbos" challenge or something but thats not really worth to take into consideration.

> I do agree, and I mentioned it earlier in discord, that vitamins could benefit from (another) rework where more options are added that modify other parts of the game rather than just maximizing BE with every pokemon having a single optimal distribution. That is a much larger undertaking though.
Also agree. I wish there was a more meaningful way to assign vitamins but currently its "big number go brrr" and before that changes there needs to be a better and less time consuming way to assign those.

RedSparr0w:
> > While I do like the automation, I feel like this should at least have some requirements to it, either by unlocking something, or by the optimisation having a cost.
> 
> Thats pretty reasonable. Everything in this game that does something automatically or saves the player time (helpers, autobattle) costs some kind of currency. I can't think one of that is a) fitting b) accessible enough and c) not so irrelevant that its not really a cost at all (looking at you, coins)

Only thing I can really think of is having a one off Vitamin cost to unlock it
(just random numbers)
1k Carbos - Unlocks apply max carbos to all (low egg steps, good for shiny hunting)
1k Protein - Unlocks apply max protein to all (idk why you would)
1k Calcium- Unlocks apply max calcium to all (idk why you would)

After all 3 unlocked - Unlocks apply max efficiency to all

AevitasDragonkin:
In my opinion, even with an unlock cost, it should also require a cost for each auto-apply. Not sure how harsh the cost would be, let's take 10% as an example.

**Example** - Gyarados in Galar with 10% cost
Optimal: 8 Calcium + 32 Carbos
10% cost: 0.8 Calcium rounded to 1 + 3.2 Carbos rounded to 4
Total needed: 9 Calcium + 36 Carbos.

When auto-applying vitamins to Gyarados it would check if you have 9 Calcium and 36 Carbos. It would apply 8 Calcium and 32 Carbos, and then also remove 1 Calcium and 4 Carbos from your inventory.

PR #4958, opened 2023-12-08
Consistent shiny image hiding

### LINKED ISSUE #4977: [Bug]: Hide Shiny option in pokedex does not work
### Confirmations

### What happened?

Toggling the Hide Shiny option in a pokedex entry appears to not do anything.

All images of the pokemon, including the one in the pokedex entry would appear to be not shiny.

### Steps to Reproduce

Open pokedex. Open a pokemon that has been caught shiny. Turn on "Hide Shiny" option.

Windows & Android

Other (Tell us in Additional Information)

### Relevant log output

_No response_

### Additional Information

I use Brave browser. It is Chromium-based.

umbralOptimatum:
The settings to hide shiny images are confusing and inconsistently labeled. One setting only works in the pokedex, a second setting only works in the statistics modal but is unlabeled, and the per-pokemon setting is clearly labeled in the statistics modal but does _not_ work there. This is a mess.

- Merged the pokedex and statistics shiny settings into an actual Setting
- Party pokemon images now (almost) always use both the per-pokemon and partywide hide-shiny-image settings
- Removed non-obvious "click pokeball to toggle setting" button in statistics modal
- Cleaned up `PokemonHelper.getImage()` a bit cause why not

Closes #4977.


src/components/pokedex.html
```diff
@@ -182,14 +182,17 @@ <h4>Filters and Settings</h4>
                                     <!--Hide shiny images-->
                                     <div class="col-12">
                                         <div class="form-group my-2">
-                                            <span>Hide shiny images</span>
+                                            <!-- let: { partyHideShinyImages: Settings.getSetting('partyHideShinyImages') } -->
```
CypherX, line 185: I don't think this `let` is used/needed, or formatted right

CypherX:
Changes look good other than the one thing I left a comment on. A couple other notes:

- I don't think the global "Hide Party Shiny Images" toggle is needed in the individual statistics modal
- Thoughts on changing "image/s" in `Hide Shiny Image` and `Hide party shiny images` to "sprite/s"?

CypherX:
Makes sense, I do see how some players might get confused why the toggle isn't doing anything when they have the party toggle enabled.

A possible alternative is a tooltip on the individual shiny toggle switch when the party setting is enabled warning the player that it's being overridden. Do you think that could work or would the party toggle still be preferable in your opinion?

CypherX:
Let's do the tooltip for now, if we end up getting a lot of questions or confused players it can always be swapped back.

CypherX:
Sure, just pushed it, take a look

CypherX (approved):
Looks good. Definitely an improvement to how the sprite toggles previously worked. Thanks!

PR #4960, opened 2023-12-09
Fix : Baby Pokémon calculation

RegisCoaxans:
## Description
Pokémon and their baby are now correctly mapped and calculated.
Breeding the last stage will always give the corresponding baby now.
This also prevents the wiki from showing mid stage Pokémon as babies from their last stages.

## Motivation and Context
`PokemonHelper.getPokemonParents` currently returns mid stage Pokémon as baby from their evo.
It also implies the player could erroneously get those Pokémon from breeding their last stages in some cases.

## How Has This Been Tested?
~~This is untested but so little has changed that no testing should be required.~~
Loaded the game and checked the calculation function returned the expected values.
I also verified eggsteps, just in case...

## Types of changes
- Bug fix


CypherX (approved):
Looks good, thanks

PR #4961, opened 2023-12-09
Prevents the player from completing daily tasks before reset hits on loading

RegisCoaxans:
## Description
Made the `Save.counter` reach its activation tick instantly when loadig a file.

## Motivation and Context
Players reported being able to do daily stuffs twice a day by doing them before the daily check happens.
Worked on quest refreshes and daily tempbattles, for instance.

## How Has This Been Tested?
Just loaded the game...

## Types of changes
- Bug fix


CypherX (approved):
Seems to do the trick, thanks

PR #4964, opened 2023-12-10
Add sound to max flow notification

AevitasDragonkin:
## Description

Adds sounds to the notification when reaching max flow in the purify chamber. Since we can't use original soundbites, I tried to use a similar chimes sound. Got it from here if sources are needed: https://pixabay.com/sound-effects/mystical-wind-chimes-transition-fx-6327/

## Motivation and Context

A lot of discord requests to make it have a sound

## How Has This Been Tested?

Make sure you have at least 1 pokemon available that can be purified. Make sure sounds are enabled.
App.game.purifyChamber.gainFlow(-1000)
App.game.purifyChamber.gainFlow(1e12)

## Types of changes

- UX improvement


Qwertypop04 (approved):

PR #4978, opened 2023-12-18
Fix: Use setInterval instead of setTimeout for game loop.

Candor-Zofain:
Prevents the callstack from overflowing over time.

## Description

I've changes the main `tick()` loop to use setInterval instead of a self-referencial setTimeout to prevent long-term lag.

## Motivation and Context

I've noticed that the game can lag if played for too long. (Mobile/Browser). While I was messing around in the code I noticed that when break in the `tick` function the callstack becomes endlessly large. This is essentially a memory leak.
Using `setInterval` instead of `setTimeout` that the `requestAnimationFrame` uses prevents that issue and is functionally the same.

## How Has This Been Tested?

After starting up a file in my brower, I ran `App.game.stop()` to stop the current implementation.
I then ran the following code to emulate what this new code will do:
```javascript
{
    let lastFrameTime = 0;
    let ticks = 0;
    App.game.frameRequest = setInterval(() => {
        if (document.hidden) { return; }
        const currentFrameTime = Date.now();
        const delta = currentFrameTime - lastFrameTime;
        ticks += delta;
        lastFrameTime = currentFrameTime;
        if (ticks >= GameConstants.TICK_TIME) {
            if (ticks >= GameConstants.TICK_TIME * 2) { ticks = 0; }
            else { ticks -= GameConstants.TICK_TIME; }
            App.game.gameTick();
        }
    }, 1000 / 60);
    console.log('test');
}
``` 
I checked various parts of the game and they still ran as expected.
Using the log to set a breakpoint, I could verify that the callstack remained constant and that the game ticks were properly set to 0 if the loop ran too slow. (e.g. if the game was paused or lagged in some other way)

## Types of changes

- Bug fix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "pokeclicker",
-  "version": "0.10.14",
+  "version": "0.10.17",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "pokeclicker",
-      "version": "0.10.14",
+      "version": "0.10.17",
       "license": "ISC",
       "dependencies": {
         "bootstrap": "^4.5.3",
--- a/src/scripts/Game.ts
+++ b/src/scripts/Game.ts
@@ -273,13 +273,13 @@ class Game {
         // requestAnimationFrame (consistent if page visible)
         let lastFrameTime = 0;
         let ticks = 0;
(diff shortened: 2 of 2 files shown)


CypherX:
I'm going to close this, I don't believe it's something we want

PR #4979, opened 2023-12-19
Replace Pink Bow with Fairy Feather

### LINKED ISSUE #4804: [SUGGESTION]: Replace Pink Bow with Fairy Feather
### Current Behavior

Pink Bow is the current fairy boosting held item, which is custom and unofficial.

### Improved Behavior

It should be changed to the official Fairy Feather. 

### Reason for change

Using official items is better than using unofficial ones.

### Detailed Description

The Teal Mask DLC finally added an official fairy boosting held item that isn't the pixie plate with the Fairy Feather. The Pink Bow we currently use should be changed into this official one.

kittenchilly:
## Description
Replaces the unofficial fairy held item, Pink Bow, with the new official fairy type held item, Fairy Feather.
Works identically, just cosmetic.

## Motivation and Context
We should use official content instead of fan-made content where we can.
Closes #4804 

## How Has This Been Tested?
I tested everywhere Pink Bow used to be and Fairy Feather seems to look and work fine.

## Types of changes
- UI improvement


CypherX:
You also need to add code to the 0.10.18 update function in Update.ts to handle renaming the item in the save file. The item name is used as the key for the item list and equipped held items are stored as the item name. Without this players will lose any they have in their inventory or any currently equipped to a pokemon.

kittenchilly:
> You also need to add code to the 0.10.18 update function in Update.ts to handle renaming the item in the save file. The item name is used as the key for the item list and equipped held items are stored as the item name. Without this players will lose any they have in their inventory or any currently equipped to a pokemon.

Honestly not sure what goes into held item save data, is what I did enough?

CypherX:
That works for the item list but you'll also need something like this to update pokemon held items
```
// Update pokemon held item Pink Bow -> Fairy Feather
saveData.party.caughtPokemon.forEach(p => {
    if (p[10] === 'Pink_Bow') {
        p[10] = 'Fairy_Feather';
    }
});
```

kittenchilly:
> That works for the item list but you'll also need something like this to update pokemon held items
> 
> ```
> // Update pokemon held item Pink Bow -> Fairy Feather
> saveData.party.caughtPokemon.forEach(p => {
>     if (p[10] === 'Pink_Bow') {
>         p[10] = 'Fairy_Feather';
>     }
> });
> ```

Alright, thanks

kittenchilly:
Should be reviewable again.

CypherX (approved):
Looks good, thanks 👍

PR #4982, opened 2023-12-21
Use Sport Ball asset in National Park

kittenchilly:
## Description
Replaces the Safari Ball asset with the Sport Ball asset only in Johto's Bug Catching Contest. This is entirely cosmetic and doesn't effect how catching works.

## Motivation and Context
We had the svg for it, and I don't think we'll ever find another use for it. I think this just polishes the game more.

## How Has This Been Tested?
Tested both national park and the other safari zones to make sure the pokeball sprite is correct in all of them.

## Types of changes
- UI improvement


Qwertypop04 (approved):

PR #4984, opened 2023-12-23
BattlePokemonGender type fixes

umbralOptimatum:
Changing a few helper functions to consistently use the actual BattlePokemonGender types instead of random `number`s and `boolean`s. This way it's possible to just pass in a pokemon's data without extraneous conversions. Also cleaned up `PokemonHelper.getImage()` in the process.


CypherX (approved):
Looks good, thanks

PR #4990, opened 2023-12-28
Setting to toggle pokeball filter priority order

CypherX:
## Description
Adds a setting to toggle the catch filters priority order between top-to-bottom and bottom-to-top.

Default priority order changed to top-to-bottom as most players assume this is the standard priority.
Existing players will have the setting enabled by default to preserve the current bottom-to-top functionality.
Added text to filters UI indicating the current priority order.
Filters will be reversed upon toggling the setting.

## Motivation and Context
We get a lot of questions and bug reports about filters not working because the assumed priority is top-to-bottom.

## How Has This Been Tested?
Loaded the game, toggled the setting, re-arranged the filters.

## Screenshots (optional):
image
image
image


wormania:
Should `PokeballFilters.createFilter()` shift or unshift depending on what the priority order is?

CypherX:
> Should `PokeballFilters.createFilter()` shift or unshift depending on what the priority order is?

Considered that, but I think the new filter showing up right under the button is more intuitive.

RedSparr0w (approved):
Changes look good to me, thanks for the PR! 👍

PR #4992, opened 2023-12-29
Update SeededRand to fix poor randomisation

wormania:
## Description

Changed SeededRand from using an LCG to a hybrid of xorshift and MCG - xorshift for  high-quality randomisation when `next()` is called in succession, MCG to expand the date seed to a wider range of variables (and to hide some of the rough edges of javascript's bitwise operator handling)

## Motivation and Context

From 2024 the existing LCG ends up producing near identical results with each daily seed, only breaking the pattern once many days or months have passed. Even before 2024, the strictly linear nature of the current generation means that things like Enigma hints always follow a set A -> B -> C -> D pattern, often getting stuck on the same hint for many days

## How Has This Been Tested?

I've compiled 2020-2030s RNG output for Enigma/Unown here to compare the randomisation: https://docs.google.com/spreadsheets/d/1CiB5J2AF2QKyABfE88sq7rR4DExXApICT6fCPotgd3M/edit#gid=0

Other things using SeededRand like Underground Deals have also been generated with a wide selection of dates to check that it functions correctly and produces random "feeling" results - no comparison for these as the current generation method doesn't appear obviously un-random.

## Types of changes

- Bug fix


Aegyo (approved):
The code seems fine, but looking at a list of numbers to figure out if this is really random is kind of difficult. Instead, I've used some code to generate images which are a little easier to use for judging randomness.

## Code

```js
// draw bar chart with heights indicating frequency
// Useful for checking how uniform the output is
function testUniform(size, times, res = 1) {
  const data = Array(size).fill(0);

  const canvas = document.createElement('canvas');
  canvas.width = size * res;
  canvas.height = size * res;
  canvas.style.position = "absolute";
  canvas.style.zIndex = 999;
  document.body.append(canvas);

  const ctx = canvas.getContext('2d');
  ctx.fillStyle = "white";
  ctx.fillRect(0, 0, canvas.width, canvas.height);

  for (n = 0; n < times; n++) {
    const cur = SeededRand.intBetween(0, size-1);
    data[cur]++;
  }

  const high = Math.max(...data.flat())

  for (i = 0; i < data.length; i++) {
    const v = data[i];
    const height = (v / high) * canvas.height
    ctx.fillStyle = `rgb(0,0,0)`;
    ctx.fillRect(i*res, canvas.height - height, res, height);
  }
}
```

```js
// Draw 2D image with opacity at (x,y) indicating frequency
// Useful for checking if there is some bias in consecutive outputs
function testCorrelation(size, times, res, next = (max) => SeededRand.intBetween(0, max)) {
  const data = Array(size).fill(null).map(_ => Array(size).fill(0));

  const canvas = document.createElement('canvas');
  canvas.width = size * res;
  canvas.height = size * res;
  canvas.style.position = "absolute";
  canvas.style.zIndex = 999;
  document.body.append(canvas);

  const ctx = canvas.getContext('2d');
  ctx.fillStyle = "white";
  ctx.fillRect(0, 0, canvas.width, canvas.height);

  let last = next(size-1);
  for (n = 0; n < times; n++) {
    const cur = next(size-1);
    data[last][cur]++;
    last = cur;
  }

  const high = Math.max(...data.flat())

  for (i = 0; i < data.length; i++) {
    for (j = 0; j < data[0].length; j++) {
      v = data[i][j];
      ctx.fillStyle = `rgba(0,0,0,${v / high})`;
      ctx.fillRect(i*res, j*res, res, res);
    }
  }
}
```

## Tests

Using these I've generated four images to compare the old and new rng
- Large uniform check
  - `testUniform(128, 100_000, 4)`
- Uniform check for "10 years" of single polls 
  - `testUniform(32, 365*10, 16)`
- Large correlation check
  - `testCorrelation(128, 100_000, 4)`
- Correlation check for 10 years from today (Dec 31 2023), using date seed
  -  `d = new Date(); testCorrelation(32, 365*10, 16, (m) => {d.setDate(d.getDate()+1); SeededRand.seedWithDate(d); return SeededRand.intBetween(0, m)})`
  
## Results

Result images are in the same order as the list above

### Old

old_unform_100000

old_uniform_3650

old_correlation_100000

old_correlation_3650

<hr>

### New

new_uniform_100000

new_uniform_3650

new_correlation_100000

new_correlation_3650

## Conclusion

- It looks like we lose a little on uniformity in the 100_000 test, but it isn't hugely bad.
- The "10 year" uniformity looks slightly better imo.
- The 100_000 correlation test appears to have more locations which are not hit at all, however these are spread seemingly randomly so I don't think it is a huge deal.
- The 10 year correlation test has obviously resolved the very clear issue in the old rng without introducing any obvious new pattern.

Aegyo:
Those `testUniform` images are basically just black squares, so here's a slightly different representation which focuses on deviation from the expected value
```js
// draw bar chart with heights indicating frequency
// Useful for checking how uniform the output is
function testUniform(size, times, res = 1) {
  const data = Array(size).fill(0);

  const canvas = document.createElement('canvas');
  canvas.width = size * res;
  canvas.height = size * res;
  canvas.style.position = "absolute";
  canvas.style.zIndex = 999;
  document.body.append(canvas);

  const ctx = canvas.getContext('2d');
  ctx.fillStyle = "white";
  ctx.fillRect(0, 0, canvas.width, canvas.height);

  for (n = 0; n < times; n++) {
    const cur = SeededRand.intBetween(0, size-1);
    data[cur]++;
  }

  const expected = times / size;

  for (i = 0; i < data.length; i++) {
    const v = data[i] - expected;
    const height = (v / expected) * canvas.height / 2
    ctx.fillStyle = `rgb(0,0,0)`;
    ctx.fillRect(i*res, (canvas.height / 2) - height, res, height);
  }
}
```
## Old
old_uniform_dev_100000
old_uniform_dev_3650

## New
new_uniform_dev_100000
new_uniform_dev_3650

CypherX (approved):
Looks great, thanks!

I guess players _won't_ have to wait until 2024-12-15 for Unown (P) after all

PR #4994, opened 2023-12-30
Fix : day-related requirement to correctly recalcultate on new day.

RegisCoaxans:
## Description
Provided GameHelper with two observables for « today » and « tomorrow ».
Max Raid tempbattles automatically turn on and off on new day. Available Unown properly swaps as well.

Also, `player._lastSeen` only updates on save tick now.

## Motivation and Context
Fix Raid Dens not refreshing without reloading the game.
Saving on closing used to change `player._lastSeen`. As a side effect, reloading after midnight, before the new day check hit, was preventing it entirely.
Note : I had to take day computation out of `GameHelper.tick` so it stays synchronised with `Save.tick`.

## How Has This Been Tested?
Timetravelled asf.

## Types of changes
- Bug fix


CypherX (approved):
Looks good. Raid dens all appear correctly at midnight now, other time and day related things all seem to work as expected.

PR #4995, opened 2023-12-30
Show map tooltip when hovering over routes, to match dungeons/towns/etc.

wormania:
## Description

Map hover tooltip now shows when hovering routes, and the Kanto-Johto connecting fake routes have their tooltips updated to show the Region name, to be consistent

## Motivation and Context

All other clickable map elements show the tooltip on hover, this makes it consistent, and also makes it easier to decipher routes in later regions where they are acronyms rather than just numbers

## How Has This Been Tested?

Hovered over all types of map icons and checked the tooltip shows correctly

## Screenshots (optional):

Standard route:
image

Named route:
image

Updated Johto connection route:
image

## Types of changes

- UI improvement


CypherX (approved):
Have definitely wondered why this wasn't a thing in the past.

Looks good, thanks!

PR #4997, opened 2024-01-01
[Accessibility] Location Navigator

CypherX:
## Description

**See comment further down**

~~Adds a module to the main view (disabled by default, toggled with a setting on Gameplay tab)~~ to assist with map navigation. This module will display the towns, routes, and dungeons in the players current (sub)region with the ability to click on a location to navigate to it, as if the location was clicked on the map.

Locations that are locked are indicated by dimmed text and a lock icon, and if clicked will display the same hint as if it was clicked on the map.

In areas where the map is not available, such as Battle Frontier and while running a dungeon, the module cannot be used and will display a message indicating such.

~~One side-effect is the hidden pixel hunt locations are listed, but I'm not too worried about that (and pixel hunts are kinda meh)~~

## Motivation and Context
Some players have difficulty navigating with the map. This can be a result of a smaller screen size, locations on the map being hard to identify, varying degrees of blindness or color blindness, or other disability. This (hopefully) helps those players navigate the various areas of the game.

## Screenshots (optional):
image
image
image
image
image

## Types of changes
- UI improvement
- Accessibility


AevitasDragonkin:
Could potentially include the Key Item [https://bulbapedia.bulbagarden.net/wiki/Bicycle](url) with this?

BalmungGame:
I suggest the addition of a button to show completion badges (similar to map color legends):

image

None filter will only show by default Locked Location, Incomplete Area (Orange text), and Quest Location (Green Text). Text color uses what is set in the settings.
When switching to others, they will show additional badges/icons:

- Pokémon: Uncaught Pokémon, Uncaught Shadow, Uncaught Shiny
- Pokérus: Infected, Contagious, Resistant
- Achievement: Completed all achievements (green), Missing at least 1 (red)

CypherX:
This has been revamped. Instead of a main screen module it is now a dropdown on the Town Map header. Secret locations do not appear in the list until the player has visited them once via normal means. The issue with the various Plasma Frigates in Unova appearing has been resolved.

image  


There are likely other quest-related towns that need updating with the new flags. Can someone also double check the added requirements to the frigates, these things are a mess and need to be reworked with cleaner requirements.

CypherX:
> I suggest the addition of a button to show completion badges (similar to map color legends):
> 
> image
> 
> None filter will only show by default Locked Location, Incomplete Area (Orange text), and Quest Location (Green Text). Text color uses what is set in the settings. When switching to others, they will show additional badges/icons:
> 
>     * Pokémon: Uncaught Pokémon, Uncaught Shadow, Uncaught Shiny
> 
>     * Pokérus: Infected, Contagious, Resistant
> 
>     * Achievement: Completed all achievements (green), Missing at least 1 (red)

I like this idea, but don't want to stuff everything into one PR. Will likely implement additional things in another PR once (if) this is merged

PR #4998, opened 2024-01-01
Fix: Modified the requirements for Distortion World during A New World

RegisCoaxans:
## Description
Properly locked Distortion World behind the corresponding questline step of A New World.

## Motivation and Context
The Distortion World was locked behind completion and uncompletion of dungeons, which was fairly insecure due to few old saves not meeting the requirements for A New World nor Zero's Ambition as a fallback.

## How Has This Been Tested?
Loaded normal and bugged saves and went through both questlines with no issue.

## Types of changes
- Bug fix ? Bug fix fix ?


Qwertypop04 (approved):

PR #5000, opened 2024-01-02
Dungeons can have their own requirements

osherwott:
Added the optional dungeon parameter of `requirement`, which lets us use dungeon towns while locking the dungeons themselves.

## Description
Dungeons can now have separate requirements from their towns. This lets us use dungeon towns without unlocking dungeons. This can be used for story towns like the Plasma Frigates or Slumbering Weald, where it is preferred to have npcs or temporary battles in the dungeon town be accessible, but not the dungeon itself.

## Motivation and Context
#4997 introduces a list of all towns, but some of them are only used for story purposes, and are otherwise filler. It's best to cut down on duplicate locations, since it will cause the list to bloat, as well as confuse the player.

## How Has This Been Tested?
Set an optional requirement for a dungeon that is separate from its town, went there, and confirmed the dungeon itself was locked.

## Screenshots:
Proof of Concept (the requirement is wrong and only there for testing purposes)
image

## Types of changes
- UI improvement


Qwertypop04 (approved):

PR #5002, opened 2024-01-03
Clean up and module-ify MapHelper

umbralOptimatum:
Created a couple MapHelper functions for changing regions so less of the logic is in the HTML.

~~I also noticed that `moveToRoute()` and `moveToTown()` don't check if the player can safely move across regions. I didn't touch them as I don't know if either is intentionally used to change regions, but both should be changed to either use the new functions for appropriate safety checks or to not allow cross-region movement at all.~~


src/scripts/worldmap/MapHelper.ts
```diff
@@ -248,6 +245,26 @@ class MapHelper {
...
+    public static accessToShip() {
+        return player.highestRegion() > 0 && TownList[GameConstants.DockTowns[player.region]].isUnlocked();
+    }
+
+    // For moving between already-unlocked regions
+    public static moveToRegion(region: GameConstants.Region) {
+        if (MapHelper.accessToShip()) {
+            if (region <= player.highestRegion() && region <= GameConstants.MAX_AVAILABLE_REGION && region !== GameConstants.Region.none) {
+                player.region = region;
+                player._subregion(0);
```
RegisCoaxans, line 257: `moveToTown` already does [that](https://github.com/pokeclicker/pokeclicker/blob/8ff2820f3672f316748dc3fbec4f5590d92ee3c7/src/scripts/worldmap/MapHelper.ts#L212) so that line can be removed safely. Same comment applies to line 287.

src/scripts/worldmap/MapHelper.ts
```diff
@@ -248,6 +245,26 @@ class MapHelper {
...
 
+    public static accessToShip() {
+        return player.highestRegion() > 0 && TownList[GameConstants.DockTowns[player.region]].isUnlocked();
+    }
+
+    // For moving between already-unlocked regions
+    public static moveToRegion(region: GameConstants.Region) {
+        if (MapHelper.accessToShip()) {
+            if (region <= player.highestRegion() && region <= GameConstants.MAX_AVAILABLE_REGION && region !== GameConstants.Region.none) {
+                player.region = region;
```
RegisCoaxans, line 256: Should we move that line to `moveToTown` so we do not have to worry about it anytime we want the player to move to another region ? Or is it just too niche of an occurrence to be worth it ?

RegisCoaxans (approved):
Looks good overall, no issue while testing.

PR #5003, opened 2024-01-04
EV decimals displayed in Pokémon Statistics

RegisCoaxans:
## Description
Self explanatory.
The text is muted because decimals are ignored in the multiplier computation, but it hints that some progress happened.

## Motivation and Context
When discovering EV and Pokérus, some players get confused at that 0EV seemingly not going up.

## How Has This Been Tested?
My cat told me it worked.

## Relevant screenshots [from cycling themes](https://github.com/pokeclicker/pokeclicker/pull/4368#issuecomment-1567424486)
image
image
image
image
image

## Types of changes
- UI improvement


AevitasDragonkin:
Maybe the rounding here should be dependant on the ratio that's being used? As EV's are calculated based of effort points (which are rounded to the nearest whole number) and then divided by 1000 to get the EV, so it should probably show up to 3 decimal digits.
Or 4 with the slow EV challenge active.

Thoughts?

RegisCoaxans:
Following your comment, I changed it to 3 significant digits, so even slow EV can see that non-integer mulitpliers are effectively accounted for, and also because it matches the significant digits of the EV bonus below.
Although, I do not think dynamically computing how many significant digits we use with the challenge is relevant enough.

wormania:
I'm not sure the meaning of the muted decimals is obvious unless you already know that only full EVs count - maybe better to explain in a tooltip?

RegisCoaxans:
Can the players not figure this out with how the EV bonus only accounts for full EVs ?
My issue with a tooltip is that it is way less obvious as it requires the player to hover over the row.

wormania:
> Can the players not figure this out with how the EV bonus only accounts for full EVs

Sure, but if they can figure it out then we don't need the muting either - my issue is that the muting looks somewhat out of place (it's not done anywhere else) but is also not explicit about what it means, so seems like a poor middle ground.

Either a straight standard display, or an information tooltip like:
image
Depending on whether you want it explained or not

RegisCoaxans:
You are right... It will feel inconsistent.
I opt for the standard display, then.

CypherX (approved):
Looks good, thanks

PR #5008, opened 2024-01-06
Added dummyEvolutions for the gmax that didnt have them

RedzepR:
## Description

Added evolution data for gmax snorlax and all 4 forms of urshifu

## Motivation and Context

The data should be consistent even if it has no effect on the actual game

## How Has This Been Tested?

Went into game and saw that the mons still had the same amount of egg steps

## Types of changes

- idk? data i guess


Qwertypop04 (approved):
Sure, why not. Thanks for the PR

PR #5009, opened 2024-01-06
[quests] Fixup loading quests that no longer exists

RedSparr0w:
## Description
This should fixup loading any Gym quests that no longer exist
Should also fix if any dungeons are removed too (not tested though)

## Motivation and Context
Allows saves to load if they had a quest to defeat a removed Gym
https://discord.com/channels/450412847017754644/1193200676688367677

## How Has This Been Tested?
Loaded save pre changes, would not load
Loaded same save post changes, loaded without the quest instead replaced with:
image

## Types of changes
- Bug fix

## Save file
[save_pokeclicker.txt](https://github.com/pokeclicker/pokeclicker/files/13851321/save_pokeclicker.txt)
Used save file above for testing, `Ilima's Trial` was the quest causing the failed load


CypherX (approved):
Does the trick, thanks! Tested the dungeon part and it also works.

PR #5010, opened 2024-01-06
Update Necrozma catch chance

Qwertypop04:
## Description
Updates Necrozma's catch chance from 255 to 3 as it was updated in the most recent games

## Motivation and Context
Up to date statistics

## Types of changes
- Stat update


CypherX (approved):
🪦

PR #5012, opened 2024-01-08
Improved dungeon shadow pokemon indicator & purify chamber notification changes

CypherX:
## Description
Changes the dungeon missing shadow pokemon indicator from an image in the encounter list to an icon next to the other caught/shiny/pokerus completion indicators. The shadow icon will appear for dungeons that have available shadow pokemon, it will be grayed out if the player is missing a shadow pokemon and the normal shadow color when all shadow pokemon in that dungeon are acquired. Also includes a tooltip on the icon.

Also changes the purify chamber maximum flow notification to last for 15 minutes instead of 1 minute and slightly tweaks the language.

I changed the town/dungeon view name bar to use flex instead of absolute positioning, and to also always have the location name be properly centered even when the left and/or right content has different widths.

## Motivation and Context
There are often players confused about the missing shadow indicator in the encounter list, they think it's a trainer or boss that hasn't yet been unlocked.

The notification timeout increase was a popular request.

## How Has This Been Tested?
Loaded the game.
Went to a normal dungeon and the shadow icon was not present.
Went to a dungeon in the Orre region and the normal color shadow icon was shown.
Removed the shadow status from one of my pokemon and the icon changed to grayscale.

## Screenshots (optional):
old indicator:
image

new indicator:
image
image

## Types of changes
- UI improvement


Qwertypop04 (approved):

PR #5017, opened 2024-01-11
Use translation for logbook filters

DreamNya:
## Description

Use translation for logbook filters.
And make the type label vertically center for multi-line log.

## Motivation and Context

i18n.
The type label uses translation, but the filter label does not.

## How Has This Been Tested?

See screenshots.

## Screenshots (optional):

before:
before
after:
after

## Types of changes

- UI improvement


RedSparr0w (approved):
Looks good to me!
Thanks for the PR 👍

PR #5018, opened 2024-01-11
Add shadow Pokemon Encounters to the Dungeon list, refactor some Dungeon Encounter code

wormania:
## Description

Update the Dungeon Encounter List to handle shadow Pokemon encounters as well, and cleaned up some of the existing code in the area, hopefully offsetting part of the increased complexity this change brings.
This is an accompaniment to #5012 rather than a replacement

Posting this PR for UI feedback - not expecting it to get merged as-is, it's a (polished) first draft.
Some specific things for feedback: Do we want to show the trainers with their Shadow Pokémon? If so, how to handle multiple Pokémon in one encounter? See screenshots for current implementation

## Motivation and Context

Currently you are forced to external sources to find what shadow Pokemon you are missing and where they are, which is not the case for other dungeons. This brings them in line, and especially helps for those who are going for all EVs, given that some mons can only be EV'd through Shadow encounters.

## How Has This Been Tested?

Checked through all types of Shadow Trainer/Bosses, non-shadow Boss Trainers, Pokemon Boss trainers, Mimics, etc. all still show correctly in the Dungeon Encounter list.

## Screenshots (optional):

Showing a variety of encounter situations, uncaught shadow, entirely uncaught, purified, boss:
image

Dungeon with no shadows or Pokemon just showing a standard boss only list:
image

Citidark Isle/Dome are now Max Lair v2, and the boss has a lot of encounters:
image

## Types of changes

- UI improvement


wormania:
What it could look like with encounter grouping:
image

osherwott:
Looking good! 👍  I really like that grouped display too.

One thing I have to say: it might be unnecessary to show the regular uncaught status of a pokemon here. If a shadow pokemon is caught then the regular version is also caught. Players want to know which shadow pokemon they are missing in these dungeons, not which pokemon they can catch elsewhere as non-shadows. It might be less confusing if we show shadow pokemon in this display with the generic caught/uncaught image filter, rather than relegating it to a small icon that blends in with the background.

I believe this will also let you show trainers with multiple shadows more easily. Your first image shows a greyed out trainer that is sharing a filter with the pokemon. If that grey filter only applies to uncaught shadow pokemon, it might mean for a cleaner application too.

wormania:
> It might be less confusing if we show shadow pokemon in this display with the generic caught/uncaught image filter

There is already confusion about Shadows not being separate forms, and I think showing them as fully uncaught would likely exacerbate that. However we could make the Uncaught Shadow more prominent, for example by changing whether the shadow status shows on the sprite (same as we do for shiny), i.e.:

Uncaught:
image

Caught, but Shadow Uncaught:
image

osherwott:
Oh, yes. That looks so good!

wormania:
With shadow background and spacing restored:
image

wormania:
Showing the greyed out aura version can make it hard to distinguish them:
image

Could mess with the opacity filtering to make it clearer:
image

wormania:
Reopened for review, resolved issues with locked shadows/trainers not being handled, and updated PokemonLocations so that it correctly returns all the shadow encounters (e.g. Voltorb, Dragonite, legendary birds are missing in the live version of the game)

CypherX (approved):
Works well, thanks

PR #5023, opened 2024-01-13
Shiny hatches give 5x damage

RedzepR:
## Description

Hatching a shiny now gives 5x the amount that a hatch usually gives. 

## Motivation and Context

Shinys give 5x EVs but shinys in breeding are pretty much useless past the first one (and for shiny quest) so lets give them something meaningful. Also someone in chat suggested it and people generally liked it.

## How Has This Been Tested?

Hatched a few shinies and non-shinies. With and without vitamins, always got the expected attack boost

## Types of changes

- New feature? Technically?


CypherX (approved):
Looks good, thanks. Made a small tweak to use a constant.

PR #5024, opened 2024-01-14
Fix shop egg pokerus infection bug

umbralOptimatum:
Pokemon bred through shop eggs could become infected with pokerus, but would not be set to contagious upon hatching. Fixed this bug and added an update check to set any incorrectly "infected" pokemon to contagious.

## How Has This Been Tested?

Bred various pokemon via shop eggs, updated a save with currently-infected pokemon.

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -2599,8 +2599,9 @@ class Update implements Saveable {
 
             // Fix pokerus status for party members infected via shop eggs
             saveData.party.caughtPokemon.forEach(pokemon => {
-                if (pokemon[PartyPokemonSaveKeys.pokerus] === GameConstants.Pokerus.Infected && !pokemon[PartyPokemonSaveKeys.breeding]) {
-                    pokemon[PartyPokemonSaveKeys.pokerus] = GameConstants.Pokerus.Contagious;
+                // PartyPokemonSaveKeys.pokerus and .breeding
+                if (pokemon[8] === GameConstants.Pokerus.Infected && !pokemon[4]) {
+                    pokemon[8] = GameConstants.Pokerus.Contagious;
                 }
             });
         },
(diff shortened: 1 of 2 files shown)


src/scripts/Update.ts
```diff
@@ -2596,6 +2596,13 @@ class Update implements Saveable {
                     helper.sortOption = 5;
                 }
             });
+
+            // Fix pokerus status for party members infected via shop eggs
+            saveData.party.caughtPokemon.forEach(pokemon => {
+                if (pokemon[PartyPokemonSaveKeys.pokerus] === GameConstants.Pokerus.Infected && !pokemon[PartyPokemonSaveKeys.breeding]) {
+                    pokemon[PartyPokemonSaveKeys.pokerus] = GameConstants.Pokerus.Contagious;
```
CypherX, line 2603: These `PartyPokemonSaveKeys` need to be magic numbers in case the enum values ever change in the future

src/scripts/breeding/Egg.ts
```diff
@@ -143,12 +143,14 @@ class Egg implements Saveable {
...
-                    partyPokemon.pokerus = GameConstants.Pokerus.Resistant;
-                }
+            }
+
+            // Update pokerus status
+            if (partyPokemon.pokerus == GameConstants.Pokerus.Infected) {
+                partyPokemon.pokerus = GameConstants.Pokerus.Contagious;
+            }
+            if (partyPokemon.evs() >= 50 && partyPokemon.pokerus == GameConstants.Pokerus.Contagious) {
+                partyPokemon.pokerus = GameConstants.Pokerus.Resistant;
```
CypherX, line 153: I know it was already there, but is this even needed anymore with the `evs` subscriber in `PartyPokemon`?

CypherX (approved):
Someone finally fixed this bug 🎉

PR #5025, opened 2024-01-14
Battle performance improvements

umbralOptimatum:
A couple small in-battle performance improvements:
- `PurifyChamber.gainFlow()` now uses a pureComputed to check for shadow pokemon instead of scanning the entire party every time the player gains exp
- Extracted the parts of the `Party.calculateClickAttack()` improvements from #4846 that don't depend on Hisui content, so they can make it into the next update

## How Has This Been Tested?
Did a few battles, played around with party state, verified everything works as before.


CypherX (approved):
Looks good, thanks. Benchmarking shows some definite performance gains.

PR #5026, opened 2024-01-14
Group Held Items by type in the Item Bag

wormania:
## Description

Sort/Group the held items in the Item Bag by the type of item, and updated Pokémon-restricted items to a new `PokemonRestrictedAttackBonusHeldItem` to accommodate this

## Motivation and Context

Currently the Held Item section is a wall of 41 (and growing) items, by splitting them into groups it is easier/quicker to find the one you want, though they do now take up more space in the Item Bag interface

## How Has This Been Tested?

Checked that the UI still works fine for giving held items, and that old saves load fine with the new Held Item type without needing an update function

## Screenshots (optional):

File with every held item available:
image
image

If you don't have access to any of a specific type, the header doesn't show (Johto file):
image

## Types of changes

- UI improvement


Qwertypop04 (approved):

PR #5027, opened 2024-01-14
Don't send player anywhere after beating a gym or dungeon

osherwott:
## Description
Removed the line that sends the player back to the parent town of a gym after beating it.
`gymLost()` doesn't use this line, and winning a gym without this sends the player back to the town the gym was initiated in.

Replaced `MapHelper.moveToTown(DungeonRunner.dungeon.name);` with `App.game.gameState = GameConstants.GameState.town;` in DungeonRunner.ts so that it doesn't break when there is no town to return to.

## Motivation and Context
I'd like for some gyms to be battled in multiple locations for my Alola rework. This includes the motivations of:
- We can use gyms more liberally (for quests or otherwise) without sending the player to a locked location afterwards
- Players can enjoy some variety in terms of backgrounds by choosing where to fight certain gyms

Also fixes New Island bug (described in the comments below) and allows for potential dungeons outside of dungeon towns.

### Gym
1) Commented out the code
2) Checked if it broke in the following circumstances:
   - Winning a gym for the first time
   - Winning in a town with multiple gyms
   - Winning in a gym in multiple towns
### Dungeon
1) Replaced code in DungeonRunner.ts
2) Changed requirement for a dungeon to be less than 2 dungeon clears
3) Clear the dungeon 2 times
4) Check that the game state is returned to a town state and that the dungeon and town are properly locked


RegisCoaxans:
Could this change apply to dungeons too ? If yes, it would solve the bug in New Island (or later similar content) and make the first clear reward of Distortion World unnecessary.

osherwott:
> Could this change apply to dungeons too?

Likely yes, replacing `MapHelper.moveToTown(DungeonRunner.dungeon.name);` with `App.game.gameState = GameConstants.GameState.town;` seems to not break the game either. I am unaware of the New Island bug you are referring to, but for the Distortion World:
1. That is a questline step reward, not a first clear reward.
2. I believe dungeons need to be Town Content before they can be accessed outside of their own dungeon towns. This change alone would not suffice.

RegisCoaxans:
You are right, it would not work with the Distortion world at all because it TP the player to another town...
The Bug in New Island is the player being stuck in the dungeon if the event ends before the dungeon and prevents them from returning to the town.

osherwott:
> stuck in the dungeon if the event ends

Would it be possible to send the player to a starting town in that case?

osherwott:
Actually, `App.game.gameState = GameConstants.GameState.town;` leaves you in the same town even if it's locked after completing or failing to clear the dungeon.

image

RegisCoaxans (approved):
Tested winning and losing some gyms and dungeons. It worked well.

CypherX (approved):
Seems fine? No issues with the handful of gyms and dungeons I tested. Hopefully there isn't some obscure thing this breaks.

PR #5030, opened 2024-01-16
Rename Sunny to Harsh Sunlight

wormania:
## Description

Rename Sunny to Harsh Sunlight, update quest texts that mention Sunny - open to any suggestions on the new wording, I've split the wording changes into its own commit for easy viewing

## Motivation and Context

Sunny is not the canon name for the weather effect, and Sunny is easily and frequently confused with Clear given that the icon for clear is also a sun

## How Has This Been Tested?

Checked Weather display in Dock, Castform app, and Town Map, and checked that requirements such as Castform (Sunny) in the Weather Institute still work

## Screenshots (optional):

image

## Types of changes

- UI improvement


CypherX (approved):
☀️

PR #5032, opened 2024-01-16
Fix : Mystery of Deoxys to auto complete on last step

RegisCoaxans:
## Description
The quest now checks for the BF Milestone being obtained instead of global statistics.

## Motivation and Context
The last step was auto completing if the player entered the BF and reached stage 100 early. Therefore, players would not get Deoxys, yet the quest line would be completed.

## How Has This Been Tested?
With my computer.

## Types of changes
- Bug fix


CypherX (approved):
Looks good, thanks

PR #5033, opened 2024-01-16
Changed Eternapantatus image for the correct Eternapants

EatPant2nd:
## Description

The Lair of Giants quest line was using regular Eternapantatus instead of the Eternapants version. This PR fixes it so it uses the right image as the text mentions the name plus the Pokémon fleeing to the south but the one that does is the Eternapants version and not regular one (which becomes unlocked in Energy Plant). Like this it's less confusing.

## Motivation and Context

Wrong image needed to be addressed.

## How Has This Been Tested?

Beated Eternapants Eternapantatus and saw that the right image displayed.

## Screenshots (optional):

Before:
eternapants

Now:
true-eternapants

(Ignore the different themes).

## Types of changes

- Image fix


Qwertypop04 (approved):
Good catch, that's an oversight on my part. Thanks for the PR!

PR #5035, opened 2024-01-16
The "Let's not add a 0 to the images" PR

EatPant2nd:
## Description

Not added a 0 to all Dynamic Background image files. Also deleted a duplicated image (Psyduck was both in a 54.png and a 054.png).

## Motivation and Context

Something, something, no leading 0s is better.

## How Has This Been Tested?

Checked Dynamic Background and Safari since both use them.

## Types of changes

- Image update


wormania:
Is there a dependency somewhere else that requires all image filenames have the same digit length? The existing code handles a mix of 3 and 4 digit file names just fine

RedSparr0w:
I'm unsure about this change, the existing code should work fine for 3 and 4 digit names.
Not sure if I'm for or against the change though, just used to the 3 digits maybe 🤷

EatPant2nd:
A mixture is gonna look weird. I prefer my filenames in order. Personal preference. However, I understand if you feel it's unnecessary.

DraKay:
As someone who have work with this images, having a mixture with or without 0 would end in a mess when trying to find a specific image since they wouldn't have a coherent criteria. If we weren't going to use 0 bulbasaur would be 1.png, but it was 001.png before this change

DraKay:
Also, next paldea dynamic background images are being named, and have been already committed to my branch, with this coherent criteria

CypherX:
> A mixture is gonna look weird. I prefer my filenames in order. Personal preference. However, I understand if you feel it's unnecessary.

They still appear in order, no?

CypherX:
No one said anything about removing the zeroes.

DraKay:
But I'm sure we don't want to see
Bulbasaur
Gholdengo
Who-chien
Chien-pao
Ting-lu
...

This change was never for the current images but for the incoming ones

CypherX:
What does that have to do with changing numerical file names?

Even images/assets/pokemon display in the "correct' order everywhere except github.

DraKay:
To avoid more mistakes and be able to find what you are looking for. If pant had to remove an extra copy of psyduck was because that image weren't following the order criteria. 
Have a coherent classification system helps to reduce change of mistakes and allow better human interaction with the files. If there is no criteria any new person who try to work with that images would have a hard time.
I'm not a coder, but wasn't consider good practice to have all system properly order and indexated to allow any newcomer to interact with it? That's exactly what this change is. Ordering and easing things for newcomers

CypherX:
Renaming 1,835 files to solve a problem that doesn't exist is not good practice. The images are named and ordered by pokemon #, that's your system, nothing is confusing about that.

CypherX:
The images are still ordered correctly with 3 zeroes or 4, there is no difference. I don't see a benefit to this.

DraKay:
> I wouldn't say it's a very good reason, but it's definitely not a bad reason either. Consistency is always good. Then again, it's not the end of the world either way

My experience working with this images shows is very simple to mess things as they are a lot, having more than one ordering criteria would end in new people managing them making mistakes, also might give problems if in the future you want to use that images for something else too

CypherX:
I'm sorry, but if you have trouble working with clearly numbered file names with an already defined naming convention, then no matter of "criteria" will help you. This is what's called a superfluous change.

Again, there is no difference in the ordering or "ordering criteria" of the images whether they are padded to 3 or 4 characters. Like I said, solving a problem that does not exist.

If you are manually scrolling through hundreds of files to find a specific one rather than using various tools at your disposal to quickly find an image, that's on you and this change would not help anyway.

DraKay:
You don't scroll the list. You check each image individually, for some info you need to fully watch the image, so the quicker is take https://github.com/pokeclicker/pokeclicker/blob/develop/src/assets/images/dynamic-background/pokemon/001.png and going image by image to see if there is fails, pokemon backwards, flying pokemons set as not flying... Obiusly if you have different criteria you miss images

PixLSteam:
IMO we should either go for a 4-digit padding or none at all, there should not be a weird mix.
I'd go for the latter for consistency, but that's just my opinion.
GitHub's "incorrect" sorting is not a valid argument for the padding, if possible you should always work with a local clone of the repo.

RegisCoaxans (approved):
I tested it.
The sprites are working well in game.
I also ran the following script to check sprites names and every pokémon up to 917 get their sprite correct.

CypherX (approved):
Looks good, thanks

PR #5039, opened 2024-01-19
Change default ItemHandler evo stone observables from Fire Stone to Leaf Stone

osherwott:
Fire stone -> Leaf stone
Vulpix -> Gloom

## Description
Changes default stone and mon selection in evolution stones to be the first ones in the current selection.

## Motivation and Context
- ItemHandler wasn't updated after changing the item order in #1690
- Was bothering me

## How Has This Been Tested?
It hasn't

## Types of changes
- UI improvement


Qwertypop04 (approved):

PR #5040, opened 2024-01-19
Nerfed Miror B. in Gateon Port Battles dungeon

EatPant2nd:
## Description

Miror B. combined HP was 348,000,000 which was higher than Eternapants Eternapantatus' 176,361,964. That seemed ridiculous so I made the combined HP be 84,000,000 instead, which is very close to XD001 and Greevil at 83,753,718 in Citadark Isle Dome.

## Motivation and Context

The HP value was too high and it needed fixing.

## How Has This Been Tested?

Checked Gateon Port Battles dungeon.

## Types of changes

- Bug fix
- `Balance`


Qwertypop04 (approved):

PR #5041, opened 2024-01-19
Reduce Money Given by Orre 2 Challenge Battles

Razorflamekun:
## Description

Currently, the Orre 2 Challenge Battles give too much money.  This PR rebalances the amount of money they give by syncing them up with the Kalos E4.  The first Challenge Battle starts off by giving slightly more than the first four Kalos E4 battles, and each subsequent ramps up from there until the final Challenge Battle, which gives the same as the Champion of Kalos E4.

## Motivation and Context

See above.

## Types of changes

-Balance


Qwertypop04 (approved):

PR #5007, opened 2024-01-05
Indigo Disk Pokemon Data and Placeholder Paldea Sprites

Qwertypop04:
## Description
Adds data for the Pokemon added in Indigo Disc, and adds placeholder sprites for all Paldea Pokemon.
The placeholder sprites are resized menu sprites from SV. They don't all look fantastic, but it's infinitely better than a bunch of MissingNo.

## Motivation and Context
Data for Pokemon is needed, better sprites are desirable

## How Has This Been Tested?
Ran game, caught some pokemon with placeholder sprites

## Types of changes
- New Pokemon
- Sprite improvements

PR #4949, opened 2023-12-05
Mega evolution to ignore EC increase

RegisCoaxans:
## Description
Megas now have the same eggsteps as their base Pokémon.
Could not save Mega Beedrill from itself, though.

## Motivation and Context
Megas are hard to get and meant to be strong, having them all* better than their base form is fair.
\* Almost...

## How Has This Been Tested?
Just checked base Pokémon had as many EC as their megas.

## Types of changes
- Balance

PR #5044, opened 2024-01-21
Cap Evolution Egg Steps to 4800 (120 cycles)

wormania:
## Description

Maximum egg cycles is currently defined reactively based on the highest known egg cycle Pokémon in the list. This means that adding a new highest-ever egg cycle Pokémon retroactively nerfs all those who would have gone over the previous cap.
In the last update (#5007), placeholder Pokémon were added with 404 egg cycles, triggering this issue to occur.

Also, snuck in a wording fix for the changelog

## Motivation and Context

Egg Steps for evolutions are effectively uncapped right now. While the change could be to just put the 404s down to 120, this change protects against legitimate >120 egg cycle Pokémon coming from the games

## How Has This Been Tested?

Exported a `name: eggcycle` list from v10.18 with this fix, and v10.17, and confirmed the only changes were the Mega Pokémon that were changed in #4949

## Types of changes

- Bug fix

### DIFF
--- a/src/modules/GameConstants.ts
+++ b/src/modules/GameConstants.ts
@@ -157,6 +157,7 @@ export const PLATE_VALUE = 100;
 
 // Breeding
 export const EGG_CYCLE_MULTIPLIER = 40;
+export const MAX_EGG_CYCLES = 120;
 export const BREEDING_ATTACK_BONUS = 25;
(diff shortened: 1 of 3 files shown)


CypherX (approved):
What, ya'll don't like 16,160 step Lunala (Full Moon)???

(Looks good, thanks for the fix)

PR #5048, opened 2024-01-22
Added an option in the settings to remove the queue clear popup

RedzepR:
## Description

Asking the player wether or not they want to clear the queue is annoying. So let's give the players an option for themself to prompt that question or not.

## Motivation and Context

Its a QoL update that doesnt hurt

## How Has This Been Tested?

Cleared queue with and without the setting enabled

## Types of changes

- new option in the settings


CypherX (approved):
Looks good, thanks. I moved the setting to the Gameplay tab to be with the other confirm settings, and updated the language.

PR #5049, opened 2024-01-22
Changed Holo Caster's unlock requirement

EatPant2nd:
## Description

Changed Holo Caster (Key Item that unlocks the Achievement Tracker) from 110 caught Pokémon to checking if the player has at least 10% Achievements Boost.

## Motivation and Context

Since it was deemed too overwhelming at the start to have the Achievement Tracker unlocked, I decided to make it check Achievements Boost instead of caught Pokémon as this requirement seemed a bit arbitrary. This way it will still require the player to progress a bit before unlocking the tracker which hopefully will happen at a point where it's less overwhelming.

## How Has This Been Tested?

Started a new file, got to 10% Achievements Bonus, saw the item unlock. Also loaded some random files without the caught Pokémon requirement but yes with enough % to unlock the tracker and the Key Item unlock triggered correctly.

## Types of changes

- Balance (?) I dunno.


CypherX (approved):
Looks good, thanks

PR #5050, opened 2024-01-22
Breeding Efficiency and Steps/Att fraction digits

AevitasDragonkin:
## Description

Updated the Breeding Efficiency and Step/Att display from `maximumSignificantDigits` to `maximumFractionDigits` and increased the maximum digits available from 2 to 3, to match other fraction digits.

## Motivation and Context

With more and more pokemons getting to the higher breeding efficiency, it's more helpful to see the fraction numbers.
`maximumSignificantDigits` caused numbers above 1 to at most have 1 fraction digits, while numbers above 10 would never have any fraction digits.

## How Has This Been Tested?

Opened the displays and checked if the value was displayed correctly.

## Screenshots (optional):

**Hatchery**
image
image

**Statistics**
image
image

**Vitamins**
image
image

## Types of changes

- UX improvement


RegisCoaxans:
> to match other fraction digits

Are you referring to EV, here ? If so, would you make these three values `maximumSignificantDigits: 3` instead ?
I know that some people complained about decimals making large numbers harder to read after EV display changed, and seeing a BE of 10.787 does not look appealing neither.

AevitasDragonkin:
Was mostly referring to the Farm, which has been updated to use 3 fraction digits.

I don't think EV benefits as much from the 3 fraction digits as BE would, as EV scales up a lot, while BE stays relatively small.

Having it with significant digits would mean that a number like 0.00012 would be correctly displayed though (with fraction digits it would be 0.00.
However, with only 2 significant digits number like 10.5 and 11.49 are both displayed as 11. Increasing it to 3 might be better, I just liked the fraction digits more, and it was more consistent even if numbers grew bigger.

RegisCoaxans:
I did not know about the farm.
I sure agreed that 3 digits are needed most of the time, though.
Regarding EV, I meant the display in the PokemonStatisticsModal, having significant digits might help, it kind of has fraction digits for now and that does not show well with high EV amounts.

CypherX (approved):

PR #5052, opened 2024-01-23
Replacing multiple notifications when removing all Vitamins with just one notification

Crobat4:
## Description

When you remove all vitamins (or all of a kind of vitamin), you'll get up to 3 notifications per each Pokémon in the Hatchery/Queue, this PR changes that so it only shows a single notification if a Pokémon's vitamins couldn't be modified.

## Motivation and Context

There is no point for multiple notifications with the same message, plus you won't be able to see them all if you have several Pokémon in the Hatchery.

## How Has This Been Tested?

In the "All Vitamins" modal, I clicked every "Removed All" button.

## Screenshots (optional):

Before:
image
After:
image

## Types of changes

- UI improvement


RedSparr0w (approved):

PR #5053, opened 2024-01-23
Fix broken Knockout extenders (and hatchery filters)

### LINKED ISSUE #5043: [Bug]: Can't select pokemons to breed since the update
### Confirmations

### What happened?

When opening the hatchery for breeding my pokemon, it doesn't show any pokemon to select. I have a lot of pokemon LV100 to breed, so it's not a me issue. Changing the filters doesn't change anything. I refreshed the page a few times, but the issue is still here.
breed

I expected to see the pokemon I can breed and choose them.

### Steps to Reproduce

1. go to "list" on tha hatchery
2. select the "pokemon" tab
3. see the issue.

Windows 8

Chrome

### Relevant log output

_No response_

### Additional Information

_No response_

umbralOptimatum:
Fixed hatchery filters initially not working after loading the game. As it turns out, this wasn't related to the hatchery improvements at all! The custom extenders the filters use had broken completely.

The culprit was the new bindingHandlers added in #4894. The handlers work as intended, but importing from Settings inadvertently loaded most of the game's modules before the custom extenders got created. And for some stupid reason Knockout doesn't throw any errors if you try to use an extender that doesn't exist (yet).

Fixes #5043.

## Description

I moved all the bindingHandlers to their own separate file, so they can use imports without interfering with the top-priority extenders. I also imported `koExtenders.ts` in all the modules that currently use custom extenders; this way the dependencies are visible if we ever want to add a build step that rejects circular dependencies.

## How Has This Been Tested?

Checked the new dependency tree, confirmed the game loads, confirmed the BreedingFilter extenders are now present and the filters operating correctly.

### DIFF
--- a/src/modules/DataStore/StatisticStore/getRouteKillsProxy.ts
+++ b/src/modules/DataStore/StatisticStore/getRouteKillsProxy.ts
@@ -1,4 +1,5 @@
-import { Observable } from 'knockout';
+import type { Observable } from 'knockout';
+import '../../koExtenders';
 import * as GameConstants from '../../GameConstants';
 
 const failedSetValue = () => 0;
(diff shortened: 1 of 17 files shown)


RedSparr0w (approved):
Seems to be working as expected AFAICT, will let someone else review too though, thanks for the PR! 👍

CypherX:
Everything seems to be working for me as well, and the hatchery filters are indeed fixed. Thanks!

PR #5054, opened 2024-01-23
Fix hatchery helper sort-by-attack oversight

umbralOptimatum:
Fixes hatchery helpers still having the option to sort by attack rather than attack at level 100.


RedSparr0w (approved):

PR #5058, opened 2024-01-24
Complete trainer sprite audit

kittenchilly:
## Description
Updates almost every trainer sprite to use their counterpart from [Pokemon Showdown](https://play.pokemonshowdown.com/sprites/trainers/)
Since we already use sprites mostly from there, the most this results in is tweaking the position. However, some sprites were completely overhauled (for the better imo). Also reduces files size without compromising quality overall. Credits were changed accordingly.

Some notes:

- If anyone doesn't like a sprite change I did, I can revert it or find another sprite to use.
- We're kind of inconsistent if we want to use the gen 3 designs or the gen 6 designs for team magma and team aqua, so I just standardized it to the gen 6 designs. I can change this if we don't want it.
- I didn't change any Kanto sprites (like the gym leaders) since it seems we specifically want to have their frlg designs instead of the b2w2 designs or lgpe designs.

## Motivation and Context

I think the quality of some trainer sprites were rough and not in the same style as the b2w2 sprites (which it seems we prefer to use), and Pokemon Showdown had higher quality ones we can freely use with credit. If you go through the diffs you can compare them and see what I mean.

## How Has This Been Tested?

I went through dungeons, gyms, and talked to npcs to check if they displayed properly.

## Screenshots (optional):

image

## Types of changes

- UI improvement


DraKay:
Personally I dislike the idea of remove the sprite made specifically for the proyect. People spend their time and put effort only to be here, is like throw it...

kittenchilly:
> Personally I dislike the idea of remove the sprite made specifically for the proyect. People spend their time and put effort only to be here, is like throw it...

I didn't replace every sprite, just some that had counterparts so we can be more consistent... I would like more opinions on it though

DraKay:
Also do we really have showdown permission to stole all their assets? Or we take them and pray they don't get mad at us?

kittenchilly:
> Also do we really have showdown permission to stole all their assets? Or we take them and pray they don't get mad at us?

Like I said, we already use showdown sprites in places. Their trainer sprite page says we must credit the creator if we use them, which we already do.

DraKay:
> > Personally I dislike the idea of remove the sprite made specifically for the proyect. People spend their time and put effort only to be here, is like throw it...
> 
> I didn't replace every sprite, just some that had counterparts so we can be more consistent... I would like more opinions on it though

If they were asked is they didn't have a counterpart at that moment or they weren't good enough or liked, like peonia sprite was added in showdown in 2020 and asked for this proyect in 2023

Someonealive-QN:
I somewhat agree with draky, you shouldn't just go on and replace sprites made specifically for the project even tho their quality isn't that good. You could've contacted the artists atleast when replacing them and not just've gone and replaced them without a singular warning. Some of the artists spent quite some time on them so even so showdown has them done it's not right so simply swap them without any notice or asking if they want

kittenchilly:
> I somewhat agree with draky, you shouldn't just go on and replace sprites made specifically for the project even tho their quality isn't that good. You could've contacted the artists atleast when replacing them and not just've gone and replaced them without a singular warning. Some of the artists spent quite some time on them so even so showdown has them done it's not right so simply swap them without any notice or asking if they want

Ah, I didn't think anyone would be bothered by it since we get sprites from a bunch of sources anyways. If more people insist we keep the custom made ones I'll revert it.

kittenchilly:
I do think we need a list of sprites we should keep and a list we should change though

DraKay:
I 100% agree with soul(someone), contact with the artists and point the problems you have seen in sprites would be really appreciated and better.

Also about the compression, don't we have a bot for that?

Razorflamekun:
> > I somewhat agree with draky, you shouldn't just go on and replace sprites made specifically for the project even tho their quality isn't that good. You could've contacted the artists atleast when replacing them and not just've gone and replaced them without a singular warning. Some of the artists spent quite some time on them so even so showdown has them done it's not right so simply swap them without any notice or asking if they want
> 
> Ah, I didn't think anyone would be bothered by it since we get sprites from a bunch of sources anyways. If more people insist we keep the custom made ones I'll revert it.

You won't need to revert it since I do not believe this PR has a chance of leaving the current stage that it is in.  I agree with DraKay.  There is no need to change any of the currently used sprites in the project at this point in time.  Keeping a list of backups is fine, but right now, there is no need to change anything.

src/assets/images/npcs/Sage.png
osherwott, line 1: Showdown sprites are sometimes strangely colored in a way that negatively impacts the sprite. This is a good example of that. Notice the details of his robe being lost in the darker colors that showdown uses. It'd be preferable to not replace official sprites (gens 3-5) with showdown sprites if they are otherwise the same, like here.

src/assets/images/npcs/Actor.png
osherwott, line 1: This sprite is used for Teacher (gen 7). This change would give us two sprites that are the same, but with different filenames, which we try to avoid. Actors are a weird trainer class that use other trainers' sprites, It might be best to leave this one alone for now.

src/modules/GameConstants.ts
```diff
@@ -2018,7 +2018,6 @@ export const TemporaryBattles = [
     'Gym Leader Bede',
     'Rampaging Gigalith',
     'Rampaging Froslass',
-    'Gym Leader Marnie',
```
osherwott, line 2021: This is a constant for a temp battle, not a credit! 😰 Removing this would cause glitches in the game! 😟

src/scripts/towns/TownList.ts
```diff
@@ -2343,7 +2343,7 @@ const FortreeRanger = new NPC('Pokémon Ranger Catherine', [
     'Have you seen this? It\'s horrific. Absolutely despicable.',
     'Disposable Dowsing Machines! Who even came up with this?! People leave these everywhere. It\'s absolutely terrible for the environment. Some poor innocent Pokémon could choke on it!',
     'Please recycle your used Dowsing Machines.',
-], {image: 'assets/images/npcs/Pokemon Ranger (female).png'});
+], {image: 'assets/images/npcs/Pokémon Ranger (female).png'});
```
osherwott, line 2346: These aren't typos, but deliberate spellings. We have two female Ranger sprites, differentiated by that `é`.

osherwott (commented):
For aqua and magma sprites, it'd be best to add both gen 3 and gen 6 ones. We can call upon either one via the code. It would help to name the gen 6 ones something like `Team Aqua Grunt ORAS (female)`, or so, in the same vein as the Cipher Peons. Some sprites are chosen more carefully than others, so it's not a bad idea to look over them. I have commented my thoughts on some individual sprites to use as examples. Many of the gen 3-5 sprites shouldn't be changed imo, but of course you are still free to defend your changes if you disagree with any of my reasons!

General note: some Showdown sprites have been replaced with newer ones since the time we have used them. I'm unsure whether we want to replace them or not, and it might be case by case.

osherwott:
Closing because most of these changes aren't needed. And any smaller scale changes pertaining to this should have their own PR, because lots of these images only differ in filesize.

PR #5059, opened 2024-01-24
Changed Holo Caster unlock requirement 2

EatPant2nd:
## Description

Changed Holo Caster unlock requirement again due to issues with Achievements proccing when they shouldn't. Now it unlocks after completing Victory Road (Kanto).

## Motivation and Context

The other one was causing bugs (sorry). This unlock requirement, hopefully, achieves a similar idea of unlocking at a certain point late in Kanto so as to not overwhelm the player.

## How Has This Been Tested?

Loaded files. Checked better. This time achievements didn't proc and the item got unlocked correctly.

## Types of changes

- Bug fix


CypherX (approved):
Looks good, thanks

PR #5060, opened 2024-01-25
Fix tutorial pokeball selector prompt

CypherX:
## Description
During the tutorial the player would be prompted to select the "Caught" pokeball filter and choose a pokeball. When #4990 was merged the default order for new files changed from bottom-to-top, with Caught at the top, to top-to-bottom, with Caught at the bottom. This changed cause the prompt to continue when the player selected the 'New Shiny' ball rather than the 'Caught' ball.

## How Has This Been Tested?
Started a new game and progressed in the tutorial

## Screenshots (optional):
image

## Types of changes
- Bug fix


RedSparr0w (approved):
Nice fix, thanks for the PR! 👍

Issue #4374, opened 2023-05-28
[TASK?]: Jaaslet's SpecialEvent Wishlist


Aegyo:
### Task Description

Issue for some things Jaaslet mentioned wanting to do in #4329 ([comment link](https://github.com/pokeclicker/pokeclicker/pull/4329#issuecomment-1553320150))

> I have a list of stuff i want to do. But they are all nice to have, and not required for the event calendar to work. And because i want this PR to be finished, and not abandoned (like the first one) and i want it to be easier to review, they will be added in PRs after this one is merged. The list is (people are allowed to help, if they want to):
- [ ] Refactor notifications to use ticks instead of setting the timer
- [ ] Refactor the SpecialEvents class to not have anything static methods/variables
- [ ] Add a setting to disable special events (i promissed someone to do it, after i closed their PR attempting to do it)
- [ ] Change the halloween event to fit the new code, so that can be used by the event calendar too
- [ ] Make keyitems clickable (event calendar should open events and gemcase should open gems)
- [ ] Move the event button up in the start menu
- [ ] Add an enum with event names, and use that instead of the title, as identifier, in the code
- [ ] Add a warning, if you are starting more than one event at a time

### Notes

Each of the task points can be separate PRs. None of them have been explicitly approved but they all look pretty fine to me, double check in discord if you want to be sure.

### Issues Required For This Task

None

PR #5069, opened 2024-01-28
Special event enum and timers on event modal fixed

RegisCoaxans:
## Description
That is it.

## Motivation and Context
Using an enum prevents typos. Requested in #4374.
The timer was only ticking if the event got started throught the event calendar.

## How Has This Been Tested?
Mispelled some event names in the QuestlineHelper.
Played with event activation to confirm it was ticking as it should.

## Types of changes
- Bug fix
- Code improvement


CypherX (approved):
Looks good, thanks

PR #5071, opened 2024-01-29
Changed Route 10 to be accessible before rock runnel

RedzepR:
## Description

Rock tunnel is no longer required to access route 10 in kanto but the other way around
Rock tunnel is now required to access route 8 and lavandia

## Motivation and Context

The only grass on route 10 is before rock tunnel, afterwards its only a trainer gaunlet. Also it saves electric mono files from christmas hell

## Types of changes

- Route accessibility


src/modules/routes/RouteData.ts
```diff
@@ -160,7 +157,10 @@ Routes.add(new RegionRoute(
     new RoutePokemon({
         land: ['Pidgey', 'Ekans', 'Sandshrew', 'Vulpix', 'Meowth', 'Growlithe'],
     }),
-    [new RouteKillRequirement(10, Region.kanto, 10)],
+    [
+        new RouteKillRequirement(10, Region.kanto, 10),
+        new ClearDungeonRequirement(1, getDungeonIndex('Rock Tunnel')),
+    ],
```
osherwott, line 163:
```suggestion
    [new ClearDungeonRequirement(1, getDungeonIndex('Rock Tunnel'))],
```
If route 10 is needed to access Rock Tunnel, which is required to access Lavender Town, then the requirement for route 10 here isn't needed.

osherwott (changes requested):

RedSparr0w (approved):

PR #5075, opened 2024-02-01
Setting to disable auto saving

CypherX:
## Description
As stated, adds a setting to the gameplay tab that disables the periodic and on exit auto save functionality.

Previously rejected in #4555 for fear of players enabling the setting without thinking, I am reviving this with additional safeguards and warnings in place. If the other devs still think it's a bad idea we can close this.

Upon enabling the setting a confirmation dialog will be displayed warning the player and asking them to confirm. Additionally, when a file is loaded with auto save disabled a warning notification is displayed. Added a new button to manually save the game next to the mute/unmute button when auto save is disabled (button is not visible when auto saving is enabled).

The floppy disk save icon SVG is from Font Awesome Free.

## How Has This Been Tested?
Toggled the setting, verified the game did not save while it was enabled and continued to save normally when disabled.

## Screenshots (optional):
image
image
image

## Types of changes
- New setting


RedSparr0w (approved):

PR #5077, opened 2024-02-02
Unexplored regions in Ship Modal

AevitasDragonkin:
## Description

Instead of only listing the regions you have unlocked. It will display all regions that are currently available, and display the ones you have not unlocked yet as "Unexplored region".
There was the option to add "More regions coming soon", but the general gist in discord was that this would probably raise more questions of when the new content would be coming. So we're leaving it out for now.

## Motivation and Context

Lots of bug reports recently of people completing Galar and asking why they can't get into Hisui. This should hopefully make it clear that there currently is no region implemented after Galar. As soon as Hisui is released, this will automatically add Hisui as "Unexplored region".

## How Has This Been Tested?

* Opened the ship modal on a save file with only Hoenn unlocked, showing the Unexplored regions.
* Openend it with a completed save file, showing all regions fully unlocked. 
* Openend it with Hisui enabled on a completed save file and it showed correctly as Unexplored region, and switching to Hisui correctly after first traveling to Hisui.

## Screenshots (optional):

Went with the second option
image

## Types of changes

- UI improvement


src/components/shipModal.html
```diff
@@ -9,29 +9,47 @@ <h5 class="modal-title">Regional Travel</h5>
...
+                            <!-- /ko -->
                         <!-- /ko -->
-                    </div>
                     <!-- /ko -->
+<!--                    <div class="btn-group w-100 my-1">-->
+<!--                        <button style="position: relative;" class='btn btn-block btn-secondary ship-modal-region-button'-->
+<!--                                data-bind='attr: { disabled: true }'>-->
+<!--                            <knockout>More regions coming soon™</knockout>-->
+<!--                        </button>-->
+<!--                    </div>-->
```
RegisCoaxans, line 52: Can you remove that, please ?

RegisCoaxans (approved):
I tested it and looked to the code. Seems good to me.

src/components/shipModal.html
AevitasDragonkin: Done

RedSparr0w:
What option did you end up going with in this PR?
Don't remember what was discussed on the Discord 😅

RedSparr0w (approved):
nvm, just re-read the description.

AevitasDragonkin:
Went with option 2
image

PR #5082, opened 2024-02-04
Added Lucky Punch as the last missing pokemon specific item

RedzepR:
## Description

Added Lucky Punch as a hold item for Happiny, Chansey, Blissey
In the canon games only Chansey can use it but it wouldnt break Happiny or Blissey by being affected too imo. Otherwise a real evo player might lose this rare item by evolving their chansey
In order to become available (by defeating chansey similar to leek/galarian farfetch'd) Chansey had to give up Oval Stone as a rare drop

## Motivation and Context

It was missing so I've added it

## How Has This Been Tested?

Defeat a bunch of chansey for it to drop - equipped it to the 6 possible mons

## Screenshots (optional):

image

## Types of changes

- new hold item


osherwott:
Held items aren't given to Kanto pokemon because they are part of overall balancing and meant to be introduced in Johto. You can give it to Happiny instead, since it is a Sinnoh pokemon, and you can find it holding one in the Diamond, Pearl, and Platinum games. 👍

RegisCoaxans:
There is no such thing as wild Happiny before Galar, though.

RedzepR:
Can safari mons drop their hold items on catch? Otherwise the first wild chansey would be in sinnoh either way

AevitasDragonkin:
Seems like it's only via BattlePokemon.defeat, which isn't called in the Safari.

osherwott:
Cool! Shouldn't be a problem then 👍

Qwertypop04 (approved):
Looks goof, thanks for the PR!

PR #5083, opened 2024-02-04
Wandering Pokémon but better

RegisCoaxans:
## Description
I reworked how wanderers work.
I will eventually make a separate PR for a revamp of the Roseli aura, so, except this not to be fully balanced on its own.

## Motivation and Context
That system is boring. Wandering Pokémon are a completely passive thing the player has almost no control over, they get almost no reward for that either. For instance, base wanderers are just meant to be an annoyance.
Now, it will actually provide something, ~~and these weird people out there who only want to catch only Pokémon of a given type, will serenely be able to use the Farm.~~

## Mechanics
The player now have to engage a bit more with the Farm. Wanderers are not automatically caught anymore, they are not guaranteed catches, neither they are free captures.

When a Pokémon wanders onto the farm, it sticks around the plot. It appears in the Farm modal using the Dynamic Background sprites.
The player can then click on it and throw a Pokéball at it based on the Pokéball Filters they set up. Therefore, caught wanderers provide Dungeon Tokens, they also provide a few Farm Points depending on the berry they got attracted by, caught or not.

Base wanderers give EV again; as many EP as routes encounters.
The Pokémon will eventually flee in a couple of minutes after the berry is harvested or dies.

Since this mechanic more heavily relies on active gameplay, one can outsource it to farmhands !

## How Has This Been Tested?
Got sure all Pokéballs still gave their bonuses.
Attracted few buddies, verified that catching worked fine.
Ensured the other farm interactions still behaved as expected.
~~Double checked the eggsteps, you know, just in case.~~

## Screenshots (optional):
LOOK, THEY ARE SO CUTE.
image

## Types of changes
- Feature rework


UnicornSnuggler:
I genuinely enjoy this idea. Both in a general sense and from a challenge runs point of view. Two things that immediately appeal from the perspective of someone who's actively in the midst of their first Grass Monotype challenge run:
* Eradicating the stress about the prospect of an unwanted Scyther ruining my day is a huge boon.
* Getting to acquire extra Dungeon Tokens off of Tangela on the farm while I'm on a route that otherwise would not have Tangela is very nice. Makes idling for Rare Candies feel less bad, because it's another fun little minigame to focus on.

osherwott:
If base wanderers give EVs, the base wanderer section in the berrydex modal needs to be changed too. It currently doesn't show the EV status for them and says they don't give any.

RegisCoaxans:
Good catch.

AevitasDragonkin:
Can we buff EP more towards dungeon amounts? I know it doesn't require the cost of a dungeon. But it takes a lot longer than routes to get the wanderers. Also, not guaranteed to catch them anymore.

AevitasDragonkin:
Can farmhands help with catching based on the pokeball filter setup you have?

RegisCoaxans:
> Can we buff EP more towards dungeon amounts? I know it doesn't require the cost of a dungeon. But it takes a lot longer than routes to get the wanderers. Also, not guaranteed to catch them anymore.

Considering you get base wanderers no matter where you are in the regions, no matter which setup you use, even though you get them less often than route encounter, would 0.1 still not be enough ? Keep in mind Roseli will get buffed next too.

> Can farmhands help with catching based on the pokeball filter setup you have?

Whether you catch the wanderers yourself or have a farmhand do it for you, Pokéball Filters are used and Pokéball and Magic Ball bonuses count.

RegisCoaxans:
> If base wanderers give EVs, the base wanderer section in the berrydex modal needs to be changed too. It currently doesn't show the EV status for them and says they don't give any.

Changed ! I also filtered out paldean wanderers as it causes confusion to the players.

Pushing the branch the my GH page so it can be tested.

src/components/berryDexModal.html
```diff
@@ -174,31 +174,13 @@ <h5 class="modal-title text-light" data-bind="text: BerryType[$data]">Berry Stat
...
-                              placement:'top',
-                              html: true,
-                            }">ⓘ</td>
+                            <th colspan="2" data-bind="text: 'Attracted by all Berries'">Base Wanderers</th>
                           </tr>
                           <tr id="wanderRow">
                             <td id="wanderPokemonList">
                               <ul class="list-inline">
-                              <!-- ko foreach: Berry.baseWander -->
-                              <li class="list-inline-item no-pokerus">
```
osherwott, line 191: `no-pokerus`  isn't being used anymore, so it can be removed from farm.less, unless you want to go the extra mile and disable it after getting the virus key item 🤔

src/components/farmModal.html
```diff
@@ -325,28 +325,9 @@ <h5 class="p-0 modal-title text-muted" data-bind="text: `#${(Number($data) + 1).
                                         <tbody>
                                             <tr>
                                                 <td class="p-0">
```
osherwott, line 327: `class="p-0"` was added here to disable padding for the berrydex preview. It's not working since you've changed the list to a template.

Robbers-Daughter:
I worry about people accidentally clicking on the berry and harvesting it when trying to click on the wanderer. Is there a way to prevent that (other than using helpers) or do the players just have to be really careful?

RegisCoaxans:
> I worry about people accidentally clicking on the berry and harvesting it when trying to click on the wanderer. Is there a way to prevent that (other than using helpers) or do the players just have to be really careful?

Interacting with the wanderer has priority over interacting with the berry. Moreover, you can still have your plot safelocked and catch the Pokémon on it. In fact, a wanderer on a plot does not prevent you from safelocking your plot if you worry about accidentally double clicking.

src/components/farmModal.html
RegisCoaxans: It is fixed now. Thank you for checking.

Razorflamekun:
I personally see no point in changing wanderers.  This is an IDLE GAME, and we don't need this change.

RegisCoaxans:
We will test the balance in due time.
My next ideas are buffs to make attracting and catching easier. If those are still not enough, I will boost the EV rate.
The Starf berry keeps appearing at the time the shiny wanderer is generated, though.

CypherX:
Copying this from discord -

Wanderer rework EV test - getting Exposed Applin resistant, full field of Haban in both tests

Live (back when it first released) - 30+ hours
Setup: power bracer, blue flute (23.92% bonus)

Rework - 7 hours 20 minutes
Setup: power bracer, no flute (forgot to enable it), repeat ball; using Kerry to handle wanderers
net gain of 3k farm points
don't think I missed a catch at all, either

wormania:
Worth noting that Applin has a very high catch rate, and Repeat Balls naturally make things take <1/5th the time.
The 24 starters have a catch rate of 17%, here is a comparison for those using Power Bracer, no Flute:

| Ball | Catch Rate | EVs Per Catch | Time Taken | Cost |
|------|------------|---------------|--------------------|------|
| v0.10.19 | 100% | 1 | 1x (Baseline) | 0 |
|  UB+MB  |  37% | 1 | ~2.7x | ~0 |
| Repeat+MB | 42% | 5 | ~0.47x | 24 balls per resistant = 576 Repeat Balls in total |
| Masterball | 100% | 1 | 1x | 50 balls per resistant = 1200 Master Balls in total |

RedSparr0w:
I think we do need to add some sort of feedback for when the Pokémon isn't caught, with them just disappearing is a bit jarring, makes it feel like they left just before you attempted to catch them.

RegisCoaxans:
Alt sprites are now displayed for alt form wanderers.
Bonus in catch rate calculation : the lowest gets + 10%, the highest gets + 5%.

N.B. : Some sprites are in the making and use Psyduck as a filler.

src/scripts/farming/Farming.ts
```diff
@@ -1737,6 +1737,11 @@ class Farming implements Feature {
 
         //#endregion
 
+        // REMOVE THIS BEFORE MERGING. THIS IS FOR TESTING PURPOSE.
+        this.berryData.forEach(b => {
+            const witherTime = b.growthTime[PlotStage.Berry] - b.growthTime[PlotStage.Bloom];
+            b.growthTime = [30, 60, 90, 120, 120 + witherTime];
+        });
```
RedSparr0w, line 1744: Just commenting here so it's not forgotten 😓

RegisCoaxans: Removed, thanks for the reminder !

src/scripts/farming/Farming.ts
```diff
@@ -2266,4 +2266,74 @@ class Farming implements Feature {
...
+
+        const farmPoints = Math.floor(berry.farmValue / (4 + berry.growthTime[PlotStage.Bloom] / 1800));
+        App.game.wallet.gainFarmPoints(farmPoints);
+
+        const pokeball = App.game.pokeballs.calculatePokeballToUse(pokemonData.id, wanderer.shiny, false, EncounterType.wanderer);
+        if (pokeball !== GameConstants.Pokeball.None) {
+            wanderer.pokeball(pokeball);
+            wanderer.catching(true);
+            // Halved catch time in farm, it does not matter in the balance
+            setTimeout(() => this.attemptCatchWanderer(plot), App.game.pokeballs.calculateCatchTime(pokeball) / 2);
```
CypherX, line 2285: Can we pass `pokeball` as a parameter to `attemptCatchWanderer` instead of having it as a property on the wanderer?

src/scripts/farming/WandererPokemon.ts
```diff
@@ -0,0 +1,49 @@
+class WandererPokemon {
+    public farmPoint: number;
```
CypherX, line 2: `farmPoint` doesn't appear to be used anywhere

src/scripts/farming/Farming.ts
RegisCoaxans: We need the pokéball property for the catching animation anyway.

CypherX: Ah, ok, didn't see it was also used there

src/modules/routes/Routes.ts
```diff
@@ -20,8 +20,8 @@ export default class Routes {
         return this.regionRoutes.find((routeData) => routeData.region === region && routeData.number === route);
     }
 
-    public static getRoutesByRegion(region: GameConstants.Region): RegionRoute[] {
-        return this.regionRoutes.filter((routeData) => routeData.region === region);
+    public static getRoutesByRegion(region: GameConstants.Region, ignoreIgnoredRouteInCalculations = false): RegionRoute[] {
+        return this.regionRoutes.filter((routeData) => routeData.region === region && !(ignoreIgnoredRouteInCalculations && routeData.ignoreRouteInCalculations));
```
CypherX, line 24: For now I think we should keep this function as-is and just have the extra filter in `wandererToRoute`, it isn't called often enough to have any impact on performance and this new override isn't used anywhere else at this time.

RegisCoaxans: Addressed.

src/scripts/farming/WandererPokemon.ts
RegisCoaxans: I forgot to remove it. Now it is done.

CypherX (approved):
Looks great, thanks

PR #5089, opened 2024-02-06
Fix a couple tooltip display issues

CypherX:
## Description
Changes the boundary of the quest pause button and locked dungeon boss tooltips to be the window rather than the immediate container to fix display issues with both in some cases. This allows the tooltip to appear and expand outside of the container.

## Motivation and Context
When a tooltip displays and its container is not large enough it will appear under the mouse, which then causes it to immediately hide leading to an endless loop of displaying/hiding and prevents interacting with any elements underneath.

Discord bug report: https://discord.com/channels/450412847017754644/1202993536694489118

## How Has This Been Tested?
Loaded the game, checked that the changed tooltips now correctly display.

## Types of changes
- Bug fix


RedSparr0w (approved):

PR #5092, opened 2024-02-06
Made Farm Hands regain energy while fired

EatPant2nd:
## Description

Farm Hands were too greedy and would only recover energy while hired. This allows them to also be able to rest at home for free instead of always needing a paid vacation to recover.

## Motivation and Context

Farm Hands being able to recover energy while fired was part of the code but was not working as intended. This fixes that.

## How Has This Been Tested?

Hired some Farm Hands. Waited for energy to go down. Let them rest and then hired them again to see if they would have some energy recovered.

## Types of changes

- Bug fix


Qwertypop04 (approved):

PR #5101, opened 2024-02-09
Change town map to use 'if' bindings for performance

umbralOptimatum:
The town map originally used `if` bindings for each region's map, but was changed to `visible` bindings as a workaround for tooltip issues (see #1184). I've switched them back to `if` bindings and added a better solution to the tooltips. This should be slightly more performant as observable updates won't reevaluate bindings in all ten regions.


RedSparr0w (approved):
Nice improvement, thanks for the PR! 👍

PR #5102, opened 2024-02-09
Reword Repeat Ball description

umbralOptimatum:
Reworded Repeat Ball description to not imply scaling EV gains.


Qwertypop04 (approved):

PR #5104, opened 2024-02-10
Fix : achievement storage to prevent them from triggering multiple times

RegisCoaxans:
## Description
Rework of #3493.
I actually provided a cleaner piece of code than last time in a future proof implementation.
Instead of manually setting which achievement has to be stored, the game automatically stores the achievements which are unlocked but whose requirement is not completed anymore at the time of the save.

## Motivation and Context
Real evo could trigger most Pokémon collection achievement (Pokédex, Shiny, Shadow or custom Dex) more than once.

## How Has This Been Tested?
Gained an attack achievement with xAttack, waited for the boost the wear off and verified it was stored and retrieved.
Similarly, I 🪄 the Kanto Pokédex and verified I still had it after reload.
I forgot to check the eggsteps !

## Types of changes
- Bug fix


CypherX (approved):
Looks good, thanks. This adds some additional overhead when saving, hopefully it doesn't cause any noticeable stutters for anyone

PR #5107, opened 2024-02-11
Removing Shuckle as Dungeon Boss in Hall of Origin

Razorflamekun:
## Description

Shuckle is only listed as being able to be found on Route 224 with Emerald inserted into the DS.  Therefore, I would like to remove Shuckle as a Dungeon Boss in Hall of Origin.

## Motivation and Context

My motivation is this isn't Canon.

## Types of changes

-Dungeon Boss Change


wormania:
Wouldn't "canon" be to have absolutely nothing in this dungeon apart from the Arceus boss? _Everything_ here is a fabrication to make it fit with how dungeons work in this game

Razorflamekun:
> Wouldn't "canon" be to have absolutely nothing in this dungeon apart from the Arceus boss? _Everything_ here is a fabrication to make it fit with how dungeons work in this game

Yes, that is true, however, the developers of this game have made it clear that they don't want single dungeon bosses that are Legendary.  In terms of the actual dungeon theme, it seems as though the bosses are all Normal type, while Shuckle is not.

Qwertypop04:
Needs a boss to replace it.

Razorflamekun:
Added Staraptor as a replacement boss.

Qwertypop04 (approved):

PR #5111, opened 2024-02-14
Convert /scripts/gems to modules

umbralOptimatum:
Converted gem scripts to modules as part of the glacial transition to modules. As part of the process I separated `GemDeal` into multiple files and moved the flute item definitions into `ItemList`. Also ended up converting `PokemonItem` to a module so it could load in time to be used for the gem deals.

Basically all of this was fiddling with types and imports and copy-pasting so the actual Javascript is basically the same. I did some cursory testing and confirmed that all the gem and PokemonItem shops are still there and behave normally when purposed.

All this was even more tedious and time-consuming than I expected. How does TypeScript manage to throw so many inexplicable errors? Why does `declare class` work perfectly in one module and cause a "Duplicate identifier" error in another??


CypherX (approved):
Everything seems to be working, thanks for the module conversion!

PR #5112, opened 2024-02-16
PokemonHelper module conversion work

umbralOptimatum:
Separated the PokemonHelper location-detection bot into a separate class to make the module conversion easier. Which meant there was only one function left in PokemonHelper, so I moved that to the module.

Which means PokemonHelper is fully converted! Wow! Amazing what can be accomplished through creative renaming! /j


src/scripts/pokemons/PokemonHelper.ts
```diff
@@ -1,789 +1,3 @@
...
-    DungeonReward,
-    Trade,
-    GiftNPC,
-    ShadowPokemon,
-    DreamOrb,
-    BattleCafe,
-    SafariItem
-}
-
 class PokemonHelper extends TmpPokemonHelper {
```
RedSparr0w, line 1: Do we still need this file?

RedSparr0w (approved):

src/scripts/pokemons/PokemonHelper.ts
RedSparr0w: Ah if we remove this file then we need to add a bunch of refs:
```ts
/// <reference path="../declarations/pokemons/PokemonHelper.d.ts" />
```
Will merge as is for now as there is a lot of refs for it.

PR #5113, opened 2024-02-17
Fix Safari battle text

wormania:
## Description
Fix the safari battle text being 1 turn out of sync with the catch rate/flee rate percentages, now it correctly returns to "is watching carefully..." at the same time the percentages change, rather than on the turn after.

## How Has This Been Tested?
Checked that the status text aligned with the change in catch rate

## Types of changes

- Bug fix


Qwertypop04 (approved):
looks good, thanks for the PR!

PR #5117, opened 2024-02-19
Move DamageCalculator to modules

umbralOptimatum:
Pretty straightforward as the only files referencing `DamageCalculator.ts` are non-typechecked html.

Did have some trouble trying to `declare` a /scripts class exists. The current approach to compiling types makes converting to modules quite a bit messier...


RedSparr0w:
Any reason `DamageCalculator` is no longer a class?

RedSparr0w:
Yeah, it can still remain a class within modules, feels like it makes sense to keep this as a class as majority of the stuff in the file is only used within itself.
and when importing from other files may get confusing trying to import `region` or `weather` for example, may auto fill from this module instead of wherever else it's supposed to come from.

RedSparr0w:
The PokemonHelper one kind of makes sense because of `TmpPokemonHelper` being the actual class within modules and then being extended from our scripts folder for the actual `PokemonHelper` class.

RedSparr0w (approved):
Looks good, thanks for the module conversion ❤️

RedSparr0w:
Yeah could be worth moving all the functionality over to the `TmpPokemonHelper` but figured can be left until it's not needed.

PR #5120, opened 2024-02-20
Alola rebalance and story quests

osherwott:
Part of the broader Alola rework. This does not include changing distribution, nor were any missing routes added.

## Description
Added temporary battles and gave Alola a more gradual difficulty curve
Added story questlines; `'Eater of Light'` is now Necrozma's dedicated questline, noted by the quest step requirement
Fixed some details in `Alolasvg.html`
Added a few NPCs to liven the region up
Added new bosses for the quests
Changed Hapu's gym to change depending on the quest step
Changed Molayne's gym and Dusk Mane Necrozma battle to have less HP to account for Steel resistance
    - E4 gyms have been increased for Hau to have no more than 25% more HP than them. They were then tweaked to be more equal to each other with type differences taken into account (this part was brute-forced, thanks to everyone that gave me feedback ❤️ )
Included saveKeys for sun and moon flute npcs to be used for the first Necrozma fight, but also a future Solgaleo and Lunala quest.
Added a way to access gyms independent of their requirements.
Added a way to check if a trainer is on their last pokemon. This is only used for swapping trainer images during that time for now.

An extra QuestLine that is being developed for Trials is meant to be playable alongside the story. A few lines for it have been included to stash Mina's Trials' temp battles for now.

## Motivation and Context
I was told to make a quest for the legendaries and it got out of hand... 😅 

Intended to meld the cool aspects of Alola games, so the story is a combined version of Sun/Moon and Ultra Sun/Ultra Moon

This helps Alola feel less like a giant wall, reflected through splitting the story quests up by island, and eases general player progression. Story quests also make Z Crystals available through Trial steps.

## How Has This Been Tested?
- Played through normally
- Played through after update a file that had been at every step in the old quest

## Types of changes
- Balance
- New QuestLine


Qwertypop04 (approved):

PR #5121, opened 2024-02-21
Added mimic encounter type for catch filters

RedzepR:
## Description

Added the ability to set up catch filters for mimics

## Motivation and Context

More and more mimics are added and with the changed EV gained to them a player might wanna use different balls for catching than regular dungeon encounters

## How Has This Been Tested?

Went into relic castle 
caught darmanitan zen with the new filter
didnt catch regular dungeon encounters with the new filter

## Types of changes

- extension of catch filters


RedSparr0w (approved):

PR #5123, opened 2024-02-24
Make Flutes unaffected by Challenges

osherwott:
- Better tooltips for flutes (credit to crobat4, thanks for never deleting your forks 🙏 )
- New tooltip to show flute type boosts that are never mentioned anywhere for some reason
- Made flutes available for all

## Description
- Changed flutes to not be disabled by any challenge
- Added tooltips to better show flute effects
  - Added new info tooltip to show that flutes boost pokemon type damage (wouldn't be surprised if no one knew this was already a thing...)
- Added a row above flutes to show the min amount of gems, similar to how we show the amounts of battle items. It has a tooltip for more specific information.
- Added a button to link to the Gem modal ~~because I stole the html from crobat~~ because we are doing that with other main screen modules.
  - One new thing this tooltip allows is a left click effect but we only have one gem modal so it's useless for us.
- Allowed the "time left" tooltip to show upon hovering of the cell rather than the data.

## Motivation and Context
Gems have no core gameplay function other than being spent on shops or upgrades. Barring flutes from expanding upon gems is not good gameplay design. With this change we can also add more unique gameplay around flutes, including requirements or other effects that aren't pure `+x%` effects.

## How Has This Been Tested?
Opened and tested the hidden row of battle items before and after having the Heat Badge (visible requirement for flutes), and checked the flute tooltips.

## Screenshots:
image

## Types of changes
- Challenge run


RegisCoaxans (approved):
Everything works well and the changes look good to me. Except for a few minor comments.

CypherX (approved):

PR #5135, opened 2024-02-29
Simplify the module conversion process

umbralOptimatum:
As it stands, converting script files to modules is a huge pain in the neck. This PR makes the conversion process much simpler and easier by improving the approach to cross scripts-modules code. There are three major differences in this new approach:

1. Define types for all `/scripts` code used by modules in `TemporaryScriptTypes.ts`

One of the biggest obstacles to the conversion process is untangling the web of dependencies. Almost all the smaller files that would be easier to convert depend upon one of the huge Helper or Controller files, which are difficult and tedious to extricate. The solution is types for all dependencies that haven't yet been converted. This was already partially done in `globals.ts`, but with less type accuracy, and separating the definitions from the global declarations makes them easier to use in `scripts` files. 

2. Type-check all `/scripts` files against their temporary type equivalents from `/modules`

The other weakness of the existing `globals.ts` approach is that it leaves the temporary types disconnected from the objects they describe. The classes described are (hopefully) accurate at the moment definitions are written, but there's no guarantee those descriptions will remain accurate forever. If a module uses a specific definition of a `scripts` method and that method's signature changes, that's an extra hoop to jump through while converting.

With all the temporary type definitions in their own side-effect-free file, each `scripts` file with a corresponding temporary type can reference that file and use it for type checking. Instantiable classes can use the typical `implements TempWhateverType` syntax. Typescript doesn't have equivalent syntax for static class properties yet, so static properties require a separate type and a `StaticClass satisfies TempStaticClassType;` statement after the class definition. (Typescript does turn this into an extraneous `StaticClass;` line in the output JavaScript but this hopefully won't affect anything.)

3. Automatically generate `GameConstants.d.ts` instead of manually updating a second version

I still don't understand why the definitions are done manually in the first place, but defining the enums differently means the types work slightly differently between `/modules` and `/scripts`. I changed the build process to automatically generate `GameConstants.d.ts` as normal, then wrap all the definitions in a `GameConstants` namespace as the manually-generated version did. The result works identically for `scripts` files, aside from a few errors from having used inaccurate enums, which I fixed by folding in #5118.

## Motivation and Context
I don't expect the module conversion to be finished anytime soon, but these changes make it much easier to move files over with fewer errors and without needing to find the exact right root dependency.

I do expect this will cause some merge conflicts, and other folks working locally will presumably have to `npm install` again as the build script changes added another dependency. Short term disruptions to make things easier long-term. I'm hoping this can be reviewed and merged pretty quickly to minimize the amount of problems it'll create (and so I can start on another module conversion next time I have a slow day).

## How Has This Been Tested?
The game compiles and loads as normal. As this is all juggling types, I don't expect any meaningful differences when the game actually runs. It's theoretically possible the difference in enums definition will have some weird niche side effect, but 1) I doubt it 2) that's not going to show up in the lower level of testing this PR warrants.


Aegyo (approved):
All really good changes, sorry it took me so long to take a look at this. I'll leave it to someone else to actually merge because I'm not sure if there are other things going on that this might cause merge conflicts for, but I don't see anything that needs changing here 👍 

---

> I'm hoping this can be reviewed and merged pretty quickly to minimize the amount of problems it'll create

👀

PR #5138, opened 2024-03-01
Battle Cafe UI redesign

CypherX:
## Description
Redesigns the Battle Cafe UI. Part of a larger project with @RegisCoaxans to improve Battle Cafe accessibility, have more information available in-game, and reduce the need to rely on the wiki or other tools to obtain all Alcremie forms.

## How Has This Been Tested?
Tested at various screen sizes

## Screenshots (optional):
image
image

## Types of changes
- UI improvement


AevitasDragonkin:
Can I suggest the following:

`text-danger` class when you don't have enough of a certain berry
image


A way to see how many spins you can actually do, and how many berries that would cost in total
Not sure if the total berries is that important, but it would be nice to see how many spins you can do with your current berry amount
image


Maybe an indicator of when this pokemon can be obtained, as well as how many spins are needed? Unless it's hidden for a reason?
image
image


Any way do display Milcery (Cheesy), or is it intentionally hidden?

CypherX:
> `text-danger` class when you don't have enough of a certain berry

Sure, good idea

> A way to see how many spins you can actually do, and how many berries that would cost in total
Not sure if the total berries is that important, but it would be nice to see how many spins you can do with your current berry amount

How about something like this?
image

> Maybe an indicator of when this pokemon can be obtained, as well as how many spins are needed? Unless it's hidden for a reason?

Not a part of this PR but there are plans to indicate the time of day and spins needed.

> Any way do display Milcery (Cheesy), or is it intentionally hidden?

Was thinking about it but not really sure how or where to display it. We probably want to hint towards it's existence and how it's obtained but dunno if we want to explicitly display it somewhere. Maybe a problem for the future.

Qwertypop04 (approved):

PR #5143, opened 2024-03-03
Fix Brambleghast ID

RedSparr0w:
## Description
Brambleghast currently has the ID 946 instead of 947

## Motivation and Context
Correct Pokédex ID

## Types of changes
- Bug fix


CypherX (approved):
Looks good, thanks

PR #5144, opened 2024-03-03
Tweak Underground trade amount input UI

umbralOptimatum:
Entering a new number into the Underground trade amount input can be a little awkward as it requires a minimum value of 1. If you want to enter, say, 5 you can't just delete the 1 that's already there; you have to select and overwrite it or delete it after typing. I changed the minimum to 0 to avoid this issue and set the trade buttons to lock accordingly.

I discovered and tested this alongside #5006 and it worked perfectly.


RegisCoaxans (approved):

CypherX (approved):
Definitely an improvement, thanks

PR #5145, opened 2024-03-03
[WIP] Tiered Quests

AevitasDragonkin:
# DISCLAIMER
* PROOF OF CONCEPT
* WORK IN PROGRESS
* NOT APPROVED

## Motivation and Context

There have been talks in discord about a system to allow for bigger and longer quests. I've kept most the system working the way it has been before with the addition that quests can now have a tier (values listed below). Tiers will unlock by achieving a certain level.
Treat this still as a proof of concept.

The way it works:
### Starting - Claiming
Quests still work in the same way. You select a quest based on a tier, which you can freely select between the ones you have access to. After starting the tier is locked in. If you want to select a new tier, you should quit the current quest and start it again with a new tier selected.

### Level requirement
Higher tiers of quests will require a high quest level. Still to be decided if this will unlock by just reaching the level, or if there will also be a QP cost to unlock the option.
There will also be a setting where you can select the default tier (still to be developed)

### Amount multiplier
Quests now generate an amount that's a float rather than an integer. This is then multiplied by a value (see table below) based on the tier. The resulting number is then rounded up to the nearest integer. This makes it so that very small ranges like the shiny quest still work. For example, Catch Shiny will now generate a number between 0 and 1. The easy tier will multiply this by 1 and round up to 1. The insane tier will multiply this by 100 and round up to get a number from 1 to 100.

### Quest Points currency
The base currency reward is now calculated based on the quest amount of 1. This is then multiplied by the amount needed to complete the quest including the tier amount multiplier. After that we can multiply it again. This is currently set to x1 for all tiers. I have not yet decided if I want to award longer quest by giving it like 10% more QP. Or if I want to reduce the QP reward because longer quests means less time needed to invest in changing quests as well as less pokedollars needed.

### Experience multiplier
Currently the base experience is calculated from the base amount (same as the easy tier). Getting a higher quest means you currently get a lot more QP without needing to start as many quests. The downside will now be that the experience gain is much slower per hour.

### Contribution multiplier
In the live game, each quest gives 10% contribution to getting a free reset. In this tiered system, higher tiers will give a higher contribution percentage. Which will allow you to skip certain quests if you complete harder ones.

### Refresh cost
The reset cost will also change in the future based on your highest region.

**ALL VALUES ARE STILL TBD**
| Tier  | Level requirement | Amount multiplier | QP Multiplier | Experience multiplier | Contribution | Reset cost |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Easy  | 1 | x1  | x1 | x1 | 10% | 1,000,000 |
| Medium  | 20 | x10  | x1 | x2 | 15% | 1,000,000 |
| Hard  | 30 | x50  | x1 | x5 | 20% | 1,000,000 |
| Insane  | 40 | x100  | x1 | x10 | 25% | 1,000,000 |

## How Has This Been Tested?

Only a few tests done by myself to make sure most of the things keep working. Not in an actual testing phase yet.

## Screenshots (optional):

image
image

## Types of changes

- New feature

### DIFF
--- a/src/components/questModal.html
+++ b/src/components/questModal.html
@@ -64,7 +64,7 @@
                                         </button>
                                         <button type="button" data-bind="text: $data.tier" class="btn btn-success dropdown-toggle dropdown-toggle-split active" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                                         </button>
-                                        <div class="dropdown-menu" data-bind="foreach: ['Easy', 'Medium', 'Hard', 'Insane']">
+                                        <div class="dropdown-menu" data-bind="foreach: App.game.quests.availableQuestTiers">
                                             <button class="dropdown-item" type="button" data-bind="click: function(){ App.game.quests.changeQuestTier($parent.index, $data) }, text: $data"></button>
                                         </div>
                                     </div>
--- a/src/components/questModal.html
+++ b/src/components/questModal.html
@@ -165,6 +165,7 @@
 
             <div class="modal-footer">
                 <div id='refreshQuests'>
+                    <span data-bind="text: App.game.quests.getCompletionProgress().toLocaleString('en-US', { style: 'percent', minimumFractionDigits: 2, maximumFractionDigits: 2 })"></span>
                     <!-- ko if: App.game.quests.freeRefresh() -->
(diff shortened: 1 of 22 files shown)


AevitasDragonkin:
Closing as I've gotten feedback. Next will be properly implementing it, which will be on a new PR.

PR #4955, opened 2023-12-07
Changed Printed name of Rare Candy

VeryConcernedSigh:
Fixed issue with Rare Candy name having an underscore instead of a space

Changed the printed name when you get a Rare Candy from the dream orbs
## Description

It's all in the summary it is super basic

## Motivation and Context

To fix a bug in the game.
Previous Issue that was opened by me: https://github.com/pokeclicker/pokeclicker/issues/4898
Fixes #4898 
## How Has This Been Tested?

this is untested but so little has changed that no testing should be required

## Types of changes

- Bug fix

PR #5148, opened 2024-03-06
fix(DreamOrbController): use item displayName

### LINKED ISSUE #4898: [Bug]: Underscore present in item name
### Confirmations

### What happened?

when opening Dream Orbs I got some rare candies and it said 3 rare_candy
Screenshot 2023-11-21 8 32 45 AM
Dream Orbs

I expected it to say 3 rare candy without the underscore

### Steps to Reproduce

1. Click on open Dream Orbs area
2. Press open Dream Orbs
3. See bug

Google Chrome Os Version 118.0.5993.124 (Official Build) (64-bit)

Chrome

### Relevant log output

_No response_

### Additional Information

this is the first time I have seen this bug and have not yet tried to reproduce it.

xDivisionByZerox:
Fixed issue with Rare Candy name having an underscore instead of a space

## Description

This PR fixes an issue with Rare Candies having the incorrect display name when opening dream orbs. This is caused because dream orb items are completly decoupled from the global item pool (`ItemList`). 

While fixing the bugs I noticd that dream orb items are initialized with an "id". This id did not match the global item id (key in `ItemList`) for items that containing a space.

Keep in mind that there is currently the dream orb item for "Enamorus (Therian)" which seems to not be implemented. This could lead to potential problems in the future. Although, If the naming convention of the project continues being used, this might not be a problem as ids for the other forms of Therian seem to be working consistently.

## Motivation and Context

Fixes #4898

- #4898

## How Has This Been Tested?

I'm not sure how tests work in this projects, so I would apprechiate some help if tests are in fact necessary.

I did test this that the current changes would fix the issue with the rare candy underscore buy leaving the dream orb modal open after receiving the rewards for this day. This allows the `App.game.dreamOrbController.itemsReceived` observer to still hold the value of the current received items.

## Screenshots (optional):

image

## Types of changes

- Bug fix


cartamundi-tvae:
Testing definitely needed, someone also attempted this at #4955 . I have not read everything there, I just know that just changing the ID does not solve the issue according to that PR. Others will know more about it.

xDivisionByZerox:
> Testing definitely needed, someone also attempted this at https://github.com/pokeclicker/pokeclicker/pull/4955 . I have not read everything there, I just know that just changing the ID does not solve the issue according to that PR. Others will know more about it.

I read through the other PR before opening this one. Thats why I noticed as well that chanig the ids will not do anything because dream orb items are completly decoupled from the actual global item list. Thats why you first need to normalize the item ids in the dream orbs to mathc the ones of the gloabl item list and then use said item list to retrieve the actual display name, You can see the in line 107 that now uses the `ItemList` to get the display name:

https://github.com/pokeclicker/pokeclicker/blob/1a8a3687780c67a3a9b037615cb745cd5121c95b/src/scripts/towns/DreamOrbController.ts#L107

xDivisionByZerox:
I'm not that familiar with the entire code base, so if the current changes should still not enought I'll gladly apply more in the future.

xDivisionByZerox:
Ok, I did check the code base and found the `BagHandler` class that they were talking about in #4955. It has a method to retrieve the display name for a `BagItem`:

https://github.com/pokeclicker/pokeclicker/blob/4ca7969335d2a16dcd7a3730257eedf38302fb66/src/scripts/items/BagHandler.ts#L10-L24

I guess, this is the one that should be used to retrieve the item's display name instead of my custom implementation.

---

As a person who looks through the codebase from the outside, this is a bit confusing. If `ItemList` is available globally a developer (like myself) might encouraged to use this directly, while another developer might find the `BagHandler` first and be encouraged to use this instead. From an architectural point of view, this is bad, as there are multiple ways to do the same thing. My current guess is that `ItemList` should not be accessed by anybody globally, but is an implementation detail to the same handlers. I'd suggest applying the [dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) to make sure that access to specific code points is only granted to the ones that should have access in the first place. 

That's just a thought from a third-party person.

RegisCoaxans:
The BagHandler is used to manage items from different features (see `ItemType`) and therefore stored differently. Its usefulness varies depending on the context. In this case,  the BagHandler is not required anymore, you can either keep using it or drop the underground ItemType and use ItemList.

xDivisionByZerox:
> [...] the BagHandler is not required anymore, you can either keep using it or drop the underground ItemType and use ItemList.

I'm generally touching as less files as possible when making a PR. If the `BagHandler` is not required anymore, I'd say that should be a different PR in which the feature gets removed and related iteam types (`DreamOrb`'s) get changed.  
That's why I'd say that this PR will simply use the `BagHandler`, if that's fine for you.

RegisCoaxans (changes requested):
The changes can not work because of the underground item.
You use their regular item names while pointing to their underground counterparts. It causes the game to crash.
Can you either convert their `ItemType.underground` to `ItemType.item` (preferred) or change the name back again, please ?
Also, for testing, you should be following [this](https://github.com/pokeclicker/pokeclicker?tab=readme-ov-file#building-from-source).

xDivisionByZerox:
> Also, for testing, you should be following [this](https://github.com/pokeclicker/pokeclicker?tab=readme-ov-file#building-from-source).

Thank you 🙏

---

Tested with 10k Orbs each:

<table>

<tr>
<th> Pink </th>
<th> Green </th>
<th> Orange </th>
</tr>

<tr>
<td></td>
<td></td>
<td></td>
</tr>

</table>

CypherX (approved):

PR #5149, opened 2024-03-06
Item price consistency fixes

umbralOptimatum:
- Changed `Item` prevent base prices of 0, which the class methods can't handle. Items that had base prices set to 0 now use Infinity instead, consistent with the other items that can't be directly purchased with currency.
- Base pokeballs having fixed price is now set on the item rather than being a hardcoded exception.


CypherX (changes requested):
Game crashes when loading, probably has to do with removing that Currency import from the global namespace.

```
Uncaught (in promise) ReferenceError: Unable to process binding "template: function(){return { name:'currency-display-template',data:{ currency:Currency.money,tooltip:'Pokédollars<br /><br />Gained by defeating Pokémon and trainers in battle.'}} }"
Message: Currency is not defined
```

CypherX (approved):

Issue #4605, opened 2023-07-29
[Bug]: Nestball giving +15% in all dungeons


AevitasDragonkin:
### Confirmations

- [X] I have searched open and closed issues for any existing report of my issue

### What happened?

1. Using a Nestball in a dungeon will always use the +15%, as the currentRoute is 0, and the maxRoute / currentRoute returns Infinity.
2. The minimum value will always be equal to the player's highestRegion, in the case of Galar, this ball will always have a minimum catch bonus of +7% (not sure if this is a bug or intended)

```
new Pokeball(GameConstants.Pokeball.Nestball, () => {
                const highestRegionRoutes = Routes.getRoutesByRegion(player.highestRegion());
                const maxRoute = MapHelper.normalizeRoute(highestRegionRoutes[highestRegionRoutes.length - 1].number, player.highestRegion());
                const currentRoute = MapHelper.normalizeRoute(player.route(),player.region);

                // Increased rate for earlier routes, scales with regional progression
                return Math.min(15,Math.max(1,player.highestRegion()) * Math.max(1,(maxRoute / currentRoute)));
            }, 1250, 'Increased catch rate on earlier routes', new RouteKillRequirement(10, GameConstants.Region.johto, 34)),
```

### What did you expect to happen?

Incorrect catch bonuses

### Steps to Reproduce

1. Select a route / dungeon
2. Run the part of the code that calculates the NestBall catchbonus as a console log

```
{
    const highestRegionRoutes = Routes.getRoutesByRegion(player.highestRegion());
    const maxRoute = MapHelper.normalizeRoute(highestRegionRoutes[highestRegionRoutes.length - 1].number, player.highestRegion());
    const currentRoute = MapHelper.normalizeRoute(player.route(),player.region);

    // Increased rate for earlier routes, scales with regional progression
    console.log(Math.min(15,Math.max(1,player.highestRegion()) * Math.max(1,(maxRoute / currentRoute))));
}
```

### What browsers are you seeing the problem on?

Chrome

PR #5150, opened 2024-03-09
Fix nest ball catch rate inside dungeons

umbralOptimatum:
Fixed nest balls using their maximum catch chance inside dungeons, as per #4605. They now use the current dungeon's `difficultyRoute`.

## How Has This Been Tested?

Tested by running the nest ball `catchBonus` function inside several routes and dungeons to confirm the new calculations were right.


RegisCoaxans (approved):

CypherX (approved):
Looks good, thanks

PR #5153, opened 2024-03-09
Hatchery notification flavor text for fossils

umbralOptimatum:
The hatchery notifications used the same generic "An egg is ready to hatch!" message for fossils, which was a bit boring, so I changed it to identify the specific fossil. I also made both the notification and the hatch button use "revive" instead of "hatch" for fossils.


CypherX (approved):
Looks good, thanks

PR #5156, opened 2024-03-10
Notification duration settings

### LINKED ISSUE #3868: [SUGGESTION]: Notification timer setting
### Current Behavior

Notificiations appear and then quickly disappear, frequently before they can be read

### Improved Behavior

I would love a setting probably in the notifications tab that allows for the time a notification displays to be modified.  A simple text field to enter a number in seconds would be fine.

### Reason for change

Notifications are frequently removed before I have a chance to read them, requiring me to go over to the logbook negating the usefullness of the notification pop-ups

### Detailed Description

Provide a setting option to the notifications settings tab that allows for the editing of how long notifications display on screen.

AevitasDragonkin:
## Description

- Updated the NotificationSetting to also support duration setting
- The notifier will now check based on priority: 1. Value is provided in the function call 2. Value is in the settings 3. Default value of 3000
- All default values have been taken from the current live values
- All settings names are now left aligned for better readability
- Flutes now have their own setting and is no longer shared with battle items
- Purify chamber now also has it's own setting

Note:
- Event start cannot be customised because it's calculated
- Duplicating items in the underground with the Oak Item now takes the base setting and adds the same amount of seconds as previously
- All values displayed are listed in seconds

## Motivation and Context

Closes #3868 

This allows the player to customise their notifications more so that they key choose what's important to stay longer on the screen and what's not.

## How Has This Been Tested?

Changed a few settings and checked if it changed the duration correctly.

## Screenshots (optional):

image

## Types of changes

- New feature
- UI improvement
- QoL


RegisCoaxans:
Are people not complaining only because the default notification duration can be too short in some cases ?
I feel like this is an overkill and adds too much to the setting menu. Can we not add a setting for a minimal notification duration instead ?

If few specific notifications are too short or too long for idling or active time reasons, we can change them individually to match their use.

AevitasDragonkin:
I think complaining itself was mostly because of the 1 that was already fixed before, which was the max flow. I believe this just adds a lot of customisation with minimal effort (and personally I don't think it adds a lot to the menu).

It would be mainly for active/idling time reasons, as well as what content you're doing and what stage of the game you are in.
Right now I'm thinking of:
* Battle items: once you have days or weeks active it can be helpful to have a longer message if they have ran out rather than the 3 seconds.
* Farm: If you have farms that run for hours it might be more helpful to have it stay up longer versus just planting Cheri
* Resistant: This is also one I would set a lot longer

But feel free to close if it's just overkill

PR #5160, opened 2024-03-21
Items icons and search by item name in Held Item modal

Crobat4:
## Description

Adds icons to the items in the Pokémon list in the held Item modal and the ability to search Pokémon who have certain items equipped.
Also increases the modal width.

## Motivation and Context

Visually ease the search of Pokémon with certain items equipped (i.e.: finding Pokémon with Power Bracer to change said item for a damage boosting item)

## How Has This Been Tested?

* Check in the modal if the items icons show up
* Use the search bar and change the search dropdown to see if searching by item works and searching by Pokémon didn't break.

## Screenshots (optional):

image
image
image

## Types of changes

- UI improvement


RegisCoaxans:
Should we remove the "Show only Pokémon using this item" setting, then ?

AevitasDragonkin:
Maybe replace it with "Hide pokemon holding this item", which in my opinion is a more useful option than the show only holding this item?

Crobat4:
Replaced "Show <ins>only</ins> Pokémon using this item" with "Hide Pokémon holding <ins>this</ins> item".
Also moved both checkboxes to two columns on desktop screens.

Crobat4:
Made the requested changes

Crobat4:
Changed the images width from 40% to 25%, they seems to look fine now (except for the scents but those are because they are small).

CypherX (approved):

PR #5170, opened 2024-03-30
Rare Candy levels and level evolutions

### LINKED ISSUE #5078: [SUGGESTION]: Rare Candies to act like Gen 8+ candies
### Current Behavior

so in many circumstances, a breed is required to evolve certain Pokemon (for example, Yanma into Yanmega in Gen 4+ or Eevee into Sylveon in Gen 6+). In Sword and Shield (and later games) the ability to evolve a lv100 Pokemon with a rare candy was added (before then, a lv100 doing a level up evolution was impossible). I would like to see this possible in Pokeclicker since it synergizes well with the real evo challenge.

### Improved Behavior

Rare Candy item evolving or making a level up evolution prompt immediately true on a lv100 Pokemon.

### Reason for change

A lot of people ask how to evolve mons when they get to Gen 4, this adds an alternative; alongside the aforementioned real evo challenge benefits.

### Detailed Description

see above

umbralOptimatum:
Changed Rare Candies to level up pokemon alongside the attack gain, as you might expect them to. This means they can activate level evolutions, including for pokemon at max level as in the latest games.

Tested by using on pokemon at various levels and confirming their exp and levels were as expected, and that level evolutions triggered when all conditions were met.

Closes #5078.


EatPant2nd:
Considering how this game works... I wonder if it should roll for shiny too.

osherwott:
Do you think you can update the description in ItemList.ts to reflect this new change?

CypherX (approved):

PR #5179, opened 2024-04-04
Added Purified to Click Attack

EatPant2nd:
## Description

Made click attack calculation count Pokémon with a shadow status of Purified.

## Motivation and Context

It was mentioned in Crew channels at one point. Also, the difference isn't that drastic since the number of Purified Pokémon a player can have today is only 131 so max click attack (as in, with all bonuses) doesn't go up by much: from 3,922,463 to 4,095,886.

## How Has This Been Tested?

Click. Click. Click. Also tested that purifying Pokémon did make click attack go up. Then clicked some more.

## Types of changes

- Balance (?)


Qwertypop04 (approved):

PR #5184, opened 2024-04-10
Slightly improve applyRouteBindings() performance

umbralOptimatum:
Change GameController.applyRouteBindings() to only look for tooltip bindings inside the town map element, which should be slightly faster.


RedSparr0w (approved):

PR #5185, opened 2024-04-11
Everstone

Sinonea:
## Description

I've modified the item "Everstone" to prevent evolutions (by leveling and by stone) and babies (by hatching). The item can be hold only if the Pokémon has <strong>available</strong> evolution/baby and the player haven't the evolution/baby.

## Motivation and Context

My first motivation are for the weird runs. I'm thinking at the Normal Run : Fletchling is Normal and lose it's type Normal by evolving. It's my first motivation. The item exists in the video games and I find this very fun. For the weird runs also, the babies may be annoying, so the everstone prevents the babies.

## How Has This Been Tested?

I've tested the code with gulp. I've began a new game and I've catch Bulbasaur, Pidgey and Rattata. I've give the Everstone and the 3 Pokémons haven't evolved. When I remove the Everstone from Pidgey, he evolves to Pidgeotto. For the ItemEvolution, I've tested on a Haunter, I've use a Linking cord on him and he doesn't evolve. When I've decided to enlarge the item to all Pokémons, Metapod was available to hold the item despite Butterfree being caught, I dug through Metapod's evolutions and saw that there were 4 evolutions without triggers, so I fixed the code. I've seen Pikachu is available despite Raichu being caught, Pikachu can evolve in Alolan Raichu in Alola, so I have limited the use of the item at the highest region of the player. Eevee to Leafeon in Etherna Forest to test the environment evolutions : in the case it's level evolution, it works ! Works too for the Burmy forms (Thanks to EatPant for the tests). I've tested for the mégas, I've seen Pidgeot when I'm in at Hoenn, I'm thinking it's a bug because I'm not at Kalos. EatPant told me I can mega-evolve only if I've the megastone and no need to be at Kalos. I've added the Pigeotite and one Key Stone for test and I could mega-evolve Pidgeot despite I'm only at Hoenn. So I made it so that the Everstone is only available if the player have the megastone associated to the Pokémon. For the babies, I've tested with Baby Kangaskhan. Kangaskhan and Méga-Kangaskhan were available. I've give an everstone at Kangaskhan, no at Méga-Kangaskhan, i've make an egg from Kangaskhan : no Baby Kangaskhan. I've make an egg from Méga-Kangaskhan : Baby Kangaskhan appears. Amazing !

## Screenshots (optional):

image

## Types of changes

- New feature : Everstone prevents evolution and babies.

PR #5189, opened 2024-04-13
Everstone (fixes)

Sinonea:
fixes for #5185


CypherX (approved):
Looks good, thanks for the fix. In the future please look into using branches for your changes

PR #5191, opened 2024-04-13
Map color priority setting

klpkt:
Added a dropdown for each map color, to set it as High, Normal, or Low priority, thus allowing the player to choose whether they want, for example, missing Pokérus resistance to show on the map over missing shinies. (Locked location is always high priotity, and Completed location is always low priority, ensuring they are always first and last in the order, respectively)

Would potentially be better to make it as a completely customizable system like the catch filters (what if I want a separate color to show that I'm missing *both* shiny and resistant) but as that would be a more involved change, I started with this.


CypherX:
superseded by #5614

PR #5195, opened 2024-04-20
Added Random Environment to Battle Frontier

EatPant2nd:
## Description

Battle Frontier was using default environment and it was boring. I made it cycle between different environments every 25 stages. Environment related evolutions do not trigger. Thanks TimeSpazm for helping and simplifying the code.

## Motivation and Context

Grass is boring.

## How Has This Been Tested?

Went to Battle Frontier. Tested some environment evolutions to check that it didn't trigger them.

## Types of changes

- I dunno, visual flair.


EatPant2nd:
I was under the wrong impression that Default shared the background image with another environment. That's why I had removed it originally. Oops! 

In any case, I made it go back to Default once you started a new run and I made it not pick the same environment twice in a row. Since I'm not that good at this, by all means tell me if there are better ways to achieve the same. Thanks!

EatPant2nd:
Leaving it as is since even if I did address the Checkpoint issue, you could still reload the game to clear it since reloads also put it back to Default and I'm not going to find a way to store the current Environment as this serves no real gameplay purpose besides giving you something else to stare at. Besides, re-entering a Checkpoint makes new encounters show up so I don't think it's that jarring.

EatPant2nd:
I know it gets fixed with that but then you can still get Default by reloading the game and that annoys me.

CypherX (approved):

PR #5204, opened 2024-04-24
Adds PKRS and Caught status to all ShardShop and GemMaster modals

Mashdon:
## Description

- Added a **_Pokerus_** and a **_caught_** indicator to all Shard shops (similar as in Gem Masters and other Shops)
- This will apply to all ShopMon, Evolution Stones and Eggs. It will also impact the Fossil Master (as it is coded as a Shard Shop).

- Also preventively added the PKRS indicator for EvolutionStone in GemMaster (even if there is no such use case for now)

## Motivation and Context

I was annoyed that there is a difference between the Shard Trader and Gem Trader in Parfum Palace in Kalos.
So I unified the Modals to both have the indicators. This gives a clear view on which Pokémon are already caught and resisted.

## How Has This Been Tested?

I checked different Shard Shops, Fossil Master, and verified that the indicators are correctly set.

## Types of changes

- UI improvement

### DIFF
--- a/src/components/gemMasterModal.html
+++ b/src/components/gemMasterModal.html
@@ -59,6 +59,13 @@
                                                             'evs': App.game.party.getPokemonByName($parent.item.itemType.type)?.evs() }}">
                                                     </span>
                                                 </knockout>
+                                                <knockout data-bind="if: ($parent.item.itemType instanceof EvolutionStone)">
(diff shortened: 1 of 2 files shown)


AevitasDragonkin (approved):
Looks good

PR #5206, opened 2024-04-25
Knockout deferUpdates performance tweak

umbralOptimatum:
Knockout's deferUpdates setting can substantially improve performance when many observables are updated simultaneously. Setting the global deferUpdates flag only affects observables created after that point. Currently the game sets global deferUpdates when it begins loading, so any observables created before that (like static class properties) do not receive the potential performance benefits. I've moved the setting to the start of initialization so that it will affect every observable.

Marked as draft as this will need in-depth testing. This change could have subtle effects across the whole game. My priorities:

1. Does this break anything? It's certainly possible that some code relies on intermediate updates or a particular order of notifications.
2. Does this noticeably improve performance anywhere? If there's breakage for no clear gain, this might not be worth the trouble.
3. Does this noticeably worsen performance anywhere? Very unlikely in my understanding, outside actual breakage.

As this will be subtle and difficult to test, I've also created a control build on #5207 to compare against without any chance of confounding variables. If anyone does find substantial effects from this PR please verify (using the same save) that it does not occur on that build.


CypherX:
Several achievements seem to be re-notifying on every load

PR #5209, opened 2024-04-26
Small multicategory fixes

umbralOptimatum:
Fixed a couple issues with multiple category support, like resetting not affecting the None category's name or removing a category not removing it from all pokeball filters. Also adjusted the PartyPokemon methods to handle the None category more safely.


CypherX:
Looks good, just one small issue. Resetting categories while category mode is enabled in the dex/hatchery and any category other than None is selected will cause an Uncaught TypeError in the console

CypherX (approved):

PR #5225, opened 2024-05-03
Increase Pokédollar earned in BattleFrontier

### LINKED ISSUE #3775: [SUGGESTION]: Give more Pokedollars in the Battle Frontier
### Current Behavior

The Battle Frontier currently gives a very small amount of Pokedollars after losing.

### Improved Behavior

I think the Battle Frontier should give way more Pokedollars than it currently does.

### Reason for change

I literally got the idea for this change a few minutes ago, after I spent 2 hours in the battle frontier, getting to stage 2752, and only getting 7.5 million Pokedollars. Now I know that the main point of the Battle Frontier is not to gain Pokedollars, but this amount is so miniscule that it's almost better to just not give any Pokedollars at all. (You can make this much money in 5 seconds with Amulet Coin and a singular clear of Peony.)

### Detailed Description

I think a few hundred million Pokedollars for getting this far in the Battle Frontier is a fair ask, especially with how long it takes to reach this stage in the first place.

Mashdon:
## Description

Changes the money earned in BattleFrontier. 
The general idea is to reward a sum of Pokédollars that would be in the same order of magnitude than what farming on a route would be with the current attack power. 

To calculate the specific amount, I try to deduce what would be the normalized route number we can one-shot based on what is the Stage cleared. Then I approximate how much time was spent to reach this Stage level and multiply the payout. 

Copy/pasting the code and detailed comments : 
```
        // we want to know what would be the route we can one-shot based on the stage beaten
        // assuming we ignore variability due to types and weather : if we can beat a certain stage, it means we can one shot a route with Pokémon that has 10 times less HP than each Pokémon in the stage
        // So this *not* super elegant piece of code here will calculate the normalized route number that would match those HP by inverting the PokemonFactory.health() method for the Stage health divided by 10
        const approximatedNormalizedOneShotRoute = Math.ceil(Math.pow(Math.pow(PokemonFactory.routeHealth(stageBeaten, GameConstants.Region.none) / 10, 1.0 / 1.15) * 12 / 100, 1.0 / 2.2));

        // it's impossible to know how much time was spent to be able to reach this stage,
        // but we want to give a reward that would be in the same order of magnitude than what farming on a route would be.
        // So we can approximate by multiplying the payout for "one" Pokémon by the number of stages times 1,5 (1,5 seconds is the minimum time spent on a stage)
        let moneyEarned  = Math.floor(PokemonFactory.routeMoney(approximatedNormalizedOneShotRoute, GameConstants.Region.none, false) * stageBeaten * 1.5);
```

Additionally, we also apply the Money multiplier bonus to that amount now. I think it make sense to do so, but feel free to challenge it.

Currently the implementation is not capped, so for people that have very high attack power (so who can one-shot every routes in Galar), we will reward an amount of Pokédollars greater than what they would have from farming "Path to the Peak", we can see if a cap is needed, but my assumption is that money is not a problem anymore for those kind of players and that they probably don't want to spend 10+hours in BF just for money.

Some stats (before money multipliers) : 
Stage 100 : 72k $
Stage 250 : 485k $
Stage 500 : 2M $
Stage 1000 : 9M $

## Motivation and Context

There is an approved suggestion that is open since a long time about it, I decided to give it a try at solving it.
'Closes #3775'

## How Has This Been Tested?

I tested with my personal save and the payout was satisfactory and matched my expectations : 
Stage 382 cleared with Money multiplier 4,5x rewarded 5.3M $
image

This needs more testing. I'm looking for more feedback based on more saves, please start some BF runs on this PR deployment and let me know your remarks.

## Types of changes

- Reward balance ?

### DIFF
--- a/src/scripts/battleFrontier/BattleFrontierRunner.ts
+++ b/src/scripts/battleFrontier/BattleFrontierRunner.ts
@@ -74,7 +74,16 @@ class BattleFrontierRunner {
         // Give Battle Points and Money based on how far the user got
         const battleMultiplier = Math.max(stageBeaten / 100, 1);
         let battlePointsEarned = Math.round(stageBeaten * battleMultiplier);
-        let moneyEarned = stageBeaten * 100 * battleMultiplier;
+
+        // we want to know what would be the route we can one-shot based on the stage beaten
+        // assuming we ignore variability due to types and weather : if we can beat a certain stage, it means we can one shot a route with Pokémon that has 10 times less HP than each Pokémon in the stage
+        // So this *not* super elegant piece of code here will calculate the normalized route number that would match those HP by inverting the PokemonFactory.health() method for the Stage health divided by 10
+        const approximatedNormalizedOneShotRoute = Math.ceil(Math.pow(Math.pow(PokemonFactory.routeHealth(stageBeaten, GameConstants.Region.none) / 10, 1.0 / 1.15) * 12 / 100, 1.0 / 2.2));
+
+        // it's impossible to know how much time was spent to be able to reach this stage,
+        // but we want to give a reward that would be in the same order of magnitude than what farming on a route would be.
+        // So we can approximate by multiplying the payout for "one" Pokémon by the number of stages times 1,5 (1,5 seconds is the minimum time spent on a stage)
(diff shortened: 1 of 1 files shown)


Mashdon:
This PR is not supposed to be the final solution, there are a few points we need to clarify like 

- Where should we place that code that calculates the "potential one shot route"
- Should we cap the reward somehow ?

PR #5228, opened 2024-05-03
Fix Egg Hunt quest Egg Hunter NPC & remove champion requirement from Easter boss

CypherX:
## Description
Changes the requirements on the Egg Hunter NPC so it is visible while on the appropriate quest step when the event is not active. It will also be visible at all times during the event after completing this quest step.

Also removes the regional champion requirement from the 3 Easter dungeon bosses. It wasn't specified anywhere and was causing confusion.

## Motivation and Context
A step in the Easter Egg Hunt quest line requires talking to this NPC. Currently, the NPC has a `SpecialEventRequirement` for Easter so it becomes unavailable when the event ends leaving the player stuck on this quest step.

## How Has This Been Tested?
Went through the quest line and checked that the NPC was available at the appropriate time.

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #5238, opened 2024-05-05
Swapped Underground with Farm

EatPant2nd:
## Description

Swapped the locations of the Underground and the Farm. With this, I also changed the requirements making the Farm unlock as soon as Route 6 is cleared (or as soon as the player can access Route 11). This makes the Farm unable to be skipped but I doubt this will be problematic.
I removed Farming's "has route access" check since the KeyItem already has this check and I found it unnecessary ~~unless it's linked to something else and I can't see it.~~

## Motivation and Context

With the Farm rework awarding Dungeon Tokens per capture, unlocking the Farm earlier will allow for more weird runs to start ~~and for Professor Oak Challenge people to suffer even more~~. Also, the Underground is unlocked way later and its location could be swapped without really altering anything. Finally, people will hopefully no longer confuse the Underground Entrance with Diglett's Cave since the Farm and this dungeon look vastly different.

## How Has This Been Tested?

Checked the map and its functionality. Started a new file to check that the Farm unlocked correctly. Did the same for the Underground (went to purchase the item to check that it unlocked correctly). Checked old files to see that they didn't lose access to the Farm or the Underground, somehow. 

## Types of changes

- Map "update"


Qwertypop04 (approved):

PR #5241, opened 2024-05-07
First Attempt at Underground Rebalance

EatPant2nd:
## Description

Daily Deals are now always 1:1. Due to this new ratio, I removed the `randomAmount()` function since it's now unused. To compensate for this, most Diamond prices have been cut down by 5 except:
- Underground Upgrades weren't touched.
- Probably (Chimecho) now costs 358 (down from 1,187). 

Also, since these Deals were used to acquire Shards for a dog, this deal is now 50% cheaper.

## Motivation and Context

Daily Deals allowing the existence of these exponential chains have been a big issue for a long time. This is an attempt to slowly mitigate some of the damage while still letting people use these deals to get certain stuff. Be it Fossils, be it future Evolution Stones. This is only a first attempt to do so since a later one should also look for ways to increase Diamond gain in a way that doesn't make people reach `MAX_SAFE_INTEGER` for more than just Diamonds.

## How Has This Been Tested?

Since it's mostly price changes I doubt anything broke in the process. I did check some stores.

## Types of changes

- Balance


Qwertypop04 (approved):

PR #5242, opened 2024-05-08
Allow resuming paused quest lines regardless of requirements

CypherX:
## Description
Follow-up to #5228. Some players paused the bugged Egg Hunt quest line to wait for a fix, but would be unable to resume it until the next time the event came around or by using the event calendar to activate the event early.

This allows paused quest lines to be resumed even if the requirements to start the quest line are no longer met.

## How Has This Been Tested?
Activated and paused the Egg Hunt quest line, checked the bulletin board and it was not available to resume. Reloaded with this fix, and now the quest line was available to resume.

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #5248, opened 2024-05-09
Quick fix for region travel bug

umbralOptimatum:
Fixes changing regions from a subregion sending you to a town outside the destination region. A more general fix can wait until daylight hours, pretty sure I've addressed this in #5002 already and can pull those changes out


CypherX (approved):
Moving to a new region from a sub-region appears to be working correctly now. Using +/- while on a route and moving between sub-regions is still busted but that can be fixed in a different PR

PR #5249, opened 2024-05-09
Quick fix for subregion move issue

umbralOptimatum:
Fixed switching routes via hotkey across subregions causing an invalid route state. More complete fixes and prevention in #5002.


CypherX (approved):
Fix looks good, thanks. Had to remove the added route check and `throw` or the game wouldn't load. `Battle.generateNewEnemy()` gets called on game initialization and it was causing the game to not load.

PR #5255, opened 2024-05-12
Add double battles

RedzepR:
## Description

Temp battles, dungeon battles and gym battles can have the double battle flag
Similar to the mainline game dealing damage to two mons makes it that both of them take 0.75x of the usual amount. When only one pokemon is alive it takes the full 100%
To not change the difficulty of each double battle i multiplied their health by 1.5x (as you do 0.75x to 2 pokemon at the same time), which may not be desireable and/or needs further look into

## Motivation and Context

Double battles are a core mechanic in the game for about 20 years now (first appearing in RSE) but pokeclicker didnt implement them yet. 

## How Has This Been Tested?

Cleared temp battles with double battles
Cleared temp battles with double battles with shadows
Cleared dungeons with double battle (regular and boss)
Cleared dungeons with double battle with shadows
Cleared gym battles with double battles

Would greatly appreciate further testing as a lot of internal battle logic has been touched

## Screenshots (optional):

image

## Types of changes

- type of battle


RedzepR:
> IMO it would be far better to make `enemyPokemon` an array and handle it with loops.

Oh yeah i like that. It would require a bit of refactoring but result in overall better code. (Also it gives the base for inevitable triple battle /j)

> I can absolutely see salt from encountering a shiny or roamer and having it faint first.

Both mons get caught on the same time (shadow trainers with 2+ shadows can make this happen)

RedzepR:
Closed this PR to reopen at another time, wanna start with a fresh branch and cleaner mindset

PR #5259, opened 2024-05-13
Fixed Farm submodule breaking eggsteps again...

RegisCoaxans:
## Description
Kidding !
I just added Farm, UG and Z-Crystal modules to the collapsible module list for settings as they got forgotten.

## Motivation and Context
Collapse status was not saved across loadings.

## How Has This Been Tested?
Reloaded the game a bunch of times with different configurations.
~~Ensured eggsteps stayed the same.~~

## Types of changes
- Bug fix

### DIFF
--- a/src/modules/GameConstants.ts
+++ b/src/modules/GameConstants.ts
@@ -2477,6 +2477,9 @@ export const ModalCollapseList = [
     'pokemonListBody',
     'shortcutsBody',
     'currencyBody',
+    'undergroundCard',
+    'plotListCard',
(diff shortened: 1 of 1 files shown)


AevitasDragonkin:
Should we add the map as well, or is that one deliberately always active on load?

CypherX:
> Should we add the Town Map as well, or is that one deliberately always active on load?

I think it's fine to leave out, not sure who's collapsing the map anyway. Probably shouldn't be collapsible or moveable in the first place

CypherX (approved):

PR #5260, opened 2024-05-14
Click-oriented Z-Crystals

RegisCoaxans:
## Description
Changed how Z-Crystals work. They now help deal more damages from clicks.
The player can activate multiple of them. The best type against the enemy is then picked to deal damages with.
Z-Crystals have an « energy » attribute, which gets consumed when they are used and replenishes from defeating Pokémon of corresponding types.
Not decided on the UI but I did not feel like adding three more rows to the Battle Item Container. Not fully decided on the balance neither. Needs feedback.

## Motivation and Context
Z-crystals were boring. Now, at least it has a unique use. :^)

## How Has This Been Tested?
Checked the click attack, verified energy consumption

## Screenshots (optional):
image

## Types of changes
- New feature


osherwott:
Can you uncomment the following lines of code as well? To make the Z-Moves available.
- Update.ts [Lines 2682 - 2687](https://github.com/pokeclicker/pokeclicker/blob/develop/src/scripts/Update.ts#L2682)
- TemporaryBattleList.ts [Lines 3898 - 3901](https://github.com/pokeclicker/pokeclicker/blob/develop/src/scripts/temporaryBattle/TemporaryBattleList.ts#L3898)
- TemporaryBattleRunner.ts [Lines 124 - 131](https://github.com/pokeclicker/pokeclicker/blob/develop/src/scripts/temporaryBattle/TemporaryBattleRunner.ts#L124)

PR #5267, opened 2024-05-17
Added the amount of underground upgrades available

RedzepR:
## Description

Added the amount of underground upgrades (how many you got, how many there are) to the UI

## Motivation and Context

Too much info is hidden in the wiki, the players shouldnt need to consult another app to learn about the game. So lets change that bit by bit

## How Has This Been Tested?

Looked at the UI

## Screenshots (optional):

image

## Types of changes

- UI improvement


wormania:
I think the upgrades could be clearer, possibly by splitting out to their own column
`+10 (0/10)` sort of implies that a single upgrade would take you to 10/10

CypherX (approved):

PR #5268, opened 2024-05-17
Hide outdated contest town entries in dev

CypherX:
## Description
Contests will most likely be changing entirely from the current dev implementation. This removes the Pokemon Contest and Contest Shop entries from Lilycove. They were only available in dev but there's no point leaving outdated content accessible. The contest pokemon were also appearing as available from the shop on the wiki.

## How Has This Been Tested?
Checked Lilycove


Qwertypop04 (approved):

PR #5272, opened 2024-05-20
Reduced Fossil Pieces underground upgrade

CypherX:
## Description
Adds an additional toggle-able underground upgrade to reduce fossil piece spawns. It is visible in the upgrade list but locked and cannot be purchased until the player has reached Galar (when fossil pieces begin appearing).

## How Has This Been Tested?
Used 🪄 to generate layers and check how many fossil pieces spawned with the option enabled vs disabled

## Screenshots (optional):
image

## Types of changes
- New underground upgrade


Qwertypop04 (approved):

PR #5273, opened 2024-05-21
Fix crash caused by the category option in a pokeball filter when attempting to catch an uncaught pokemon

CypherX:
## Description
Bug from the multi-category implementation. When a pokeball filter is configured with a category option, and that filter is checked while attempting to catch an uncaught pokemon, the game will crash.

## How Has This Been Tested?
Set up a filter with a category option and attempted to catch a pokemon before and after the fix. Crash no longer happens.

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #5277, opened 2024-05-23
Added Certain Johto Pokémon to Kanto Routes

EatPant2nd:
## Description

Added Houndour, Sunkern, and Slugma to specific Kanto routes. They have "Caught + Johto Champion" requirements to both follow canon and to also force the player to acquire them as egg Pokémon first. As for the encounters themselves:
- Houndour: Route 7.
- Slugma: Route 16, 17, and 18.
- Sunkern: Route 24.

## Motivation and Context

There is precedence to re-add egg Pokémon to their canon routes with a caught requirement. Besides, it helps with resisting and it looks neat.

## How Has This Been Tested?

Went to Kanto before having the requirements fulfilled. No Johto. Then went with the requirements. Yes Johto.

## Types of changes

- Route Update


Qwertypop04 (approved):

PR #5281, opened 2024-05-24
Refactor MapHelper for safer movement

umbralOptimatum:
MapHelper movement safety pulled out from #5002. I want to make sure this gets in with plenty of time before the next version releases, so this is all the functional changes without variable renaming, function reorganizing, module conversion, etc. Should make it much easier to tackle.

## Description

- Movement function changes:
  - MapHelper now has `moveToRegion()` and `moveToSubRegion()` functions that confirm that movement is allowed and handle moving to the new region/subregion's default location.
   - MapHelper's `moveToRoute()` and `moveToTown()` functions now check whether the player has access to the corresponding region/subregion and move them there if necessary.
- Access function changes:
  - Added `MapHelper.accessToShip()` to check whether the player can leave the current region.
  - MapHelper now has `accessToRegion()` and `accessToSubRegion()` functions. The former makes sure the player has access to the ship, both confirm the region's starting location is unlocked.
  - The `accessToRoute()` and `accessToTown()` functions also now check region and subregion accessibility.
  - `ableToMove()`, which checks ability to begin the next region, now requires having ship access
- Player setter changes:
  - Moved subregion wrapping-around logic for the prev/next subregion buttons out of the player setter (**why. pain.**) and into dedicated functions
  - The player region setter now won't go higher than MAX_AVAILABLE_REGION.
- `SubRegions.getSubRegions()` no longer only returns unlocked subregions. That's now handled by a new `getUnlockedSubRegions()` function for clarity (almost every use was re-filtering for unlocked status).
- Clean up shipModal
- Add unexplored subregion buttons to subregionModal to match shipModal
- TemporaryBattleRunner is more careful when sending the player back to a town

## How Has This Been Tested?
Tested all the modified functions from various region/subregion distances. Confirmed the modified buttons and menus were behaving as expected. Exhaustive testing will of course take full beta tester attention, but for now this covers every edge case I can think of.


CypherX:
There seems to be a noticeable delay when moving regions that doesn't happen on the dev build

PR #5285, opened 2024-05-26
Settings refactor: pokedex filters

umbralOptimatum:
Refactoring PokedexFilters to exist solely as settings instead of being stored in multiple places at once. Another chunk of #4885, thankfully much smaller than the equivalent BreedingFilters PR.

- Converted pokedex filters back to settings. No more FilterSettings and inexplicable duplication of state.
- Pokedex searchbar setting now fully supports regex and saves across sessions.
- Added translation support to pokedex menu
- Tweaked pokerus filter to have "Uninfected" only show *caught* uninfected pokemon
- Caught status filter now waits to unlock shadow and purified status options until Orre quest has started

## How Has This Been Tested?
Changed all the settings a few times, confirmed the expected pokemon were shown, confirmed the settings persisted after a refresh.


CypherX:
> Caught status filter now waits to unlock shadow and purified status options until Orre quest has started

I don't think this is working right, I can see the Caught Not Shadow / Caught Shadow / Caught Purified options on a brand new file

CypherX (approved):

PR #5295, opened 2024-05-31
Add new PC profiles

### LINKED ISSUE #5294: [SUGGESTION]: Allow Trainer model for Green
### Current Behavior

Allows you to be a character in the overworld map

### Improved Behavior

Allow a selectable trainer model to be green

### Reason for change

Green is cute and she is not in the selection

### Detailed Description

Add green to the list of selectable trainers, Story characters like may are already in and some popular gym leaders and elite four. Why not green from the original pokemon?

DraKay:
Add 3 new profile characters for player personalization 

## Description
Add green and ghensin sprites, fix jostav duplicate credits. As green have 2 different outfits I made both just in case. 

## Motivation and Context
This characters were missing while other similar are avaibale. (red was but not green, other bad guys where) 
Closes #5294
Pant asked to fuse jostav credits 

## How Has This Been Tested?
E... 

## Types of changes

- UI improvement


Qwertypop04 (approved):

PR #5296, opened 2024-06-01
Make sevii island roamers be determined by starter

RedzepR:
## Description

Changed the sevii island roamers so only one of the beasts is roaming. If Pikachu is chosen Entei is roaming (same as bulbasaur)

## Motivation and Context

5 roamers are a bit too much. EVs can be obtained from shadow bosses in orre. Kalos handles it the same with the three birds. Its ✨canon✨

## How Has This Been Tested?

Checked the roaming NPC on island 6

## Screenshots (optional):

image

## Types of changes

- roaming pokemon


osherwott:
I believe this isn't something that we want, since this is an optional subregion and no progress is blocked by too large of a roamer pool. The philosophy behind Kalos' roamers change was different from Sevii's because the birds felt unfair since they diluted the pool by completing an optional dungeon. Conversely, the Butterfree here are the new pokemon, and they are unlocked as a reward upon completing the Orange Island gyms.

PR #5302, opened 2024-06-03
Made Exposed Applin Bug / Dragon

EatPant2nd:
## Description

Made Exposed Applin, previously mono Dragon, be Bug / Dragon instead.

## Motivation and Context

Qwerty said it was fine and alternate forms tend to have more unique typings.

## How Has This Been Tested?

Opened the PokéDex, Damage Calculator, and other things that show typing. Pokémon appears as expected.

## Types of changes

- Type update


Qwertypop04 (approved):
Looks good, thanks for the bug fix!

PR #5304, opened 2024-06-05
Shortcut module updates

CypherX:
## Description
Updates the Shortcuts module to display immediately upon unlocking one of the options (Underground, Farm, Poke Mart) rather than hiding it until Johto. Also displays the configured hotkey next to the shortcut button, and fixes the Safari battle button hotkeys not updating when changed.

## Motivation and Context
With the addition of the farm and underground modules to the main view, along with the hotkeys working as soon as the features are unlocked anyway, it doesn't make sense to hide the shortcuts module until the player reaches Johto.

## How Has This Been Tested?
Tested that each shortcut becomes visible/hidden when expected, tested that the entire module is hidden if no shortcuts are visible.

## Screenshots (optional):
image


RedSparr0w (approved):

PR #5305, opened 2024-06-05
Ability to toggle pokeball filters from main view

### LINKED ISSUE #4760: [SUGGESTION]: Faster way to toggle catch filters
### Current Behavior

To activate or deactivate a catch filter, you need to click on the settings icon and the on the toggle button.

### Improved Behavior

To activate or deactivate a catch filter, you would just need to click on the catch filter in the list.

### Reason for change

Toggling a catch filter would be (a tiny bit) faster. Also, consistency with how you can change the catch filter ball directly from the list.

### Detailed Description

You can change the ball for the catch filter directly from the list, or also from the setting menu. You can only toggle the filter from the setting menu.
The clickable area for toggling the filter could the name area, since the ball icon area is already a button.

CypherX:
## Description
Adds the ability to right click (or long press on mobile) to toggle pokeball filters enabled/disabled from the main view without opening the filter menu. Also added text to mention this functionality.

Closes #4760 

## Motivation and Context
QoL

## How Has This Been Tested?
Right clicked a few times

## Types of changes
- UI improvement


RedSparr0w (approved):
Works well, thanks for the changes!

PR #5309, opened 2024-06-06
Refactor and standardize hatchery egg types

umbralOptimatum:
The hatchery has a lot of redundant and inconsistent code to handle different kinds of eggs. I've reworked the types to remove redundancy and standardized + simplified the methods for fossils and shop eggs.

## Description
- New `Breeding.addItemToHatchery()` method that supports both Fossils and EggItems
  - Neither need to concern themselves with internal hatchery details anymore
- Fossils now have their own item subclass with a hatchery method and shared interface with EggItem
  - This replaces "selling" them through the underground to hatch them
  - Though maybe they should just be made a subclass of EggItem, since they're almost the same thing? At least they should have some relation to CaughtIndicatingItem.
- `EggType` no longer distinguishes between different EggItems, that's now solely handled by `EggItemType`
- Fossils now go by their item name rather than display name in the hatchery and associated GameConstants
- Removed unused `Pokemon_egg` item

## Motivation and Context
This will make future improvements to the hatchery much less of a headache.

## How Has This Been Tested?
Tested updating save files with all four EggTypes of egg, adding and hatching various EggItems and Fossils, changing caught status for EggItems and Fossils


src/modules/GameConstants.ts
```diff
@@ -1082,12 +1082,25 @@ export enum StoneType {
...
-    'None' = -1,
-    'Fossilized Bird',
-    'Fossilized Fish',
-    'Fossilized Drake',
-    'Fossilized Dino',
+    'Fossilized_bird',
+    'Fossilized_fish',
+    'Fossilized_drake',
+    'Fossilized_dino',
 }
```
CypherX, line 1104: `FossilType` and `FossilPieceType` don't appear to be used anywhere now, is there any reason to keep them?

CypherX (approved):

PR #5317, opened 2024-06-11
Use Moon phase requirement for lunar duo bosses and roamers

osherwott:
## Description
Added moon phase requirements to Darkrai and Cresselia. Cresselia roams during all phases except Crescent and New Moons, when it "rests" at Fullmoon Island as a dungeon boss. Darkrai behaves the opposite way.

## Motivation and Context
Adds some relevancy to moon phases, and also provides an example use as a PR. The amount of phases they roam were chosen to not interfere too much with their obtain methods. Cresselia is still mostly a roamer, and Darkrai is still mostly a dungeon boss. The phases themselves were chosen based off pokedex entries, where Darkrai is active around the new moon, and Cresselia is associated with the quarter moon. We can always change these later, if needed.

Also thought this would be a nice substitute for their questline, until it's made.

## How Has This Been Tested?
Added the requirements and checked the dungeon list and roamer npc.

## Screenshots:
image

## Types of changes
- Fun


Qwertypop04 (approved):

PR #5325, opened 2024-06-16
Changed the "and" in some of the Jessie & James npc names to &

aheadfrogg:
## Description

Changed the "and" in the Jessie & James NPC names in the Pinkan quest line and on New Island to &.

## Motivation and Context

I wanted the all the names for Jessie and James to be the same and I like & more than and.

## How Has This Been Tested?

I started the quest line and visited New Island to see if the names were updated and they were.

## Types of changes

- Text update


Qwertypop04 (approved):

PR #5341, opened 2024-06-19
Use Pokerus Progress for routes and dungeons instead of default icons

DreamNya:
## Description

Use Pokerus Progress for routes and dungeons instead of default icons, and it can be toggle in settings.

## Motivation and Context

The default PKRS icons are so boring, players can't know the specific progress from them.
So I wanna change it, and I think it will be very cool, if players can know at a glance the specific pokerus progress of routes and dungeons or even the total numbers of pokemons on the routes.

The default icons are PNG that can't be modified easily, so I used CSS to simulate it. But I'm poor at CSS, I have no idea if there's better way to handle it. Maybe I need some help.

## How Has This Been Tested?

Toggle the new setting and then observe whether the changes are the same as below.
`Uninfected`: None.
`Infected`: Only icon. Whether enabled or not.
`Contagious & Resistant`:  Icon if disabled or Progress if enabled.

## Screenshots (optional):

dungeonView
dungeon
route

## Types of changes

- UI improvement

### DIFF
--- a/src/components/dungeonView.html
+++ b/src/components/dungeonView.html
@@ -122,12 +122,12 @@
                   <!--Pokérus image-->
                   <knockout data-bind="using: DungeonRunner.dungeon.allAvailablePokemon()"">
                       <knockout data-bind="if: RouteHelper.minPokerusCheck($data)">
-                          <!-- ko let: {minPokerus: RouteHelper.minPokerus($data)} -->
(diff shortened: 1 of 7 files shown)


RegisCoaxans:
Closing this.
With the addition of the route info modal, I believe routes and dungeons are both clear enough as to which Pokémon still need to be EV-trained and how many they are.
Thanks for the PR, though.

PR #5345, opened 2024-06-19
Hide Z-Power Ring key item from key items menu

CypherX:
Z-Power Ring is not currently obtainable but is displayed in the Key Items list


Qwertypop04 (approved):

PR #5358, opened 2024-06-20
Prevent mistakenly leaving a dungeon while a guide is hired

CypherX:
## Description
Prevents the player from mistakenly leaving a dungeon and firing a guide while one is hired. While a guide is hired the Leave Dungeon button will be hidden and ability to manually leave a dungeon disabled. The player will have to fire the guide to then leave the dungeon.

## Motivation and Context
Due to how guides work the "Leave Dungeon" button and interaction is active immediately upon (re-)starting a dungeon while using a guide. This leads to players mistakenly exiting the dungeon, firing the guide, and losing the spent currency if they happen to be clicking during the boss fight.


RedSparr0w (approved):
Looks good thanks 👍

PR #5359, opened 2024-06-21
Update translations for settings refactor

umbralOptimatum:
Updates translations for the changes made in #5271 and #5285, which renamed several setting keys. Also fixes `breedingDisplayTextSetting` showing the internal name instead of a human description.


CypherX:
Translation changes get pulled in automatically, this shouldn't be needed. I checked dev and the hatchery label was showing correctly now

PR #5370, opened 2024-06-23
Fix PokemonLocations returning too-high-region pokemon

umbralOptimatum:
Fixed PokemonLocations incorrectly showing certain pokemon as available with a `maxRegion` before their native region:
- Pokemon obtained through evolution where the base pokemon's native region is at most `maxRegion` but the evolved pokemon is native to a later region
- Farm wanderers showing as always available

## How Has This Been Tested?
Compared the output of `PokemonLocations.getPokemonLocations()` for every combination of pokemon and maxRegion and confirmed that only the intended changes had occurred.


CypherX:
I think something might be wonky with the `getPokemonWandering` change, calling `PokemonLocations.getPokemonLocations('Treecko')` says it's only available from an egg rather than both an egg and wanderer.

CypherX (approved):

PR #5371, opened 2024-06-23
Fix multi-category update error for very old files

CypherX:
## Description
Found this while testing a different PR with a very old file. Game would error upon updating, then again when opening the pokedex or hatchery.

## How Has This Been Tested?
Loaded a v0.6.7 save and everything proceeded to work as expected

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #5377, opened 2024-06-24
Updated Some Farm-related Stuff

EatPant2nd:
## Description

Updated Colbur's and Petaya's descriptions to include their effects: Kasib mutation prevention and the timer pause, respectively. 
Added more information to the Mulches part of the Farm's Help tab.

## Motivation and Context

Colbur and Petaya were a bit vague with their descriptions, especially Colbur where the player had no way of knowing it prevented Kasib from mutating. As for the mulches... Many times it's asked how long they last so I added that missing information to the Farm's Help tab.

## How Has This Been Tested?

Checked the respective modals.

## Types of changes

- Text update

### DIFF
--- a/src/scripts/farming/Farming.ts
+++ b/src/scripts/farming/Farming.ts
@@ -1179,7 +1179,7 @@ class Farming implements Feature {
             BerryFirmness.Very_Hard,
             [
                 'This Berry is surrounded by mystery. It is rumored to be imbued with the power of all living things.',
-                'This power resets and pauses the wither timer of other Berries indefinitely.',
+                'This power revitalizes and prevents other berries from withering.',
             ],
             undefined,
             ['Mew']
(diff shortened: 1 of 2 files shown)


src/scripts/farming/Farming.ts
```diff
@@ -1178,7 +1179,7 @@ class Farming implements Feature {
             BerryFirmness.Very_Hard,
             [
                 'This Berry is surrounded by mystery. It is rumored to be imbued with the power of all living things.',
-                'This power keeps other Berries alive for longer.',
+                'This power resets and pauses the wither timer of other Berries indefinitely.',
```
RegisCoaxans, line 1182: It seems too straightforward and technical for in-game information. What do you think of this :
```suggestion
                'This power indefinitely saves other Berries from withering.',
```

EatPant2nd: I mean, it's better than the original but I wanted some way to convey that it resets timers too as that was the sole reason I edited it.

CypherX: > Revitalizes and prevents other berries from withering

EatPant2nd: That's perfect, thanks!

CypherX (approved):

PR #5380, opened 2024-06-25
Fix farmDisplay collapse contents breaking after disabling module

umbralOptimatum:
Fixed disabling a collapsible module breaking DisplayObservables' tracking of the collapse's state. This mainly affected the farmDisplay, breaking its contents if it was disabled while collapsed or breaking the performance-saving bindings if it was disabled while open.

## How Has This Been Tested?
1. Collapse `#plotListCard`
2. Disable the showFarmModule setting
3. Enable the showFarmModule setting

On develop this consistently breaks the contents of the farm module as DisplayObservables thinks it's still hidden. Works as expected after these changes.


CypherX (approved):
Seems fine, appears to fix the issue and didn't notice any other problems while testing besides the small thing I mentioned in discord. Gonna merge this now to be sure it gets into the update

PR #5385, opened 2024-06-26
Make Spiritomb a trade instead of a shopmon

RegisCoaxans:
## Description
Costs 108 diamonds and an Odd Keystone.
Changed shard trades to not only use QP but the currency of the traded item.

## Motivation and Context
More canon. :^)
With the changes in the UG, 108 diamonds seem to be a fair price for a required Pokémon.

## How Has This Been Tested?
traded for a bunch.of Spiritomb and still tried few other trades.

## Types of changes
- New feature


Qwertypop04 (approved):

Issue #4917, opened 2023-11-29
[SUGGESTION]: 2 mega stone dont have any clue ingame to find it


milotic:
### Current Behavior

if you dont check the wiki no exist a clue ingame in how get Aggronite and Tyranitarite because the way to get both megastone isnt tie with any normal gameplay because both require beat a gym leader 2000 times and the max achievement of beat any gym leader is 1000 times

### Improved Behavior

both megas stones need a clue ingame
### Reason for change

Better documentation

Qwertypop04:
Resolved by #5389

PR #5389, opened 2024-06-28
Cyllage npc

hikimoi:
## Description

Added an NPC to Cyllage City that gives a hint how to obtain the Tyranitarite and the Aggronite. 

## Motivation and Context

Until now there was no hint in the game on how to obtain said Mega Stones. Players have to check the wiki to find out how to obtain them.

## How Has This Been Tested?

Made the changes, loaded the game with a file that has both Aggron and Tyranitar and one that doesn't have both and checked if the NPC is there.

## Types of changes

- "Bug" fix
- UI improvement
- It's just a new NPC for completion sake


milotic:
the NPC need a hint about beat the leader 2000 times because in a normal gameplay the max achievement is 1000: https://github.com/pokeclicker/pokeclicker/issues/4917

hikimoi:
> the NPC need a hint about beat the leader 2000 times because in a normal gameplay the max achievement is 1000: #4917

The NPC hints at beating the leader "enough" times, Don't have to tell the exact solution to every riddle, a hint in the right direction suffices, in my opinion.

I would prefer to lower the requirement from 2000 to 1000, though. But I don't know if that is something that we want.

CypherX (approved):

PR #5392, opened 2024-06-28
Changed Key Stone price

RegisCoaxans:
## Description
Base price is 250 and get multiplied by 1.1 on each purchase. It never goes down again, reaching its peak of usual 25000 at 49 purchases.
The price multiplier is reset for everyone, just in case.

## Motivation and Context
That makes for a much more linear progress toward obtaining all the Mega Pokémon, without making it too easy.

## How Has This Been Tested?
I bought a bunch of Key Stones. Checked the price increase, and made sure it does not decrease.

## Types of changes
- Feature improvement


Qwertypop04 (approved):

PR #5394, opened 2024-06-29
Improved save selector context menu access

CypherX:
## Description
Adds a vertical ellipsis to the top left of save selector trainer cards that can be clicked/tapped to access the context menu.

## Motivation and Context
Recently we learned that certain mobile devices/browsers (iOS / Safari) don't handle the `contextmenu` event so accessing the context menu to download a copy of a save from the save selector screen is not possible on these devices.

## Screenshots (optional):
image
image

## Types of changes
- UI improvement


RedSparr0w (approved):

PR #5395, opened 2024-06-29
Fix dungeon guide next floor timer bug

CypherX:
While a dungeon guide is hired if the player manually progresses to the next floor the guide will continue to explore but the timer will remain paused.

Also includes a small update fix for a multi-category bug for very old files

## How Has This Been Tested?
Manually clicked the ladder tile and next floor button and verified the timer continued.

## Types of changes
- Bug fix


RedSparr0w (approved):

PR #5399, opened 2024-07-02
Fix galar DLC encounters

RedzepR:
## Description

The encounter tables (both weathered and regular) for almost all galar DLC routes were wrong so i fixed them (data source bulbapedia)

## Motivation and Context

Found this issue during gameplay and wanted to fix it

## How Has This Been Tested?

Look at possible encounters depending in different weatehr

## Types of changes

- Bug fix


Qwertypop04 (approved):

PR #5400, opened 2024-07-03
Set farm wanderer to higher priority than mulch

umbralOptimatum:
Makes catching farm wanderers take priority over adding mulch when clicking. Interacting with a berry already behaves this way, so this makes the UI more intuitive.


RegisCoaxans:
But then we can not put Gooey Mulch anymore before attempting to catch the wandering Pokémon.

DraKay:
Would it possible to make player choose the action order similar as pokeball? So they can order even mulch individually?

RegisCoaxans:
At this point they could just pick the berry tool, lmao.

PR #5410, opened 2024-07-07
Workaround? Fix? For the obscure achievement, flute, locale bug

CypherX:
## Description
Implements a workaround for the crash some players experience on load while the pokemon attack flute is active during certain periods of time.

No idea what the root cause is or why it only affects some locales and only happens during certain periods of time, but this seems to take care of it.

## How Has This Been Tested?
The two files I have that are affected by this bug both load successfully with these changes.

## Types of changes
- Bug fix????


RedSparr0w (approved):

PR #5414, opened 2024-07-09
Updated Attack Modifier Setting to Include Attack Bonus

EatPant2nd:
## Description

Attack Bonus was currently not displaying attack multipliers (Held Item, EVs, Shadow status) like Breeding Efficiency was so the value was unhelpful. I moved the Settings' check to Attack Bonus and now enabling or disabling the corresponding checkbox will affect both values.

## Motivation and Context

Attack Bonus is used in Breeding Efficiency but the multipliers were only making Breeding Efficiency update correctly, leaving Attack Bonus with just the raw value that, at most, had Vitamins. Making both change is more helpful for the player.

## How Has This Been Tested?

Checked the values. Bred some Pokémon to see that the value displayed was correct.

## Types of changes

- Bug fix (?)


Qwertypop04 (approved):

PR #5431, opened 2024-07-13
Changed EventCalendar to correctly calculate the QP needed 

RedzepR:
## Description

With the old system you would always pay for one less day than actually needed, resulting in a free activation when the event is less than 24h away

## Motivation and Context

It being free on the day prior to the event seemed more like an oversight than a feature

## How Has This Been Tested?

Activated an event, saw the correct number 

## Types of changes

- Bug fix


Qwertypop04 (approved):

PR #5432, opened 2024-07-13
Reduced Fossils toggle

CypherX:
## Description
Adds a Reduced Fossils upgrade toggle that's unlocked and enabled by default. Some weirdos might want to find more fossils so why not let them?

## How Has This Been Tested?
Checked that the fossil weights were calculated correctly with and without the pokemon/fossil and while the toggle was enabled/disabled


CypherX:
No longer relevant with the underground rework

PR #5440, opened 2024-07-16
Changed Roamer Base Rate so each SubGroup goes from 8192 to 4096

RedzepR:
## Description

Changed it so that every SubGroup of Roamers has their own gradually increasing Roamer Base Rate regardless of other SubGroups in the region.

## Motivation and Context

Roamer Base Rate looked at each region as a whole, despite different subgroups not having the same roamer pool. This resulted in stuff like MKJ region or sevii 4567 to be all max roamer rate (1/4096) which is just a byproduct of how the roamer code was developed.

## How Has This Been Tested?

Checked every SubRegion so that the first route has a roamer rate of 1/8196, the last route a rate of 1/4096 and every other route something inbetween.

## Types of changes

- QoL


Qwertypop04 (approved):

PR #5443, opened 2024-07-17
Added MegaMedicham + NPC to the game

### LINKED ISSUE #5290: [SUGGESTION]: Method to obtain the Mega Stone for Medicham
### Current Behavior

As I understand, the Medichamite is now the only Mega Stone that does not have a definite obtain method. As it stands, the last Mega Stones should be rather uniquely obtained, and not just be another quest or temporary battle, as so many other Mega Stones have been before. 

### Improved Behavior

There's Mega Stones in the underground, the battle frontier, in quest lines and temporary battles locked behind a variety of different requirements, and as dungeon loot and available to be bought in shops. The normal quests to gain QP and the Achievements remain untouched by Mega Stones so far.

### Reason for change

This will complete the set of Mega Pokemon, as Mega Altaria and Mega Banette have been discussed before and have methods to obtain them in future updates. 

### Detailed Description

One possibility to handle the acquisition of the Medichamite would be as a additional reward when handing in quests for QP. This could have a similar requirement as the underground Mega Stones, meaning they will become available at the same time.
Similar to [#5114](https://github.com/pokeclicker/pokeclicker/pull/5114), this would be a random event, happening only once.
As quests can be spammed rather quickly, the chance for this reward should be rather low, probably 1% or 0.5%. The number of required quests with bad luck wouldn't be too bad, and give players another reason to work towards the achievement for 5000 quests.

Another possibility is a specific quest, which will be able to turn up with similar requirements, or possibly once the player has become the Champion of Kalos. This quest is guaranteed to reward the Mega Stone, but has a very low chance of appearing.
This quest could simply be "Defeat xxx Fighting-type Pokemon", "Defeat xxx Psychic-type Pokemon", "Defeat xxx Medicham in Victory Road Hoenn/Trainer Encounters", "Defeat Gym Leader xxx a few times" or "Catch xxx Pokémon in Dungeon/Safari/whatever specific location".
To prevent the player from refreshing their quest list until they discover the new quest, it could for example only appear if the quest list refreshes itself once all quests have been completed, although I see an issue with many players not realizing that this is a mandatory requirement.

A last possible acquisition method could be an automatic reward once a specific threshold of achievement% has been reached. Unfortunately I do not have any numbers of normal achievement values of a normal playthrough in Kalos, but for the sake of the example, let's say 500%. As soon as the player crosses 500% achievement bonus, they will be rewarded the Mega Stone. The reward could be handled in similar fashion as Original Color Magearna. This will of course not be very interesting for players who have already completed the game to a certain point, but provide a nice reward for grinding out achievements to new(er) players.

RedzepR:
## Description

Added a way to obtain mega medicham. When claiming a quest there is a chance (Quest Level - 15)/8192 (proposed by Hayenn) to obtain the mega stone. 

## Motivation and Context

Closes #5290 

## How Has This Been Tested?

Talked to the NPC
Obtained the Mega Stone
Evolved Medicham

## Screenshots (optional):

image

## Types of changes

- new pokemon


Qwertypop04 (approved):

PR #5444, opened 2024-07-17
Nerfed dungeon trainers

RegisCoaxans:
## Description
Added a HP modifier for trainer Pokémon in dungeon based on how many Pokémon they have in total.
It does not just straightly divide HP by that number, so trainers are still overall tougher the more Pokémon they have.
Also increased base HP in Pattern Bush to now match the difficulty of nearby dungeons.

## Motivation and Context
Motivated by me losing to dungeon trainers in Unova, lul.

## How Has This Been Tested?
I did some dungeons.

## Types of changes
- Balance

## Notes
Maybe there are some other dungeons which need a new balanced base HP but I think I correctly looked over the dungeon list.


Qwertypop04 (approved):

PR #5446, opened 2024-07-19
Make the gym leaders speak about their badges properly

VocalFan:
## Description

Skimming through the GymList.ts file after experiencing this oddity in game. 
I have now made it so Gym Leaders don't sometimes say BadgenameBadge, and instead Badgename Badge
(Like "ThunderBadge" is now "Thunder Badge") as the badges are shown in the Badge Case.
I also made it so some Gym Leaders remember their badge's name
(Like "Take this Gym Badge!" is now "Take this Knuckle Badge!")

## Motivation and Context

Consistency and it also makes the Gym Leaders not seem rushed or... Forgetful, haha.

## How Has This Been Tested?

I SURE DIDN'T. It's just text changes and this is my first pr lol-

## Types of changes

- UI/Text changes

### DIFF
--- a/src/scripts/gym/GymList.ts
+++ b/src/scripts/gym/GymList.ts
@@ -11,7 +11,7 @@ GymList['Pewter City'] = new Gym(
     ],
     BadgeEnums.Boulder,
     250,
-    'I took you for granted, and so I lost. As proof of your victory, I confer on you this... the official Pokémon League BoulderBadge.',
+    'I took you for granted, and so I lost. As proof of your victory, I confer on you this... the official Pokémon League Boulder Badge.',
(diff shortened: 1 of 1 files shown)


EatPant2nd:
It pains me to write this but the Kanto badge names are like that on purpose.

VocalFan:
> It pains me to write this but the Kanto badge names are like that on purpose.

Are you uh... Sure? Because...

image
The badge case sure has it properly spaced-

Askellown:
Also "Take this Gym Badge" is the direct quote of Brawly's from RSE. He calls it the Knuckle Badge in the next line of text, which we don't include for size.

PR #5448, opened 2024-07-19
Meowth with Phanpy painted on his butt...

RegisCoaxans:
## Description
Added Meowth as disguised in [this episode](https://bulbapedia.bulbagarden.net/wiki/EP230).
It appears when the player owns Phanpy at level 20 or below, after at least one hatch, near where the events happen in the series.
Added Officer Jenny in Mahogany, which hints towards Meowth's location.

## Motivation and Context
It is a funny reference to the anime.

## How Has This Been Tested?
Tried to find Meowth under correct and wrong circumstances.

## Types of changes
- New alt form !!

## Notes
I am not sure about the name at all. I would appreciate ideas.
The used name format implies the Pokémon impersonates another one, which is not the case here...


osherwott:
It is generally not a good idea to put these kind of pokemon on routes or locations where a player may encounter them through normal gameplay. I think the requirement is a good start.
Honestly our routes are pretty limited in complexity, I'm unsure what advice to offer besides moving it to Sevii, where Phanpy also appears, or in a shop...

RegisCoaxans:
I want to explore unique obtaining methods for special Pokémon.
Encountering this one is not as easy as you think : it takes 8 encounters not to match the requirement anymore with no XP bonus at all, 5 with Lucky Egg, even less with achievement bonus. The odds of a player noticing and getting it are very low.
I can lock this behind Johto story line and talking to the NPC, though, if you prefer.

osherwott (commented):
Oh, those _are_ some good odds. Considering Phanpy is found after the quest, on Route 45, I would say it's fine as is. 👍

Qwertypop04 (approved):

PR #5460, opened 2024-07-24
Fix "infinity" farm actually being infinite

CypherX:
## Description
"Infinity" farm, as it's been dubbed, should last for ~3.5 years. Due to the precision rounding on the berry age it was actually infinite. This removes the precision so the age ticks correctly.

## Screenshots (optional):
Current:
before

After change:
after

## Types of changes
- Bug fix


EatPant2nd:
Wait, you are gonna let it keep the 3.5 years? Lol.

CypherX:
> Wait, you are gonna let it keep the 3.5 years? Lol.

I mean, that part is working as intended

Qwertypop04 (approved):

PR #5462, opened 2024-07-24
Fixed butler's location in the meta groudon questline

aheadfrogg:
## Description

Moved the butler3 npc from jagged pass to mt. chimney to match the questline text and where it is supposed to be,

## Motivation and Context

I did this so new players would be less confused on where the npc is supposed to be.

## How Has This Been Tested?

I did the questline to make sure the butler was in the correct spot and it was.

## Types of changes

- Bug fix


Qwertypop04 (approved):

PR #5464, opened 2024-07-25
Added Gyarados in Coba

aheadfrogg:
## Description

I replaced Farfetch'd with Gyarados in the Coba berry wanderer list.

## Motivation and Context

Farfetch'd is already available on a Kanto route and Gyarados being a wanderer makes certain challenge runs more possible.

## How Has This Been Tested?

I loaded up a file and checked to see if Gyarados was a wanderer now and it was.

## Types of changes

- New feature

### DIFF
--- a/src/scripts/farming/Farming.ts
+++ b/src/scripts/farming/Farming.ts
@@ -810,7 +810,7 @@ class Farming implements Feature {
             BerryFirmness.Very_Hard,
             ['This Berry is said to be a new kind that is a cross of two Berries brought together by winds from far away.'],
             undefined,
-            ['Farfetch\'d', 'Murkrow', 'Tropius', 'Chatot']
+            ['Gyarados', 'Murkrow', 'Tropius', 'Chatot']
         );
 
         this.berryData[BerryType.Payapa] = new Berry(


Razorflamekun:
I don't agree with this change.  The special wanderers for the Coba Berry are **supposed** to be Flying types, and Gyarados does not make sense for this berry, in my opinion.

Blorog1:
> I don't agree with this change. The special wanderers for the Coba Berry are **supposed** to be Flying types, and Gyarados does not make sense for this berry, in my opinion.

Gyarados is a flying type.

xslk:
Original wanderers were selected based on PRIMARY type

aheadfrogg:
Hey what's the primary typing of farfetch'd, tropius, ans murkrow

xslk:
normal, grass, dark. There are a finite amount of pokemon per generation that can be added as each berry's generation wanderer. 

Aside from that, adding one of the defacto best BE pokemon as a wonderer feel pretty broken.

Qwertypop04:
> Original wanderers were selected based on PRIMARY type

If this was true, coba wouldn't have any wanderers until Kalos, with Noibat

aheadfrogg:
How is it broken? It's faster to get gyarados from other places by the time you would want to get more of them and you can get a magikarp before Mt moon

xslk:
> > Original wanderers were selected based on PRIMARY type
> 
> If this was true, coba wouldn't have any wanderers until Kalos, with Noibat

True, gave a better reasoning since it has been quite awhile since that original PR and had to look at the notes I had when I made it. 

> How is it broken? It's faster to get gyarados from other places by the time you would want to get more of them and you can get a magikarp before Mt moon

Maybe it's not. End of the day I don't approve anything. 🤷

Razorflamekun:
> > I don't agree with this change. The special wanderers for the Coba Berry are **supposed** to be Flying types, and Gyarados does not make sense for this berry, in my opinion.
> 
> Gyarados is a flying type.

I'm very well aware of this, but my argument still stands, as well as the arguments raised by xslk, and a few arguments raised in the poke-crew chat.

Overall, this change does not help mono dark because you would still need the Halloween event in order to make it happen, so the mono dark argument is null.

PR #5465, opened 2024-07-25
Moving Mew to be a Lum berry wanderer

aheadfrogg:
## Description

I moved Mew from Petaya to Lum as a wandering pokemon.

## Motivation and Context

Lum as a berry makes more sense for Mew to be a wanderer for since it also holds the berry. Also if a player has awful luck getting a mew to spawn as a roamer, they can get it as a wanderer instead.

## How Has This Been Tested?

I opened up a pokeclicker file and looked at the lum wanderer list to check if mew was there and then I checked the petaya wanderer list to make sure mew wasn't there and both are true.

## Types of changes

- Feature Change


AevitasDragonkin (approved):
Code looks good

PR #5470, opened 2024-07-27
Added team rocket balloon as an easter egg

RedzepR:
## Description

When you have a full dex you can use the Professors blimp to move to the next region. But if you don't (disable full dex challenge) you have to resort to team rocket smuggling you across.

## Motivation and Context

The team rocket balloon is pretty iconic and DraKay did an amazing job spriting it so it would be a shame to let it go to waste.

## How Has This Been Tested?

Check the balloon position in every region.

## Types of changes

- easter egg


Qwertypop04 (approved):

PR #5473, opened 2024-07-29
Removed weather from Pokémon attack in Attack achievement

RegisCoaxans:
## Description
That is it. I also excluded some other factors. The policy here is : only things the player has full control over are taken into account.

## Motivation and Context
Reported on Discord. If the BF ignores the weather, the achievement should as well.

## How Has This Been Tested?
Toggled on and off modifiers and found no issue.

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #5480, opened 2024-08-03
Underground V2

AevitasDragonkin:
## Description

This is a complete overhaul of the current Underground system:
* Different types of mines containing specific loot
* No more energy system, but tools working on a cooldown
* Helpers
* Rebalanced diamond gains and costs
* Underground experience and level system
* Oak item changes
* Trading 
* New pokemon
* A lot of UI upgrades
* ...

Everything is explained in detail in the UG ingame Help tab.

## Motivation and Context

The initial entry point to take a look at revamping the UG was because of the value of diamond, and it's income and cost.
The rebalancing part of this update increases the income of diamonds towards the same level as QP for end-game players who will have levelled their UG to a certain level. This increase in diamonds gained also makes it possible for other content to lower the diamond cost for things that were already costing only 1 diamond.

## How Has This Been Tested?

Only personal testing has been done so far. Now open for beta testing

## Types of changes

- New/Revamped feature
- Balancing


src/components/genericTraderModal.html
```diff
@@ -0,0 +1,100 @@
...
+                    <div style="display: grid; column-gap: 0.5rem; grid-template-columns: auto auto 1fr auto auto auto 1fr auto;">
+                        <div class="font-weight-bold" style="grid-column: 1 / 3">Cost</div>
+                        <div class="font-weight-bold">Inv</div>
+                        <div class="font-weight-bold"></div>
+                        <div class="font-weight-bold" style="grid-column: 5 / 7">Profit</div>
+                        <div class="font-weight-bold">Inv</div>
+                        <div class="font-weight-bold"></div>
+
+                        <!-- ko foreach: GenericDeal.getDeals(ShopHandler.shopObservable().traderID) -->
+                        <!-- ko let: { deal: $data } -->
```
CypherX, line 25: Can this `foreach` and `let` be simplified with this? https://knockoutjs.com/documentation/foreach-binding.html#note-3-using-as-to-give-an-alias-to-foreach-items

src/modules/deal/GenericDeal.ts
```diff
@@ -0,0 +1,186 @@
...
+
+export type GenericTraderShopIdentifier = 'Palaeontologist' | 'EverstoneDealer';
+
+type DealCostProfit = {
+    amount: number,
+    hidePlayerInventory?: boolean,
+};
+
+type GemDealCost = {
+    type: 'Gem',
```
CypherX, line 17: Would it make sense to have this `type` be an enum rather than string?

src/components/genericTraderModal.html
cartamundi-tvae: Resolved

src/modules/deal/GenericDeal.ts
cartamundi-tvae: Resolved

CypherX (approved):
the children yearn for the mines

PR #5484, opened 2024-08-06
Changed Oak mutations and tweaked legendary berries

RegisCoaxans:
## Description
- New mutation pattern for formerly Oak Item mutations, equipping the Oak Item still doubles the mutation chances.
- Legendary berry mutations now only require the related questline completed, catching the related Pokémon doubles the mutation chances instead.
- Snover berry now is a rare drop item from Snover. Chances depends on the mutation layout currently used, even if the mutation isn't unlocked.

**(Description updated with information from later posts.)**

## Motivation and Context
Those mutations were outdated. No Dex Challenge came out since then, and forcing people to catch specific Pokémon is therefore not something we want much anymore as this is restrictive.
Also, No Oak Item Challenge uses an awful workaround for the related mutations.

## How Has This Been Tested?
Triggered the mutations, tested the mutation chances to make sure they were correct.

## Screenshots (optional):
No screenshot. Go find the setup. :^)

## Types of changes
- New feature


RegisCoaxans:
- The legendary berries now require the story line of their respective Pokémon to be complete (Land vs. Water / A New World), back from requiring nothing.
- Owning the Pokémon associated to a legendary berry now doubles its mutation chances. Previously, I made it so not owning it halved these.

RegisCoaxans:
Reintroduced Oak Items in their mutations : just like legendary berries, not required but having them doubles the odds.

Added Snover berry as a rare drop from Snover (as well ?). Chances vary in accordance to the mutation setup used, whether or not the mutation unlocked.
And, with that, no berry is locked behind having a Pokémon in the party anymore.

CypherX (approved):

PR #5487, opened 2024-08-08
Added rare held items to the alolan geodude line

aheadfrogg:
## Description

Added the everstone and hard stone to the alolan geodude line's rare hold item drop.

## Motivation and Context

They now have the same rare held item drop as their kantonian counterparts, like the other regional variants.

## How Has This Been Tested?

I loaded up a file and made sure that the pokemon had the correct rare held item drops.

## Types of changes

- Bug fix


Qwertypop04:
I've checked multiple sources and this is incorrect, the only item they've ever held is a cell battery

PR #5499, opened 2024-08-18
Limit first Silvally items to one purchase

umbralOptimatum:
Prevent buying more than one of each initial-price Silvally item. It's not a big deal that you can buy multiple at once at the lower price, as they can't gain EVs yet, but it's still not intended.

## How Has This Been Tested?
It hasn't, I don't have a save in the right place.


Qwertypop04 (approved):

PR #5505, opened 2024-08-20
Only create defeat dungeon quests from unlocked regions

gero3:
## Description

This PR makes sure that you only create defeat dungeon quests for unlocked dungeons by checking if a dungeon is unlocked for the highest region and if not, use the second highest region as the highest region. The dungeon quest also will not be unlocked before you can access viridian forest.

## Motivation and Context

I was quite perplexed when I just arrived at Johto and got a quest I couldn't have access to. It seems it just selected Sprout Tower dungeon 50 percent of the time. 
https://discord.com/channels/450412847017754644/1019758760589271071/threads/1275413432606265496

## How Has This Been Tested?

I loaded the following save and kept reloading the quest list. 
[[v0.10.21] PokeClicker 2024-08-20 11_16_23.txt](https://github.com/user-attachments/files/16676558/v0.10.21.PokeClicker.2024-08-20.11_16_23.txt)

## Types of changes

- Bug fix

### DIFF
--- a/src/scripts/quests/questTypes/DefeatDungeonQuest.ts
+++ b/src/scripts/quests/questTypes/DefeatDungeonQuest.ts
@@ -16,14 +16,23 @@ class DefeatDungeonQuest extends Quest implements QuestInterface {
         this.focus = App.game.statistics.dungeonsCleared[GameConstants.getDungeonIndex(this.dungeon)];
     }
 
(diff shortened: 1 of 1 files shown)


PixLSteam:
Does this actually help the player in any way? You should not be able to go back to the previous region before reaching the dock, which is usually way past the first dungeon.

PR #5506, opened 2024-08-20
Quest now check if dock is unlocked to generate quests

gero3:
## Description

Quests now check if the dock is unlocked to make sure you can do the quests in the previous regions. It will now only generate quests in the region of the player if the dock isn't unlocked. 

## Motivation and Context

I was quite perplexed when I just arrived at Johto and got a quest I couldn't have access to. 
https://discord.com/channels/450412847017754644/1019758760589271071/threads/1275413432606265496

## How Has This Been Tested?

I loaded the following save and kept reloading the quest list.
[[v0.10.21] PokeClicker 2024-08-20 11_16_23.txt](https://github.com/user-attachments/files/16676558/v0.10.21.PokeClicker.2024-08-20.11_16_23.txt)

## Types of changes

- New feature


gero3:
Hi! I wanted to follow up on this PR. It’s been a while, and I’m wondering if you’re still interested in merging it. If not, I’m happy to close it. Thanks for your time!

Issue #5393, opened 2024-06-29
[SUGGESTION]: Unown and the rotation of some letters


milotic:
### Current Behavior

Some letters have no canonical reason to find until Sinnoh or are only locked in one place. 

Example letter E are locked until sinnoh with no reason.
!? are locked to sevii but canonwise you can unlock both in HGSS and DP meeting the requisites.

![Captura de pantalla 2024-06-28 202247](https://github.com/pokeclicker/pokeclicker/assets/1099848/ed7e261b-dd82-4d7c-8c35-f611a6f0c272)

### Improved Behavior

all base letters appear in all regions and the 2 extra unowns are unlocked in jhoto/sinnoh until you catch both in sevii island

### Reason for change

locking 6 letters with daily rotation in a single place isnt funny to get pokerus.

### Detailed Description

Unown E, F, H, P appear in all places 
 
!,? appear in all places only if you already catch other letters and both in the sevii island

Qwertypop04:
Issue with Unowns not appearing for too long resolved by #5507. Removing Unown exclusivity to certain dungeons rejected

PR #5507, opened 2024-08-21
More Unowns when clearing a dungeon

AevitasDragonkin:
MultiRequirement now has the option for any requirement complete instead of only every requirement

A second unown will be present when you've cleared the dungeon 100 times
A third unown will be present when you've cleared the dungeon 250 times

## Motivation and Context

Unowns should now be easier to find. As Unown H would only occur in February 2025 (we are August 2024 today)

## How Has This Been Tested?

Cleared some dungeons, saw that more unowns were present as I cleared 100 and 250 times.

## Screenshots (optional):

image

## Types of changes

- Improvement


milotic:
hi, Can add the sign unowns in the rest of places but only if the player already catch both? canonwise both appear in sinnoh and jhoto if the player catch the other 26 to close this: https://github.com/pokeclicker/pokeclicker/issues/5393

CypherX (approved):

PR #5511, opened 2024-08-26
Island Scan Fixes

AevitasDragonkin:
## Description

Hide Marill and Marshtomp when it's not Saturday by adding hide: true so that the requirement is taken into account.
Fixed Axew now spawning on Saturday instead of Sunday

## How Has This Been Tested?

Loaded file on Monday, checked that they weren't there

## Types of changes

- Bug fix


Qwertypop04 (approved):

PR #5530, opened 2024-09-17
Give the Ice Stone at rare item to Cristal Onyx

Sinonea:
## Description

Replace the old rare item of Crystal Onix by his Evolution Item (Ice Stone).

## Motivation and Context

My first motivation is that the Ice Stone as such cannot be obtained normally in Alola. Crystal Onix being obtained in Hoenn, that makes 4 regions (Hoenn/Sinnoh/Unova/Kalos) where it cannot evolve. The only current way to evolve it before Alola is to pray for the ice stone in the berry shop in Johto. If we don't take care of the farm or have bad luck (the ice stone doesn't show up for a long time), we have to wait for Alola. I discussed it with @EatPant2nd who advised me to add the stone as a rare item to Crystal Onix himself. I asked the CC community for feedback and a developer ( @Qwertypop04 ) gave me the green light for the PR.

## How Has This Been Tested?

I looked at the Crystal Onix Pokédex sheet and saw that it actually gave the Ice Stone. I haven't done the dungeon to verify that Crystal Onix actually gives the ice stone but I assume that is the case.

## Screenshots (optional):

image

## Types of changes

- New feature : Ice Stone to Crystal Onix


Qwertypop04 (approved):

PR #5535, opened 2024-09-17
Implementation of Boxes

Sinonea:
Add the system of boxes.

## Description

Implementation of boxes.
A Pokémon can be freely placed in or removed from boxes via the Pokédex modal.
A Pokémon in a box has the following properties:
- Do not participate in the attack
- Do not participate in the click bonus
- Do not gain experience
- Cannot be hatched
- Do not gain EV
- Cannot evolve
- No modification of Vitamins / Held Item
- No Consumable use
- Continue to earn Dungeon Tokens via catch

The system can be enabled/disabled (it is disabled by default) in the options and when disabled, all Pokémon in the boxes come out. A confirmation is asked to enable the system. The icons of the boxes in the Pokédex are visible only if the system is enabled.

The Pokédex has a filter to see Pokémon depending on whether they are in boxes or not, only visible if the system is enabled.

When a Pokémon is deposited/removed, a notification is launched. It can be deactivated in the notifications menu.

Pokémon in boxes are not displayed in the Party list.

## Motivation and Context

Official video games have a system of boxes to store pokemon that you don't need. With weird runs in mind, you can decide yourself to put Pokémon in the boxes to avoid including them in the party. I initially had the idea of ​​completely removing the Pokémon from the party but @EatPant2nd (yes, them again) told me about "unique" Pokémon and that we cannot recover once we got them back. It was then that they told me about a developer who would agree to a box system which could be a good compromise. Something I did in this PR.

I also decided to hide the boxes option so as not to annoy players who don't care about it. By default, the system is disabled and it is only visible when you go to the game options and activate the system yourself. It’s really for those who want to challenge themselves. The confirmation request is present precisely to alert that the system was created for advanced players.

## How Has This Been Tested?

I tested the GUI to check that everything works well and that when you press the box button, the Pokémon is placed in the boxes. Once the Pokémon is in a box, I checked that all the properties are validated. Once everything was checked, I removed the Pokémon from the boxes and verified that the Pokémon returned to its normal status.

## Screenshots (optional):

So that the images speak for themselves, I will stay in the same configuration throughout the screenshots.

Here the config :
Roucool/Roucoups/Mew not in Boxes | Rattata/Rattatac/Machopeur in Boxes.
image

I can begin to show the screenshoots in its config.

image
image
image
image
image
image
image
image
image
image
image
image
image

## Types of changes

- New feature : Boxes


osherwott:
Closing as rejected. Boxing party members isn't something we're interested in, and it doesn't mesh well with the intended gameplay.

Sinonea:
Don't worry about that. Implementing boxes was fun and helped me understand vanilla code better. So it's not a complete waste of time. Thanks for your quick response, @RegisCoaxans !

RegisCoaxans:
All good, then.

Sinonea:
Yes. All good. You're the best @RegisCoaxans ! Thanks for all you do for the dev of the game. Stay as you are, you are perfect.

PR #5538, opened 2024-09-18
Changed No Gem Upgrades tooltip

EatPant2nd:
## Description

Changed the tooltip from "damage multipliers" to "type effectiveness". Also, capitalized the B in Gems modal when Gem Bonuses are disabled.

## Motivation and Context

Damage multpliers was confusing enough that people thought it may prevent Flute usage too.

## How Has This Been Tested?

With my eyes.

## Types of changes

- Text change


Qwertypop04 (approved):

PR #5551, opened 2024-09-20
Fix transfering held items in real evo

CypherX:
## Description
Attack boosting held items were not being correctly included in the attack calculation after being transferred to the evolved pokemon when real evo was enabled. The held items were not lost but the effect was not active until the game was reloaded.

[discord bug thread](https://discord.com/channels/450412847017754644/1286460809626517598)

## How Has This Been Tested?
Evolved a pokemon holding an attack boosting held item with real evo enabled and verified the held item bonus was calculating correctly

## Types of changes
- Bug fix


RegisCoaxans (approved):

Qwertypop04 (approved):

PR #5561, opened 2024-09-26
Add more pokemon to outside the Friend Safari

osherwott:
## Description
Added more pokemon outside the friend safari, most of them with an ObtainedPokemon requirement

- Giratina (Origin) was added to the Distortion World, and (Altered) was moved to Sendoff Spring, to give the dungeon some more importance. We should change the questline sometime to not make these spawns so awkward.
- Probopass was added to Chargestone Cave because Magnezone is also available in Unova
- Crabominable was added to Mount Lanakila with a requirement of clearing the dungeon 100 times, but we can change that if we don't like it
- Crowned Zacian and Zamazenta were added to Slumbering Weald with similar weight distributions as their base forms in Energy Plant. Following the rarity of 1/5, they should have a 1/6 encounter chance here
- Archeops was added to Path to the Peak because most other fossils are found in Crown Tundra, and that route could use some more pokemon
- Jumpluff, Darmanitan, and Lycanroc (Dusk) were added to their respective dungeons for thematic reasons.
- Pinkan Pidgeotto now appears alongside Pinkan Pidgey once obtained
- Burmy and Wormadam appear in Kalos dungeons that don't have otherwise unique pokemon to them and are also appropriate to their forms

## Motivation and Context
Makes routes and dungeons more interesting

## How Has This Been Tested?
N/A

## Types of changes
Pokemon distribution


osherwott:
- Undid the event pokemon I moved, as I feel like they'd be better dealt with separately (and not necessarily by me)
- Updated this with Burmy and Wormadam locations. I put them in Kalos dungeons that have no other unique pokemon in them, to give more dungeons some relevance.
  - This also serves to differentiate from #5562, which is a pr focused on gameplay elements instead of friend safari availability. This helps it have a more focused discussion

Qwertypop04:
- I dislike the changes to Giratina since it makes the current method for Origin Giratina completely pointless, but if this were to be changed the quest line would need to be changed alongside it, not just "sometime"
- The Burmy and Wormadam locations seem completely random. Why are they being added to these locations?

osherwott:
- Sure I can undo those Giratina changes. Btw what about it makes the obtain method pointless? I thought all I had to do was add an ObtainedPokemonRequirement to it so players would still need to grind for the griseous orb before they EV it. I don't know what else I missed.
- Burmy and Wormadam dungeons were chosen through the following thought process:
  - They can't be on routes, because the farm is already a passive way of getting Burmy (Plant)
  - They can't be in Sinnoh, because part of the design philosophy for EVs is to motivate players to advance
  - There are too many pokemon in Unova's dungeons, and I didn't want to dilute the pool
  - Kalos has some dungeons that don't have anything unique to them, like Connecting Cave, so I put Sand Cloak there first, and the rest followed

We can choose something else for Burmy and Wormadam, like adding them all to the farm somehow (Burmy could be a color wanderer, but Wormadam would need something more specific by virtue of being an evolution), but since they are forms and players only need one to progress, I thought it would be more interesting to figure out something in #5562, and have them in dungeons as a plan B.

Qwertypop04 (changes requested):
It's taken me quite a while to get round to this, since I struggled with where to draw the line. I'll keep it simple and give you a list of which I'm accepting and which are denied

Accepted: Pinkan Pidgeotto, Archeops, Jumpluff, Darmanitan, Zacian (Crowned Sword), Zamazenta (Crowned Shield)

Denied: Burmy (All), Wormadam (All), Probopass, Crabominable, Lycanroc (Dusk)

Additional note, could you remove the weight from the trainer boss in Relic Castle?

osherwott:
Changes applied 👍

Qwertypop04 (approved):

PR #5562, opened 2024-09-26
Add requirements for wanderers

osherwott:
## Description
Wanderers can now have requirements. I'd appreciate if we used this sparingly, and not copy-paste requirements from wanderer pokemon's other obtain methods, to keep some variety.

## Motivation and Context
I wanted to include the burmies in a fun way

## How Has This Been Tested?
Locked the wanderers and then unlocked them.

## Types of changes
Pokemon distribution


CypherX:
Not a fan of environment requirements on wanderers, or the way this is coded.

Why would you separate the requirements from where the wanderers are defined? A better way would be to create and use a class similar to `SpecialRoutePokemon`

osherwott:
Can finally take this off my to do list thanks to someone else tackling it.
Closing in favor of #6300

PR #5567, opened 2024-09-30
Super Secret Starter Skip

CypherX:
## Description
Already possible today through the use of hotkeys, this implements an "official" way to skip the starter at the beginning of the game. Triggered the same way as unlocking Pikachu as a starter, but more (20).

## How Has This Been Tested?
Skipped the starter


Qwertypop04 (approved):

PR #5573, opened 2024-10-03
Make keyItemModal display behave nicely with other modals

umbralOptimatum:
The keyItemModal now waits its turn when the NPC, gym badge, and tempbattle reward modals are open.

## Description
Previously, the keyItemModal would by default open instantly upon obtaining a key item, closing any modals that were already open. This obviously didn't work well with key items that are rewards for story progression, so many key items used workarounds to manually display the modal later. This is no longer necessary.

## Motivation and Context
No more janky inconsistent workarounds! We can (and should) now use requirements for all key item rewards. I already converted the two unlocked by badges, which also means the ancient startup checks for those are now handled by the `KeyItems` startup checks. I don't really want to fiddle with the quest ones though; that should be done by someone with more knowledge of those quests.

## How Has This Been Tested?
Confirmed the modal was patient for the badge (Gem Case after Giovanni), tempbattle (DNA Splicers after Ghetsis), and NPC (Reins of Unity after Calyrex) modals.


umbralOptimatum:
Closing until I can clean this up.

PR #5581, opened 2024-10-06
Add Xerneas (Active)

RedzepR:
## Description

Add a new Pokemon (Xerneas (Active)). Adds a new evolution method which takes 100x the base attack. 

## Motivation and Context

Xerneas (Active) wasnt in the game yet. Could be used as an evolution method for the "walk 1000 steps" paldean pokemon

## How Has This Been Tested?

Went ingame, bred Xerneas and depending on attack it evolved or not. Checked AZ out.

## Types of changes

- new Pokemon


src/modules/requirements/AttackEvolveRequirement.ts
```diff
@@ -0,0 +1,25 @@
...
+
+        return partyPokemon?.attack >= pokemonMap[this.name].attack * ATTACK_EVO_REQUIRED_ATTACK_MULTIPLIER ? 1 : 0;
+    }
+
+    hint(): string {
+        const attackRequired = pokemonMap[this.name].attack * ATTACK_EVO_REQUIRED_ATTACK_MULTIPLIER;
+        if (this.getProgress()) {
+            return 'Level up to evolve.';
+        } else {
+            return `Needs at least ${attackRequired.toLocaleString('en-US')} attack to Mega Evolve.`;
```
CypherX, line 22: `Mega Evolve`
copy & paste strikes again

CypherX (changes requested):
Instead of having `AttackRequirement` use a constant (`ATTACK_EVO_REQUIRED_ATTACK_MULTIPLIER`), I think it should rather accept an additional parameter to specify the multiplier. That'd allow it to be used in other cases with different requirement amounts

Qwertypop04 (approved):

PR #5583, opened 2024-10-07
Update tutorial skip code

aheadfrogg:
## Description

I make it so the tutorial skip code gives you the dungeon ticket item.

## Motivation and Context

I did this because some people asked for it to be added and it makes it so people don't have buy the ticket every time they start a file.

## How Has This Been Tested?

I opened up a file and used the code to make sure the item was given to the player and it was.

## Types of changes

- Redeemable code change

### DIFF
--- a/src/modules/codes/RedeemableCodes.ts
+++ b/src/modules/codes/RedeemableCodes.ts
@@ -190,7 +190,6 @@ export default class RedeemableCodes implements Saveable {
                 return refund;
             }),
             new RedeemableCode('tutorial-skip', -253994129, false, async () => {
-                App.game.keyItems.gainKeyItem(KeyItemType.Dungeon_ticket, true);
                 const quest = App.game.quests.getQuestLine('Tutorial Quests');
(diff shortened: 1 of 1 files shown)


src/modules/codes/RedeemableCodes.ts
```diff
@@ -190,6 +190,7 @@ export default class RedeemableCodes implements Saveable {
                 return refund;
             }),
             new RedeemableCode('tutorial-skip', -253994129, false, async () => {
+                App.game.keyItems.gainKeyItem(KeyItemType.Dungeon_ticket, true);
```
CypherX, line 193: This should be moved to after the `gainDungeonTokens` line. Can you also update the notification text to indicate it was given to the player along with skipping the tutorial and finding some dungeon tokens?

CypherX (approved):
One side effect of this is the player will have an extra 100 QP from not having to purchase the dungeon ticket doesn't seem like it would affect anything though

PR #5589, opened 2024-10-12
Added magmar to the kanto fire egg pool

aheadfrogg:
## Description

I replaced Ponyta in the fire egg with Magmar.

## Motivation and Context

I did this so  it would be similar to electabuzz in the electric eggs and so the fire eggs has a stronger pokemon in it.

## How Has This Been Tested?

I loaded up a file and hatched fire eggs to make sure ponyta wasn't in it and magmar was.

## Types of changes

- Egg pool change


Qwertypop04 (approved):

PR #5593, opened 2024-10-13
Removed Base Meloetta Requirement from Anville Town

EatPant2nd:
## Description

Removed the Meloetta (Aria) requirement from Anville Town. Now it only has the Unova Champion requirement.

## Motivation and Context

Alternate forms are fun. For those that are auto-rewarded to the player, I understand why the base form requirement exists ~~people doing weird stuff, yes, I'm self-aware.~~ But for shop Pokémon, I think it's fine to allow the player to spend their Quest Points (or whatever currency) to skip getting the base if they so desired. 

## How Has This Been Tested?

Checked that the town was accessible without having base Meloetta.

## Types of changes

- "Balance"


Qwertypop04 (approved):

PR #5594, opened 2024-10-14
Fix farm shortcut visibility

CypherX:
## Description
Fixes the farm shortcut button never being visible even when the farm module is disabled. The settings were simplified in #5423 and the shortcut visibility needed to be updated.

## How Has This Been Tested?
Checked that the shortcut was visible when the module was disabled, and not visible when the module was enabled

## Types of changes
- Bug fix


RegisCoaxans (approved):

Qwertypop04 (approved):

PR #5618, opened 2024-11-01
Change dummy evolutions to fix some botcoins question to be more logical

RedzepR:
## Description

Changed some dummy evolutions around.

## Motivation and Context

Botcoins is better if the answers are more intuitive (e.g. Gigantamax Gengar evolves from Gengar instead of Haunter)

## How Has This Been Tested?

Went ingame, saw no changes to eggsteps.

## Types of changes

- data changes

### DIFF
--- a/src/modules/pokemons/PokemonList.ts
+++ b/src/modules/pokemons/PokemonList.ts
@@ -2992,7 +2992,10 @@ export const pokemonList = createPokemonArray(
         'levelType': LevelType.slow,
         'exp': 70,
         'eggCycles': 20,
-        'evolutions': [StoneEvolution('Hisuian Growlithe', 'Hisuian Arcanine', StoneType.Fire_stone)],
+        'evolutions': [
+            StoneEvolution('Hisuian Growlithe', 'Hisuian Arcanine', StoneType.Fire_stone),
+            DummyEvolution('Hisuian Growlithe', 'Noble Arcanine'),
+        ],
         'base': {
             'hitpoints': 60,
             'attack': 75,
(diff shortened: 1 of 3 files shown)


CypherX:
I feel like there's a reason it was done that way.

And you also have changes other than the DummyEvolution stuff in there

RedzepR:
The reason was to let a mon and their gmax form have the same eggsteps since a dummy evolution didnt have the IgnoreECChange flag before

CypherX:
What about the eggCycle changes that aren't mentioned?

RedzepR:
e.g. Ash Greninja was a standalone pokemon and to have the same eggsteps as regular Greninja it had an increased egg cycle, now that ash Greninja is a third stage pokemon it needs the smaller egg cycle (the same as Greninja) to not have way more steps

src/modules/pokemons/PokemonList.ts
```diff
@@ -14384,7 +14412,7 @@ export const pokemonList = createPokemonArray(
         'catchRate': 50,
         'evolutions': [
             StoneEvolution('Munchlax', 'Snorlax', StoneType.Soothe_bell),
-            DummyEvolution('Munchlax', 'Gigantamax Snorlax'),
+            DummyEvolution('Munchlax', 'Snorlax (Snowman)'),
```
osherwott: I'm not sure if this should be the case. If this is for bot coins then it should follow a train of logic, and this seems a bit questionable.

osherwott (commented):
Looks good, left a comment. Can you also change the title of the PR to make it more descriptive?

PR #5619, opened 2024-11-03
Changed Type1 Ice Colour (Eggs)

EatPant2nd:
## Description

Changed main colour Ice in the eggs to distinguish it from main colour Water.

## Motivation and Context

They were very similar (check screenshots).

## How Has This Been Tested?

With my eyes.

## Screenshots (optional):

So, currently, main colour Ice and main colour Water look almost identical. Go ahead. Guess which is which. One egg is Ice / Ice and the other is Water / Ice.
old-ice-water

I changed the colour by using the Dex's Ice colour and giving it a bit more saturation. First one is Ice / Ice, second one is Ice / Water.
new-ice-water

The saturation was made to distinguish main colour Ice from main colour Flying as shown here. Left side is Flying / Flying, right side is Ice / Flying.
flying-new-ice

## Types of changes

- Colour edit

### DIFF
--- a/src/styles/breeding.less
+++ b/src/styles/breeding.less
@@ -216,7 +216,7 @@
   "Ghost" #dbdbdb,
   "Grass" #a1f173,
   "Ground" #ad5840,
-  "Ice" #00aaff,
+  "Ice" #69d9d5,
   "Normal" #e8e0a0,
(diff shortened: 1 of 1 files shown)


Qwertypop04 (approved):

PR #5587, opened 2024-10-11
Give Chilan an aura

aheadfrogg:
## Description

I made it so Chilan has an aura that debuffs egg steps.

## Motivation and Context

I wanted to give the berry an aura and I thought this aura would be interesting thing to mirror chople.

## How Has This Been Tested?

I loaded up a file and made sure chilan had the correct aura.

## Types of changes

- berry aura added

PR #5620, opened 2024-11-03
fix to pr #5587

aheadfrogg:
## Description

This is a fix to pr #5587. The aura multiplier on chilan has been changed from tanga's numbers to micle's numbers. The multipliers are now x0.99, x0.98, and x0.97 for the respective stages that berries give auras.

## Motivation and Context

I did this because the old numbers were wildly unbalanced and resulted in the eggstep gain being awful.

## How Has This Been Tested?

I loaded up a file and fully grew a farm of the berry and made sure the aura number was correct. I also loaded up another file with nothing on the farm to see how much the aura affected breeding so I had both files use the same hatchery helpers with the same filters autobattling cynthia to see how much damage each would gain over an hour.

## Types of changes

- Aura fix


CypherX:
It looks like you used the same branch as the previously merged PR. Please create a new branch with the fix and open the PR from there

PR #5622, opened 2024-11-03
Changed the obtainment method for Aggronite

aheadfrogg:
## Description

I changed the method for obtaining Aggronite from being given from the temp battle in kalos to being in Rusturf Tunnel as a legendary drop.

## Motivation and Context

I did this to seperate the obtainment methods for Tyranitarite and Aggronite so if a player wanted to obtain only one they could. Also canonically trainers cant have more than 1 mega pokemon at a time so that temp battle having both doesn't make sense.

## How Has This Been Tested?

I loaded up a file and made sure that the Aggronite was in the dungeon and that the temp battle would only appear if you had Tyranitar.

## Types of changes

- Mega stone move


RegisCoaxans:
Why not just remove the Obtained Pokemon Requirements ? They only are remnants of when megastones were added to their Pokémon in the party.

Qwertypop04:
> Why not just remove the Obtained Pokemon Requirements ? They only are remnants of when megastones were added to their Pokémon in the party.

I had no idea that Mega Stones are no longer connected to their respective Pokemon, but if that is the case, they should probably all be removed in a separate PR.

Qwertypop04 (approved):

PR #5623, opened 2024-11-04
Spelling fix

aheadfrogg:
## Description

I changed all instances of Pokemon HQ Lab to Pokémon HQ Lab.

## Motivation and Context

I did this to match all other instances of the word Pokémon.

## How Has This Been Tested?

I opened a file that was on the location and made sure it was still on the location and that it was the new correct name.

## Types of changes

- Spelling fix


CypherX:
You missed a couple, which breaks the shard trader in that location

image

CypherX (changes requested):
The town image also needs to be renamed
`src/assets/images/towns`

CypherX (approved):

PR #5624, opened 2024-11-04
Clarify Amulet Coin quest text

umbralOptimatum:
Changed the Amulet Coin `UseOakItemQuest` text to specify that it only activates when earning bonus money. Slightly less confusing!


Qwertypop04 (approved):

PR #5627, opened 2024-11-04
Fix Attack Evolution

RedzepR:
## Description

Fixed Attack Evolution because Evolutions are very confusing.

## Motivation and Context

Follow up to #5581 

## How Has This Been Tested?

Evolved the Xerneas.

## Screenshots (optional):

image

## Types of changes

- Bug fix


src/modules/requirements/AttackEvolveRequirement.ts
```diff
@@ -11,7 +11,7 @@ export default class AttackEvolveRequirement extends Requirement {
     getProgress(): number {
         const partyPokemon = App.game.party.getPokemonByName(this.pokemon);
 
-        return partyPokemon?.attack >= pokemonMap[this.pokemon].attack * this.requiredValue ? 1 : 0;
+        return Math.min(Math.trunc(partyPokemon?.attack / pokemonMap[this.pokemon].attack), this.requiredValue);
     }
 
     hint(): string {
```
wormania, line 17: Hint now returns `Level up to evolve` as soon as current attack >= base attack

RedzepR: Good catch 👍

CypherX:
After some discussion we'd like `AttackEvolveRequirement` changed to a more generic `PokemonAttackRequirement` that also supports attack lower than the given value

RedzepR:
If I understand you correctly that would just require a renaming? Since it also accepts a multiplicator as its first parameter

RegisCoaxans:
It would also need the requirement to work with an AchievementOption passed to the constructor, and therefore a more versatile hint as well.
Also, could you rename the file accordingly, please ?

CypherX (approved):

PR #5632, opened 2024-11-07
Various fixes for 0.10.22

AevitasDragonkin:
* Rename Inv -> Inventory on the Generic Deals screen
* Fix Palaeontologist Token.png naming to Capitalized Case
* Underground Treasures alignment center to be consistent with the rest of the UI
* Pokedex detail view should use the same style as Gem Modal UI for typings, but text should be UPPERCASE
* Underground Chisel Achievement is not tracking and completing correctly
* Explain triggering new mines behaviour (player vs helper) in the Help tab
* Reduced Survey Achievement from 100 to 50
* Renamed UG helper achievements to mention 'all 5' instead of 'at least 5'

## How Has This Been Tested?

Tested each commit separately to see if it fixed what it was supposed to fix.

## Types of changes

- Bug fix


src/components/underground.html
```diff
@@ -339,6 +339,19 @@ <h5 class="card-header">
                                 The Underground lexicon can be found at the bottom of this page.
                             </p>
 
+                            <h4>Finding a new mine</h4>
+                            <p>
+                                A new mine can be manually found be clicking the Fine mine button in the Dig tab.<br/>
```
CypherX, line 344:
```suggestion
                                A new mine can be manually found be clicking the Find mine button in the Dig tab.<br/>
```

CypherX (approved):

PR #5634, opened 2024-11-09
Fix html related bugs for Key items

osherwott:
## Description
The htmls related to key items still used enums to go through the key items itemList

## Motivation and Context
Finished #5565

## How Has This Been Tested?
- Bought new key item, worked ✔️ 
- Hovered over new key item, worked ✔️ 
- Switched around key items _and_ hovered over them, worked ✔️ 

## Screenshots:
image

## Types of changes
- Bug fix


CypherX (changes requested):
Use a knockout `with` or `let` binding so you only have to call `find` once per section

CypherX (approved):

PR #5643, opened 2024-11-13
Condense pokemon statistics modal's vitamin table into one row

osherwott:
## Description
The vitamins table was taking too much space for something optional, so I made it into a row with a tooltip.

## Motivation and Context
~~Wanted room for new statistics~~ I wanted to make the table more informative and streamlined, and added a way to see if you have max vitamins on your pokemon, since the vitamin modal is pretty much always used to check the spread anyways.

## How Has This Been Tested?
- Applied and removed protein to see the numbers go up
- Went to new region to see the max amount go up

## Screenshots:
Screenshot 2024-11-09 185934

## Types of changes
- UI improvement


AevitasDragonkin (approved):
Code looks good

If the max amount of vitamins become used in multiple places it might be better to extract this into a separate static property somewhere

umbralOptimatum (approved):

PR #5646, opened 2024-11-13
Disable trade interface for special underground treasures

CypherX:
## Description
Disables the trade interface modal for Special underground treasures as it currently serves no purpose


RedSparr0w (approved):

PR #5661, opened 2024-11-14
Fix SaveSelector import button CSS

umbralOptimatum:
Adjusted CSS for the import override button in the save right click context menu so that it's no longer possible to miss the functional part of the button.

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -2830,6 +2830,17 @@ class Update implements Saveable {
                 playerData._townName = 'Pokémon HQ Lab';
             }
         },
+
+        '0.10.23': ({ playerData, saveData, settingsData }) => {
+            // Remove easier-to-fix locale misformatting from underground grid item tiles
+            saveData.underground?.mine.grid.map(t => t.reward).filter(r => r).forEach(r => {
+                if (!r.backgroundPosition.match(/^\d+% \d+%$/)) {
+                    r.backgroundPosition = r.backgroundPosition.replaceAll(',', '.');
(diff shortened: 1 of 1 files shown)


src/modules/SaveSelector.ts
```diff
@@ -29,7 +29,7 @@ export default class SaveSelector {
             $('#saveSelectorContextMenu').html(`
                 <a class="dropdown-item bg-success" href="#" onclick="Save.key = '${key}'; SaveSelector.Download('${key}')">Download (backup)</a>
                 <a class="dropdown-item bg-info" href="#" onclick="Save.key = '${key}'; document.querySelector('#saveSelector').remove(); App.start();">Load</a>
-                <a class="dropdown-item bg-warning" href="#"><label class="clickable my-0" for="import-save" onclick="Save.key = '${key}';">Import (overwrite)</label></a>
+                <a class="dropdown-item bg-warning p-0" href="#"><label class="clickable my-0" style="padding:.75rem 1.5rem;" for="import-save" onclick="Save.key = '${key}';">Import (overwrite)</label></a>
```
PixLSteam, line 32:
```suggestion
                <a class="dropdown-item bg-warning p-0" href="#"><label class="clickable my-0 w-100" style="padding:.75rem 1.5rem;" for="import-save" onclick="Save.key = '${key}';">Import (overwrite)</label></a>
```
The label doesn't fill the whole width of the dropdown item by itself, it requires some encouragement.

RedSparr0w (approved):

PR #5669, opened 2024-11-17
Lights! Camera! Pika!

Jaaslet:
## Description

Adds a questline based on the anime episode Lights! Camera! Pika!.
The questline is found in Pikachu Land, found a secret place north east of Anistar City.
It requires a lot of breeding, as the quests are based on attack, like mega evolutions.

There is no sprite for Frank yet, but DraKay is drawing one at some point. Not sure if you want that one added before the PR is merged, or it can be added later?

## Motivation and Context

I thought it would be fun to make a questline based on that episode. And it had two new alt-forms that looked cool (i know. Yet another pikachu... But that is what they give us!).

Secret place to match the other Anime alt-forms.

The attack requirements might be a bit high, but the players needs something to do while waiting for Kalos! (it can be changed, if it's either too hard or too easy. I don't mind).

I feel like the texts are a bit boring. It was hard for me to write like Frank... If someone more talented at writing than me wants to spice up the texts, feel free!

## How Has This Been Tested?

I played through the questline. Used Rare candy to train my pokemon

## Types of changes

- Questline


src/scripts/quests/questTypes/TrainSpecificPokemonQuest.ts
```diff
@@ -0,0 +1,29 @@
...
+        }
+        return `Train ${this.pokemon.name} to ${this.amount} attack.`;
+    }
+
+    claim(): boolean {
+        if (this.customReward !== undefined) {
+            this.customReward();
+        }
+        return super.claim();
+    }
```
RegisCoaxans, line 28:
```suggestion
```
`super.claim` also takes care of custom rewards just fine now.

RegisCoaxans (commented):
Quest type classes have been improved since then to better help make quests.
Quest.ts is now directly provided with functions to set all these custom things only a few quest types had.
Here are suggestions based on these changes :

Jaaslet:
I just copy/pasted from CaptureSpecificPokemonQuest.ts (which has not been refactored), so did not notice the refactors. Thanks! Will fix it

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -2472,6 +2472,43 @@ class QuestLineHelper {
         App.game.quests.questLines().push(unrivaledPowerQuestLine);
     }
 
+    public static createLightsCameraPikaQuestLine() {
+        const lightsCameraPikaQuestLine = new QuestLine('Lights! Camera! Pika!', 'Frank is making a new movie, and needs your help with finding his stars.');
```
osherwott, line 2476: This should probably have requirements for having Hawlucha and all the Pikachu needed for the quest. Also the lack of bulletin board makes this quest unpausable. I like the gift npc giving it out, but it's something we should note.

src/scripts/quests/QuestLineHelper.ts
```diff
@@ -2472,6 +2472,43 @@ class QuestLineHelper {
...
+
+        const trainHawluchaQuest = new TrainSpecificPokemonQuest('Hawlucha', 50000).withCustomReward(() => {
+            App.game.party.gainPokemonByName('Hawlucha (Stuntman)', PokemonFactory.generateShiny(GameConstants.SHINY_CHANCE_REWARD));
+        });
+        lightsCameraPikaQuestLine.addQuest(trainHawluchaQuest);
+
+        const talkToFrankQuest3 = new TalkToNPCQuest(DirectorFrank4, 'Return to Frank');
+        lightsCameraPikaQuestLine.addQuest(talkToFrankQuest3);
+
+        const trainPikachusQuest = new CustomQuest(15, undefined, 'Train 15 Pikachus to 25.000 power',
```
osherwott, line 2498: Should `power` be attack? Also I think these quests aren't working. They all increased every time my Pikachu leveled up.

src/scripts/quests/QuestLineHelper.ts
Jaaslet: It kinda fits the theme that you might need to catch the pokemon after you have started the quest, i think.
And did not want to put it on a bulletin board, as it's an anime questline, and should be a pixelhunt to be consistent.
But I can change it, if people think it should be changed

osherwott (commented):
I tried testing it again but the PR preview wouldn't load. Console says it might be a party pokemon issue?

src/scripts/quests/QuestLineHelper.ts
Qwertypop04: Can always just add a new bulletin board in the pixel hunt town. There's no limit on having more of those, we just haven't had a reason to add more than 1 in a region until now

src/scripts/Update.ts
```diff
@@ -2841,6 +2841,10 @@ class Update implements Saveable {
                 }
             });
         },
+
+        '0.10.24': ({ playerData, saveData, settingsData }) => {
+            saveData.statistics.temporaryBattleDefeated = Update.moveIndex(saveData.statistics.temporaryBattleDefeated, 224);
+        },
```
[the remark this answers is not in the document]
Jaaslet: We are no longer sorting the temp battles?

src/scripts/quests/QuestLineHelper.ts
Jaaslet: It's now a bulletin board!

src/scripts/Update.ts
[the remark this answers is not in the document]
CypherX: Not in the `GameConstants.TemporaryBattles` array, no, to remove the need to update indexes. The battles in `TemporaryBattleList` are still ordered as they've always been

RegisCoaxans:
I put the new Pokémon on routes, so they won't be found in Friend Safari anymore.
I'll need someone to double check my changes as I altered a good chunk of the original PR.

PR #5674, opened 2024-11-18
Fossil revives

AevitasDragonkin:
## Description

* Fossils can no longer be put in the hatchery
* Create Revive centers in their respective maps according to [https://bulbapedia.bulbagarden.net/wiki/Fossil#In_the_games](https://bulbapedia.bulbagarden.net/wiki/Fossil#In_the_games)
* Added a little bit of functionality to Generic Trades to support button overrides
* Renamed Galar Fossil Master to Cara Liss according to [https://bulbapedia.bulbagarden.net/wiki/Cara_Liss](https://bulbapedia.bulbagarden.net/wiki/Cara_Liss)
* Update function removes old fossils from the Hatchery and refunds the fossil

## Motivation and Context

With the new generic trade system and the underground changes, it made sense that Fossils could get more use other than a 1 time breed via the Hatchery.

## How Has This Been Tested?

* Used the new trades to acquire some new mons or get shinies for existing ones
* Checked the hatchery when updating from an older file, they were correctly refunded, and the other pokemon fix themselves when you gain some steps

## Screenshots (optional):

image
image
image

## Types of changes

- New feature


AMB1008:
I like this but now its weird that the galarian mons use QP while the others don't. Not sure what sort of fix could be made however

aheadfrogg:
Someone in the discord had the idea to add the evolutions of the fossils to these shops. That could be a way to also ev them before galar and would completely remove all fossils from the friend safari. This could be balanced by having the evolved forms cost more fossils to purchase.

AevitasDragonkin:
Added the feature with the following requirements:
* Cost is 10x the fossil needed for the 1st pokemon in the evo line
* You can only use this trade once you've obtained the pokemon through it's regular evolution method

image

CypherX (approved):

PR #5675, opened 2024-11-19
Add QuestLines translation support

umbralOptimatum:
Added translation support to QuestLine. 

## Description
Translations now support:
- `QuestLine.displayName` (a new accessor for `name`)
- `QuestLine.description`
- `Quest.description`, for Quests that are a step in a QuestLine

As proposed on Discord, each translation key includes a hash of the translated text, so that any changes to the English text will change the key and require updates to the translation repo.

## Motivation and Context
Expanded translation support. A couple people have asked about contributing to the translations recently, so...

## How Has This Been Tested?
Loaded the game and confirmed it complained about many, many missing translation keys.


src/modules/translation/Translation.ts
```diff
@@ -39,9 +43,8 @@ export default class Translate {
         });
     }), getTranslatedMemoResolver);
 
-    private languageUpdated: Observable<number>;
-
     constructor(languageSetting: Setting<Language>) {
+        const namespaces = ['pokemon', 'logbook', 'settings', 'questlines'];
```
Aegyo: Not important for this PR, but this could be an enum for better type safety wherever we reference a translation namespace

src/scripts/quests/Quest.ts
```diff
@@ -49,7 +50,24 @@ abstract class Quest {
     }
 
     get description(): string {
-        return this.customDescription ?? 'Generic Quest Description. This should be overriden.';
+        const description = this.customDescription ?? this.defaultDescription;
+        if (!this.inQuestLine) {
+            // Quest translations currently only supported for questlines
+            return description;
```
Aegyo, line 56: Was there some difficulty having these translated too, or are they left out for now to avoid bloating this PR?
Only asking because defaultDescriptions from quests feel like they should have a dedicated translation key with variable replacement.

I imagine defaultDescription would give you the key and variables and if we aren't in a questline we translate it directly, but questlines can do the same hashing but the tl file references the generic key directly (and hopefully passing the variables still works?)

Aegyo (approved):
Everything looks good. I forced a new save to begin all quests and stepped through each, all english text appeared as normal and no console complaints about missing translation keys with your translation-repo branch

src/modules/translation/Translation.ts
umbralOptimatum: Added a type for it. Long term maybe we'd want to load it from the translation repo alongside the list of languages.

src/scripts/quests/Quest.ts
umbralOptimatum: There's no technical obstacle to translating defaultDescriptions to my knowledge, I'm just not familiar with i18n variable replacement. It's a good subject for a future PR.

PR #5614, opened 2024-10-25
Allow changing the map coloring priority

Aegyo:
## Description

The map color settings can now be re-arranged to change the priority.
No reset button because it wasn't working :(

## Motivation and Context

Fixes #5167 

## How Has This Been Tested?

Manually, on a limited set of save files.

## Types of changes

- New feature
- UI improvement

PR #5677, opened 2024-11-19
Map : Add an option to prior Pokérus to Shiny and add the state "Missing Pokérus And Missing Achievement"

Sinonea:
## Description

Modification of the states (css) of a location. Added the state 'Missing Pokérus and Missing Achievement' to the states. The important point of this PR is the fact of being able to prioritize Pokérus over Shiny. By default, if you wanted to see the 'pokérus' state of a location, you had to first capture all the shinys of the location in question. Here, you can switch between the two: you can stay by default or see the pokérus first and only show the shiny once the pokérus is completed.

For it to work correctly and for there to be a real duality between Pokérus and Shiny, I had to add the state 'Missing Pokérus and Missing Achievement' to have a perfect balance.

## Motivation and Context

This Shiny/Pokérus duality was inspired by my frustration at having to capture all the shinys in a location before being able to see the Pokérus state. With this option, I no longer have to capture all the shinys (which could take a long time) before being able to tackle the pokérus.

I had confirmation from @EatPant2nd that some players are requesting it like me, so I programmed it to cover this problem.

Thanks to them for helping me! Little info: they are my admiration in the game, without them, I will not move forward, so thank you Pant!

## How Has This Been Tested?

I tested with different statuses of roads/dungeons/cities/safaris/etc... to check that everything is fine in both cases.

## Screenshots (optional):

image
image

## Types of changes

- UI improvement : possibility of targeting the pokérus before the shinies

### DIFF
--- a/src/components/settingsModal.html
+++ b/src/components/settingsModal.html
@@ -94,11 +94,29 @@
                                 <!-- /ko -->
                                 <tr data-bind="template: { name: 'ColorChoiceSettingTemplate', data: Settings.getSetting('--uncaughtShinyPokemonAndMissingAchievement')}"></tr>
                                 <tr data-bind="template: { name: 'ColorChoiceSettingTemplate', data: Settings.getSetting('--uncaughtShinyPokemon')}"></tr>
-                                <tr data-bind="template: { name: 'ColorChoiceSettingTemplate', data: Settings.getSetting('--missingAchievement')}"></tr>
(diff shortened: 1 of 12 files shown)


Sinonea:
My beloved Pant that I love so much showed me the PR of someone MUCH more talented than me. Here it is: #5614. I just look ridiculous in comparison. So I close the PR and hide six feet underground.

PR #5683, opened 2024-11-20
Changed Unova Gym Leader Level to represent BW2 challenge mode

RedzepR:
## Description

Changed the Unova Gym Leader Level to represent BW2 challenge mode.

## Motivation and Context

They were all set to 58. Other regions use the mainline levels.

## How Has This Been Tested?

Checked a gym to see that i gain the correct exp.

## Types of changes

- data fix


Qwertypop04 (approved):

PR #5696, opened 2024-11-25
Allows the click on the chisel and hammer in the reduced underground module.

hippolythe:
## Description
Allows you to click on the chisel  and hammer in the reduced underground module.

Chisel and hammer dig coordinates are randomly generated by clicking on the reduced module. If the excavation zone has already been explored, the function is restarted until excavation is possible.

## Motivation and Context
I found it strange to be able to use the bomb and the clue but not the other 2 tools. Plus, it's more practical to simply click from the reduced module than to open the large one.

## How Has This Been Tested?
I've used the chisel and hammer from the reduced module to completely empty several mines without any problems.

## Types of changes
- New feature


AevitasDragonkin:
This is not something we'd want to be available from the main screen module. 
If Bomb and Survey were dependant on the clicked tile, they would not be available as well.

Combined with the changes from #5805 items will only be gained with targeted clicks in the modal.

PR #5704, opened 2024-11-30
Changed berry master lum hint

aheadfrogg:
## Description

I changed the Lum berry hint to be more clear on which berries it needs for the mutation.

## Motivation and Context

I did this so people could actually figure out the mutation for lum using the hint.

## How Has This Been Tested?

I tested this by loading up a file and looking at what the hint for lum had said.

## Types of changes

- Text Change


Qwertypop04 (approved):

PR #5706, opened 2024-12-01
Made Shinies always drop Rare Hold item

EatPant2nd:
## Description

Made shinies always drop a Rare Hold item (if they have one).

## Motivation and Context

Shinies are uncommon enough that it shouldn't break anything. At most, people will be able to evolve their Galarian Farfetch'd earlier or they will take slightly less to resist a given Pokémon. Also, with how rare Roamer + Shiny it is, I wouldn't worry about this change altering the pain of getting those manes / Zenith.

## How Has This Been Tested?

Got some shinies. Got some Rare Hold items. Got some regular Pokémon. Defeated a bunch until the item dropped to see that they still could.

## Types of changes

- Balance (?)


src/scripts/pokemons/PokemonFactory.ts
```diff
@@ -35,9 +35,9 @@ class PokemonFactory {
         const catchRate: number = this.catchRateHelper(basePokemon.catchRate);
         const exp: number = basePokemon.exp;
         const level: number = this.routeLevel(route, region);
-        const heldItem: BagItem = this.generateHeldItem(basePokemon.heldItem, GameConstants.ROUTE_HELD_ITEM_MODIFIER);
         const money: number = this.routeMoney(route,region);
         const shiny: boolean = this.generateShiny(GameConstants.SHINY_CHANCE_BATTLE);
+        const heldItem: BagItem = this.generateHeldItem(basePokemon.heldItem, GameConstants.ROUTE_HELD_ITEM_MODIFIER, shiny);
```
CypherX, line 40: Instead of modifying `generateHeldItem` you can just do:
```suggestion
        const heldItem: BagItem = shiny ? basePokemon.heldItem : this.generateHeldItem(basePokemon.heldItem, GameConstants.ROUTE_HELD_ITEM_MODIFIER);
```

RegisCoaxans: Doing so would bypass the item requirement.

CypherX: Right, requirements, ignore me

CypherX:
In `generateHeldItem` after the requirement check you can just do:
```
if (shiny) {
    return item;
}
```
Instead of going through the rest of the function

EatPant2nd:
Ok. Moved the check to post-requirements. Thanks.

CypherX (approved):

PR #5707, opened 2024-12-02
Changed Nomel's Mutation

EatPant2nd:
## Description

Changed Nomel's mutation from requiring just Pinap to needing Pinap and Wepear.

## Motivation and Context

Currently, Nomel only needing Pinap and nothing else, made for a very boring mutation and caused issues when this berry was involved in any mutation. On the other hand, Wepear is used in 0 mutations. Conveniently, their flavour profiles work well since Wepear is 10 Bitter / **10 Sour**, Pinap is **10 Spicy / 10 Sour**, and Nomel is **10 Spicy / 20 Sour**.

## How Has This Been Tested?

Planted some Pinap. Planted some Wepear. Waited for mutations to happen.

## Types of changes

- Farm


AMB1008:
This would be super useful and buffs many other berries so I hope this gets added

Qwertypop04 (approved):

PR #5710, opened 2024-12-04
Added Eevee as a Rowap berry wanderer

hikimoi:
## Description

Added Eevee to the wandering pokemon pool of the Rowap berry as a Kanto wanderer.

## Motivation and Context

Rowap is the EV berry. Eevee and EV sound very similar, so much that people on the Discord have started to make the connection (and this joke) already.
A berry with an EV increasing aura is therefore also a berry with an Eevee-increasing aura, lol. It just works.

Another motivation is challenge runs.
Vaporeon + Let's Go Eevee can beat Blue2 and start the general 9C challenge procedure. But as Eevee and its evolutions are unavailable to be caught until very late into the game, an Eeveelutions-only challenge can't easily get enough DT to be feasable. With this change, an Eeveelutions-only challenge can be played.

Does this affect balancing?
No, I don't think it will.
Files without Real Evo only need 1 Eevee anyway, and they can get that one in Celadon City, or for free from Bill's Grandfather. Files with Real Evo would need 3 Eevees. Normally, they need to buy 2, now they also have the option to wander it onto the farm. Helps the dedicated player a bit. However, most players will probably not have the Rowap berry unlocked this early into their run anyway, so it doesn't change much about Eevee's availability in Kanto. If anything, it makes obtaining more Eevee for the evolutions in the regions after Kanto more versatile, as the player does not have to necessarily backtrack to Kanto, but can obtain Eevee as a farm wanderer as well. 
It has the potential that Eevee becomes very easy to resist. But as Eevee is also a route encounter in Kalos, it is already relatively easy to resist. Sitting on the route with a Rowap setup will probably yield more EV than catching wandering Eevees.

## How Has This Been Tested?

Loaded a save file in Kanto with Rowap berry unlocked.
Checked if the berry list shows Eevee as a potential wanderer.
Checked if Eevee actually appears as a wanderer.
It worked.

## Screenshots (optional):

p2
e1

## Types of changes

- New feature
- Fun


RegisCoaxans:
I'm closing this PR. I have two concerns :
- While it is argued that it's a harmless addition, I believe we should take the opposite approach and focus on changes that are actually meaningful to supported gameplay and coincidentally benefit unusual runs.
- It's a patchwork solution for one very (very, very) specific use case. Instead, I'd prefer more generic content so we don't end up with a handful of potentially incoherent PRs based on individual needs.

PR #5712, opened 2024-12-05
Egg Pool display

CypherX:
## Description
When clicking an egg in the hatchery a new modal will display that lists the possible pokemon that can be hatched from that egg, the egg pattern, and the odds of it being that pokemon. It will show up to the players highest region and any uncaught pokemon will display "???" instead of their name, but the egg pattern will be visible.

Also changed the egg display in the hatchery modal so that all eggs are always shown so players can click them, and added some information about eggs.

## Screenshots (optional):
image

## Types of changes
- New feature
- UI improvement


CypherX:
After discussion in discord, updated to use extra space in breeding modal rather than a new modal to display the egg pool.

image
image
image

PR #5722, opened 2024-12-12
Show Temp battle Pokemon Locations for locked pokemon

osherwott:
## Description
PokemonLocations now uses the full pokemon list from tempbattles

## Motivation and Context
Was browsing the wiki and realized Solgaleo and Lunala never showed up as obtainable through a temp battle even though they are

## How Has This Been Tested?
Used the PokemonLocation function on Solgaleo and Lunala. Unsure if this breaks anything else

## Screenshots:
image

## Types of changes
- Bug fix


RegisCoaxans (approved):
No other Pokémon affected it seems.

osherwott:
Good to know. Thanks for the review Regis and umbral!

PR #5723, opened 2024-12-15
Changed Solgaleo and Lunala's evolution to a time based one.

aheadfrogg:
## Description

I changed Solgaleo/Lunala's evolution from Cosmoem to be a day/night based evolution.

## Motivation and Context

I did this for two reasons. 1. So people who do pokemon restricted runs can pick which evolution they want, 2. So the evolution is more thematically similar to the SM/USUM games.

## How Has This Been Tested?

I loaded up a file and made sure the evolution worked properly.

## Types of changes

- Evolution change

### DIFF
--- a/src/modules/pokemons/PokemonList.ts
+++ b/src/modules/pokemons/PokemonList.ts
@@ -24068,8 +24068,8 @@ export const pokemonList = createPokemonArray(
         'exp': 140,
         'catchRate': 45,
         'evolutions': [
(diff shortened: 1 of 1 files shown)


RegisCoaxans (approved):
I believe this could need a NPC telling the players how to get the one they miss.

aheadfrogg:
> I believe this could need a NPC telling the players how to get the one they miss.

Added

PR #5724, opened 2024-12-15
Snow and event bgs

osherwott:
## Description
Added a Snow battle background
Events can now push css classes to the html to change backgrounds

## Motivation and Context
Snow: Wanted more variety of backgrounds. The current ice one is too icy for routes like Sinnoh 217
Event bgs: Wanted events to have cosmetic differences

## How Has This Been Tested?
Snow: Went on routes specified in GameConstants
Event bgs: Activated event and went on routes and in dungeons

## Screenshots (optional):
image

## Types of changes
- New battle background
- Event visuals


src/modules/GameConstants.ts
```diff
@@ -983,6 +983,17 @@ export const BattleBackgrounds: Record<string, BattleBackgroundData> = {
         [Region.galar]: new Set([49]),
     },
 
+    Snow: {
+        [Region.kanto]: new Set(),
+        [Region.johto]: new Set(),
+        [Region.hoenn]: new Set(),
+        [Region.sinnoh]: new Set([216, 216]),
```
DreamNya, line 990:
```suggestion
        [Region.sinnoh]: new Set([216, 217]),
```

src/modules/specialEvents/SpecialEvent.ts
RegisCoaxans, line 1: Manually starting an event with the use of the event calendar does not trigger the starting and ending functions. Therefore, the background is not applied in this case.
Anyway, I would recommend such thing is directly handled in `getBattleBackground` function from [MapHelper](https://github.com/pokeclicker/pokeclicker/blob/d66429aae824cf1e43df5ef7cc08d3ddd09032c1/src/scripts/worldmap/MapHelper.ts#L129) using `event.isActive()`. Without having to change anything to the SpecialEvent and SpecialEvents classes, except maybe to mention the MapHelper file in the header comment of the respective event.

src/modules/settings/index.ts
```diff
@@ -496,6 +496,7 @@ Settings.add(new Setting('discord-rp.large-image', 'Discord main image',
         new SettingOption('Ice Environment', 'background-ice'),
         new SettingOption('Mansion Environment', 'background-mansion'),
         new SettingOption('Power Plant Environment', 'background-power-plant'),
+        new SettingOption('Snow Environment', 'background-snow'),
```
RegisCoaxans, line 499: Can we have a generic code for the background option list instead of manually maintaining it ?
```js
...Object.entries(GameConstants.BattleBackgroundImage).map(([name, code]) => new SettingOption(`${name} Environment`, `background-${code}`)),```

RegisCoaxans (changes requested):

src/modules/settings/index.ts
osherwott: I couldn't find where this is expressed in the code but the console makes it seem like it works so 🤷 Added, thanks!

PR #5732, opened 2024-12-26
Add an enhanced / full version of offline calculation

### LINKED ISSUE #2157: [SUGGESTION] Update offline progress
Currently being offline involves taking the base value of the current (or closest) route, multiplied by time, capped at 24 hours. The current amount of coin generated is insignificant after very early game - feels outdated and pointless, such gains are easily outstripped by a few seconds of active gameplay.

I would suggest removing this feature entirely, or buffing it to be meaningful. Can give a more balanced amount of coins, or gems, pokemon exp or low amount of attack stat gain.

omacranger:
## Description
This PR updates the offline calculation to process all 'offline' ticks 'online' since the player was last seen, meaning a player gets all / full rewards as if they had the browser open and focused. This follows a similar pattern to how Melvor Idle handles their offline calculation by processing all ticks when the user returns so 'full' credit is given.

Current notes / 'limitations':

- Offline calculation is probably slower than in an ideal world, but is by no means unusable. With the optimizations in this PR it's about 10s / hour of offline calculation on my Desktop.
- Achievement based progression isn't factored in until after offline calculation due to slow performance (comparatively) of `AchievementHandler.checkAchievements`, so gained exp / currency, etc, may be a little lower than expected.
- Similarly, saving is disabled until all offline calculation is processed for the same reason.
- Notifications are untouched, so there can be a [huge wall of notifications](https://github.com/user-attachments/assets/6ce67385-0e8c-4577-b84b-9575cc4d79f1) during each chunk of offline processing. This doesn't bother me much, but would also likely benefit from having a modal afterwards that condenses all of these into a 'review' of offline progress.

## Motivation and Context
This prevents users having to prevent their computer from sleeping in order to still gain 'progress' overnight / while a device is put away. I think this encourages planning of offline periods and just generally conserves energy 🙂 

Related to  / Fixes #2157 , maybe others.

## How Has This Been Tested?
Loaded various saves from the 'past' and observed progress made. No opposed to adding automated tests here but want to get some feedback on whether or not this will even be accepted before diving too deep.

## Screenshots (optional):

Updates the preloading splash screen to display loading progress:

image

Offline calculation notification present after loading: 

image

## Types of changes

- New feature
- UI improvement

### DIFF
--- a/src/components/splash.html
+++ b/src/components/splash.html
@@ -15,6 +15,7 @@
                     <span id="preload-progress-text">Loading...</span>
                 </div>
             </div>
+            <div id="preload-progress-label" class="progress-text"></div>
         </div>
     </div>
 
--- a/src/modules/TemporaryScriptTypes.ts
+++ b/src/modules/TemporaryScriptTypes.ts
@@ -31,7 +31,7 @@ import type CaughtStatus from './enums/CaughtStatus';
 import type SpecialEvents from './specialEvents/SpecialEvents';
 
 /*
-    These types are only temporary while we are converting things to modules. As things are converted, 
+    These types are only temporary while we are converting things to modules. As things are converted,
     we should import their types here for use, instead of these cheap imitations.
 
     When a file is converted to a module, the types for any /scripts dependencies should be added here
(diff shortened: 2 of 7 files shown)


Askellown:
"10s/hour"
sure let me load a save from 2020 - 2022, as I often do for testing purposes. surely that won't take. uh. 3 days to load. you need less accurate tick values for values over a certain amount of time, or a cap.

omacranger:
It already has a maximum of 24 hours, like the current offline implementation.

RegisCoaxans:
Closing as this isn't something we want.
Thanks for the effort, thouh.

PR #5734, opened 2024-12-28
Adjust Alcremie Spin Numbers to be only dependent on caught number

### LINKED ISSUE #5584: [SUGGESTION]: Remove Alcremie (Battle Cafe) Spin Limit
### Current Behavior

Currently, we have a limited number of "spins" we're allowed to spin for Alcremie. This is based on the number of Alcremies we've unlocked and shinies we have.

### Improved Behavior

Once we've unlocked all the Alcremies, players should be able to use the Battle Cafe infinitely, and spin to their heart's content.

### Reason for change

At this point in time, resisting Alcremies take a lot of time for both the number of berries as well as the amount of time, due to the limited number of spins available. Allowing us to spin an uncapped amount of time makes it so we're only berry-gated for resisting Alcremies, as opposed to both time and berry-gated.

### Detailed Description

Once all Alcremies have been obtained, remove the cap of the number of spins and let users spin as many times as they want in the day.

RedzepR:
## Description

Number of spins per day is now dependent on the amount of alcremie you own. 3 base + 1 for each Alcremie. Owning all Alcremie gives infinite spins. 

## Motivation and Context

Closes #5584

## How Has This Been Tested?

Went in with files that had all, none and some amount between alcremie. Simulated day change and got the expected Spin number. Can spin indefinitely if owning all Alcremie, can't spin with 0 spins remaining. 

## Types of changes

- UX improvement

### DIFF
--- a/src/components/battleCafeModal.html
+++ b/src/components/battleCafeModal.html
@@ -97,7 +97,7 @@
                 <div class="card">
                     <div class="card-header py-2 font-weight-bold">Spin!</div>
                     <div class="card-body col-lg-6 align-self-center p-2">
-                        <div>Remaining Spin Attempts: <span data-bind="text: BattleCafeController.spinsLeft()"></span></div>
+                        <div data-bind="if: BattleCafeController.spinsLeft() > -1">Remaining Spin Attempts: <span data-bind="text: BattleCafeController.spinsLeft()"></span></div>
                         <img style="width: 60px;" src="assets/images/profile/trainer-1.png" data-bind="
                             attr: { src: `assets/images/profile/trainer-${App.game.profile.trainer() || 0}.png` },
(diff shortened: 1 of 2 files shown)


CypherX:
After some discussion here are our thoughts:

Having unlimited spins after simply obtaining all Alcremie, which can be accomplished after only 5 days of optimal spinning, is too easy. It would make the spinning mechanic mostly irrelevant. However, we're fine with spins becoming unlimited after all Alcremie have been resisted so players can *go crazy* if they want.

We also think the distribution of unlocking additional spins is a bit skewed, and that the cap could be increased, so we think each unique Alcremie resisted should grant +1 additional spin.

RedzepR:
That's understandable.

I _am_ interested in how to get all alcremie within 5 days of spinning, can i find the math somewhere in the discord?

RedzepR:
Reworked the PR as of the most recent discord discussion

RegisCoaxans (approved):

PR #5738, opened 2025-01-01
Reveals the dex upon defeating the regions champion

RedzepR:
## Description

Upon defeating the champion of a region ones goal is to complete the regional dex (r2c) and the pokedex is suppose to help, yet often falls flat with providing useful information. Now upon defeating the champion of a region all pokemon with that natives region get put with silhouette into the pokedex.

## Motivation and Context

A lot of people are confused if theyre missing the last pokemon of the regional dex and the professor tells them theyre missing one more mon. Most prominent with Kalos as Volcanion is hard to find. 

## How Has This Been Tested?

- Loaded up a file that has already beaten the champ
- Loaded up a file that still had to beat the champ
- Beat the champ

## Screenshots (optional):

image

## Types of changes

- QoL


osherwott (commented):
Works fine, but if you have the dex challenge turned off and skip the next region's starter it goes back to hiding them up to the highest seen/caught again. Shouldn't be that much of an issue since that's not normal gameplay but worth noting.

Qwertypop04 (approved):

Issue #5739, opened 2025-01-03
[SUGGESTION]: Add N as a selectable character for the profile


DarthGamer22:
### Current Behavior

N is currently unavailable, either to purchase from the Discord or from the selection screen.

### Improved Behavior

I request that you please add N as a profile character. I have attached a couple options for sprites taken from the game below.

### Reason for change

I like N and would like him to be my profile character, and I'm sure other players do as well.

### Detailed Description

![4299127cc4c7a00ff378b3301ba15729213d8db5](https://github.com/user-attachments/assets/ecb12888-f093-47a9-ae23-1e396f1d3fbb)

DraKay:
Although your images are slightly different style and mostly size than the ones we use, they can be edited in 5 mins. I can make it, but would like first cc to confirm me that N is not a rival fight and thus can be added safely

WarioInParis:
> Although your images are slightly different style and mostly size than the ones we use, they can be edited in 5 mins. I can make it, but would like first cc to confirm me that N is not a rival fight and thus can be added safely

Well, according to the wiki, it currently is not a temporary battle in the game. However, it is a character you fight multiple times in the original games.

DarthGamer22:
That's because the game follows the Black and White 2 progression instead of the Black and White progression. Black and White are the games in which you fight N as a "rival" and the story goes that he and the original protagonist disappear for the events of Black and White 2. Thus, it shouldn't be an issue.

WarioInParis:
In the sequels, there is a fight with N, although it is more of a secret bossfight situation akin to Red in the Johto games or even Steven in Emerald, so it shouldn't be an issue.

Qwertypop04:
N will likely be added as a fightable character at some point. Not sure why this would be an issue as there are plenty of characters you can fight that are already available as player characters

DraKay:
Then doing it! Tomorrow or so will pr

Qwertypop04:
Closing as completed

PR #5742, opened 2025-01-05
Add N king of unova, the pokemon king, leader of plasma team, as a PC

DraKay:
## Descriptioñ
Add the 162 pc image, in this case N from Gen 5 

## Motivatioñ añd Coñtext
Asked in #5739 

## How Has This Beeñ Tested?
Ñ

## Types of chañges

- UI improvemeñt


CypherX (dismissed):
Doesn't work. Test your changes before opening a PR

DraKay:
Solved. Just forgot to change a number in the profile ts file. Now should work. _next time don't pr half sleep and with a slight headache_

CypherX:
Still doesn't work, which again tells me you haven't bothered testing

DraKay:
Both pant and martius helped to point and check the fail. Was a mistake with the max trainer number that came from months ago. THANKS for the help to both :)

Qwertypop04 (approved):

PR #5754, opened 2025-01-31
Pokemon List NPCs for uncaught babies and evolutions

osherwott:
## Description
Added npcs that can list pokemon. This helps players see which pokemon they should focus on for whatever the npc is helping out with. The two npcs added are for baby pokemon and evolutions, since players usually get confused with catching those specific kinds of pokemon.

Details:
- Pokemon list is limited by player's highest region
- All baby parents are shown, given that they are caught (ex. list shows Pikachu and Raichu as potential parents)
- Pokemon that "evolve" into another form aren't included in the list (ex. Cherrim (Overcast) can evolve into Cherrim (Sunshine), but is not included in the evolution assistant's list)

Canon influences:
- Professor Elm specializes in eggs and his assistant in gen 2 gives you one and uses a Scientist sprite
- Professor Rowan specializes in evolution and Lucas follows the alternating gender of our rivals in each region

## Motivation and Context
- We don't have anything that hints at which pokemon can have babies
- Some cross-gen pokemon don't have npcs and that causes players to miss some pokemon. Alternatively, they may have trouble finding the specific npc even if they have already read it
- I wanted to make a generic npc so that we don't have to remake it again if we don't need these anymore (RIP Safari Ranger)

## How Has This Been Tested?
- Talked to npcs after changing caught pokemon and highest player region

## Screenshots:
Screenshot 2025-01-31 100410 Screenshot 2025-01-31 100458

## Types of changes
- NPC


Qwertypop04 (approved):

PR #5755, opened 2025-01-31
Evolution Data tab for Pokemon Statistics modal

osherwott:
## Description
Created Evolution Data tab in pokemon statistics modal that includes:
- Evolution lines (with caught and pokerus status)
- Evolution methods for uncaught pokemon
- Alternate forms, if caught

### Details:
- Evolution Line
  - Filters out dummy evos unless the base stage only evolves via them like Milcery or Kubfu
  - Filtered by native region
    - Cross-gen pokemon won't appear until their region is reached
    - Evolution lines with common prevolution are separate until that prevo's region is reached
  - Uncaught baby pokemon are shown with an egg and tooltip
  - Shows caught, shiny, and pokerus status
  - Can click on seen pokemon in to view their statistics modals
  - Forms are included if they can be evolved into
    - Mega pokemon are filtered until the corresponding mega stone is obtained or the final region is reached, as endgame content
- Evolution tables
  - All requirements will be concealed with a `???` until the `basePokemon` is caught
    - Uses statistic for compatibility with mid-stages in real evos challenge
  - Filters out completed requirements. This was mostly done to hide the not-holding-an-everstone requirement
  - Level and Stone requirements are always shown unless hidden with `???` because they are the main evolution triggers
- Alternate Form section
  - Only shows caught forms that aren't part of the evo line. Uncaught silhouettes are shown after reaching the final region, as endgame content. Only available forms are included
  - Caught, shiny, and pokerus statuses are invisible until final region
  - Pure html. I tried to make a function for it but it would always lag for a bit with Milcery despite all Alcremie being part of the evo line

### Notes on some code:
- pokemonStatisticsModal.html Line 442: `<!-- ko foreach: $data.filter(e => e.trigger != EvoTrigger.NONE) -->`
  - Cherubi has a dummy evo into Cherrim (Sunshine) that needs to be filtered out otherwise it shows up alongside Cherrim (Overcast)'s evolution method into (Sunshine)
- Towards the end I also thought about regressing pokemon to their base stage during the real evos challenge but that was too complicated and is a different discussion, so I won't be attempting it. I planned for the Evo Data tab to have a more practical use during the challenge, keeping track of pokemon, so that's why I included `followDummy` in `PokemonEvoLines.getEvoLine()` and used the `pokemonCaptured` statistic instead of `alreadyCaughtPokemon` in some parts of the html

## Motivation and Context
- Players unfamiliar with the games may not know all the evolutions and evolution methods available. This provides a diegetic way to view this information.
- I like the evolution lines of the games' pokedexes and thought it would be fun to have something similar in clicker.

## How Has This Been Tested?
- Checked evolution tab when changing regions
- Checked evo line statuses when having caught, shiny, and 'rus
- Checked pokemon that have mega evolutions before and after getting their stones
- Checked pokemon with babies
- Checked pokemon with only dummy evolutions (Milcery, Kubfu)
- Checked pokemon with babies that only have dummy evolutions (Happiny, Mantyke, Manaphy)
- Checked that evolution hints show up properly depending on requirements met
- Checked pokemon with and without alternate forms, and after catching one
- Checked filtered out megas, forms, and form statuses show up in final region
- Checked PokemonEvoLines functions in console

## File breakdown (mostly lots of tiny changes)
| File | Description |
| --- | --- |
| `src/components/pokemonStatisticsModal.html` | Main addition - Added Evolution Data and Form Data under the same tab |
| `src/components/templates/caughtAllStatusTemplate.html` | Copied from CaughtStatus template and modified for plural pokemon |
| `src/index.html` | Imports `caughtAllStatusTemplate` |
| `src/modules/pokemons/PokemonHelper.ts` | Created a function for getting a mega pokemon's mega stone |
| `src/modules/requirements/InDungeonRequirement.ts` | Made `getProgress()` not return an error if you haven't entered a dungeon yet |
| `src/modules/requirements/MegaEvolveRequirement.ts` | Made `megaStone` `public` instead of `private` |
| `src/scripts/party/PartyController.ts` | Made `getCaughtStatusAll()` for arrays of pokemon |
| `src/scripts/pokemons/PokemonEvoLines.ts` | Main addition - Added functions for getting Base Stages, Evolution Lines, and Evolution Methods |
| `src/scripts/pokemons/PokemonLocations.ts` | Added functions for getting dummy evo `basePokemon` in a way that doesn't interfere with existing `PokemonLocations` |
| `src/scripts/wildBattle/RouteHelper.ts` | Made function for evo line pokerus and replaced a magic number |

## Screenshots:

Screenshot 2025-01-31 133848
Screenshot 2025-01-31 133933
Screenshot 2025-01-31 134513
Screenshot 2025-01-31 134551
Screenshot 2025-01-31 134812
Screenshot 2025-01-31 135122
Screenshot 2025-01-31 135455
Screenshot 2025-01-31 135424

## Types of changes
- New feature


RegisCoaxans:
There is a lot to cover here. Played with this tab and here are the things that bother me :

1. The alternate form container is not very consistent at all and it might reveal alt forms we would like the player to find on their own.
- Any alt of any Pokémon in the selected evoline shows up, but if you click on one of these alt, whose evoline is only partial, it doesn't mirror it (e.g., Giga Charizard only shows alt forms for Charizard, Pinkan Oddish only shows Oddish).
- Regional variations like Alolan Raichu or Alolan Exeggutor disappear from the alt form list if I then select an alt of these base Pokémon despite them being part of the main evoline.
- Selecting Gigantamax Pokémon show the evoline of the base form and the Giga is still in the list of alt forms, lol. See Toxtricity, Snorlax Melmetal or Alcremie. Maybe due to dummy evolutions ?
image
Overall this looks like a confusing section... Not very helpful when it should, probably too helpful when it should not.


2. The evolution hint card works really well. Although I think it should probably 1) display base form sprite alongside the sprite of the evo form, so it becomes clear which Pokémon in the evo line actually evolves, 2) not reveal literal hints but let players figure out based on what they need to look for (Like, Weather, if the weather is a factor, but not the actual weather).
Moreover, It's... uh, chaotic, when it comes to Alcremie (yes, that villain again !).
Lastly, I believe it would be better not to see the evo data for a Pokémon from one we do not already own. Like, we should not have these evo methods explained in the following screenshot as we do not own Pichu or Pikachu respectively, because they are not relevant yet.
image


3. Dummy evos are a mess to use. It only really helps for Alcremie (Kubfu has been changed to use regular evo methods) and it does not really work well, lol. It positively affects some baby evos as well, though, but we could just change their dummy evo to level evo instead (except for Phione, I guess).


5. The whole code looks a bit confusing and probably unoptimised. I believe it could be working clearer and smoother if it was pre processed and in the way eggcycles are calculated (from start to end of the Pokémon list, following the evo lists of each entry).
 
 
What I could recommend is an `EvoLine` class whose parameters are a list of the Pokémon in said evolution line and a list of evolution data you will need to display. All being processed beforehands. That would save many loops on the Pokémon for all the calculations. It also allows us to make up some evolines, like Milcery + Alcremie or Phione + Manaphy, or even hints if we need to.

PR #5771, opened 2025-02-25
Remove missingno on update

RedzepR:
## Description

Removes missingno within the update function. Loses attached vitamins and held item.

## Motivation and Context

Missingno isnt suppose to be owned so there is no reason for people to own it. It also doesnt do its job as bug finder well if people own it from X upgrades ago. 

## How Has This Been Tested?

Give me missingno, update the game, see no more missingno.

## Types of changes

- Bug fix

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -2841,6 +2841,10 @@ class Update implements Saveable {
                 }
             });
         },
+
+        '0.10.24': ({ playerData, saveData, settingsData }) => {
+
+        },
     };
 
     constructor() {
@@ -3011,6 +3015,9 @@ class Update implements Saveable {
                 }
             }, { playerData, saveData, settingsData });
 
+        //remove missingno
+        updateResult.saveData.party.caughtPokemon = updateResult.saveData.party.caughtPokemon.filter(p => p.id != 0);
+        
         try {
             this.automaticallyDownloadBackup(backupButton, settingsData);
             Notifier.notify({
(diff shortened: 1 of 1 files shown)


wingull7:
Missingno. is in the game for debug purposes, isn't it? I believe it's there so that if someone messes up some code while making a Pokemon spawn, it just spawns Missingno. instead.

wingull7:
> > Missingno. is in the game for debug purposes, isn't it? I believe it's there so that if someone messes up some code while making a Pokemon spawn, it just spawns Missingno. instead.
> > This is stated in [https://wiki.pokeclicker.com/#!Pokémon/MissingNo./](https://wiki.pokeclicker.com/#!Pok%C3%A9mon/MissingNo./)
> 
> You are indeed correct on the purpose of MissingNo. That is the entire point of this PR; it isn't removing MissingNo. from the game, it's removing it from the parties of those who have obtained it

Oh, I misread the original post then. My bad.

RegisCoaxans:
Considering that until now, players have been told MissingNo. would not be taken away from them, thus that they may have bred the Pokémon, I would recommend we at least give the vitamins invested back to the players.

Moreover, either in that PR or a following one, it should be fair to prevent people from breeding MissingNo, or assigning vitamins to it; similarly to how MissingNo. can not be looked at in the Pokémon modal, that would reinforce the idea that this is not a legit one.

CypherX:
This doesn't handle the player having MissingNo. as an egg in the hatchery or being in the queue

RedzepR:
> This doesn't handle the player having MissingNo. as an egg in the hatchery or being in the queue

fixed

CypherX:
Made some adjustments to the code that removes it from the party/hatchery/queue and also added a one-off update function to refund vitamins. Also got rid of the `nextEggFromQueue` part; I'm worried it might cause an issue somewhere with the game not being fully loaded at that point

PR #5776, opened 2025-02-27
First Attempt at Regional Debuff Rework

EatPant2nd:
## Description

Regional debuff rework. Instead of being a global value, it's now based on how far away a Pokémon is from its native region. It goes from 45% mantainted to 25% depending on the distance. As an example, 
A Pokémon from Kanto would have:
- 45% maintained in Johto.
- 35% in Hoenn.
- 25% in Sinnoh and onwards.

A Pokémon from Sinnoh would have:
- 25% in Kanto
- 35% in Johto
- 45% in Hoenn
- 100% in Sinnoh
- 45% in Unova
- 35% in Kalos
- 25% in Alola and onwards.

Or you can take a look at this table:
debuff-2

Regarding current and new debuff, the raw data looks like this (WIP):
Debuff

The calculated debuff was obtained by averaging the number of Pokémon multiplied by the new debuff divided by the total number of Pokémon debuffed. It looks quite harsh in the later regions but no matter what approach was made to the debuff, this would have happened regardless. 

**NOTE:** Right now only Kanto - Sinnoh* have been rebalanced.
- Johto has been buffed by 70%. Youngster Joey has not been touched though.
- Hoenn by 20%.
- *Sinnoh will receive no rebalance since the difference is not that noticeable.
- Unova and onwards are pending.

## Motivation and Context

The debuff was getting to a point where it was getting to 100% maintained and was fulfilling no purpose. This rework aims to incentivize people to breed from current region. It was chosen to be gradually decreasing instead of a flat harsh value so it's less painful.

## How Has This Been Tested?

Lots of numbers, also that's why I made the PR, so people can take a look at it and maybe give their input regarding the values from either what was done or the pending regions.

## Types of changes

- Balance


EatPant2nd:
See, I'm terrible at math so I was hoping someone would help with reaching the right numbers. I did it like this because I didn't know any better lol.

WarioInParis:
if you want to take away 20% of the value of something, you'll need to multiply by 0.8. 20% less of 45 is 0.8*45 or in other words 36. Then, 20% less of that is 0.8*36 or in other words 28.8. And so on.

PR #5781, opened 2025-03-05
Kubfu line evolution and catching locations.

aheadfrogg:
## Description

I added Kubfu and the Urshifus to the tower dungeons, with the requirement that you need to catch the urshifus for them to spawn as a boss. Kubfu is just a regular dungeon encounter. I also removed the Urshifu rewards from completing the dungeons. I also made them an actual evolution.

## Motivation and Context

I wanted to remove them from the Friend Safari.

## How Has This Been Tested?

I loaded up a file and went through the isle of armor questline to see if everything worked and it seemed to. I also checked if the pokemon had left the Friend Safari and they had.

## Types of changes

- Evolution and pokemon location change


src/scripts/dungeons/Dungeon.ts
```diff
@@ -12788,6 +12788,7 @@ dungeonList['Brawlers\' Cave'] = new Dungeon('Brawlers\' Cave',
 
 dungeonList['Tower of Darkness'] = new Dungeon('Tower of Darkness',
     [
+        {pokemon: 'Kubfu', options: { weight: 4 }},
```
RegisCoaxans, line 12791: That weight makes it half the encounters in the dungeon. Is it not overkill ? 1 in 5 or maybe 1 in 3 seems good enough to me.

src/scripts/dungeons/Dungeon.ts
```diff
@@ -12815,21 +12816,15 @@ dungeonList['Tower of Darkness'] = new Dungeon('Tower of Darkness',
     },
     28886112,
     [
+        new DungeonBossPokemon('Urshifu (Single Strike)', 160924440, 60, {requirement: new ObtainedPokemonRequirement('Urshifu (Single Strike)')}),
```
RegisCoaxans, line 12819: Could you add a ClearDungeonRequirement so Urshifu can not be battled on first clear as well ?

src/scripts/dungeons/Dungeon.ts
aheadfrogg: Done

src/scripts/dungeons/Dungeon.ts
aheadfrogg: Done

Qwertypop04 (changes requested):

Qwertypop04 (approved):

PR #5783, opened 2025-03-11
Changed 916.01.png (Pokemon)

wingull7:
Replaced old Oinkologne (Female) sprite with a new one that is much closer to the other sprites


Qwertypop04:
This is not from the smogon sprite project. The time may come where we have to consider alternatives to that project, but now is not that time. And in such a case, we would want to get many sprites from the same place, not just one sprite at a time

PR #5786, opened 2025-03-15
Currency display updates

CypherX:
## Description
- Removes the `Show Diamonds, Farm Points, Battle Points, and Contest Tokens on main screen` setting
- Farm Points, Diamonds, Battle Points, and Contest Tokens will be displayed upon unlocking their source locations or if the player has acquired any.

## How Has This Been Tested?
Checked the display with files of various progression and that each currency type appeared correctly after unlocking or being obtained


AMB1008:
I like this change but I wish there was maybe settings to disable specific currencies such as FP or Contest Tokens but keep everything else on. Also, this seems to already be partially in effect as seen on one of my challenge run files having everything *but* contest tokens visable.
uno
dos

Qwertypop04 (approved):

PR #5800, opened 2025-03-29
Makes finished or current quest lines clickable.

hippolythe:
Makes finished or in-progress quest lines clickable  from the Quests Lines menu, so you can see the list of tasks for each quest.

## Motivation and Context
Allows you to view the history of quest tasks, which can be useful for remembering steps in the case of a quest that has been paused for a long time.

## How Has This Been Tested?
Open lots of completed and ongoing quests at different stages of the quests.

## Screenshots (optional):
An exemple of a completed quest :
image

An exemple of a in-progress quest:
image

## Types of changes
- New feature
- UI improvement


AevitasDragonkin:
I would suggest some UI changes
image
image

1. `vertical-middle` on both the first and second column
2. `text-left` on the first column to keep the consistency with the main quest modal
3. I would also include the quest step index, that will match up with the wiki
4. The second column has some weird top margin on the buttons, and I would remove the buttons and just make it coloured text. If it looks like a button I would expect it to do something
5. Personally I would probably put the locked steps as `5. ???`
6. Highlight the current step
7. Add some extra padding (`div` inside `td`), so that it still has enough padding inside the cells for mobile
8. `text-wrap: no-wrap` on the last column

Proposed:
image

Right now this feature is only enabled on 'Active Quest Lines' and 'Complete Quest Lines'. 
Can we also enable this on 'Available Quest Lines' with the following reasoning:
* It will show everything as locked, but Complete Quest Lines also show everything as 'Complete'
* Paused quest lines also show up here, this will enable someone to view where they're at on their paused quest line

The changes to `src/components/pokemonListContainer.html` don't seem to belong in the PR.

No need to store the `selectedQuestName` in the statistics, I'd put that as a static property on the `Quests` class.

`getQuestDetails`
1. Create some failsafes, for example if `this.getQuestLine(questName)` doesn't return a value for some reason
2. Use `description` instead of `customDescription`
3. All tasks should be displayed and not filtered wether they have a description or not, you could use the `.map` function here instead

No need to add the `.disabled-button`, this will be fixed in #5810 

No need to add `. questListRow`, if something needs to be clickable you can add the `clickable` class

hippolythe:
@AevitasDragonki I've taken your feedback into account

src/components/questDetailsModal.html
```diff
@@ -0,0 +1,31 @@
...
+                        </td>
+                    </tr>
+                </tbody>
+            </table>
+            <div class="modal-footer">
+                <button type="button" class="btn btn-danger" data-dismiss="modal">Close</button>
+            </div>
+        </div>
+    </div>
+</div>
```
AevitasDragonkin, line 31:
```suggestion
<div class="modal noselect fade" id="questDetailsModal" tabindex="-1" role="dialog" aria-labelledby="questDetailsModalLabel">
    <div class="modal-dialog modal-dialog-scrollable modal-lg" role="document">
        <div class="modal-content">
            <div class="modal-header align-items-center">
                <h4 class="m-0" data-bind="text: App.game.quests.selectedQuestName()"></h4>
            </div>

            <div class="modal-body p-0 p-md-2 p-lg-3">
                <table class="table table-striped table-bordered table-hover table-sm m-0">
                    <tbody data-bind="with: App.game.quests.getQuestLine(App.game.quests.selectedQuestName())">
                    <!-- ko foreach: $data.quests() -->
                    <tr data-bind="css: { 'bg-info': ($parent.state() === QuestLineState.started || $parent.state() === QuestLineState.suspended) && $index() === $parent.curQuest() }">
                        <td class="vertical-middle"><div class="px-2" data-bind="text: $index() + 1"></div></td>
                        <td class="vertical-middle text-left">
                            <div class="px-2" data-bind="text: $parent.state() === QuestLineState.ended || $index() <= $parent.curQuest() ? $data.description : '???'"></div>
                        </td>
                        <td class="vertical-middle">
                            <div class="px-2 text-nowrap">
                                <span class="text-success" data-bind="if: $parent.state() === QuestLineState.ended || $index() < $parent.curQuest()">Completed</span>
                                <span data-bind="if: $parent.state() === QuestLineState.started && $index() === $parent.curQuest()">In progress</span>
                                <span class="text-danger" data-bind="if: $parent.state() === QuestLineState.started && $index() > $parent.curQuest()">Locked</span>
                                <span data-bind="if: $parent.state() === QuestLineState.suspended && $index() === $parent.curQuest()">Paused</span>
                                <span class="text-danger" data-bind="if: $parent.state() === QuestLineState.suspended && $index() > $parent.curQuest()">Locked</span>
                            </div>
                        </td>
                    </tr>
                    <!-- /ko -->
                    </tbody>
                </table>
            </div>

            <div class="modal-footer">
                <button type="button" class="btn btn-danger" data-dismiss="modal">Close</button>
            </div>
        </div>
    </div>
</div>
```

Could you try it with this? There are multiple things that do not work with your code:
* The modal does not scroll when the screen height is not big enough
* The padding does not work as it's overwritten by bootstrap using !important on table cells on small screens
* .detailQuestModal is also overwritten by bootstraps dialog CSS
* Since this is the only place that uses it, I've opted out of creating a helper function, and done it straight in HTML
* I've also used existing class names to vertical align, add color and padding

src/scripts/quests/Quests.ts
```diff
@@ -200,6 +201,40 @@ class Quests implements Saveable {
...
+                isCurrentTask: task.index - 1 === quest.curQuest() && quest.state() !== QuestLineState.ended,
+                questStatut: quest.state(),
+            };
+
+            return detailsQuest;
+        });
+
+        return returnObject;
+    }
+
```
AevitasDragonkin, line 237:
```suggestion
```
No longer needed if it's all being done within HTML

src/styles/quest.less
```diff
@@ -36,23 +36,53 @@
...
+    padding: 0px 10px;
+}
+
+.lockedTask {
+    color: red;
+}
+
+.completedTask {
+    color: green;
+}
```
AevitasDragonkin, line 88:
```suggestion
```
No longer needed as these are already existing bootstrap classes

AevitasDragonkin (changes requested):

PR #5801, opened 2025-03-29
Added Surfing Pikachu to Kanto route 19

aheadfrogg:
## Description

Surfing Pikachu is now on route 19 once it has been obtained. This is in reference to the surfing pikachu minigame in pokemon yellow.

## Motivation and Context

I wanted to remove the pokemon in the friend safari.

## How Has This Been Tested?

I opened up a file with the changes and made sure the surfchu only appeared on the route when it was caught.

## Types of changes

- pokemon location change


Qwertypop04 (approved):

PR #5802, opened 2025-03-29
Added dual types to some pikachu alt forms and pikachu related alt forms.

aheadfrogg:
## Description

I gave the psychic secondary typing to Detective pikachu and raichu. I also gave the normal typing to lets go pikachu and the electric typing to lets go eevee.

## Motivation and Context

I did this because I want alt forms to have more unique typings. The psychic type was chosen for the detective line because intelligence is often associated with the psychic typing and you need to be smart to be a good detective. The lets go typings were based on the fact that gengar pikachu has the ghost typing and its just wearing a gengar onsie.

## How Has This Been Tested?

I loaded up a file and made sure all those pokemon had the correct typings.

## Types of changes

- pokemon type change


Qwertypop04 (approved):

PR #5807, opened 2025-04-12
Vitamin UI improvement to make distributing vitamins less tedious

CypherX:
## Description
Allows editing the number of vitamins applied to a pokemon directly. Like any normal input the tab key can be used to move to the next field.

`notifySubscribers` is called after the `setVitaminAmount` function to force the UI to update if the amount of vitamins applied differs from the amount entered, or if the amount doesn't change

## Motivation and Context
While I do think vitamins need a rework, this should make it less tedious until then

## How Has This Been Tested?
Adjusted vitamins, tried to enter invalid amounts and verified it handled all scenarios correctly

## Screenshots (optional):
https://github.com/user-attachments/assets/e7111f06-b5af-4947-9cb4-ced659050d33

## Types of changes
- UI improvement


Qwertypop04 (approved):

PR #5812, opened 2025-04-19
Sprinklotad boosting mulch duration

### LINKED ISSUE #5518: [SUGGESTION]: Sprinklotad rework
### Current Behavior

Currently the Sprinklotad multiplies the chance for berries to replant.

### Improved Behavior

I believe it should multiply wanderer chance instead.

### Reason for change

Sprinklotad is one of the least used Oak Items, only used to mutate a berry and complete the achievements. The increase to replant chance doesn't help the berries that could use a replant chance the most (because a lot of those berries already have 0% replant chance).

### Detailed Description

Essentially just change it to multiply wanderer chance instead. If possible, colour and berry-specific wanderers specifically, but all wanderers is fine too.

AevitasDragonkin:
## Description

Sprinklotad Oak Item has been refactored form boosting replant chance to mulch duration.
Scaling as follows:
| Level | Boost |
|-------|-------|
| 0     | x1.25 |
| 1     | x1.40 |
| 2     | x1.55 |
| 3     | x1.70 |
| 4     | x1.85 |
| 5     | x2.00 |

XP Gain is 1 XP per minute per plot that has mulch on it.
Level 5 time should be between 6.5 hours (25 mulched plots) and 167 hours (1 mulched plot)

Previous farming points invested in levelling up have been refunded. And the current Level and XP have been reset.

## Motivation and Context

Sprinklotad was deemed very bad and not used other than levelling it up. This will bring it another purpose that will get used more.
Closes #5518 , while this was not the suggested answer, it does fulfill its purpose of being more useful

## How Has This Been Tested?

Added Mulch, saw the timer being 10:00 instead of 5:00 with a level 5.
Added Mulch without and saw the timer being 5:00
Equiped / Unequiped while Mulch was active and saw the timer adjust
Saw the Oak Item XP increase each minute there was mulch, XP gained matched with amount of mulched plots.

## Types of changes

- Rebalancing


Qwertypop04 (approved):

PR #5813, opened 2025-04-23
'New Shadow' default filter

AevitasDragonkin:
## Description

Added 'New Shadow' as a default filter. This unlocks as soon as the Shadow filter option becomes available

## Motivation and Context

We had one for Contagious, might as well add one for New Shadow. Can now remove the pinned post in discord and refer to the default filters.

## How Has This Been Tested?

Tested with a file that was just before Wallace. Defeated him, started the Shadows in the Desert quest and progressed until the Shadow filter option becomes available. At this point the default 'New Shadow' filter popped up.

## Screenshots (optional):

image


Qwertypop04:
In testing, when reaching the point when shadow filters are unlocked, I did not get a new default filter

AevitasDragonkin:
> In testing, when reaching the point when shadow filters are unlocked, I did not get a new default filter

Added update function that pushes the new filter to existing save files when they have not started, or are not yet at quest step 4 in the Shadows of the Desert quest.
Files that have already unlocked that step (and the shadow feature) are expected to already having set a filter for shadows, in which case I haven't pushed that filter.

CypherX:
I added an IntroJS overlay to alert the player to the newly available filter and also updated the update function to handle inverted filter priority. Can someone double check my changes before I merge this?

image

PR #5817, opened 2025-05-03
Add small delta to Farm Hand-planted berries to fix rounding issue

PFQNiet:
## Description

Berries planted by Farm Hands will get a small (0.1s) boost to their age.

## Motivation and Context

I was trying to use Farm Hand Jessie to aggressively farm some Leppa Berries. Jessie's Work Time is 1m and Leppa Berries take 4 minutes to grow. Perfect match, right?

Unfortunately because it takes one game tick for the Farm Hand to work, the Leppa Berries are still one tick away from being ready to harvest. This means that I have to wait an entire extra Work Cycle for the plants to be handled.

This PR just adds a 0.1s "nudge" to berries planted by Farm Hands, so that you can rely on them actually harvesting on time.

## How Has This Been Tested?

Implemented, built and run, I let Jessie harvest several rounds of Leppa Berries and they are now harvested flawlessly as soon as they are ready.

I have not tested on longer berries, so depending on how exactly Farm Hand cycles round, it's *possible* that 0.1s may not be enough. It can be trivially bumped up if needed.

## Types of changes

- Bug fix (rounding error)

### DIFF
--- a/src/scripts/farming/FarmHand.ts
+++ b/src/scripts/farming/FarmHand.ts
@@ -234,7 +234,7 @@ class FarmHand {
                     berry = berry < 0 ? BerryType.Cheri : berry;
                     // Only plant and work if the player has a berry to plant
                     if (App.game.farming.hasBerry(berry)) {
-                        App.game.farming.plant(emptyPlotIndex, berry as BerryType);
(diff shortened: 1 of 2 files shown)


CypherX:
Superseded by #6106, thanks for the contribution

PR #5821, opened 2025-05-10
Simplify logic for revealing dex after defeating the champion

CypherX:
Simplifies the logic from #5738 and adds a safeguard if a champion is undefined (shouldn't ever happen, but ya know)


Qwertypop04 (approved):

PR #5823, opened 2025-05-10
Fix Red fight during An Unrivaled Power

RegisCoaxans:
## Description
I broke the requirement a few months ago when fixing another softlock... It now only appears when the quest is active.
The update function skips the quest for people who defeated Red while the questline was paused.

## Motivation and Context
Bug bad.

## How Has This Been Tested?
Loaded a file at different steps of the quest line to ensure it was working as expected.

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #5834, opened 2025-05-22
Added XP to Solgaleo (Radiant Sun) and Lunala (Full Moon)

AevitasDragonkin:
## Motivation and Context

Solgaleo (Radiant Sun) and Lunala (Full Moon) were the only pokemon that were not giving any XP. Now they give the same XP as their base forms.

## How Has This Been Tested?

🪄 BF to check, pokemon were not levelling before fix, they are now levelling after fix.

## Types of changes

- Bug fix


Qwertypop04 (approved):

PR #5845, opened 2025-05-29
No more Enigma duplicate hints

RegisCoaxans:
## Description
This change makes it impossible to get the same direction twice. If the berry master tells about Enigma mutation, the hint always is a new piece of information. After the four directions are revealed, the berry master reads the entire hint instead.
Also fixed a bug with the master continuously revealing new hints without having to talk to him again.

## Motivation and Context
Getting these hints with high chances of repeats is awkward. The berry master should only repeat himself if the player can do something about it.

## How Has This Been Tested?
Updated files with no hint, all hints and some hints to ensure all was preserved. Unlocked all hints and ensured it would deliver a different direction at random every time the mutation of the day is Enigma. Made sure other mutation hints would not break.

## Types of changes
- Feature improvement
- Bug fix


AevitasDragonkin (commented):
Would it be better if the generation of the hints is fully done inside the EnigmaMutation object, and all the NPC has to do, is mark the hint as seen. That way the NPC doesn't need to handle any logic about the hintIndex or if all hints have been seen before.

RegisCoaxans:
Yeah. No idea why I stuck with adding code here... Implemented the asked changes and I believe it's even cleaner than before, now.

AevitasDragonkin:
Just found a bug with loading. You don't supply the last property in the Update function, this causes the lastIndex() to be undefined. And this skips the lastIndex() === null check.

That then calls this function `return `There's a mysterious berry that requires ${this.getHint(undefined)}.`;`

image

RegisCoaxans:
Fixed.

AevitasDragonkin (approved):
Looks good now

I guess it's intended that the NPC will show all 4 hints and it needs a 5th day to show them all at once. Either way, I don't think it's a big deal and can leave it as is right now. 
There's still the compiled hint at the berry dex.

src/scripts/towns/NPCController.ts
```diff
@@ -1,9 +1,17 @@
...
     public static selectedNPC: KnockoutObservable<NPC> = ko.observable();
-
+    private static modalState: any;
     public static openDialog(npc: NPC) {
         this.selectedNPC(npc);
         $('#npc-modal').modal();
         npc.setTalkedTo();
+        if (!this.modalState) {
+            this.modalState = DisplayObservables.modalState['npc-modalObservable'].subscribe((value: BootstrapState) => {
+                if (value === 'hide') {
```
CypherX, line 10:
```suggestion
                if (value === 'hidden') {
```
hide - as the modal begins closing
hidden - after it's fully closed

prevents the model content from disappearing as it closes

CypherX (approved):

PR #5847, opened 2025-05-30
Preventing illegal Pokémon from being bred or assigned vitamins or given held item

RegisCoaxans:
## Description
From now on, they are excluded from the respective Pokémon lists of these features.
The Pokémon Statistic Modal also displays a BIG DANGER message for those Pokémon.

## Motivation and Context
Continuation of #5771.

## How Has This Been Tested?
I magically obtained them. Tried to give them vits, could not. Tried to give them held items, could not. Tried to see their statistics, could not.

## Types of changes
- Bug fix
- UI improvement

### DIFF
--- a/src/components/pokemonStatisticsModal.html
+++ b/src/components/pokemonStatisticsModal.html
@@ -14,9 +14,9 @@
           <span aria-hidden="true">&times;</span>
         </button>
       </div>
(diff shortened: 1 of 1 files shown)


RegisCoaxans:
Good catch, thank you.

PR #5853, opened 2025-05-31
Track file creation timestamp

CypherX:
## Description
Saves a timestamp of when the save file is created. Existing files will be set to zero to indicate they were created before this was added

## Motivation and Context
Neat info to have

## How Has This Been Tested?
Loaded an existing file and verified it was set to zero.
Created a new file and verified the timestamp was set

## Types of changes
- New feature


RegisCoaxans:
Can it be displayed somewhere in the statistics modal ?
For people whose file creation timestamp is 0 anyway, that could read « before [...] », or something, where the date is computed using total playtime ?

Qwertypop04 (approved):

PR #5877, opened 2025-06-12
Feature: fairer offline progress

RegisCoaxans:
## Description
Based the values on total earned divided by seconds played. Seems very high but that still can't defeat online time at all and it's actually average by definition. The bonus includes DT by the way.
Could be expanded with more rewards or even more types of rewards later if we want.
Removed related settings because they are either obsolete or non-applicable, as the player now has to actively collect the bonus.

## Motivation and Context
Offline bonus was a joke, and a bad one. :^)

## How Has This Been Tested?
Loaded files at different steps through the game, with different offline times.

## Types of changes
- New feature
- UI improvement


RegisCoaxans:
Not satisfied with this design anymore.

PR #5879, opened 2025-06-14
Fix sortable binding UIs desyncing from internal order

umbralOptimatum:
Fixes UIs that use the `sortable` binding, like the full pokeball selector modal, sometimes getting out of sync with the internal setting.

Marked as draft for the moment since this issue is difficult to reproduce and will need testing on others' devices to confirm it's properly fixed.

## Description
Reworked our `sortable` custom knockout binding to keep the UI and internal model properly synced. As far as I can tell, the issue came from reordering elements while Knockout was already in the process of updating the UI; this was exacerbated by the binding creating a new Sortable instance every time an element was moved without disposing of the existing one. The binding now only creates one Sortable instance which sorts the underlying array to exactly match the UI.

## How Has This Been Tested?
Messed around with the three existing sortable menus (`pokeballSelectorAdvModal`, `categoryModal`,`settingsModal` map colors) and didn't encounter any desyncs. However this bug is difficult to reproduce consistently so this'll need testing on other people's devices to confirm this fully works.

## Types of changes

- Bug fix

### DIFF
--- a/src/scripts/Update.ts
+++ b/src/scripts/Update.ts
@@ -2938,7 +2938,8 @@ class Update implements Saveable {
             const categoryNoneIndex = saveData.categories.categories.findIndex(c => c.id === 0);
             if (categoryNoneIndex > 0) {
                 const cats = saveData.categories.categories;
-                saveData.categories.categories = [cats[categoryNoneIndex], ...cats.toSpliced(categoryNoneIndex, 1)];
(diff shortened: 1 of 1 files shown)


CypherX:
If the 'None' category is going to have its position locked to the top going forward we'll need an update function to change its position for any players who have previously re-ordered it

CypherX (approved):

PR #5906, opened 2025-07-03
Refactor a bit of the roamer chance function

RedzepR:
## Description

Seperate the rolling for a roamer and the calculation of a the roamer chance. Remove a bunch of parameters that are no longer needed

## Motivation and Context

The wiki can more easily calculate the roamer chance per route.

## How Has This Been Tested?

Checked the roamer chance of a bunch of routes and they were all the expected value.

## Types of changes

- refactor


umbralOptimatum (approved):
Confirmed that after my changes all routes still have the same roamer chance as on dev branch.

PR #5916, opened 2025-07-05
Fix breeding flag getting reset for pokemon in queue on load

CypherX:
## Description
Fixes a bug causing pokemon in the hatchery queue to have their `breeding` flag set to false on load allowing multiple of the same pokemon to be queued simultaneously

## How Has This Been Tested?
Added a pokemon to the queue, reloaded, could not add them to the queue again

## Types of changes
- Bug fix


umbralOptimatum:
While you’re making changes to the breeding flag check, it would make sense to use a Set so having a lot of pokemon in the queue doesn’t affect load speed.

Not related to the current bug, `Breeding.addPokemonToHatchery()` should get an initial safety check for pokemon already in the hatchery.

CypherX:
Updated to use a Set

Qwertypop04 (approved):

PR #5925, opened 2025-07-07
Removes the Genesect from the FS

RedzepR:
## Description

Adds the drive genesect to the p2 lab based on moon cycle. Adds the high speed drive genesects to various dungeons in unova. Adds an NPC to explain the mechanic.

## Motivation and Context

Genesect use up 8 slots of the current FS. Moon cycle and "easter dungeon jumping" mechanic are both underutilized. Idea was thought up by the discord.

## How Has This Been Tested?

Did a few dungeons, saw Genesect jumping. Saw the NPC at the p2 lab.

## Types of changes

- FS removal


Qwertypop04 (approved):

PR #5934, opened 2025-07-10
Fix more issues with "There can be only one" secret achievement

CypherX:
## Description
Turns out the setting values are unreliable on new saves so changed it to work a different way

## How Has This Been Tested?
Verified that the achievement unlocked correctly on existing and new files

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #5935, opened 2025-07-11
Change: standardized Pokémon-related map filters

RegisCoaxans:
## Description
Added a function on MapHelper that does all the comparisons to the party and returns all matching area statuses.
If a Pokémon matches the uncaught filter, it doesn't match uncaught shiny, missing resistant and uncaught shadow ones anymore.

## Motivation and Context
There were inconsistencies with how the filters calculated in some places.
Challenged people will like for unwanted Pokémon not to pollute the UI with all these map filters anymore. Harmless to R2C as they have to catch anything anyway.

## How Has This Been Tested?
Magically obtained and lost Pokémon and saw how the map behaved here and there.

## Types of changes
- Bug fix
- New feature
- UI improvement


src/scripts/worldmap/MapHelper.ts
```diff
@@ -383,6 +368,34 @@ class MapHelper {
 
     }
 
+    public static getPokemonAreaStatus(pokemon: PokemonNameType[], pokerusUnlocked: boolean): areaStatus[] {
```
umbralOptimatum: Is there a reason `getPokemonAreaStatus()` needs a `pokerusUnlocked` parameter? All of its uses here are just passing in the setting, it could just check that internally.

umbralOptimatum (changes requested):

src/scripts/worldmap/MapHelper.ts
RegisCoaxans: Oops, lol.
Should be good now.

umbralOptimatum: Looks good to me now!

umbralOptimatum (approved):

CypherX (approved):

PR #5937, opened 2025-07-11
Fix update error for files from v0.10.14 or earlier

CypherX:
## Description
`npcTalkedTo` statistic doesn't exist on files earlier than v0.10.14 so it would error on update

## How Has This Been Tested?
Loaded a save file from v0.10.0

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #4874, opened 2023-11-08
Great Quest Refactor, part 2

RegisCoaxans:
## Description
As mentioned in https://github.com/pokeclicker/pokeclicker/pull/4659, a second part would be turning some CustomQuests into their respective quest type. Few quests which should have been turned into DefeatDungeonQuest were turned into DefeatDungeonBossQuest instead.

## Motivation and Context
I wish I had not the motivation to copy and past that much quests.

## How Has This Been Tested?
Same as in https://github.com/pokeclicker/pokeclicker/pull/4659, I used a script to stringify the quests, diff between them and fix the typos. I also ran a code to ensure every DefeatDungeonQuest or DefeatGymQuest I created had a valid target.

## Screenshots (optional):
From my first PR :
260229042-8bb299ee-f2bb-4d1c-a1e9-28ccaab1e1df
Now : 
image

## Types of changes
- Refactor
- Code cleaning

PR #5940, opened 2025-07-12
fix(questline): Lower the required amount of clears of Castelia Sewers in Hollow Truth and Ideals

RegisCoaxans:
## Description
The quest initially required 1 clear, I mistakenly made it 10 in #4874, now making it 1 again.

## Motivation and Context
That change was unintended, and it's very annoying.

## How Has This Been Tested?
Hasn't, people with higher progress will just be warped to the next quest.

## Types of changes
- Bug fix


Qwertypop04 (approved):

PR #5945, opened 2025-07-14
Squirtle ! Squirtle ! Squirtle !

RegisCoaxans:
## Description
Added Squad Leader Squirtle from the anime, to the game. It's a level evo using black glasses.

## Motivation and Context
Ash's Squirtle is very cool, we had to add it !

## How Has This Been Tested?
Loaded the game and obtained the Pokémon as intended.

## Types of changes
- New Pokémon


src/scripts/items/HeldItem.ts
```diff
@@ -251,4 +251,11 @@ ItemList.Everstone = new HeldItem('Everstone', 10000, GameConstants.Currency.mon
...
         return baby !== undefined && pokemon.name != baby;
-    });
+    }
+);
+
+// Override Black Glasses' canUse function to factor in Squirtle evolution.
+const glassesRestrict = (ItemList.Black_Glasses as AttackBonusHeldItem).canUse;
+(ItemList.Black_Glasses as AttackBonusHeldItem).canUse = (pokemon: PartyPokemon): boolean => {
+    return pokemon.name != 'Squad Leader Squirtle' ? pokemon.name == 'Squirtle' || glassesRestrict(pokemon) : false;
+};
```
CypherX: I'm not exactly thrilled with doing it this way but the only alternatives I can think of are having `TypeRestrictedAttackBonusHeldItem` accept an optional function parameter for additional criteria that `canUse` calls, or extend `TypeRestrictedAttackBonusHeldItem` with a new class to do something similar

aheadfrogg:
Why was level 25 selected?

src/scripts/items/HeldItem.ts
Qwertypop04: I think it also feels unintuitive, although maybe that's not a bad thing, since it's an optional Pokémon.

RedzepR:
Instead of placing this Pokemon in the FS it should spawn at R11 to mimic the fact you get Squirtle in Pokemon Yellow in Viridian/Lets Go.
image
 
Then I'd place an NPC (Officer Jenny) in Viridian either once you move to Johto or obtain your first Black Glasses that says something along the lines "A bunch of wild squirtles terrorize the area, it would need a strong squirtle leader to unite them... i heard their love sunglasses." because having an optional mon with no hint ingame is whack.

And then I'd also like to have the squirtle *squad* as a pokemon, not just the leader. 
Preferably in their later anime outfit.
image

Askellown:
> Instead of placing this Pokemon in the FS it should spawn at R11 to mimic the fact you get Squirtle in Pokemon Yellow in Viridian/Lets Go. 

I think in the anime in happens between the Charmander episode and the Bill's lighthouse episode meaning it's on 24/25

RegisCoaxans:
> Why was level 25 selected?

No particular reason. I wanted a greater level than 16, but lower than 36. 25 looked like a good middleground.

RegisCoaxans:
Added to route 11 for EV sake.
Officer Jenny is added to Vermilion and tells the player about the Squirtle Squad.
Changed the level evolution from 25 to 10 so it can be triggered before evolving into Wartortle.

CypherX (approved):

PR #5973, opened 2025-08-04
Update farm mutation indicator

CypherX:
## Description
Updates the farm mutation indicator to always be visible when a mutation is possible, even if the player has not yet seen the hint. The tooltip will display the berry name if the player has unlocked the berry previously or has seen the mutation hint, otherwise it will display '???'

## Motivation and Context
something something where green dot

## How Has This Been Tested?
Loaded several saves with active mutation layouts and verified the indicator was visible and showed the berry or ??? 

## Types of changes
- UI improvement


RegisCoaxans (changes requested):
The indicator erroneously shows up if the mutation requirement isn't satisfied.
Tested with Ganlon without having caught any Groudon yet. It also worked on Passho + Enigma before unlocking Liechi.

CypherX:
> The indicator erroneously shows up if the mutation requirement isn't satisfied. Tested with Ganlon without having caught any Groudon yet. It also worked on Passho + Enigma before unlocking Liechi.

fixed

RegisCoaxans (approved):
Seems to be working fine now.

RedSparr0w (approved):

PR #5981, opened 2025-08-12
Fix farmhands getting all energy back upon firing

RedzepR:
## Description

Changed from tick to worktick cycle

## Motivation and Context

Was a bug

## How Has This Been Tested?

Checked ingame cycles

## Types of changes

- Bug fix

### DIFF
--- a/src/scripts/farming/FarmHand.ts
+++ b/src/scripts/farming/FarmHand.ts
@@ -170,7 +170,7 @@ class FarmHand {
             this.workTicks(0);
             if (this.hired()) {
                 this.work();
-            } else {
+            } else if (this.energy() < this.maxEnergy) {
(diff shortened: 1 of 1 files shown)


src/scripts/farming/FarmHand.ts
```diff
@@ -157,18 +157,9 @@ class FarmHand {
     }
 
     tick(): void {
```
wormania, line 159: I'd like it if there was an early return on `!this.hired() && this.energy() >= this.maxEnergy`, to avoid incrementing 18 observables every tick

RedzepR: added

wormania (approved):

CypherX (approved):

PR #6016, opened 2025-08-31
Adding Leftovers to the game

aheadfrogg:
## Description

I added the leftovers item to the game and made it an exp boosting item. It can be dropped from Snorlax and Munchlax, in the Sinnoh victory road and castelia sewer dungeons and purchased at the Hoenn battle frontier shop. It boosts exp gain by 100% and it is able to be held by any pokemon.

## Motivation and Context

I wanted to add a universal exp item that is better than the ones we currently have and this gives a use to the leftovers item.

## How Has This Been Tested?

I loaded up a save and made sure that the pokemon were dropping the item, the dungeons were dropping the item, that the shop was working and it was for the right price and that it was actually working as a held item.

## Types of changes

- Held Item add


RegisCoaxans:
Munchlax and Snorlax are route encounters at the time of Galar. Snorlax can be found wild before in Hall of Origin at best. That kind of makes the exp bonus pointless, even for weird challenge runs, doesn't it ?

PR #6024, opened 2025-09-10
Add a notification when clicking the save icon

RedzepR:
## Description

Add the same notification you get when you press Crtl + S

## Motivation and Context

Clicking something and not getting a feedback is bad

## How Has This Been Tested?

Go ingame and click save icon

## Types of changes

- notifications


umbralOptimatum:
The notifier call would be better placed in `Save.store()`.

RedzepR:
> The notifier call would be better placed in `Save.store()`.

With an parameter to make it not appear? Otherwise the game would give the notification every 10s

umbralOptimatum:
Definitely. I was thinking mostly in terms of minimizing the amount of code off in the HTML.

RedzepR:
Done

CypherX (approved):

PR #6026, opened 2025-09-11
Display dungeon loot % chance

CypherX:
## Description
Adds the % chance of each item to the dungeon loot info modal.

Can someone double check that the calculation is right? I'm not 100% confident it's fully correct.

## Motivation and Context
More info display in game

## How Has This Been Tested?
Looked at the loot of various dungeons with various amounts of clears.

## Screenshots (optional):
image
image
image

## Types of changes
- UI improvement


RegisCoaxans (approved):
Couldn't find some more oddities.

PR #6028, opened 2025-09-11
Farm shop

RedzepR:
## Description

Add a farm shop after you unlock all 25 plots and reach goldenrod. Default shortcut W.

## Motivation and Context

You can buy basic stuff after johto champ. Can't buy farm stuff after johto champ. Might change the requirement to reach slateport. 

## How Has This Been Tested?

- Was shown on full cleared save
- Was hidden on file who didnt clear goldenrod

## Screenshots (optional):

image

## Types of changes

- new shortcut


RedzepR:
Added the farm items to the regular shortcut shop with a requirement
Saves shortcut usage
better experience (find all in one place)

RegisCoaxans (changes requested):
Goldenrod is also the first location the player can buy farm stuff at.
Making the portable shop unlock at the same time means the Berry Master shop are barely relevant.
It should be at least Hoenn, ideally Sinnoh, so the player has time to discover them in towns. And probably tied to defeating the league just like the other available items are.

RedzepR:
> It should be at least Hoenn, ideally Sinnoh, so the player has time to discover them in towns. And probably tied to defeating the league just like the other available items are.

Done.

RegisCoaxans (approved):
Merging this now but it should probably be done more neatly next time we ever want to expand it.

PR #6033, opened 2025-09-16
Add new settings to filter Pokémon based on held item quality

Kelenius:
Adds two new settings when dealing with held items: "Hide Pokémon holding equivalent or better item" and "Only show Pokémon holding inferior item".

## Description
"Hide Pokémon holding equivalent or better item" hides any pokemon that holds an item that provides the same bonus and is higher or equal. E.g. if you're giving out Miracle Chest (+50% XP), pokemon holding Miracle Chest or any Scent will be hidden, but pokemon holding nothing, Wonder Chest, or +attack items will be visible.

"Only show Pokémon holding inferior item" hides all pokemon other than ones holding an item that provides a lower bonus of the same type, or no item. E.g. if you're giving out Miracle Chest, only pokemon with no item and pokemon holding Wonder Chest will be visible.

## Motivation and Context
Makes it easier to hide pokemon that are already holding something as good as the item (which there can be multiples of due to dual types and pokemon-specific items) you're giving, or quickly upgrade to a better item.

## How Has This Been Tested?
I tested that filters apply correctly.

A part of the changes involved changing how conditional items work, so I checked that ` ItemList.Power_Herb.attackBonus ` updates correctly in the console when entering a boss fight.

## Screenshots (optional):
изображение
изображение
изображение

## Types of changes
- UI improvement


Kelenius:
Updated
image

RegisCoaxans (approved):

PR #6036, opened 2025-09-17
Display both per-floor and total chest/encounter count in dungeons

Kelenius:
Multi-level dungeons will display chest and encounter count both for the current floor and the whole dungeon.

## Description
Currently, they only show totals for the whole dungeon. With this change it's displayed as `[this floor/this floor count] ([total/total count])`.

## Motivation and Context
I'd like to see the numbers for the current floor.

## How Has This Been Tested?
Checked that Kanto dungeons still display in the same way they do before this change.

Checked that Alola dungeons display both counts.

Checked that counters update correctly when picking a chest, starting an encounter, and moving to the next level.

Checked that the maps are revealed when the correct number of chests is opened.

## Screenshots (optional):
изображение
изображение

## Types of changes
- UI improvement


RedzepR:
I find this more confusing than helpful
The idea itself might be worth following but the UI does _not_ get improved by this

Kelenius:
I could see it being confusing. Maybe the information could be added to the tooltip? Or maybe the format could be just `[collected this floor]/[total this floor] ([total])`? It could also only show future floors, since you can't go back to the previous floor.

RedzepR:
Do we _need_ the total amount? Wouldnt the amount per floor be enough?

Kelenius:
New version after a discussion on discord, only shows the current floor + total amount of chests opened (since it affects enemy HP):
image

RegisCoaxans (changes requested):
While the display looks fine, the implementation confuses me.
The point of observables is to remove the need for explicit updates of the view.
I recommend to add them back.

PR #6039, opened 2025-09-18
Add a separate color for affordable gem upgrades

Kelenius:
## Description
Currently, non-maxed upgrades have the same color in the gem summary section. This PR makes affordable upgrades show up with a different color.

I think the color blends a bit with a maxed color, but that's a general problem with the theme.

## Motivation and Context
I'd like to be able to quickly tell when I can afford gem upgrades.

## How Has This Been Tested?
Opened the gem case and confirmed that the colors are correct.

## Screenshots (optional):
Maxed - affordable - unaffordable - maxed
изображение

## Types of changes
- UI improvement

### DIFF
--- a/src/components/gemsModal.html
+++ b/src/components/gemsModal.html
@@ -42,7 +42,7 @@
                                 <div class="d-flex" style="gap: 0.25rem">
                                     <!-- ko foreach: Array(Gems.nEffects) -->
                                     <div class="flex-grow-1" style="height: 2px;" data-bind="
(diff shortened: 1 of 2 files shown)


RegisCoaxans (changes requested):
You're using bg-info, which is already in use when the gem type is selected. Although it's not functionally an issue, it makes the UI looks bad.
Can you find an other colour combination that doesn't interfer ?

RegisCoaxans:
After some tests, I found out that the `flash` animation is gentle enough on small UI elements. Could save one colour while making the affordable upgrades very noticeable. Might be worth giving it a try.

PR #6042, opened 2025-09-18
Add sorting to Pokedex

ribomblebee:
## Description

Added sorting options to Pokedex
- Added new setting values (pokedexSort, pokedexSortDirection)
- Duplicated HTML code from the breeding modal
- Sort behavior is to sort all caught and uncaught pokemon when sorting by ID, Name, Base Attack, and Times Hatched. For all other sort options, only party pokemon are sorted and uncaught pokemon are sorted last by ID (regardless of sort direction). 

## Motivation and Context

Better pokedex view and filtering

## How Has This Been Tested?

- Tested new file with caught and uncaught pokemon
- Tested new file using real evo
- Tested Galar file with 1000+ pokemon
- Tested modal while a Pokemon's attack changed (sort updates dynamically, like the hatchery) 

## Types of changes

- UI improvement


RedzepR:
Hatchery:
image
Pokedex:
image

It does not use the translated text even if it exists

CypherX (approved):

PR #6044, opened 2025-09-19
Show if aura is positive or negative in berry info

Kelenius:
## Description
In the berry info, aura type will indicate if the effect is positive or negative with a plus or a minus sign.

## Motivation and Context
Makes it easier to tell what kind of aura the berry is going to provide.

## How Has This Been Tested?
Opened berrydex and checked that the signs appear where they should.

## Screenshots (optional):
изображение
изображение

## Types of changes
- UI improvement


RegisCoaxans (approved):
I love Kasib showing as positive death aura, lol.

PR #6048, opened 2025-09-20
Remove the Quest Point price on the Galar Fossils.

aheadfrogg:
## Description

I removed the quest point cost of the galar fossils.

## Motivation and Context

I did this to match with the other fossils now that they all are just shopmons. This makes them much cheaper to resist and not stick out.

## How Has This Been Tested?

I loaded up the game and made sure that the qp price was gone and it was.

## Types of changes

- shop change


aheadfrogg:
Changed it so only the first purchase costs qp, later purchases don't.

WizardSpriter:
When testing, it seems like if the fossil mon is already owned, hovering over the lock on the Quest Points option will say 'This Pokemon cannot be owned yet', or something like that. That message doesn't make sense in this scenario since the lock will only show up if you already own the Pokemon.

PR #6050, opened 2025-09-21
Goal Tracker

### LINKED ISSUE #2211: [SUGGESTION] A tracker widget box like the one for achievements or pokeballs, but that tracks Gems
**Current Behavior**

**Improved Behavior**
I'd like to see a widget tracker box for Gems like we have for pokeballs or pokemon, etc. I have to keep going to the menu to track how many Gems I have when I'm Gem.

**Reason for change**
Because we have widgets for everything else that tracks pokeballs, battle items, flutes, pokemon and everything else that can be moved around as a widget box. I think it'd be an improvement if we could also track our Gems.

**Detailed Description**
A widget box that allows you to keep track of your gems. It should consist of a box similar to Battle Items, where there's an icon for each Gem in each box, with a number indicator underneath each icon that tracks the amount of Gems you have, live. Each row should have 6 Gems, with a total of 3 rows stacked. When the number of Gems farmed reaches a certain amount, instead of showing the amount as 120,000 it would be displayed as 120k, or 120.8k (rounded to the nearest 100th place, so that anything above 50 is rounded up, and anything under 49 is rounded down) for 120, 854. However, when you hover your cursor over the Gem's icon box, it gives the full complete number without any rounding.

I went ahead and tweaked the code for the battle item widget since it already had the boxes and numbers, and what not, so that I could provide some screenshots of what it'd look like. The first screenshot is what it'd look like as a sidebar widget (left or right) and the second is what it'd look like in the center. This widget doesn't allow to spend the Gems like you would in the Gem screen accessed via Menu, this is just for tracking them.

I've displayed what they'd look like if the amounts were in single digits, or hundreds, or thousands, hundreds of thousands and millions. I assume it will allow for billions as well so it would use the appropriate symbols for display, as well as the appropriate decimals for sums requiring them as shown for the Poison Gem (Gem #8), 343, 835 displaying as 343.8k (since the tenths place is under 49 it rounds down to 800 instead of rounding up to 900).

If there are any questions about my suggestion, I am MurderMouse on the Pokeclicker Discord.

**Screenshots:**
image
image

CypherX:
## Description
Adds a Goal Tracker that allows players to create and track custom goals. There are more things that could be added as trackables, I just added some common stuff for now.

- Create & edit custom goals and objectives by clicking the cog icon at the top right of the Goal Tracker module
- Module can be hidden in the Display settings
- Both goals & objectives can be re-ordered
- Right clicking an objective on the main screen module will open the objective config
- When an objective is completed a notification will be shown and it will play the Achievement sound effect (could change to something unique later)

Potential additions & things to figure out
- ~~A method to import/export and share goals & objectives~~
- A "Notes" section for each objective?
- ~~Implement a way to make dropdowns searchable or improve how they're presented, the Pokemon and Berry dropdowns specifically are kind of annoying to find what you want~~
- More trackables
- When should it unlock? After the Tutorial? I want to available early for challenge files but also don't want to overload new players.
- The code can probably be improved
- [your suggestion here]

## Screenshots (optional):
main screen module
image

goal tracker modal
image

objective configuration
image

## Types of changes
- New feature

technically closes #2211


src/modules/goalTracker/GoalTracker.ts
```diff
@@ -0,0 +1,104 @@
...
+    public goals: KnockoutObservableArray<Goal> = ko.observableArray([]);
+    public selectedObjective: KnockoutObservable<Objective | undefined> = ko.observable<Objective>(undefined);
+
+    initialize(): void {}
+
+    canAccess(): boolean {
+        return App.game.statistics.dungeonsCleared[getDungeonIndex('Victory Road Johto')]() > 0;
+    }
+
+    update() {}
```
RegisCoaxans:
```suggestion
```
Feature doesn't need `update()` anymore.

CypherX:
Added a "Display" tracking mode that displays the current amount without tracking progress toward a goal

PR #6070, opened 2025-10-20
Add an backup option before first swapping regions

RedzepR:
## Description

Adds a Download Button on the nextRegionModal.

## Motivation and Context

Especially in the Kanto -> Johto region change unprepared players get stuck and dont have a backup. Now they have the option shoved into their face.

## How Has This Been Tested?

Prompted to change region, saw the button, clicked the button

## Types of changes

- UX improvement


CypherX:
Thoughts on adding a button to the nextRegionModal instead of automatically downloading it? I also think the file name should mention something related to it being a pre-new region backup

RedzepR:
Yeah thats reasonable

RegisCoaxans (approved):

Issue #6076, opened 2025-10-31
[SUGGESTION]: A new berry effect allowing for Wandering Pokemon in the farm to evolve, prior to being caught


rocketguy2:
### Current Behavior

Right now, there are 63 Pokemon in the Friend Safari, the majority of which are evolutions of Pokemon that are catchable in the farm.

### Improved Behavior

Introduce a new berry effect, that emits an aura that gives wandering Pokemon currently in the farm the chance to evolve into a previously caught evolution of that Pokemon

As some optional additions to further decrease the Friend Safari (most of which would only work if the main effect was added)

A) Replace Manaphy as a wandering Pokemon connected to the Leichi Berry with Phione (This would fully work as a standalone effect to get Phione off the Friend Safari)
B) Whilst in Alola/Galar, any wandering Pokemon that have an Alolan/Galarian form get replaced with said alternate form
C) Replace certain farm wanderers with their event forms, during the given event (Bulbasaur replaced with Spooky Bulbasaur during Halloween, Pikachu with Pikachu (Clone) during Mewtwo Strikes Back!, etc.)
D) Add the following Pokemon to the list of potential Wanderers.

Goomy to the Haban wanderers
Honedge to the Kasib or Babiri wanderers
Nosepass and Corsola to the Charti wanderers
Yamask to the Kasib wanderers
Burmy (Sand) to the Yellow Berry wanderers
Burmy (Trash) to the Pink Berry wanderers
Meltan and Poipole to berries currently lacking unique wanderers. (Ganlon and Apicot?)

### Reason for change

With a combination of all of these features, you'd be able to reduce the number of Pokemon in the Friend Safari down to Floette (Eternal), Handout Happiny, Giratina (Origin), Baby Kangaskhan, Magearna (Original Color), Gogoat, and Xerneas (Active). Only 7 Pokemon.

### Detailed Description

For the berry to give this effect to, I'd suggest giving it to the Lansat berry. Not only does it already have a connection with wandering Pokemon, given it uses the Roseli berry to mutate, but it also has a very long fully ripe stage, allowing it to give this effect for a very long time. This effect would act like the Petaya berry does currently. If one berry like this exists in the farm, then the effect is active, and if a wandering Pokemon tries to appear in a space currently occupied by a wanderer, it will attempt to evolve the Pokemon in that space instead.

[...]

PR #6079, opened 2025-11-05
feat(farm): Kee aura is causing Wandering Pokémon to evolve as they appear

RegisCoaxans:
## Description
As the title says, when Kee is planted and ripe, a Psyduck wandering onto the farm would be Golduck instead. Only works on level evo if evo has been obtained once. It's actually not finalized, more like a POC for now.
What's missing : updating Pokémon Locations, OW sprites for a handful of alt Pokémon, maybe removing some extra obtaining methods (route SL Squirtle and roaming Vivillion (Meadow)).

## Motivation and Context
Partial implementation of #6076.
Gets 17 (!) Pokémon out of FS and makes resisting Pokémon like Forretress and Toxtricity more comfortable.
Makes one more gen 5 berry somewhat useful....

## How Has This Been Tested?
Got a bunch of evolved and non evolved wandering Pokémon.

## Types of changes
- New feature

### DIFF
--- a/src/scripts/farming/Farming.ts
+++ b/src/scripts/farming/Farming.ts
@@ -1119,7 +1119,10 @@ class Farming implements Feature {
             BerryColor.Yellow,
             5.7,
             BerryFirmness.Very_Hard,
-            ['This Berry remains poisonous until fully ripened. Once ripe it has a spicy and sweet complex flavor.']
+            [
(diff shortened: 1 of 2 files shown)


RedzepR:
Kee berry requires Groudon and Kyogre to be caught. While no baseline for implementation this would lock out challenge files of some pokemon to EV (e.g. Charizard in fire)
Would be a great moment to merge #5484

PR #6091, opened 2025-11-14
Cell Battery Multiplier

### LINKED ISSUE #6063: [SUGGESTION]: Modification of Cell Battery Oak Item effect
### Current Behavior

Right now, the Cell Battery Oak Item decreases the required charge in order to activate the Battery in the Underground. 

### Improved Behavior

Instead of decreasing the amount of charges needed to activate the Battery, instead increase the amount of charge gained per successful action. 

### Reason for change

Force the usage of the Cell Battery item to be more of a commitment than it currently is

### Detailed Description

With the way it works currently, there's no reason to have the item equipped unless you are currently at or over the modified value, at which point you'd equip it, activate it, then remove it again. A small amount of micromanaging can mean you effectively have the item equipped at all times. By changing it to this suggestion, there would be a reason to have it constantly equipped whilst in the underground.

Right now, the Cell Battery subtracts 5 from the cost each level. I suggest that for the new version, it starts at 1.5x the charge gained at level 0, and it increases by a 0.1x multiplier for each level up, ending in a 2x charge bonus, which is effectively the same as the level 5 Cell Battery currently. This would end up buffing the item in the early game, making the new level 0 bonus the same as the old level 2 bonus, but I think this is acceptable. If not, then the exact numbers can be more precisely adjusted and balanced.

This would end up making the item work more like other Oak Items in the game, as the vast majority of them work off a multiplicative bonus.

rocketguy2:
## Description

This is a change to the Cell Battery Oak Item, which replaces the current effect of decreasing the number of charges needed to activate the discharge, to multiplying the amount of charge gained per underground action. In addition, this also adds a new farm aura to the Kee Berry, which also multiplies the Underground Battery charge. This also allows for the underground charge to showcase a charge value up to one decimal place.

## Motivation and Context

This is an implementation of, and therefore Closes #6063 and as such, my motivation for doing this is explained there. Adding the effect to the Kee Berry was to give an example of what effects making battery charge a multiplier could have outside of modifying the Cell Battery. The Kee Berry was selected due to its current lack of purpose, but I believe this effect would be better off on a berry with a shorter lifespan, with the Kee Berry being the shortest lasting berry Gen IV+ that does not currently have an effect.

## How Has This Been Tested?

This was tested by compiling the code as is, and checking whether or not equipping the Cell Battery would cause the underground charge to increase at an increased rate. When it did, I further modified the code to make it so that at the max level, the Cell Battery would work on a 1.5x multiplier, to make sure the intermediate levels would work properly. When I got that working, I set the code back to how it is now. Some minor text edits were required to ensure that no graphical errors occur, removing the spaces in the meter showing the underground charge in the underground screen.

In order to test the new Farm Aura, I applied it to the Cheri Berry, in order to make the test easier. When I was able to see that it worked, I moved the aura to the Kee Berry afterwards.

## Types of changes

- New feature
- Modified pre-existing feature


RegisCoaxans (changes requested):
The aura feels out of place, it would need to be way too OP to be relevant, yet I don't think anyone is going to go for a faster discharge setup in the farm. Please revert this change.
Also, since this makes the needed charges unvarying, we don't need to show `/ 60`; either it could be removed or the current charges could display in percentage of needed charges.

RegisCoaxans (approved):

PR #6092, opened 2025-11-15
Creating the dungeon

Ultima1990:
Hey. Hi. Guess who's back.

Very deep currency sinks are cool. But when they were created there was no incentive to buy more than one. They were definitely _not_ created with that idea in mind. Nerfing the cost is no solution, that deep sink for the initial purchase is fine IMO. So here's a way to obtain duplicate furfrous in a dungeon. The Parfum Palace Garden. Which unlocks upon obtaining any of the special doggos and any special doggo you own is added to the dungeon boss pool.

This is extremely unfinished and entirely untested, just pushing now to show some sign of life I guess.
Actually... I _might_ just have entirely forgotten how to test in the first place lol. It's been a long time.


RegisCoaxans:
We decided against this change a while ago but I forgot to update. Closing it now.
Although something for Shardfrou and Gemfrou is debatable, it shouldn't be that, and shouldn't include the other Frou forms.

PR #6103, opened 2025-11-21
Allows for gyms from optional subregions to be selected for the Defeat Gym Leader quest

### LINKED ISSUE #6101: [SUGGESTION]: Defeat Gym Leader quest does not include gyms from optional subregions
### Current Behavior

Currently, whenever a quest that requires the player to defeat a gym leader X amount of times is created, it only includes gym leaders from the main regions, and excludes gym leaders from the Orange Islands, Orre and Magikarp Jump.

### Improved Behavior

Whenever these quests are generated, the gym leaders from the Orange Islands, Orre, and Magikarp Jump are included.

### Reason for change

No other type of quest excludes these subregions, so I see no reason for the gym leader quest to do the same. In addition, it gives the player a reason to go back to these gym leaders other than for achievement reasons.

### Detailed Description

I don't think there's much more that needs explaining.

rocketguy2:
Whenever the DefeatGymQuest is ran, it now has a chance to give a gym from the Orange Islands, Orre, or Magikarp Jump.

## Description

The DefeatGymQuest now has a chance, equal to selecting a region, to pick a gym from a subregion. At this point, it then chooses from the three subregions with gyms currently in the game, Orange Islands, Orre, and Magikarp Jump. It then runs as normal, using the gyms from the subregion selected.

## Motivation and Context

This is an implementation of, and therefore Closes #6101 

## How Has This Been Tested?

I have generated a quest that picks a gym from each of the subregions on a savefile that has all of the subregions unlocked, which all worked properly.

## Types of changes

- Expansion of pre-existing feature


RegisCoaxans (changes requested):
The subregion are hard coded, and the way it picks them rather than main gyms is uncanny.
You could instead retrieve these with something like `GameConstants.regionGyms.slice(GameConstants.Region.final)`, and filter out these whose first gym isn't unlocked and already cleared. I don't mind if subregions are as likely to be picked as main regions.

RegisCoaxans (approved):

PR #6105, opened 2025-11-22
Re: Split, I'm bringing it back!

Ultima1990:
I'm not giving up on this one. I genuinely think it is a good idea. It will pass some day, I just know it!

Starring: Kanto's encounters split between land, old rod, good rod, super rod, and surf, all unlocking at the appropriate time.
Gyms being able to award multiple key items. Thanks Jessica.
Leveling key items: One fishing rod that upgrades, instead of many fishing rods. Thanks Jaaslet.
And, of course, the game not loading any save or new game because of a console error! Classic.


src/modules/keyItems/LevelableKeyItem.ts
```diff
@@ -0,0 +1,55 @@
...
+import KeyItemType from '../enums/KeyItemType';
+import KeyItem from './KeyItem';
+
+export class KeyItemLevel {
+    public unlockReqComputed: KnockoutComputed<boolean>;
+    public unlocker: KnockoutSubscription;
+    public parent: LevelableKeyItem;
+    constructor(
+        public description: string,
+        public unlockReq: () => boolean,
```
Aegyo:
```suggestion
        public unlockReq: () => boolean = () => true,
```
try this

Ultima1990:
Two old issues persist, and one new issue has popped up.
There is currently no Oak pop up on key item levelups.
There _must_ be a land encounter type. If there is no land, every other type on the route is allowed regardless of requirements. For now, that means surfing mons on water only routes are going on land. All of those routes have requirements that correspond to owning surf anyway. I'm sure seeing all the water mons on land is gonna make the RouteInfo cry.
Speaking of the RouteInfo, that's the new issue. No clue how I'm gonna make that play nice with key item levels.

PR #6110, opened 2025-11-30
Held item that greatly boosts Alcremie EV gain

Ultima1990:
Adds the MooMoo Farm to Johto, which sells MooMoo Milk to players that have obtained all Alcremie forms. A held item that boosts EV gain for Alcremie. By doing it this way the initial grind to obtain all the forms remains intact, but the grind to resist them all, which IMO is way too much currently, is easier.
The Milk is sold for BP and there is also a hidden extra tempbattle for one freebie. The town itself unlocks at the normal time during Johto progression and features an NPC that hints at Alcremie.


src/scripts/temporaryBattle/TemporaryBattleList.ts
```diff
@@ -5746,6 +5746,93 @@ TemporaryBattleList['Rampaging Zamazenta'] = new TemporaryBattle(
...
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Vanilla)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Ruby Cream)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Caramel)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Ruby Swirl)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Matcha)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Salted)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Lemon)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Mint)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Rainbow)'),
+        ]),
```
RegisCoaxans:
```suggestion
        new CaughtUniquePokemonByFilterRequirement((p: PartyPokemon) => p.name.startsWith('Alcremie'), 'All Alcremie variations must be owned', 63);
```

src/scripts/towns/TownList.ts
```diff
@@ -1538,6 +1538,80 @@ const JohtoBerryMaster = new BerryMasterShop(GameConstants.BerryTraderLocations[
...
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Vanilla)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Ruby Cream)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Caramel)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Ruby Swirl)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Matcha)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Salted)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Lemon)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Mint)'),
+            new ObtainedPokemonRequirement('Alcremie (Ribbon Rainbow)'),
+        ]),
```
RegisCoaxans:
```suggestion
        new CaughtUniquePokemonByFilterRequirement((p: PartyPokemon) => p.name.startsWith('Alcremie'), 'All Alcremie variations must be owned', 63);
```

Kelenius:
It would probably be much better to do an Id comparison and exclude the Gmax Alcremie than to make the game do a thousand string comparisons.

PR #6116, opened 2025-12-08
Free Isle of Armor boat from Workout Sea SVG color

Ultima1990:
Title speaks for itself.

image
image


RegisCoaxans (changes requested):
It looks weirder like that in my opinion.
Rather, could you just add the ship image to the ship mapPOI, please ? So it renders on top of the overlay for the route.

RegisCoaxans (approved):

PR #6132, opened 2025-12-21
Change EON twin roamer requirements

Ultima1990:
This makes finishing The EON Duo questline easier by adding only one of them to the roamer pool, the one that didn't join your party on Southern Island.
If you are on the right quest step and own Latias, then Latios will roam. If you are on the right quest step and own Latios, then Latias will roam. If you have completed the questline, they will both roam. This last check seems redundant, but it's there to catch situations where the player loses one of the twins later on, like when mega evolving them with the real evolution challenge active. Wouldn't want that to result in a roamer vanishing.
image
image


RegisCoaxans (approved):
Technically, none of them appear to roam Hoenn, with a stupid combination of Real Evo and mega evo while still on the last step. But they do once you finish the questline in Unova, I guess it's good enough.

PR #6135, opened 2025-12-25
Meloetta switchup

Ultima1990:
A three step overhaul.
1. Accessing Anville Town once again requires owning (any) Meloetta (no using QP to skip a roamer). Access is gained through a NPC interaction in Nimbasa City. A new tempbattle was added to Nimbasa to lure players there for the interaction. This tempbattle is a subway boss, so it also rewards a small amount of BP. For the inconvenience.
2. Meloetta's alternate form now costs 125k, down from 200k. The conversation to access Anville costs 75k. This means the combined cost of obtaining your first alternate Meloetta is still 200k, but the resist grind will be more forgiving.
3. To throw the players doing weird challenges a bone, there is an NPC hidden away in Celestial Tower, that will only exist if you are the Unova champ and don't own any Meloetta (no using a roamer to skip QP), that will (PERMANENTLY, choose wisely) switch their locations. The shopmon becomes the roamer, the roamer becomes the shopmon.


CypherX:
I dislike the idea of removing player options. The change to allow skipping the roamer by spending QP was approved and I see no reason to change it back now.

Ultima1990:
Approved changes can and have been overturned. For example, the original implementation of Meloetta was approved _and very intentional_. That was overturned, though clearly I think that was mostly a bad choice, _EXCEPT_ to allow players doing mono-fighting runs access to Pirouette, which this PR addresses.
Don't be too afraid to inconvenience players. That is what games _are_. A string of inconveniences. Roadblocks thrown in the player's way that they must overcome. Without them, there is no game. Roamers are such a roadblock, and I don't see why Meloetta should be some exception. More options does not always make a game better, and can make games worse.

CypherX:
Yeah, nah, I disagree with that design philosophy and with needlessly removing player options. One roamer having an alternative is no big deal.

PR #6136, opened 2025-12-26
Hatchery modal queue

rocketguy2:
## Description

The Breeding Queue is now visible inside the Modal. In addition, an option is available to disable this in case the player wants to stick with how the modal used to look. The file breedingModal INLINETEST.html is an alternative take on what the queue could look like in the modal.

EDIT: The breeding queue is now below the pokemon list in the modal.

## Motivation and Context

This change was made in order to make it so that the player does not have to leave the hatchery modal to make changes to the queue. I see no benefit to making the player have to exit and re-enter the modal if, for example, they accidentally put a pokemon in the queue that they did not want in there. I have added an alternative layout in case the other is preferred, and added a setting to disable the feature entirely if the player is so inclined.

## How Has This Been Tested?

I have filled and emptied the queue in the Hatchery successfully, and I have tested the setting to remove the queue from the modal, which worked.

## Screenshots (optional):

Desktop

image
image
image

Mobile(ish)

image
image
image

## Types of changes

- UI improvement


RegisCoaxans:
It's hard to do change in the hatchery modal. There isn't much space left to play with.
We would like it better if the list was in a footer instead. It's better not to move the Pokémon list up and down. Also, this might fix an issue with the additional scrollbar on the Pokémon list. If it doesn't, then you'll have to (it's due to hardcoded height in [breeding.less](https://github.com/pokeclicker/pokeclicker/blob/5498b2802aac11cb4d321919c997e7a18668aa35/src/styles/breeding.less#L326)).
Lastly, could you make the setting *false* as a default, please ?

RegisCoaxans (approved):

PR #6140, opened 2026-01-03
Nerf Slow EV Challenge (or buff, depending on how you look at it)

Ultima1990:
EVs have a Slow challenge, instead of a disable challenge. The point of this is that a challenge is supposed to be... well... _challenging_, and disabling mechanics entirely kinda... isn't... that. The issue is that, based on feedback, the challenge slows down EV gain so severely that, in practice, it kinda disables the mechanic anyway. In which case the original point is lost.
This PR changes the Slow EV Challenge modifier from 10 to 4. Basically, reducing EV gain down to 25%, up from 10%.


Askellown:
next time you want to make a change for challenge runs, ultima "I don't make changes with challenge runs in mind", maybe go to the actual channel and see if people actually complain about it or want this.

PR #6141, opened 2026-01-04
Adds a new notification when the timer for discovering a mine is over

rocketguy2:
## Description

When a mine is generated that has a discovery time that's not instant, a notification is provided once the discovery time has finished. This notification can be disabled in the settings. ~~Notably, in order for this to work properly, I have it set so that when the amount of time needed to generate a new mine is 0 it instead sets it to -60, as to not make this notification appear on every mine generation. As far as I'm aware, I have solved any issues that may come from this.~~ This is no longer necessary, thanks to a modification by RegisCoaxans.

## Motivation and Context

I believe the existence of this notification makes for good quality of life, as it provided the player with an easy way of telling when the underground is ready if the player doesn't have the Underground module enabled, or if they don't have the tab enlarged.

## How Has This Been Tested?

- Ensured that the notification only appears when intended (Doesn't appear with Chaos Cavern or at level 50+, but appears otherwise once)
- Checked other usages of modified code and ensured nothing breaks from that

## Types of changes

- UI improvement


src/modules/underground/mine/Mine.ts
```diff
@@ -185,6 +185,10 @@ export class Mine {
     public tick(deltaTime: number) {
         if (!this.completed) {
             this._timeUntilDiscovery(this.timeUntilDiscovery - deltaTime);
+            if (this.timeUntilDiscovery <= 0 && this.timeUntilDiscovery > -60) {
+                UndergroundController.notifyMineFound();
+                this._timeUntilDiscovery(-60);
+            }
```
RegisCoaxans:
```suggestion
        if (!this.completed && this.timeUntilDiscovery > 0) {
            this._timeUntilDiscovery(this.timeUntilDiscovery - deltaTime);
            if (this.timeUntilDiscovery <= 0) {
                UndergroundController.notifyMineFound();
            }
```
Then you don't need to return `-60` anymore in Mine.ts, and the mine doesn't tick anymore after it hits 0.

RegisCoaxans (changes requested):

RegisCoaxans (approved):

PR #6151, opened 2026-01-14
Make time flute increase flow gain

RedzepR:
## Description

While the Time Flute is active your flow gained is increased by your regular flute bonus (up to 24%)

## Motivation and Context

In Colosseum you can use the Time Flute to summon Celebi and instantly purify a pokemon. This is a reference to it. 

## Types of changes

- New feature/Reference


RegisCoaxans:
The flute description should probably reflect this feature. Can't rely on just one NPC.

RedzepR:
Personally do not agree because its more an added bonus and it could get crowded in the hint. The description is incomplete anyway bc it doesnt mention TempBattles at all

RegisCoaxans:
What about the description being `Bonus to Battle Timers and Flow gained` then ? Sure, it could make it look like it affects BF but would it actually confuse people that it doesn't ?

RegisCoaxans (approved):
Applied the suggested change and fixed a minor bug reported over Discord.

src/scripts/towns/TownList.ts
```diff
@@ -3199,6 +3199,7 @@ const RelicSage = new NPC('Relic Stone Sage', [
     'This stone has the power to cleanse and purify the spirits of Pokémon.',
     'If you train with your Pokémon, you will gain spiritual energy, or "Flow". You can use this Flow to purify your Pokémon.',
     'Purification will take more flow with each Pokémon you purify.',
+    'It\'s said that there is a Flute with a Time altering effect. Playing on it might help to purify your Pokémon a bit fast.',
```
wormania, line 3202: `Playing it might help to purify your Pokémon a bit faster.`

PR #6154, opened 2026-01-15
Fix MKJ Quest Reward bug

RedzepR:
## Description

Changed Defeat Pokemon, Defeat Dungeon and Defeat Gym quest to calc damage in that region and not active region
Removed calc attack calculations from Dungeon Quests
In Dungeon Quests calculate how many attacks you need to defeat the pokemon in the dungeon, it was hard capped to 4. Removed the hardcap which results in Dungeons in your highest region giving out more QP. 

## Motivation and Context

Having much lower attack in MKJ while resetting made the quest give out unproportional reward.

## How Has This Been Tested?

Went into game and created a bunch of quests. Compared it to the live value

## Types of changes

- Bug fix


src/scripts/quests/questTypes/DefeatDungeonQuest.ts
```diff
@@ -29,8 +29,9 @@ class DefeatDungeonQuest extends Quest implements QuestInterface {
     }
 
     private static calcReward(amount: number, dungeon: string): number {
-        const playerDamage = App.game.party.calculateClickAttack() + (App.game.party.pokemonAttackObservable() / GameConstants.QUEST_CLICKS_PER_SECOND);
-        const attacksToDefeatPokemon = Math.ceil(Math.min(4, dungeonList[dungeon].baseHealth / playerDamage));
+        const town = TownList[dungeon];
+        const playerDamage = App.game.party.calculatePokemonAttack(undefined, undefined, undefined, town.region, true, undefined, Weather.regionalWeather[town.region](), true, false, town.subRegion)  / GameConstants.QUEST_CLICKS_PER_SECOND;
```
RegisCoaxans, line 33: Why is click/s dividing the Pokémon attack ? Shouldn't it multiply the click attack instead ?

RegisCoaxans (changes requested):
Could we make these calculations uniform accross all quest types ? They all estimate their respective difficulty differently right now.

I like that it's now uncapped in dungeon quest, might be the one to base the others on. Although I can't make sense of the formula.

PR #6161, opened 2026-01-24
Make held items more available

Ultima1990:
This PR eliminates all loot competition from the legendary tier of dungeon chests, making them contain only held items.

Johto is supposed to be the introduction of held items, but currently it doesn't really work out that way. The Wonder Chest aside, held items are incredibly rare. So rare that these type boosting held items don't really become relevant until the player has progressed way past Johto. We don't really expect most players to be farming many dungeon clears at the early stages of Johto, and yet, from my own testing and also testimony from other players, it could easily take over 100 clears to get a single Miracle Seed from Sprout Tower. This PR seeks to decrease the rarity of these items to make them more available, in the hope that Johto will serve as a proper introduction for the mechanic. And hey, if you are in fact one of those players that does like to farm many dungeon clears in the early game, good news, more held items for you.

There are a couple of ways I've thought of to make this happen. Moving held items to a different tier, removing loot competition from their current tier, or maybe even both. This PR goes for a held back approach, only going for the loot competition method for now. Just gonna see how that works out and if more needs to be done afterwards.

Note that none of the other items that were in the legendary tier are actually _removed from the dungeon_. Just from that tier. They are moved to different tiers.

Also note that I do not plan on interfering with rare, Darn Zen style mimics, if they happen to be in the legendary tier. I'm gonna leave those alone. They can be the exception. I ain't touchin' that.

A couple of other miscellaneous changes in this PR include:
Removing Magnet from Radio Tower.
Adding Magnet to Olivine Lighthouse.
Adding Charcoal to Ruby Path.
Adding Silver Powder to Pattern Bush.
Sprinkling held items all over the Hoenn dungeons.

**This is a work in progress and only touches Johto, Sevii and Hoenn dungeons so far.**


rocketguy2:
Looking through this, whilst I fully agree with the concept of this change, I don't fully agree with where you've moved specific items to. Given the new version of the Underground, I don't believe any item that can be found there is rare enough that being found in a Mythic Chest is reasonable. Especially not items like Revives and Plates.

Ultima1990:
I understand, and even somewhat agree with that critique. However, this is me following the example set by other dungeons, and I'm probably just going to continue doing that. Max Revives are mythic tier loot in other dungeons. Changing that standard would require a larger dungeon loot overhaul, and a discussion on what items should be in mythic tier now, both of which are outside the scope of this PR.
Plates getting moved to mythic was probably a mistake on my end. Either that or it is Ilex Forest, where shards monopolize rare tier as is standard, spikychu monopolizes epic tier and I didn't want to change that, and so where else could they go? It was either common or mythic. I actually flipped a coin lol.

PR #6162, opened 2026-01-25
Added event achievements

RegisCoaxans:
## Description
Added event achievements. Names are meh but I'm open to suggestions. Included : New Island clears and some Pokédex / Pokérus achievements. 25% total for now.
Event evolutions aren't available anymore if the event isn't started.
Got the Clones out of the FS by adding them to New Island. We might want to do the same to the other event Bulba lines but I couldn't bother finding a solution.
Fixed the lock icon for normal encounters in dungeon encounter list.

## Motivation and Context
Achievements are good milestones for the players

## How Has This Been Tested?
Started some events, tried the evo, completed the achievements.

## Types of changes
- New feature


src/scripts/dungeons/Dungeon.ts
```diff
@@ -1501,15 +1509,28 @@ dungeonList['New Island'] = new Dungeon('New Island',
...
+                    new ClearDungeonRequirement(50, GameConstants.getDungeonIndex('New Island')),
+                ]) }},
+                new DungeonTrainer('Armored Mewtwo',
+                    [new GymPokemon(`${clones[0]} (Clone)` as PokemonNameType, 20000, 50)],
+                    { weight: 2, requirement:  new OneFromManyRequirement([
+                        new ObtainedPokemonRequirement(`${clones[0]} (Clone)` as PokemonNameType, true),
+                        new ClearDungeonRequirement(50, GameConstants.getDungeonIndex('New Island'), GameConstants.AchievementOption.less),
+                    ])}, ''),
+            ];
+        }).flat(1),
```
CypherX: Could these encounters be individually typed out instead of generated like this? It would be easier to read, and in this format, the `as PokemonNameType` type assertion suppresses compile-time checking, so a typo'd or renamed Pokémon name could cause a runtime error or silent failure instead of being caught immediately.

Also, both clone encounters use ```new ObtainedPokemonRequirement(`${clones[0]} (Clone)` as PokemonNameType),```, should one of those be `clones[1]` instead?

src/scripts/achievements/AchievementHandler.ts
```diff
@@ -564,7 +565,7 @@ class AchievementHandler {
             }
             // Dungeons
             GameConstants.RegionDungeons[region]?.forEach(dungeon => {
-                if (TownList[dungeon].requirements.some((req) => req instanceof DevelopmentRequirement)) {
+                if (TownList[dungeon].requirements.some((req) => req instanceof DevelopmentRequirement) || dungeonList[dungeon].optionalParameters.achievement) {
```
CypherX:
```suggestion
                if (TownList[dungeon].requirements.some((req) => req instanceof DevelopmentRequirement) || dungeonList[dungeon].optionalParameters.achievement === false) {
```
`=== false` is needed here, it still generates a regular dungeon clear achievement for New Island as-is

src/scripts/Game.ts
```diff
@@ -108,6 +108,7 @@ class Game implements TmpGameType {
     }
 
     initialize() {
+        this.quests.initialize();
```
CypherX: What's the reason for moving this?

src/scripts/dungeons/Dungeon.ts
wormania: It's possible to drop the `as PokemonNameType` for type safety by forcing string literals using `as const`:

```ts
...([
    ['Venusaur', 'Ivysaur'],
    ['Charizard', 'Charmeleon'],
    ['Blastoise', 'Wartortle'],
] as const).map(clones => {
    const cloneNames: PokemonNameType[] = [`${clones[0]} (Clone)`, `${clones[1]} (Clone)`];
    return [
        {pokemon: cloneNames[1], options: { hide: true, requirement: new MultiRequirement([
            new ObtainedPokemonRequirement(cloneNames[0]),
            new ClearDungeonRequirement(50, GameConstants.getDungeonIndex('New Island')),
        ]) }},
...etc
```

src/scripts/Game.ts
RegisCoaxans: The quest class was populating the questline array in fromJSON, after achievements were initialized. So the questline achievement wouldn't work.

src/scripts/dungeons/Dungeon.ts
RegisCoaxans: Fully wrote the encounter list as requested.
I meant to put the last Pokémon of the evoline as the requirement so it's not messing with the weight total.

CypherX (approved):

PR #6170, opened 2026-02-05
More balanced CT obtained from National Park

RegisCoaxans:
## Description
Changed the formula used for CT gained from catching in the BCC.
Still based on attack but the spread is smaller now.
Max is unchanged, min goes from 2 (lmao) to 20 (lmao²) and average doubles, up to 23.

## Motivation and Context
Current formula doesn't make sense. Most bugs are extremely weak, and the strong ones unlock too late. It naturally encourages the player to postpone the grind and only catch a few species in a very large pool of encounters.

## How Has This Been Tested?
Caught some bugs and confirmed that the new values matched these from my maths.

## Types of changes
- Slight balance change I guess


PearShapedProgrammer (approved):
Not sure if I can approve, but more tokens is always good

CypherX (approved):

PR #6186, opened 2026-02-27
QP cost removed from Shard Trader

davidcforte:
QP cost removed from Shard Trader

## Description
QP cost removed from Shard Trader, Updated ShardTraderModal and ShardDeal to remove the QP cost

## Motivation and Context
The QP cost seemed extraneous and unbalanced on the shard deals, so it was removed. The Shard values can be increased if required for balance, but for now left them as they were.

## How Has This Been Tested?
Tested locally, purchased items from shard trader, tested buying too many shard items and also regular QP purchases.

## Screenshots (optional):
image
image
image

## Types of changes

- UI improvement


davidcforte:
also added caught and PKRS symbol to trader page

image

RegisCoaxans (changes requested):
Now that #6254 is merged, there is no reason not to abandon shard deals for generic trades.

davidcforte:
Made obsolete by #6282

PR #6192, opened 2026-03-02
Secret Shop Hints (Featuring Specific Berry Requirements)

rocketguy2:
## Description

Added four new NPCs, each of which give hints to the location of a secret shop that didn't have any hints already. A gentleman in Pewter that only appears after Mt. Moon is defeated, a guitarist in Fortree, an Icy Guy (real trainer class) in Snowpoint that only appears if the Snover Berry is unlocked, and an Old Lady in Geosenge who has extra dialogue if Alolan Raichu is caught.

Additionally, included with this PR is a new internal Requirement, based on having a specific Berry unlocked.

## Motivation and Context

I believe that having internal clues for this mechanics is good for the game, as is most people likely would only find out about this stuff through the wiki/companion, so giving people another way to learn about the secret shops is helpful.

## How Has This Been Tested?

I have gone to all four towns that have an NPC added, and made sure that the NPC appears as intended. Additionally, to test the new requirement, I entered Snowpoint in a save that did not have the Snover berry unlocked, and confirmed that the NPC did not show up.

## Screenshots (optional):

image
image
image
image

## Types of changes

- UI improvement


src/modules/requirements/SpecificBerryRequirement.ts
```diff
@@ -0,0 +1,17 @@
+import * as GameConstants from '../GameConstants';
+import AchievementRequirement from './AchievementRequirement';
+import BerryType from '../enums/BerryType';
+
+export default class SpecificBerryRequirement extends AchievementRequirement {
```
AevitasDragonkin: I would probably rename this to `UnlockedBerryRequirement`, sounds more like what it actually checks.

src/modules/requirements/SpecificBerryRequirement.ts
```diff
@@ -0,0 +1,17 @@
...
+    constructor(public berry: BerryType, option: GameConstants.AchievementOption = GameConstants.AchievementOption.equal) {
```
AevitasDragonkin: Since it only needs to check if you have unlocked it or not, I would remove the `option: GameConstants.AchievementOption = GameConstants.AchievementOption.equal` and make that `unlocked: boolean = true`.

Then make the required value 1 or 0 based on unlocked

src/modules/requirements/SpecificBerryRequirement.ts
rocketguy2: I would, but the current requirement for having X berries unlocked total in BerriesUnlockedRequirement, and I feel like doing it this way makes sure the two don't get confused.

src/modules/requirements/SpecificBerryUnlockedRequirement.ts
```diff
@@ -0,0 +1,18 @@
+import * as GameConstants from '../GameConstants';
+import AchievementRequirement from './AchievementRequirement';
+import BerryType from '../enums/BerryType';
+
+export default class SpecificBerryUnlockedRequirement extends AchievementRequirement {
+    constructor(public berry: BerryType, public unlocked = 1) {
+        super(1, unlocked ? GameConstants.AchievementOption.more : GameConstants.AchievementOption.less);
```
osherwott:
```suggestion
    constructor(public berry: BerryType, option: GameConstants.AchievementOption = GameConstants.AchievementOption.more) {
        super(1, option);
```
You don't actually need it to be that complex. You can go into `Requirement.ts` to get a better look at how it works, but using AchievementOption is fine. In this case, you can use AchievementOption.less for when you want the berry to be locked and it will work just fine if you also have the other changes.

src/modules/requirements/SpecificBerryUnlockedRequirement.ts
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    public getProgress() {
+        return App.game.farming.unlockedBerries[this.berry]() ? 0 ^ this.unlocked : 1 ^ this.unlocked;
+        //
```
osherwott:
```suggestion
        return Number(App.game.farming.unlockedBerries[this.berry]());
```

Coders hate it when you use this one neat trick! /j
This also goes into how `Requirement.ts` works in how it recognizes a value of 0.

src/modules/requirements/SpecificBerryUnlockedRequirement.ts
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    public hint(): string {
+        return this.unlocked ? `The ${this.berry} Berry must have been unlocked.` : `The ${this.berry} Berry must be locked`;
```
osherwott:
```suggestion
        return `The ${BerryType[this.berry]} Berry must be ${this.option >= GameConstants.AchievementOption.equal ? 'un' : ''}locked.`;
```
`this.berry` is just the number, you need to use `BerryType[this.berry]` to get its name

osherwott (commented):
I figured I may as well take a look at this.

RegisCoaxans (approved):
BerriesUnlockedRequirement → TotalBerriesUnlockedRequirement
SpecificBerryUnlockedRequirement → BerryUnlockedRequirement
Also modified the latter to use boolean parameter `unlocked`.

PR #5949, opened 2025-07-16
[In Progress] New Pokemon Snap - Lental Subregion

MarksmanTyler:
Below, I have listed everything I have implemented (and tested by playing through the quest) along with a few things that still need to be done and some ideas for how they should be implemented. A few of the images are (obviously) just placeholders for now. I am looking for feedback on this and some help with the things listed at the bottom. 
(Sorry about the repeat PR, I messed something up when creating the original and accidentally closed it while trying to fix it. This will correct the mistake though.)
Related Issue: #5941 
Discord Discussion thread: https://discord.com/channels/450412847017754644/1395107541989589152

New Pokemon added to PokemonNameType and PokemonList
- 154.01 Illumina Meganium
- 208.03 Illumina Steelix (after Crystal Steelix)
- 350.01 Illumina Milotic
- 637.01 Illumina Volcarona
- 716.02 Illumina Xerneas (after Xerneas (Active))
- 746.03 Illumina Wishiwashi (after Totem Wishiwashi)
illumina_sprites

added DummyEvoution for:
- Bayleef -> Illumina Meganium
- Feebas -> Illumina Milotic
- Wishiwashi (Solo) -> Illumina Wishiwashi
- Larvesta -> Illumina Volcarona
- Onix -> Illumina Steelix
- none for Illumina Xerneas

added New Pokemon Snap Questline
- explores the Lental Region, a subregion of Galar
- explore each island and discover the Illumina Pokemon that resides there
	- Florio Island: talk to Prof. Mirror -> Nature Park -> Illumina Meganium battles -> talk to Mirror ->
	- Belusylva Island: Elsewhere Forest -> Illumina Milotic battles -> talk to Mirror ->
	- Maricopia Islands: Lental Seafloor -> Illumina Wishiwashi battles -> talk to Mirror ->
	- Voluca Island: Fireflow Volcano -> Illumina Volcarona battles -> talk to Mirror ->
	- Durice Island: Outaway Cave -> Illumina Steelix battles -> talk to Mirror ->
	- Aurus Island: Ruins of Remembrance -> Illumina Xerneas battles -> talk to Mirror
- some of the routes are not available immediately upon receiving the quest step, follow the hints on locked routes to get to the correct route
- for each of the Illumina battles, there are 2 temporary battles followed by a gym
- after each set of Illumina battles, a new dungeon unlocks with the Illumina Pokemon as the boss (not required for the quest)
- exploring Side Path unlocks more encounters for Nature Park

added 12 Temporary Battles
[...]

PR #6200, opened 2026-03-09
feat(routes): weighted encounters

RegisCoaxans:
## Description
Implemented a weight system for SpecialRoutePokemon encounters. Also made it so SpecialRoutePokemon can have no requirement.
Magikarp Jump routes now make use of it as they used to have Magikarp added three times in the encounter list to mimic weights.

## Motivation and Context
We discussed it in dev channel and we thought it would be useful later for #5949.

## How Has This Been Tested?
Played around in Magikarp Jump and noticed the weight applied and that no Pokémon was missing.
Made sure Galar routes, which have a lot of special encounters, were still working as expected.

## Types of changes
- New feature


CypherX (approved):
Not the biggest fan of `getAvailablePokemonWeightList` being a near identical duplicate of `getAvailablePokemonList`, but it's fine for now. Maybe someone will merge it into a single function

PR #6202, opened 2026-03-11
Replaced Coba with the Belue Berry in the Passho Mutation

RedzepR:
## Description

Replaced Coba with the Belue Berry in the Passho Mutation

## Motivation and Context

It was always weird Passho used another Gen 4 berry to mutate. Oran and Belue are both purple by definition but "blue enough". Also puts it more in line with Occa (uses Spelon) and Wacan (uses Grepa)

## Types of changes

- Mutation change


RegisCoaxans:
I'm going to close this PR.
While Coba is further in the berry list, it's easier to mutate than Belue.

PR #6235, opened 2026-04-18
CustomGymPokemon

CypherX:
## Description
Implements `CustomGymPokemon` to be used in place of `GymPokemon` for gym, dungeon, and temporary battle encounters replacing the need to add pokemonList entries for these one-off or otherwise unobtainable pokemon.

Updated `???` and `You hateful little Trainer!` to use `CustomGymPokemon`. Didn't touch the hisui placeholders (yet)

## How Has This Been Tested?
Checked that everything seemed to be working. Likely I missed something

## Screenshots (optional):
image
image


src/components/templates/pokemonSpriteTemplate.html
```diff
@@ -1,6 +1,6 @@
 <script type="text/html" id="pokemonSpriteTemplate">
     <img class="enemy"
-        data-bind="attr:{ src: PokemonHelper.getImage($data.pokemon.id, $data.pokemon.shiny, $data.pokemon.gender, $data.pokemon.shadow) }"
+        data-bind="attr:{ src: $data.pokemon.getImage?.() ?? PokemonHelper.getImage($data.pokemon.id, $data.pokemon.shiny, $data.pokemon.gender, $data.pokemon.shadow) }"
```
umbralOptimatum: Instead of getImage() only sometimes being defined, it would be cleaner to require it here and add it to SafariPokemon. Every other class used with pokemonSpriteTemplate is a BattlePokemon subclass so already has getImage() available. Maybe add it to EnemyPokemonInterface just in case some future addition has a separate class.

src/modules/gym/CustomGymPokemon.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export default class CustomGymPokemon extends GymPokemon {
+    constructor(
+        private displayName: string,
+        maxHealth: number,
+        level: number,
+        private type: PokemonType[],
+        requirements: Requirement | Requirement[] = [],
+        shiny?: boolean,
+        private image?: string,
```
umbralOptimatum: The image parameter should be required.

umbralOptimatum (changes requested):
I'm wary of moving the BattlePokemon generation code out of PokemonFactory into GymPokemon and its subclasses. PokemonFactory is very messy; it stays semi-comprehensible in large part through all that mess living in the same class where it's easy to compare different encounter types. Splitting that up feels like a recipe for future headaches, especially with the circular dependency.

Customizing the BattlePokemon generation via overriding the base class method also leads to a lot of redundant code, obscuring the actual differences in behavior.

IMO a better approach would be to make `GymPokemon.getBattlePokemon()` into a PokemonFactory method. (`gymPokemonToBattlePokemon()` maybe?) It could have an if-else for the parts where CustomGymPokemon needs different handling, or it could work from a DataPokemon retrieved from a GymPokemon method that CustomGymPokemon would override.

CypherX:
Updated w/ changes from feedback

umbralOptimatum (approved):

umbralOptimatum:
Changes look good and everything's working properly on local build.

PR #6239, opened 2026-04-23
Dungeon town cleanup

osherwott:
## Description
Combined towns that had a dungeon town equivalent. Requirements for dungeons were moved to Dungeons.ts. This was done to:
- Orre "Battle" towns
- Plasma Frigate
- Exeggutor Island
- Slumbering Weald

Also changed possible dungeons in quests to use the dungeon's `isUnlocked()` status instead of the town's
The townView now only shows dungeon pokemon when the dungeon itself is unlocked
Most of these have updated town images

## Motivation and Context
Cleaner integration of requirements for affected towns
Simplifies player navigation
Fixes dungeons being available before properly unlocked

## How Has This Been Tested?
Played through a couple of quests and requirement unlocks
Checked dungeon clears after renaming them
Checked how this affects players already in a dungeon
Checked how this affects dungeon related quests and questline steps that are in progress prior to update

## Types of changes
- Code cleanup


umbralOptimatum:
Most of the dungeons changed make sense to me, but I don't see the logic for Plasma Frigate. That wasn't a nested DungeonTown so player navigation isn't really affected. If removing those temporary questline towns is desirable, we could add dynamic town background images to preserve the flavor of the different Plasma Frigate encounters, but I'm mainly concerned about the unlock hints:

<p>Screenshot 2026-04-26 at 13-31-57 PokéClicker</p>
<p>Screenshot 2026-04-26 at 13-33-37 PokéClicker</p>
<p>image</p>

With the DefeatDungeonQuest change, Dungeon.isUnlocked() should always check both its requirements and the parent town's requirements.

src/assets/images/towns/Phenac City.png
RegisCoaxans, line 1: I feel like the zoomed-in shot better compensates for the lack of scenery outside the city walls... Even when looked at in game, it really breaks the immersion.

src/modules/requirements/QuestLineStepCompletedRequirement.ts
RegisCoaxans, line 1: No. That should stay a wiki thing.

src/scripts/Update.ts
```diff
@@ -2972,6 +2972,32 @@ class Update implements Saveable {
                 saveData.farming.berryInventory = savedBerries;
                 delete saveData.farming.berryList;
             }
+
+            // Towns renamed
```
RegisCoaxans: For the record, this isn't required for the player is moved to the starting town whenever the game loads and it couldn't find a town with the name pulled from player data.
Although, the questlist does need to be updated to account for the name changes as the dungeon quests store the dungeon name.

RegisCoaxans (changes requested):
I agree, the hint is too confusing. It should be changed. Especially when on one of the last steps , it automatically triggers the hint message from trying to teleport the player back to the frigate.

PR #6240, opened 2026-04-24
Separate berry list from player data

osherwott:
## Description
Moved berry data so that it can be accessed without needing to "initialize" or go through the player data.

## Motivation and Context
Allows the use of the berry data list (BerryType, flavors, etc.) without needing to account for player data. Can be used in #5576 to have everything in modules instead of scripts.

## How Has This Been Tested?
Checked how berryData was generated in Farming.ts and what it affects and it's just a list.
Booted up the game and nothing changed.
Would appreciate feedback on if there's a better way to do this. I opted for an array to keep it similar to the existing version, but had trouble making it a const like in PokemonList.

## Types of changes
- Code cleanup


umbralOptimatum (changes requested):
BerryList doesn't need to wrap the array in a static class, it can just export the array directly. You should also change any references to Farming.berryData to reference BerryList directly. (And maybe rename Farming.berryList to avoid confusion. berryInventory?)

Everything else looks good.

umbralOptimatum (approved):

PR #6243, opened 2026-04-25
A few change on route logics

RegisCoaxans:
## Description
Reorderered some routes and reworked how route normalization computes. Had to remove the sorting, though.
It fixes a bug where optional routes have abnormally high difficulty index.
Makes Sevii 4567 able to give the player Hoenn DT / money / XP / eggsteps.

## Motivation and Context
The bug poses a long term issue that already showed when testing #5949.
Sevii 4567 giving stuff in accordance with its difficulty is neato.

## How Has This Been Tested?
Confirmed that the money gains were as expected everywhere.
Used +/- to make sure the order still matched that of live version.

## Types of changes
- Bug fix


src/modules/routes/Routes.ts
```diff
@@ -7,14 +7,22 @@ export default class Routes {
 
     public static add(route: RegionRoute): void {
         this.regionRoutes.push(route);
-        // Sort the routes so we can normalize the route number
-        this.sortRegionRoutes();
+        this.checkRouteOrder();
```
CypherX, line 10: Could this route order check be moved into a unit test, rather than validating on every `add()` and throwing at runtime?

src/modules/routes/RegionRoute.ts
```diff
@@ -9,10 +9,10 @@ export default class RegionRoute {
...
         public pokemon: RoutePokemon,
         public requirements: Requirement[] = [],
-        public orderNumber?: number,
+        public orderNumber = number,
         public subRegion?: number,
         public ignoreRouteInCalculations = false,
-        public routeHealth = undefined,
+        public routeHealth?: number,
     ) {
         this.orderNumber = orderNumber || number;
```
CypherX: This line can be removed now

CypherX (approved):

PR #6246, opened 2026-04-28
Added a Rare Hold Item boosting farm aura

rocketguy2:
## Description

Added a new type of farm aura, called "Pickup" to the Rindo berry. This aura increases the chance that wild pokemon drop a held item when being defeated. The multiplier is the same as the Jaboca berry's boost, being 1.03 times more likely per berry.

The Dowsing Machine now interacts with the drop chance using this multiplier as well, as opposed to how it used to work.

## Motivation and Context

There are two main use cases I had in mind when making this.

1. This notably speeds up rare candy farming in the early game of certain challenge runs. At the cost of keeping a farm up and running, this can speed up rare candy gains by nearly 3.5 times.
2. Makes it faster to collect certain rare hold items that can only be obtained this way. Currently this would not speed up the Black/White mane hair, or the Crystallised Shadow, but all other unique held items would see a noticeable speed up.

## How Has This Been Tested?

While having the valid berry and dowsing machine enabled/disabled, the multiplier connected to it increased as well. The rates of held items dropping noticeably increased as well.

## Types of changes

- New feature


RegisCoaxans (approved):
Nerfed the aura from ×1.03 to ×1.015 (×2.543 at most).

PR #6265, opened 2026-05-22
Update pr-preview-action

umbralOptimatum:
Update to the latest version of pr-preview-action, which will hopefully fix the PR preview deployments failing.

## Description
The action's main repository now supports setting the deployment path, so we no longer need to use a custom version. I updated the parameters which they've renamed in updates since.

## Motivation and Context
PR preview deployments aren't working. Not clear if this will fix it, but it can't get more broken.

## How Has This Been Tested?
It hasn't; changes to github actions don't take effect until merged to develop. If it doesn't work... well, it's already broken.

### DIFF
--- a/.github/workflows/deploy-pr.yml
+++ b/.github/workflows/deploy-pr.yml
@@ -53,16 +53,17 @@ jobs:
           ln -s ../../libs build/libs
       
       - name: Deploy to GH-Pages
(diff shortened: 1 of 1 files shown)


RedSparr0w (approved):
Will merge and see how it goes 🤞

PR #6149, opened 2026-01-14
Changes on hatchery helpers

RegisCoaxans:
## Description
Hatchers now have a proper level system which improves their only stat *efficiency* as per a *bonus per level*, up to level 50. No more step efficiency, everyone is at 100%.
I increased the price on most helpers to better reflect their cost. Initial and final overall efficiencies remain the same except for some (the first, shitty ones are buffed basically).
Current progress isn't lost : hatches for +X% (as integer) convert to level X.

## Motivation and Context
The current system has flaws :
- Step efficiency and attack efficiency confuse players as to how efficient it is, truly.
- The 500 pseudo levels are a needlessly too precise progress system; no one is going to benefit from +0.1% eggsteps.
- The bonus being +50% on both stats and for every helper makes it hard to balance between helpers and even between initial and final stats on one helper. Sam becomes 36 times better, lmao.

## How Has This Been Tested?
Got a few hatchery helpers to various levels and ensured the bonuses were as expected, in the display as well as when hatching a Pokémon.

## Notes
- ~~I still need to upate the help tab.~~ Done.
- attack bonuses are rounded, making half the efficiency points pointless. Another PR will tackle this.
- Values are thought to be final but can be changed if needed.

## Types of changes
- Feature improvement

PR #6269, opened 2026-05-25
refactor(partyPokemon): merged attack bonuses and lightened the rounding applied when breeding

RegisCoaxans:
## Description
AttackBonusPercent and AttackBonusAmount are now one AttackBonus value. It's also rounded to the 1000th instead of the integer. That's a change I mentioned in #6149.
It seems to decrease the file size but that's a marginal side effect.

## Motivation and Context
No use for two values when storing them in one is doable and more future proof.
I changed the rounding because helpers have a precision down to a tenth of a percent, yet in practice the efficiency is ultimately rounded to the nearest multiple of 4, lol.

## How Has This Been Tested?
Made sure the update function broke nothing in the save. Bred a few Pokémon, evolved them with Real Evo, bred with helpers, used candies. Everything seemed to work as in current version.

## Types of changes
- Bug fix
- Code change


src/scripts/breeding/Egg.ts
```diff
@@ -132,8 +132,9 @@ class Egg implements Saveable {
         if (partyPokemon) {
             // Increase attack
             const shinyMultiplier = shiny ? GameConstants.BREEDING_SHINY_ATTACK_MULTIPLIER : 1;
-            partyPokemon.attackBonusPercent += Math.max(1, Math.round((GameConstants.BREEDING_ATTACK_BONUS + partyPokemon.vitaminsUsed[GameConstants.VitaminType.Calcium]()) * (efficiency / 100)) * shinyMultiplier);
-            partyPokemon.attackBonusAmount += Math.max(0, Math.round(partyPokemon.vitaminsUsed[GameConstants.VitaminType.Protein]() * (efficiency / 100)) * shinyMultiplier);
+            const prots = partyPokemon.vitaminsUsed[GameConstants.VitaminType.Protein]();
+            const calcs = partyPokemon.vitaminsUsed[GameConstants.VitaminType.Calcium]();
+            partyPokemon.attackBonus += ((GameConstants.BREEDING_ATTACK_BONUS + partyPokemon.vitaminsUsed[GameConstants.VitaminType.Calcium]()) / 100 * partyPokemon.baseAttack + prots) * shinyMultiplier * efficiency / 100;
```
CypherX:
```suggestion
            const protein = partyPokemon.vitaminsUsed[GameConstants.VitaminType.Protein]();
            const calcium = partyPokemon.vitaminsUsed[GameConstants.VitaminType.Calcium]();
            partyPokemon.attackBonus += ((GameConstants.BREEDING_ATTACK_BONUS + calcium) / 100 * partyPokemon.baseAttack + protein) * shinyMultiplier * efficiency / 100;
```
Didn't replace `partyPokemon.vitaminsUsed[GameConstants.VitaminType.Calcium]()` in the calculation, and let's use the full vitamin name to be more descriptive

src/scripts/party/evolutions/EvolutionHandler.ts
```diff
@@ -46,14 +46,13 @@ class EvolutionHandler {
             const basePartyPokemon = App.game.party.getPokemon(PokemonHelper.getPokemonByName(data.basePokemon).id);
             // Calculate and transfer the flat attack bonus rather than bonus percent.
```
CypherX: Let's update or get rid of this comment

PR #6273, opened 2026-05-27
Update the Safari Bait

RedzepR:
## Description

Swap the effects of Razz and Nanab.
Add Pinap bait that increases the drop chance of items.

## Motivation and Context

In PoGo Razz is the catch rate increase berry and Nanab the escape berry, reverse of what Pokeclicker did.
In PoGo there is also the Pinap berry which doubles the amount of candies gotten from a pokemon.

## How Has This Been Tested?

Go to the Safari and throw bait at the pokemon

## Types of changes

- safari update


RegisCoaxans (approved):
Took the liberty of modifying your Pinap effect.
It adds a second roll for an item, instead of doubling the chance for an item. Makes it so its effectiveness wouldn't feel partly wasted after just level 11.

PR #6283, opened 2026-06-11
Ability to view which pokemon hatchery helpers will hatch next

### LINKED ISSUE #2099: [SUGGESTION] Hatchery Helper Suggestion
**Current Behavior**

Hatchery Helpers are great, but it can be a bit confusing which pokemon they will next try to hatch based on the sorting, as unlike with the queue, we can't really see the sorted list the helper will see.

**Improved Behavior**

I'd like to see the Hatchery Helper's information page have a info box on what the current sorting setup would select as the next pokemon to hatch.  This should make it clearer to everyone whether they've set their sorting option up correctly based on which pokemon they're expecting to hatch with a given sort setup.  Could be made even better with the two or three 'next pokemon' shown, or by showing the information that the Display Value option from the Filter would overlay on a pokemon when selecting for the queue next to the next pokemon to hatch.

**Reason for change**

This would add clarity to what a hatchery helper would be hatching.

**Detailed Description**

**Screenshots:**

Mockup of what it might look up, excuse the poor paint job.  the (2.4) in parentheses would be the Breeding Efficiency, or could be some other info used by the Display Value filter on the pokemon queue tab.
untitled

CypherX:
## Description
Implements the ability to see the next 5 pokemon a hatchery helper will grab based on the current settings. Also verified the changes to how helpers determine their next pokemon doesn't add any additional overhead.

 Screenshots (optional):
image
image

## Types of changes
- New feature
- UI improvement

closes #2099


RegisCoaxans (changes requested):
Could it say that there is no Pokémon available to breed instead of an empty line ?

Image

CypherX:
> Could it say that there is no Pokémon available to breed instead of an empty line ?

done

RegisCoaxans (approved):

PR #6291, opened 2026-06-29
Fix layout shift causing shake due to scrollbar

ozgurerturk:
## Description

Adds a stable scrollbar gutter after the game starts to prevent shaking caused by the vertical scrollbar appearing and disappearing at certain window sizes.

## Motivation and Context

This problem can block certain layouts users prefer and is a nuisance in general. Suboptimal for the user experience

## How Has This Been Tested?

Tested with different windows sizes, different device layouts (pc and mobile sizes), Full width (5 columns) game display style on.

## Screenshots:

Before:
shakey-ezgif com-video-to-gif-converter

After:
after_fix-ezgif com-video-to-gif-converter

## Types of changes

- Bug fix

### DIFF
--- a/src/scripts/App.ts
+++ b/src/scripts/App.ts
@@ -48,6 +48,9 @@ class App {
             App.game.start();
             GameLoadState.updateLoadState(GameLoadState.states.running);
 
+            // Add stable scrollbar gutter class to prevent layout shaking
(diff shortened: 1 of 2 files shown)


CypherX:
While this does fix the issue, I'm not a fan of the permanent scrollbar gutters this adds even when scrolling isn't needed. That's mainly the reason I haven't bothered pushing this myself

PR #6310, opened 2026-07-25
Updated regex to also remove whitespace before digit when winning/losing a Gym Battle

mennockramer:
## Description

The regex that removes the digits from the end of Gym Leader names when winning/losing a Gym Battle now removes the trailing space also.

For example:
Before: Kareign 2 -> "Congratulations, you defeated Kareign !"
After: Kareign 2 -> "Congratulations, you defeated Kareign!"

## Motivation and Context

This tiny bug annoyed a friend of mine.

## How Has This Been Tested?

Loaded friend's saves and won and lost against:
- Fast League Kareign (Kareign 2)
- Ultra League Jump Champ (Jump Champ 2)
- Elite 4 Jump Champ (Jump Champ 3)

Messages are fixed in all cases.

## Types of changes

- Bug fix

### DIFF
--- a/src/scripts/gym/GymRunner.ts
+++ b/src/scripts/gym/GymRunner.ts
@@ -90,7 +90,7 @@ class GymRunner {
         if (GymRunner.running()) {
             GymRunner.running(false);
             Notifier.notify({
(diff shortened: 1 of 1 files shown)


RegisCoaxans (approved):
Made it work for Johto gyms again as well.
Thanks for the PR.

PR #6314, opened 2026-08-02
Hide starter skip during No Click Attack challenge

ilkecan:
## Description

If starter capture is skipped while No Click Attack is active, the selected starter is recorded without adding a Pokemon to the party. Click attacks are then disabled, while the empty party deals no automatic damage. This leaves progression soft-locked until the challenge is permanently disabled.

This change prevents the Capture Starter option from appearing when No Click Attack is already active.

The conflicting state can still be reached through a less common sequence:

1. Unlock and disable Capture Starter.
2. Select a starter.
3. Enable No Click Attack before defeating the starter.

I considered forcibly re-enabling the capture filters in this case, but that would override an explicit player choice. The state is also recoverable by disabling No Click Attack. Players who want to preserve the challenge can instead restart the save, since this situation can only occur during the initial setup.

## Types of changes

- Bug fix
- New feature
- UI improvement

### DIFF
--- a/src/scripts/StartSequenceRunner.ts
+++ b/src/scripts/StartSequenceRunner.ts
@@ -86,7 +86,7 @@ document.addEventListener('DOMContentLoaded', () => {
                            onclick="StartSequenceRunner.pickStarter(GameConstants.Starter.Special)">
                     </div>`);
             }
-            if (StartSequenceRunner.noStarterCount == 20) {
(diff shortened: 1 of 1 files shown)


RedzepR:
Skipping starter is a hidden expert feature for challenge runs. Youre suppose to use the shiny code to get passive damage

ilkecan:
Ah, OK. So the state this PR prevent is something supported by the developers or expected by the community.

PR #6316, opened 2026-08-10
Minor AchievementHandler cleanup & optimizations

CypherX:
Some minor cleanup and optimizations in AchievementHandler

- Unified dungeon clear achievement creation
- Deduplicated work/logic in several places, removed redundant code
- `findByName` now uses an achievement map instead of scanning the entire list
-  Bonus calculation does one pass over the achievement list instead of one per category

see comment further down for additional changes


DreamNya:
I took a look at the current call sites of `findByName()`, and I'm wondering if keeping an additional `Map` is worthwhile here.

As far as I can tell, `findByName()` is currently called very infrequently:

* `AchievementHandler.fromJSON()` — only during initialization, although it may perform multiple lookups for the saved achievement names.
* `AchievementTracker.fromJSON()` — once during initialization.
* `unlockAchievement()` — only used by a small number of special achievements.
* `ProfNPC.dialogHTML` — when interacting with the relevant NPC.
* `MapHelper.ableToTravel()` — when changing regions.

With ~2000 achievements, `Array.find()` is of course O(n), while the map gives O(1) average lookup, but given how infrequently these lookups happen I wouldn't expect the difference to be measurable in practice.

The map also introduces a second representation that needs to stay in sync with `achievementList`. Currently that's safe because the map is built after the achievements are added, but `addAchievement()` / `addSecretAchievement()` themselves don't update it, so this adds an invariant that future changes would need to preserve.

Unless this is intended to support some future code that will call `findByName()` much more frequently, would it be simpler to keep the list as the single source of truth?

Since achievement names are unique, `findByName()` could still avoid scanning past a matching but currently unachievable achievement by separating the lookup from the `achievable()` check:

```ts
public static findByName(name: string): Achievement | undefined {
    const achievement = AchievementHandler.achievementList.find(
        achievement => achievement.name === name
    );

    return achievement?.achievable() ? achievement : undefined;
}
```

I may be missing some planned use for the map, though, so feel free to ignore this if it's laying groundwork for something later.

CypherX:
Fair - that change by itself likely wouldn't bring any noticeable benefits, however I believe it's worth keeping.

I bring additional optimizations, with noticeable benefits:

### Achievement location indexes
Replaced three `achievementList.every(...)` scans with route/gym/dungeon achievement maps built once. This primarily impacts when map and area bindings invalidate and/or recalculate offering a significant decrease in overhead. Noticeable improvements when catching a new pokemon, refreshing the quest list, and anything that causes map/area bindings to re-evaluate

### `getGymIndex` / `getDungeonIndex` memoization
Instead of calling `RegionGyms.flat()` or `RegionDungeons.flat()` and doing a linear scan every call, a name → index map is now built once. Ties into the above change to offer further improvements as well as improvements in other areas as these two functions are called often

### `allAvailableShadowPokemon` rewrite
Currently this function goes through the `normalEncounterList` / `bossEncounterList` getters, which rebuild a full `EncounterInfo` per enemy (party lookup, caught checks, etc.) then discarded everything except the name. Now walks `enemyList` / `bossList` directly, significantly reducing overhead. Again, further improves map binding re-evaluations, and dungeonView / townView rendering

### Duplicate shadow pokemon in `calculateTownCssClass`
`possiblePokemon` was `[...allAvailablePokemon(), ...shadowPokemon]`, however `allAvailablePokemon()` already pulls the shadow list

### `isEncounterHidden` / `isTrainerEncounterHidden`
Cleanup only - the encounter/boss visibility rules existed in multiple places, this just puts it in a common location

### Benchmarking 
Results for when map/area bindings re-evaluate (late game save):

| Stage | Time |
| ------------- | ------------- |
|  Baseline | 157ms |
| + achievement index maps | 105ms |
| + `allAvailableShadowPokemon` rewrite | 77ms |

PR #6330, opened 2026-08-20
Fix map overlay for early Pokérus

RegisCoaxans:
## Description
Setting was using dungeon clear requirement.

## Motivation and Context
Setting was using dungeon clear requirement.

## How Has This Been Tested?
Purchased the virus, it added the map filter for Pokérus.

## Types of changes
- Bug fix


src/modules/requirements/ObtainedKeyItemRequirement.ts
```diff
@@ -0,0 +1,19 @@
+import { AchievementOption } from '../GameConstants';
+import Requirement from './Requirement';
+import KeyItemType from '../enums/KeyItemType';
+
+export default class ObtainedKeytemRequirement extends Requirement {
```
CypherX: `ObtainedKeytemRequirement` → `ObtainedKeyItemRequirement`

RegisCoaxans: I'm so blind...

CypherX (approved):
There are a few instances of CustomRequirement being used for this same purpose, those should be swapped out for this new requirement, but that can be done in a different PR by anyone.

e.g. `new CustomRequirement(ko.pureComputed(() => +App.game.keyItems.hasKeyItem(KeyItemType.Safari_ticket)), 1, 'Obtain the Safari Ticket')`

